PR #1, opened 2017-06-23
README: Eclipse instructions

kasemir:
We also need maven instructions.
Or ant build file.


shroffk:
Will add maven instructions...not sure about ant

Issue #10, opened 2017-08-22
Selection results in NoClassDefFoundError


kasemir:
1) Build and run via run_maven.sh
2) Open 'Greetings' to see the current selection
3) Open 'Probe'
4) Enter PV name, press enter

-> stack trace shown below.

Would expect to see the PV name in the list of selections.

Push 'search' still works OK, PV connects and probe updates

```
Exception in thread "JavaFX Application Thread" java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at javafx.fxml/javafx.fxml.FXMLLoader$MethodHandler.invoke(FXMLLoader.java:1787)
	at javafx.fxml/javafx.fxml.FXMLLoader$ControllerMethodEventHandler.handle(FXMLLoader.java:1670)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
	at javafx.base/com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
[...]

PR #11, opened 2017-08-22
Eclipse 'Java Project' Files

kasemir:
Modify .product and .classpath files to be plain Java projects with project dependencies.

At least for me that allows a simple "Import Existing Project" and then "Run" on the Launcher.java.

If you will, there are two downsides:
* Lengthy dependencies/.classpath, but then that is the "target platform". In here, one does need to edit version numbers (maybe in pom.xml and then in the .classpath). Rest of the application code doesn't care.
* There are now META-INF/services files. In a way, that may be more obvious in the long run compared to annotations.


shroffk:
I will walk through a complete import and build within eclipse, @claudio-rosati do you wish to test this too?

I will merge this PR once the above step is complete.

kasemir:
Note you do need to once build 'dependencies' with maven, or have the dependencies/target/lib in place by other means, before the import!

kasemir:
I now tried it on Linux and Mac as described in the README.MD:

Get Java 9, Eclipse 4.7 with added J9 support.

`git clone https://github.com/kasemir/phoebus.git`

```
# Build target platform
cd phoebus
```
Then import General/Existing project (NOT: Maven/Existing Maven Project!!), run.

kasemir:
Did you try to still import it as Maven/Existing Maven Project?
Presumably that would update the .classpath and .project files to use the maven info, and on a good day, on your computer, that'll actually work.

shroffk:
Was able to import them as pure eclipse projects.

Was able to import them in another workspace as maven projects too.

kasemir:
Very good, that's the best we could hope for.
I still can't get the maven build from the command line to fully work, there's always issue #10 .
For the core-ui addition I tried to add the pom.xml, but I'm really giving up on maven, life's to short.

shroffk:
Ok, I will look into this and maintain the maven build and the packaging of the product with it.

I have been looking into the target creation with maven...It seems like the way to do this would be to use the  -Dmaven.repo.local and then taring that for distribution.

shroffk:
I tried importing the project as ```existing projects``` from the latest master branch however I still have errors related to the java compliance

image


The jdk installed is jdk 9 and the java compiler preferences have been set to compliance level 9

shroffk:
The above issue was resolved with a manual resetting of the project properties

properties -> java compiler 
and unchecking the "enable project specific settings"

kasemir:
On Linux and Mac OS X I have not run into that problem again.
The only thing that happens: When I run "ant clean run" or "sh run_maven.sh" to rebuild, using the same sources that Eclipse has open, I later get errors in Eclipse because the class files have changed.
This is fixed vie Project/Clean.. (*) Clean all projects.

I wonder what's selecting project specific settings in your case.

Issue #8, opened 2017-07-17
How to manage docking


shroffk:
@kasemir says"

> 
> Hi:
> 
> Claudio, I think you mentioned in the last hangout that Java 9 would include updated Tab/TabPane/.. support that includes docking.
> Do you have any reference for that?
> The closest I found was a request for docking tabs, ending in "targeted JDK10", https://bugs.openjdk.java.net/browse/JDK-8092098 
> 
> I briefly looked at the following
> 
> http://berry120.blogspot.co.uk/2014/01/draggable-and-detachable-tabs-in-javafx.html
> https://github.com/alexbodogit/AnchorFX
> https://github.com/RobertBColton/DockFX
> 
> The first one has a trivial amount of code. The blog includes a video which just shows a window with tabs, you can drag tabs out into individual windows, then add tabs to them. I think that's perfect, just like a web browser.
> 
> AnchorFX and DockFX seem similar, in fact https://github.com/alexbodogit/AnchorFX/issues/20#issuecomment-263724971 suggests that one copied the other when it should have been a clone.
> They support the idea to 'split' a window area into subpanels, like Eclupse where dropping a tab to the left/right/top/bottom re-sizes the original tab that occupied the complete area. That's the type of operation most CSS users never quite understand and thus hate.
> 
> The draggable-and-detachable-tabs-in-javafx.html demo is nice on Mac OS X, and again I think it does all that we need.
> Trouble is that none of the demos work on Linux. I tried RedHat 6 and 7, with Java 8 and 9.
> Starting to drag any tab results in a tiny new window, but can't 'drop' anywhere.
> For AnchorFX & DockFX there's an addition window with docking icons, I guess to drop  left/right/top/bottom within the existing window, but again I can't 'drop' there.
> 
> -Kay

claudio-rosati:
My answer:

> Hello Kay,
>
>your reference is correct. I’ve seen this page probably before it was re-targeted for JDK10. That is very bad.
>
>I like AnchorFX/DockFX, and I prefer them , the problem is that these project seems inactive.
>
>There are also the following:
>
>https://github.com/andy-goryachev/FxDock
>http://www.drombler.org/drombler-fx/0.10/docs/tutorial/docking-framework.html#dockingFramework
>
>Claudio

kasemir:
There is basic docking: New Probe, PVTree, ... opens in tab. Tabs can be rearranged, detached.

[...]

PR #13, opened 2017-08-23
core-ui, Docking

kasemir:
Adding a core-ui module, for now with the docking library and demo.

Next would be moving the PhoebusApplication to the core-ui module (because it _is_ UI related) and then using the docking tabs in the PhoebusApplication, probe, greeting, ..


kasemir:
Updated PhoebusApplication (now in core/ui), Probe, Greetings to use docking:

docking

For now every Probe etc. opens a _new_ stage, because it's not passed a parent stage/scene/DockPane to which it could append itself.
User can then move the probe into an existing window.

Better would be another change to the way actions are called that allows them to get the parent stage/scene/DockPane. Same when starting an application.
For what it's worth, the Application isn't used at all.
Instead of the LaunchProbe action opening a stage, it should send a "Start the probe application" message to a framework, which then handles starting the probe application with the current parent window.

kasemir:
Updated `ToolbarEntry` to receive the parent `Stage`.
This allows Probe etc. to open as a new tab in the parent stage.
User can then detach, move around etc.

kasemir:
When pressing the "Control" key while pushing the tool bar button, "Probe" etc. will open in a new standalone stage.

kasemir:
Fixes #8

PR #16, opened 2017-08-30
Creation of a main parent

shroffk:
goals:
reduce size of child poms
single place for declaration of properties and repository locations

@kasemir @claudio-rosati @willrogers


kasemir:
Well, the child poms actually get even larger, but I guess that's Maven..
Still, fundamentally a good idea, go for it.

willrogers:
I agree this is a good idea.

kasemir:
All-right, now they're smaller! Good.

PR #18, opened 2017-08-30
PVtable

kasemir:
Beginnings of PV Table.

Has the model, opens with some predefined loc:// and sim:// content.
Runs, can change PVs, take snapshot.

Still missing a lot, but I'd like to have this merged because if we now rename/move things, like move core components into core/ui, the refactoring will automatically update this code.


kasemir:
pvtree

PR #19, opened 2017-08-30
Issue 17

shroffk:
removing the annotation and adding the manual files.


shroffk:
@kasemir oops merged by mistake...but you can still check

PR #20, opened 2017-08-30
Issue_12

shroffk:
WIP: do not merge

I have started updating the application interface...

@claudio-rosati @kasemir @willrogers any ideas?


kasemir:
I suggest you merge the PVTable at its current state and then check LaunchProbe, PVTreeToolbarEntry,  PVTableToolbarEntry.

The toolbar entry call() receives a Stage. Should be DockPane, because that's what's really needed, but can't until toolbarentry moves into core/ui.
In any case, the toolbar entry just like the application needs the parent DockPane (or Stage from which you can hack down to the DockPane) to open in the right place.

PVTree and PVTable already have an "Application" with start(DockPane). That or open(Dockpane), something that receives the parent, would need to be in the Application interface.

shroffk:
hmm...I also tried to port the javafx logging configuration application.

Was going to talk to you about the Toolbar and Menu interfaces next.

kasemir:
Should we have a hangout tomorrow, 10:30 or later?

I think the logging configuration is similar: It would open in a tab, so it also needs to know the parent

-> Application, MenuEntry, ToolbarEntry all move into core/ui and receive the parent DockPane when called. They can ignore that and open a new Stage, or just a Dialog, but at least they have the information to add themselves as a tab to the existing window.

kasemir:
You added logConfiguration to phoebus-product/.classpath, I added applications-pvtree.
Resolve by including both, and then I suggest to merge this, so we can both see it.

kasemir:
The master merge joined both log and tree application in the phoebus-product/.classpath, so go ahead and merge.
Then we should have a hangout to see if we add the DockPane as an arg to all 'start/open/..' type calls, or the Stage, or ..

kasemir:
The log configuration works for me once I removed the missing test src folder from the classpath.

I did that on my pvtable development branch like this:

https://github.com/kasemir/phoebus/commit/adf26f606873953ed649d5348696f7484c37badb

PR #24, opened 2017-09-06
Context menu sets active pane

kasemir:
On Linux and Mac it is otherwise sometimes possible to invoke a context
menu in window B while window A is still 'active'.


kasemir:
Obsolete, included in #26

PR #25, opened 2017-09-06
Issue 12

shroffk:


shroffk:
Clean up interfaces using ```stage```

Additionally updated the MenuService to create a tree view on menu contributions which can then be used by the phoebus application to create a menu/submenu structure in the top menu

kasemir:
maven build error, looks like missing 

```
   <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
```

kasemir:
If some pom.xml adds a new dependency, like core/framework/pom.xml:
```
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <version>3.4</version>
+      <scope>test</scope>
+    </dependency>
```
then we should also add that to the dependencies/pom.xml.

That way, building the dependencies module will download all external dependencies once.

Also, the dependencies/pom.xml already has comments on which combination of junit, mockito, hamcrest "work", so adding easymock there allows us to keep track.

shroffk:
oops, this is my bad...this was just a test thing which should not have been committed since I am not actually using mock objects

kasemir:
OK, let's merge the pull requests

shroffk:
Thanks!

PR #26, opened 2017-09-06
Dock item with input, Singleton app, Drag/Drop of PV name

kasemir:
Adds a DockItemWithInput:

Similar to DockItem, but has a URL as input (or a File that's wrapped as a URL).

When closing the stage, it closes the tabs.
When closing a tab, and the tab is 'dirty', it prompts for a file to save.

To try, open PV Table, click one of the check boxes in the 1st column, then close the tab.
The file save dialog is of course on the UI thread.
When writing the file, this is done in a background job.
On success, the tab is then closed back in the UI thread.


kasemir:
Adapted to interfaces that no longer pass the stage.

Note core-ui ContextMenuHelper, used by PVTable and Tree to add the context menu entries.
ContextMenuService in core-framework provides the information, UI agnostic, and ContextMenuHelper then turns that into JavaFX MenuItems, appending that to an application context menu.

kasemir:
Added a menu entry Debug/PV List.
It's useful to check if applications release their PVs.

In addition, it demonstrates a 'singleton' application. There's at most one instance of the PV List running. Trying to open a second one will raise/select the existing tab.

kasemir:
LoggingConfiguration should also be a singleton

kasemir:
@kunal I suggest we merge this because it's as far as I can go with the PV Table until there's a way to open it with an input from the command line.

PR #27, opened 2017-09-08
Add Appveyor build support

mattclarke:
Just for fun I added Appveyor support for a Windows build.
I have added the build badge too, but it points to the build of my fork so will need to be changed if you decide to merge the PR.


willrogers:
Good idea.

kasemir:
Fine w/ me since it doesn't change any of the build setups, it just invokes them.
Do you want to change the build badge right now so it'll "work" when merged?

shroffk:
Great, I like this...we allow as much diversity as possible with various CI technologies.

PR #28, opened 2017-09-08
Docking: Hide single tab header; drag tab 'out' to detach

kasemir:
When a dockpane has only a single tab, the tab headers are hidden.
Once you drag in another tab, the tab headers are again shown.
#8

Could extend this a little more to update the title of the window instead of the (now hidden) label of the tab when there's only one tab.


kasemir:
When a window contains only a single pane, there's no tab, and what used to be the tab label moves to the window title. When another pane is dragged in, tabs reappear and the window title reverts to just "Phoebus".
It's pretty neat how compact the secondary windows, those without the menu, then get.

screen shot 2017-09-08 at 10 47 03 pm

kasemir:
@shroffk OK to merge?

Issue #4, opened 2017-07-05
Managing preferences


shroffk:
Include the use of the java preferences API

kasemir:
Basic API is
```
    // Get preferences for the _package_ of MyClass
    final Preferences prefs = Preferences.userNodeForPackage(MyClass.class);

    // Read, write
    int calls = prefs.getInt("calls", 0);
    prefs.putInt("calls", calls+1);
```

Suggestions:

Each module that needs to read (and in some cases write) preferences uses some class in its base package to access preferences.

The launcher has command line options to export and import preferences, basically calling
```
    Preferences.userRoot().node("phoebus").exportSubtree(new FileOutputStream("export_file.xml"));
```
resp.
```
    Preferences.importPreferences(new FileInputStream("import_file.xml"));
```

This provides the essential functionality.

Downside:
 * The exported/imported settings are in a rather verbose XML format. Not a show stopper, could eventually support export/import in other formats.
 * The default implementation stores the settings in ~/.java or Windows registry, depending on OS. This means there's one set of preferences per user. Each instance of Phoebus for that user will have the same preferences. Same user cannot run instances with different settings. To support this, we would have to implement our own `PreferencesFactory` service.

kasemir:
The PV Tree and core/pv library use `Preferences`.
Reading them is easy, and it overall "works", with some concerns:

### Knowing what's available
With RCP, a `settings.ini` in each plugin can list all preference settings, define the defaults, and ideally add some explanatory comments.

The closest equivalent is now for example the `org.phoebus.applications.pvtree.Settings.java` source code.

You can "export" the current settings, but if code only reads and never writes the settings, that export would be empty, not presenting the available preferences and their current (default) values.

I thus coded all access like this:
```
Preferences prefs = Preferences.userNodeForPackage(...);
value = prefs.get(THE_NAME, default_value);
prefs.put(NAME, value);
```

So once the code runs, the preferences are populated, initially with the default values.

When then exporting them to a file, you can at least see the values.
There is no explanation on their meaning in the XML, though.
Still, you can edit the exported file and then import it.

### One set of Preferences per User
Once you run
```
./phoebus.sh -settings settings.xml     
```
this imports the preferences from `settings.xml` for that user.
From then on, you can run without the command line arg.

This can be an advantage, but also a problem if you want to run different instances with different settings, for example different "top displays".
Would have to use different user accounts, or implement our own `PreferencesFactory`.

kasemir:
Questions:

[...]

PR #30, opened 2017-09-11
#4 test version of a file based multi system preferen…

shroffk:
…ces implementation""

WIP:
TODO persistence

### DIFF
--- a/core/framework/.classpath
+++ b/core/framework/.classpath
@@ -3,6 +3,7 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
 	<classpathentry kind="src" path="src/test/java"/>
+	<classpathentry kind="src" path="src/main/resources/"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/4"/>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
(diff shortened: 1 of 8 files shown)


kasemir:
After I pull this, then try to start the LogManager, it ends with
```
Exception in thread "main" java.lang.NullPointerException: inStream parameter is null
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at java.base/java.util.Properties.load(Properties.java:364)
	at java.logging/java.util.logging.LogManager.readConfiguration(LogManager.java:1556)
```

kasemir:
Sorry, ignore, I had deleted the phoebus-product/src/main/resources/logging.properties. All fine.

kasemir:
My Eclipse IDE reported "Updates availalble" for the Java 9 support. I installed those, and ever since I keep getting strange errors.

kasemir:
Compared to the Linux FileSystemPreferences, this one always creates a new, random-number based user root, so I cannot actually read the settings that were previously saved.
Can you remove the `createUniqueInstanceRoot()`?

kasemir:
Regarding the IDE, the update to "Eclipse JDT (Java Development Tools) Patch with Java 9 support (BETA) for Oxygen development stream", 1.1.1.v20170826-0521_BETA_JAVA9, was a bad idea.

After I installed that, I suddenly got errors like "Can only iterate over an array or an instance of java.lang.Iterable error " for this type of code:
```
List<String> list = ...;
for (String item : list)
```

A Project/clean didn't help. Just typing that same code again would fix it, until the error suddenly reoccurred. Reverting back to the Java 9 support 1.1.1.v20170627, which was thankfully easy via Help/ About, Installation Details, .., Revert, the problems disappeared.

shroffk:
> Can you remove the createUniqueInstanceRoot()?
Yes, this is a todo. I wanted to also figure out how the preferences should be persisted.

If you restart phoebus which preferences are remembered? especially when you have multiple instances how do you (if at all) do you pick and choose the preferences to be remembered between restarts.

PR #34, opened 2017-09-13
Memento

kasemir:
@shroffk  It saves the basic location of all stages on shutdown and restores that on startup.


kasemir:
For what it's worth, I started to track the open windows (so they can be persisted), when I noticed that Java 9 added `Window.getWindows()` that already has that information.
To identify each stage, I used the `Window.getProperties()` to add an "id".
Similarly, the existing JFX properties map of each Tab could be used to track the application name, which can then be saved and later restored.

shroffk:
> For what it's worth, I started to track the open windows (so they can be persisted), when I noticed that Java 9 added Window.getWindows() that already has that information.
To identify each stage, I used the Window.getProperties() to add an "id".
Similarly, the existing JFX properties map of each Tab could be used to track the application name, which can then be saved and later restored.

oh great...this sounds good

PR #35, opened 2017-09-13
Issue 32

shroffk:
making PR's to ensure we minimize conflicts...this still needs work


kasemir:
Right, will pull as soon as travis finishes

shroffk:
I need to fix the URI usage...

kasemir:
It's OK to try several things as long as we merge frequently.

Issue #32, opened 2017-09-12
ResourceHandler service


shroffk:
Goal: Maintain a list of applications and the resources it can handle
```
phoebus -app "probe"
phoebus -app "probe?pvs=pv_name1,pv_name2"
phoebus -resource /tmp/example.opi
phoebus -resource /tmp/example.opi -resource /tmp/example.plt
```

shroffk:
TODO:
Update the Applications interface
 - rename Application to AppDescriptor
 - remove the openWithSelection method (this should be handled but the contextMenu contribution)
 - move the open(resource... resource) and all other resource related methods to the resource handler
 - the only methods in the AppDescriptor would be open() and a default open(String... args)
 - have a methods to get display name and app name

shroffk:
@kasemir 
I like the direction of the last commit, it mainly used core java libraries
java.net, java.nio, etc...

I am still ironing out the details but I feel that this can help us support all the above operations in a nice standard way.

shroffk:
Can you think of more test cases for resources location
relative paths, absolute paths, ... I think we should add more unit test for the utility class

kasemir:
It would be good to have tests for
```
parseQueryArgs(createAppURI("some_app"));
parseQueryArgs(createAppURI("some_app?file=x"));
parseQueryArgs(createAppURI("some_app?pv=x&pv=y"));
```

Related: I'm not sure we're "done" with this API.

Right now `AppResourceDescriptor.create(String resource)` receives the complete "some_app?file=x".
This means that `PhoebusApplication.openResource(path_to_resource)` which used to just call `app.create(path_to_resource)` now needs to call that as
`app.create(app.getName() + "?file=" + path_to_resource)` .

Otherwise `AppResourceDescriptor.create(String resource)` needs to have more code to decide: Was I passed a URL where I have to parse the query, or is it just a plain path?

Maybe it's fine as is, `AppResourceDescriptor.create(String resource)` will always get the URL, so then we should change it to `AppResourceDescriptor.create(URL app_resource)`.

In any case, I need to update the way arguments are passed to an existing phoebus instance. That used to only send resource names. Maybe it should send the complete "-app ..." or "-resource ..." command line text, so when the 'server' instance receives additional requests from another command line they're parsed the same way as actual command line arguments.

kasemir:
This is done:
```
./phoebus.sh -help
...
Command-line arguments:

-app probe                              -  Launch an application with input arguments
[...]

Issue #15, opened 2017-08-25
UI state persistence


kasemir:
When applications stop, they need some way to persist information,
and when applications start, they need some way to retrieve it.

The location & size of each stage as well as the current set of tabs should be persisted the same way on shutdown, then automatically starting those applications on restart.

This likely means every application instance needs a unique ID to associate it with its persisted information.
Do apps need a way to opt out, so they're not restored after phoebus restart?

kasemir:
The location of all windows, the tabs that were open, the input file for tabs that do have an input are persisted.

By making the .phoebus/memento read-only, or by always putting a "golden" file in place before starting phoebus one can get the behavior where it always starts in a known layout.

"Perspectives" would be a separate issue, but the basic mechanism for saving and then restoring a layout are there, just for now using one setting automatically, instead of switching between multiple settings at runtime.

PR #38, opened 2017-09-14
Issue_32

shroffk:
Address issue #32 #15 #8


kasemir:
@shroffk I'll turn the LoggingConfiguration into a singleton app similar to PV List: Will be saved/restored. Trying to open more than one will just show the existing one.

kasemir:
@Kunal I'll remove the last 2 calls to the deprecated DockItem, the one without app descriptor: Greeting, Welcome.

kasemir:
@shroffk The DockItem is now only available as `DockItem(AppInstance, content)`, so all tabs get saved & restored.

What'll be helpful now is a one-stop place to get all the applications, because we have several places where we iterate over all the `ServiceLoader.load(AppDescriptor.class)` and `ServiceLoader.load(AppResourceDescriptor.class)`:

 ```
ApplicationService:
    // Get app for the name
    // (May return AppDescriptor or AppResourceDescriptor)
    AppDescriptor getApp(String name);

   // Get apps for resource
   List<AppResourceDescriptor> getApplications(URL resource);
```

If  you want to keep the latter in `ResourceHandlerService`, or move that into a more generic `ApplicationService` doesn't make a difference to me.

**Comment by kasemir** [comment]
I think I found all places where we used to create another AppDescriptor instance.
Those now all get the one and only AppDescriptor for a name from the ApplicationService.
That's also used to start an app via "-app .." from the command line, but it's not yet passing any resources from the command line to that app.

Can do `-app pv_tree` but not `-app pv_tree?SomePVName`.

**Comment by kasemir** [comment]
Can do
```
-app pv_tree
-app probe
-app pc_table
```
=> AppDescriptor.create()

```
-resource /some/file.pvs
```
=> AppResourceDescriptor.create(that_resource)

TODO:
```
-app probe?pv=SomePVName
-app probe?file=SomePVName
-app pv_table?pv=a,pv=b
-app pv_table?file=/some/file.pvs
```
=> AppResourceDescriptor.create(that_resource)

shroffk:
This might be becoming a bit too big
@kasemir should this be merged when the current build succeeds?

shroffk:
@kasemir can we have a hangout?

Issue #22, opened 2017-09-06
Managing the migration from cs-studio to phoebus


shroffk:
Create a list of requirements for the transition from cs-studio to phoebus.

Technical Issues

 - [ ] the reusing of the same application configuration files
 - [ ] transition from BOY to Display Builder (more below)
 - [ ] handling of saved layouts in Phoebus (more below, #115)

User Experience/Branding issues

@willrogers @kasemir @claudio-rosati

willrogers:
I was previously considering a partial transiton from BOY to Display Builder within CS-Studio.

If Display Builder is not going to be supported in CS-Studio after a period, this is a risky approach.  BOY won't be supported in Phoebus.  So moving from CS-Studio+BOY to Phoebus+Display Builder would have to be done in one big transition.

willrogers:
Despite all of Eclipse's limitations, we have now got a nice way of saving and sharing layouts using perspectives at DLS.  I'd like to be able to do a similar thing in Phoebus.

kasemir:
I believe you that an EDM to XXX transition is no fun. Are you now 100% on BOY displays in the DLS control room?

Of the 20+ SNS beam lines, 13 have been converted to EPICS. Of the 13, 4 use the display builder. All can be viewed with the display builder, but some details of local operation tied to scripts still require BOY. The remaining beam lines to be converted to EPICS will use the display builder, and those that are still on BOY will also have their scripts transitioned. The BOY -> Display Builder conversion is much simpler than going EDM -> XXX.

For beamlines, I need display builder, data browser, PV Table, scan client UI, alarm UI, logbook & email submission as essential applications. Apart from scan and alarm UI, these are all generic applications that most everybody would use. Probe, PV Tree, job progress UI, Web Links, logbook viewer, launcher for external application would be nice to have.

If Kunal and I continue at the current pace, I think we can have the essential framework (docking panels w/ and w/o 'input', jobs, preferences, persistence) in place by end of this year, and about one year from now the essential applications for SNS beam lines can be ported, where I will likely concentrate on those essential for SNS while Kunal would concentrate on what's essential to BNL and so on. Once phoebus is sufficient for SNS beam lines, I would naturally devote less time to maintaining the non-phoebus code. DLS and others might prefer to stay with the existing CS-Studio source code for longer, but I honestly don't see how you'll be able to do much beyond bugfixes. For BOY, the core tool, the way it's tied to GEF and thus SWT and the UI thread is a dead end.

[...]

PR #42, opened 2017-09-19
Splash screen; read memento file and start apps off UI thread

kasemir:
Adds a splash screen that's displayed as soon as possible.
Then reads the persisted state from memento off the UI thread,
and returns to the UI thread to open the tabs etc.

(Branding, #22)


claudio-rosati:
Are you using the [Preloader](https://docs.oracle.com/javase/8/javafx/api/javafx/application/Preloader.html) class?
These could be useful: [link1](https://stackoverflow.com/questions/35849930/how-to-handle-java-web-start-jnlp-downloading-progress-in-a-preloader), and [link2](https://stackoverflow.com/questions/31882784/how-to-create-splash-screen-as-a-preloader-in-javafx-standalone-application).

kasemir:
I thought about it, and if you check the javadoc you'll see I also considered the even faster java.awt.SplashScreen and associated JRE command line arguments.
But I saw no advantage in using those, because the splash shows just fine and in a timely manner by calling it in Application.start().
Besides just showing an image this is about restructuring the code to include a startup job that starts the applications and reads the file with persisted state off the UI thread, then returns to the UI thread for restoring previously visible tabs.

claudio-rosati:
So, peobably, the startup job should be a worker, being able to notify the (splash screen) UI thread of the read and startup progress.

shroffk:
I like the idea of the splash listing the app's as they are being loaded. Since the App start is not lazy this could be a cool thing to have...plus if the startup in the future becomes slow, the update info can be used to pin point the problem.

shroffk:
loading logbooks...
loading display builder...
restoring /tmp/screen.bob
restoring /tmp/xyx.plt

kasemir:
Yes, that's on my mind, but it needs to be considered with the case of a 'server' instance later opening additional files as requested from another command line. In that case, we won't open another splash screen, but simply use the JobManager and updates then appear in the Job List view (if you care to open it).

The first step is this:
1) Have a splash screen at all
2) Perform application startup (that global AppDescriptor.start(), loading the memento file, ..) in the background, with a JobManager that has a monitor to which the splash could listen.

So I suggest to merge this an move forward.

kasemir:
To track the `AppDescriptor.start()` calls, we could change that to `AppDescriptor.start(JobMonitor)`.
Apps thus have a chance to update their progress.

shroffk:
@claudio-rosati @kasemir I am merging this PR for now

It seems like there are some open issues here should we put them in a new issue ticket? Can we have a hangout next week to define the requirements for application startup and accordingly decide the technologies we would need.

kasemir:
Issue: The 'server' mode is broken since we now pass around "app?file=/path/to/resource" instead of just "/path/to/resource". I'm about to fix that.

Then I'll add the progress update in the splash screen.

claudio-rosati:
@shroffk @kasemir 
No problem for the hangout.
I think we should also talk about documentation. For the ICALEPCS paper I walked (again) through some other frameworks and all of them have documented:

- how to create a simple, hello-world-like application, mostly used to explain the build system, the module system, how to add libraries, etc.
- how to create a more complex application, exploiting all the aspects of the framework (windows management, actions, menu, toolbars, status bar, selection/context passing, etc);
- API (javadoc);
- (most important) the architecture of the framework!

I think the last item is the most important one, and it had to be provided before any development. We are very badly placed in this.

Issue #51, opened 2017-09-23
'Help' menu


kasemir:
If phoebus-doc is checked out in parallel to phoebus source code, ``Product`` build should include the doc in html form as well as the javadoc of the source code.

From 'help' menu, these are then accessible via a web browser.

Could either display the content in a `WebView`, i.e. inside phoebus, or use `javafx.application.Application.getHostServices().showDocument("path/to/phoebus/help/index.html")` to display in the platform web browser (Safari, Firefox, ...).

External web browser has the advantage that it's a full web browser. Users can access any other web pages, download files, play video, .. without using JVM CPU & memory. But especially on windows we could just get a "cannot open web browser" error message.

Embedded web browser will always open up, but the web browser will use JVM CPU & memory, and its features a a little limited.

berryma4:
Although help pages are usually simple, it's worth noting that the Java webkit hasn't been kept up to date in Oracle.  I'm not sure if this will improve in openfx.  It's worth keeping an I on.

kasemir:
Plan:

 - [ ] Place help for applications into the source code for the application, like phoebus/applications/the_app/doc, with a top-level index.rst. So the material for the manual section of a specific application feature is kept with the source code for that application. 
 - [ ] Have phoebus-doc fetch all phoebus/applications/**/doc/index.rst files, auto-creating a top-level applications page that links to all of them. This gives us an overall 'manual' that includes sections for each application feature. 
 - [ ] When building phoebus product, if the phoebus-doc/build/html is available, copy that into product/doc. 'Help' menu opens web browser (likely the platform web browser) with product/doc/html. This way, the manual is also available inside the product, no network access required.

kasemir:
The phoebus 'docs' branch has doc/index.rst files for PV Table, Tree, PV List, Log Config, ..

The phoebus-doc/source/conf.py checks for a ../../phoebus folder with sources. If not found, it calls git to clone the phoebus sources.
Thus having the sources one way or the other, it then includes the application docs in the manual.

That "works" even with readthedocs, except readthedocs would sometimes fail in its initial `pip install` phase (http://readthedocs.org/projects/phoebus-doc/builds/6029866), or at other times spend 15+ minutes in the initial `cloning` phase (http://readthedocs.org/projects/phoebus-doc/builds/6029879/), which look like general network problems unrelated to this.

kasemir:
[...]

PR #54, opened 2017-09-27
'Help'

kasemir:
Help/Help now opens the manual.
Also adds a Help/About dialog.
( @shroffk , those are basically MenuEntrys, but right now a MenuEntry can only show up under Applications, can't have a getmenuPath() return "/Help/About" or "/File/Open")

Implements online help, #51

If you have both phoebus-doc and phoebus cloned into parallel directories,
building the manual in phoebus-doc will include the application help (PV Tree, ...) from the phoebus sources.
When starting phoebus from the IDE or building it, that will read resp. bundle the manual from phoebus-doc/build/html.

The http://phoebus-doc.readthedocs.io setup clones the phoebus source code to include its manual pages.

When you only have the phoebus sources on your computer and then try to invoke 'Help', it will not find the local manual pages and go out to http://phoebus-doc.readthedocs.io .

For more, see http://phoebus-doc.readthedocs.io/en/latest/help_system.html

Related:
One key issue here is that the "Help" menu needs to find the manual.
Similar to the application startup that needs to check for a memento.
==> Created org.phoebus.framework.workbench.Locations to determine a phoebus.install and phoebus.user location, similar to the RCP/OSGi install and instance locations, see http://phoebus-doc.readthedocs.io/en/latest/locations.html


kasemir:
Maven cannot find the dependencies.

Meanwhile, the result of "ant dist" which I just got to run on Windows as well as Linux is on http://ics-web.sns.ornl.gov/kasemir/phoebus-0.0.1.zip

shroffk:
Well it seems source forge is down...if the server hosting the dependency zip were down the same problem would result. I don't think this is a maven issue...in the future (when the repos are restructured) we can use the -Dmaven.local.repo so have the same local setup as ant. 

In the meanwhile I will merge this PR and ping Ralph Lange and the epics v4 team

PR #56, opened 2017-09-27
Email support

shroffk:
Not complete but ready to merge

### DIFF
--- a/applications/.project
+++ b/applications/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>applications</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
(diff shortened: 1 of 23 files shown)


kasemir:
I can open Display, Utility, Email App.

What/how do I need to configure?
SMTP host?

kasemir:
OK, I found the email.properties, and managed to send an email.

Even with HTML content, but the HTML editor doesn't fully work on Linux.
I can only enter one line. Pressing "Enter" does not go to the next line.
Also tried Ctrl-Enter etc.: Has no effect.
Does _not_ trigger the "Send" button (which is good), but also doesn't insert a new line in the email body.

I've merged it into the 'help' pull request, no conflicts there.

kasemir:
Also, changing between text/plain and text/html toggles between totally different texts.
I can enter "Hi\nthere" as plain text, and "Hello, Dolly!" (no newline possible) as HTML.
I assume people would expect that you can "upgrade" from plain text to HTML without loosing the entered text.

kasemir:
I can paste text into HTML that contains newlines.
I can then also 'delete' some of the newlines, but not add more newlines via 'Enter' key.

shroffk:
> "upgrade"

Ya ggreed

PR #58, opened 2017-09-27
Pre properties 4

kasemir:
#4 : 

Load a preferences.properties file for defaults.
That file lists and also explains the properties.

Launcher -settings can still load an *.xml file for exported prefs, but more useful it can now load a settings.ini file of the format "package/setting=value".


kasemir:
Travis was happy twice, now it's back to "Connect to epics.sourceforge.net:80 failed: Connection timed out" during the initial "dependencies" build..

kasemir:
Merge?

PR #63, opened 2017-09-29
Ui monitoring

kasemir:
Adds `ResponsivenessMonitor`.

Similar to the RCP idea, this monitors the UI thread.
When it detects a freezeup, it logs a stack trace.

Compared to RCP implementation, it's simpler:
Fixed check rate of 500ms,
always logs just one stack trace when the freeze is detected,
then another message once UI thread resumes.

Added Menu entry Debug/Freeze UI to test/demo.

The log does add a header to the JFX thread, since that's the key one which blocks.
In this example, it points to FreezeUI invoked from the test/demo menu entry.

```
2017-09-29 16:16:35 SEVERE [org.phoebus.ui.application.PhoebusApplication] UI Freezeup

"PVMgr Worker 7" daemon prio=5 Id=42 WAITING on java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@4bdd5bba
	at java.base@9/jdk.internal.misc.Unsafe.park(Native Method)
	-  waiting on java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@4bdd5bba
	at java.base@9/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
	at java.base@9/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2062)
	at java.base@9/java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1119)
	at java.base@9/java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:848)
	at java.base@9/java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1092)
	at java.base@9/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
	at java.base@9/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	...

...

*********************************
*** JavaFX Application Thread ***
*********************************
"JavaFX Application Thread" prio=5 Id=19 TIMED_WAITING
	at java.base@9/java.lang.Thread.sleep(Native Method)
	at app//org.phoebus.ui.monitoring.FreezeUI.call(FreezeUI.java:33)
	at app//org.phoebus.ui.monitoring.FreezeUI.call(FreezeUI.java:1)
	at app//org.phoebus.ui.application.PhoebusApplication.lambda$10(PhoebusApplication.java:317)
	at app//org.phoebus.ui.application.PhoebusApplication$$Lambda$338/513002640.handle(Unknown Source)
	at platform/javafx.base@9/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at platform/javafx.base@9/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at platform/javafx.base@9...

...


2017-09-29 16:16:37 SEVERE [org.phoebus.ui.application.PhoebusApplication] UI Updates resume


```

### DIFF
--- a/applications/pvtable/src/main/java/org/phoebus/applications/pvtable/PVTableApplication.java
+++ b/applications/pvtable/src/main/java/org/phoebus/applications/pvtable/PVTableApplication.java
@@ -10,6 +10,7 @@ package org.phoebus.applications.pvtable;
 import static org.phoebus.framework.util.ResourceParser.createAppURI;
 import static org.phoebus.framework.util.ResourceParser.parseQueryArgs;
 
+import java.net.URL;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
@@ -18,6 +19,8 @@ import org.phoebus.applications.pvtable.persistence.PVTableAutosavePersistence;
 import org.phoebus.applications.pvtable.persistence.PVTableXMLPersistence;
 import org.phoebus.framework.spi.AppResourceDescriptor;
 import org.phoebus.framework.util.ResourceParser;
+import org.phoebus.ui.docking.DockItemWithInput;
+import org.phoebus.ui.docking.DockStage;
 
 import javafx.scene.image.Image;
 import javafx.stage.FileChooser.ExtensionFilter;
@@ -88,8 +91,19 @@ public class PVTableApplication implements AppResourceDescriptor
         {
             for (String file : files)
             {
-                instance = create();
-                instance.loadResource(file);
+                final URL input = ResourceParser.createResourceURL(file);
+                // Check for existing instance with that input
+                final DockItemWithInput existing = DockStage.getDockItemWithInput(NAME, input);
+                if (existing != null)
+                {   // Found one, raise it
+                    instance = existing.getApplication();
+                    instance.raise();
+                }
+                else
+                {   // Nothing found, create new one
+                    instance = create();
+                    instance.loadResource(input);
+                }
             }
         }
         else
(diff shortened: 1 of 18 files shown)


kasemir:
@shroffk Can we merge this?

shroffk:
ok, unfortunately I have not gotten a chance to test this but as agreed that with a successful build we should move ahead.

kasemir:
Two things have accumulated on this branch.
One is the UI monitoring, which triggers a log message when the UI thread becomes unresponsive.
I see that occasionally when JavaFX applies the (built-in modena) CSS style sheet to a new scene. Don't think we can do anything about that.


The other is in the PV Table: Opening PV Table x.pvs does just that. When then trying to once more open x.pvs, the PV Table recognizes that it already has a tab open with the same input, and it selects that existing tab instead of opening another one.
Should be possible to handle that in the framework, so the PV Table would never know that somebody tried to open another "pv_table" app for the same input.

shroffk:
> Should be possible to handle that in the framework, so the PV Table would never know that somebody tried to open another "pv_table" app for the same input.

So, to clarify, when the pv_table is opened with the same input another instance should be created or should the same input result in the already open application instance being brought into focus.

I think we should do the former, while cs-studio was smart enough to do this I feel it results in too much complexity. We would have to check input equality, then check that the existing instance is not dirty, ... why not treat it like the browser, when you launch chrome from command line with the same URL it reuses the existing instance of chome and open the same URL in new new tabs.

PR #72, opened 2017-10-20
Add icon to context menu entries

kasemir:
`ContextMenuEntry` had `Object getIcon()`, unclear what that object would be.
Changed that into
```
    /** @return Stream for icon, or <code>null</code> */
    public default InputStream getIcon()
    {
        return null;
    }
```
and adding that icon - if one is provided - to the context menu entry.

For now PV Tree and PV Table hand out icons.

### DIFF
--- a/applications/email/src/main/java/org/phoebus/applications/email/actions/ContextCreateEmail.java
+++ b/applications/email/src/main/java/org/phoebus/applications/email/actions/ContextCreateEmail.java
@@ -5,7 +5,6 @@ import java.util.List;
 
 import org.phoebus.applications.email.EmailApp;
 import org.phoebus.framework.selection.Selection;
-import org.phoebus.framework.spi.AppDescriptor;
 import org.phoebus.framework.spi.ContextMenuEntry;
 import org.phoebus.framework.workbench.ApplicationService;
 
@@ -27,11 +26,6 @@ public class ContextCreateEmail implements ContextMenuEntry {
     @Override
     public Object callWithSelection(Selection selection) {
         ApplicationService.findApplication(EmailApp.NAME).create();
-        return null;
-    }
-
-    @Override
-    public Object getIcon() {
         return null;
     }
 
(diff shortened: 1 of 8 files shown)


claudio-rosati:
Using a stream means that each time an icon has to be realized it is read from the stream. Probably returning an Image (JavaFX) it is better, allowing caching strategies

kasemir:
`ContextMenuEntry` is in `core-framework`, i.e. a non-UI package and should consequently not use any UI-specific API like JavaFX Image or AWT Image or Swing Image or SWT Image or ...

As for caching the image, if necessary, this could be done in other ways.  `ContextMenuService` or `ContextMenuHelper` could cache the underlying stream into a `ByteArray` or UI-specific image.

kasemir:
@shroffk : Alternatively, should I move `ContextMenuEntry` from `core-framework` to `core-ui`?

shroffk:
I think the `ContextMenuEntry` can be moved to core-ui.

kasemir:
OK, I'll update as follows and then create a new pull request:

 * ContextMenuEntry moves to core-ui
 * ContextMenuEntry.getIcon() can then return an Image
 * Implementor may cache the image, i.e. always return one and the same icon Image
 * ContextMenuHelper will, as now, always ask the ContextMenuEntry for its image, so in case the implementation wants to change the image at runtime for some reason, that'll be possible.

claudio-rosati:
@kasemir @shroffk 

As suggested here ([JavaFX Tip 27: HiRes / Retina Icons](http://dlsc.com/2017/08/29/javafx-tip-27-hires-retina-icons/)): never use input streams to load icons, only URL or CSS, and if you provide the right set of icons... et voilà les jeux sont faits!

So, returning a URL could be the solution because it doesn't require UI knowledge. Caching strategies could be realised using an utility class in core-ui to load the image from the URL.

kasemir:
Note updated pull request is #73 .
The previous discussion ended in `ContextMenuEntry.getIcon()` returning a JavaFX `Image` (or `null` for default).
I still consider that a good decision because it doesn't limit the implementation to loading a `*.png`. The implementation could also draw an in-memory icon, for example generate it based on some internal state.

When loading an icon from a *.png, the implementation could indeed use  
```
new Image(MyClass.class.getResource("icon-name.png").toExternalForm());
```
instead of
```
new Image(MyClass.class.getResourceAsStream("icon-name.png"));
```
I think you mentioned implementing a helper class for that.

claudio-rosati:
OK, understood.

PR #73, opened 2017-10-23
Context menu icons

kasemir:
Adding icon to context menu:

 * ContextMenuEntry moved from core framework.spi to core ui.spi
 * Same with ContextMenuService
 * ContextMenuEntry.getIcon() now returns JavaFX Image (or null for default)
 * Update dependencies in Eclipse, ant, pom files, and renamed all the service files for the new location of ContextMenuEntry
 * PVTable and PVTree provide one and the same, cached, icon for each call

.. but implementations are free to return a varying icon, for example based on internal state.

In principle, core framework.spi.MenuEntry and ToolbarEntry are also UI elements, so they should move to core ui.spi and support icons.
@shroffk  if you agree, I'll do that after this one is merged.


kasemir:
@claudio-rosati  Updated the icon helper class to use URL, see ee3571c

kasemir:
@shroffk Can we merge this now?

kasemir:
Thanks!

PR #75, opened 2017-10-25
NLS: Move to framework, add test

kasemir:
This moves the "NLS" class from core-ui to core-framework.

Starting the display builder import, I noticed that the 'model' doesn't have a UI, but it still has localized messages for widget names etc.
So placing the NLS class in the UI module was wrong, just like in the OSGi case NLS can also be used for "System.out.println" and is thus a non-UI idea
--> With that, the display builder "model" can depend on only core-framework, not core-ui.

While at it, I added a unit test and found out that *.property files use ISO 8859-1/Latin1 character encoding, not UTF-8. It's documented in Properties#load(). Eclipse also enforces it: Even if you generally select UTF-8 for all source code, the *.property files are edited with ISO 8859-1/Latin1, so it's best to enter for example Umlaut characters as "\u1234" unicode escapes.
Anyway, now there's a test that demonstrates proper loading of default (US) messages.properties vs. messages_de.properties.

### DIFF
--- a/applications/pvtable/src/main/java/org/phoebus/applications/pvtable/ui/Messages.java
+++ b/applications/pvtable/src/main/java/org/phoebus/applications/pvtable/ui/Messages.java
@@ -7,7 +7,7 @@
  ******************************************************************************/
 package org.phoebus.applications.pvtable.ui;
 
-import org.phoebus.ui.nls.NLS;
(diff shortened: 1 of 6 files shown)


kasemir:
@shroffk OK to merge?

PR #85, opened 2017-11-14
Resource uris 79

### LINKED ISSUE #79: URI, URL, command line, input inconsistencies
The DockItemWithInput uses a URL as its input, for example
 * `file:/path/to/pv_table.pvs`
 * `http://server/path/to/display.bob?Macro1=Hello+Dolly&M2=abc`

There are valid URLs, they're stored in the memento with proper XML encoding.

The rest of the code often uses a URI (no 'schema:'), for example for opening tools from the command line:
 * `-app probe?pv=Fred`
 * `-app pv_table?file=file:/path/to/pv_table.pvs`

The `AppResourceDescriptor.create(String resource)` doesn't exactly say what type of resource text it expects. Could be a channel finder query, but most existing code now calls `ResourceParser.parseQueryArgs` in there and thus expects a command line type URI. Can't pass just "pv=Fred", must pass "probe?pv=Fred". So `PhoebusApplication.openResource` constructs that as `application.getName() + "?" + ResourceParser.FILE_ARG + "=" + resource`.

With the Display Builder runtime #78 there's now the issue that each instance is uniquely identified by the display file (local file or http://site/some_display.bob) combined with macros.
On the command line that would be
 * `-app display_runtime?file=http://server/path/to/display.bob&Macro1=Hello+Dolly&M2=abc`
In the DockItemWithInput, it turns into the input
 * `http://server/path/to/display.bob?Macro1=Hello+Dolly&M2=abc`

Note how one has "&Macro1" while the other has "?Macro1", so converting between the two formats, as right now would be required to re-open that display with same macros from the memento, isn't trivial.

Simply using
```
application.getName() + "?" + ResourceParser.FILE_ARG + "=" + resource
```
to turn a resource path into a URI is naive, since the resource could itself be "https://site/rest_api/displays/get?device=magnet&instance=2", so need to use
```
application.getName() + "?" + ResourceParser.FILE_ARG + "=" + URLEncoder.encode(resource, "UTF-8")
```

==> Turning into a mess, need a few better ideas

kasemir:
This fixes #79 :

The 'input' to a DockItemWithInput is now a URI.
It's the same URI that can be passed via the "-resource /some/file.opi?MACRO=value" command line.
The "app=.." is treated as a special query parameter in case the resource needs to be opened with a specific application: "-resource /some/file.opi?app=some_app&MACRO=value".

The memento saves and restores that URI without changes, and the save/restore of display files with macros now works.

Documentation needs to be updated, and as I work further on the display builder there may be more tweaks, but this snapshot seems to be at least as good as what we had before.

### DIFF
--- a/applications/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayInfo.java
+++ b/applications/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayInfo.java
@@ -10,15 +10,15 @@ package org.csstudio.display.builder.runtime.app;
 import static org.csstudio.display.builder.runtime.WidgetRuntime.logger;
 
 import java.io.UnsupportedEncodingException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLDecoder;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URLEncoder;
 import java.util.Objects;
 import java.util.logging.Level;
 
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.macros.Macros;
+import org.phoebus.framework.util.ResourceParser;
 
 /** Information about a display
  *  @author Kay Kasemir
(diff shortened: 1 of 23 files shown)


kasemir:
Updated the '-help' output to show some of the now supported examples.

PR #87, opened 2017-11-17
Fixed test failure because of missing default locale.

claudio-rosati:


kasemir:
'Files changed' shows no difference, since that change is already on 'master'.

Issue #47, opened 2017-09-21
Split Job API into basics and UI


kasemir:
Right now org.phoebus.ui.jobs contains the basics of submitting a monitored, cancellable job (JobManager, JobMonitor, JobRunnable, ..) as well as the UI that displays running jobs and allows cancelling them (JobViewer*).

Is that fine, because the main use for Jobs is loading saved state, loading PV Table file, reading archived data, ... which all happen in UI tools?
Or should that be split, because in principle you can submit a Job from a non-UI tool, and there could be an ASCII interface where you press 'l' to list jobs, 'c1' to cancel Job # 1 etc.

PR #91, opened 2017-11-27
Move 'Jobs' API from core.ui to core.framework

kasemir:
Handles #47


kasemir:
Obsolete, is included in #92

PR #99, opened 2017-12-14
Completion

kasemir:
Adds PV name completion to display builder, probe, PV table, PV Tree.

org.phoebus.framework.autocomplete has the basic ProposalProvider API.
There's a generic implementation for History, and PV-name implementations for Local PV and Sim PV.
The ProposalService (and the PV-specific PVProposalService) query a list of providers, returning their data as it arrives.
The PVProposalService allows adding site-specific PV name completion via SPI.

Finally, org.phoebus.ui.autocomplete has the UI components which attach to a TextField.

The SNS-specific product, https://github.com/kasemir/phoebus-sns , adds an SNS-specific PV name lookup which is currently bogus, using a fixed list, until I figure out how to best add the Oracle JDBC library.

There's always room to split this into more sub-packages, but the basics are there:
PV name completion for all tools with example of site-specific addition.


kasemir:
Screenshot of the SNS product which adds PV name lookup from local database:

completion

shroffk:
wow!!!

PR #103, opened 2017-12-18
Add RDBConnectionPool

kasemir:
Common API for connecting to any RDB (Oracle, MySQL, Postgres) in a way
that existing connection can be re-used.

Used by SNS product's implementation for PV name lookup from local RDB.
To be used by RDB archive data reader.


kasemir:
Test is specific to site since it needs to connect to actual RDBs.
So for now there's just a test in the SNS product, https://github.com/kasemir/phoebus-sns/blob/master/product-sns/src/test/java/org/phoebus/framework/rdb/RDBConnectionTest.java

Issue #110, opened 2018-01-26
"Transparent" Property for Group Widget


pavel-ch:
For my tests with Display Builder on Phoebus, I need this "Transparent" Property for Group Widget.
As I have understood, you are going to be compatible with BOY, where this property is available.
I tried to add it.
Hope this can be usefull for you, too.
And generally, I hope that some small contributions from "public" are not unwanted.
Pull request follows.

kasemir:
Thanks, has also been imported in the Eclipse-based Display Builder.

PR #111, opened 2018-01-26
Group Widget: Transparent Property

pavel-ch:
As mentioned in #110


kasemir:
Thank you, this is perfect, even keeping with the general source code formatting of existing files!

PR #129, opened 2018-02-28
Display Builder: DisplayInfo.toURI() tweak for Windows platform

pavel-ch:


pavel-ch:
Abandoned due to test failure, lookig further

PR #131, opened 2018-02-28
Display Builder: DisplayInfo.toURI() tweak for Windows platform 

pavel-ch:
... improved a bit, passes test and works for me on Windows and Linux


kasemir:
Thanks!
I think this is fine. Will check with my displays next week, am currently out of the office.

kasemir:
Could you add something to `DisplayInfoTest` for the "C:\..." type path that caused problems on windows?

PR #134, opened 2018-03-15
File browser

shroffk:
While trying to do a demo/training I was stuck with the issue of how the file diloag looks different for everyone and it brought back memories of SWT's native dialog.

While we should keep keep Phoebus simple...a thin simple file browser in phoebus would be good to have.
image

It defaults to home dir (we can decide on another location)
Shows a simple file/dir tree with open and openWith operations only.
User can type or select a new dir root.

The tree is dynamically populated.

TODO:
momento to remember the last root.


shroffk:
I also pulled out the openFile and openResource from the PhoebusApplication as a separate service. 

TODO:
refactor the PhoebusApplication to use the ApplicationLauncherService

PR #140, opened 2018-03-30
Issue 136

shroffk:


kasemir:
The scan server pull request #139 was earlier, but merging this first so I can then update the scan server pull request in case of conflicts.

PR #146, opened 2018-04-26
Alarm snapshot

kasemir:
Adds kafka to dependencies.
Uses the jackson dependency that was already there from channel finder work.

For now just a standalone demo of alarm tree with demo code to create initial config and send bogus state updates.

alarm1


kasemir:
Alarm Server skeleton has proof of principle, connects to PVs, and directly sends the alarm state to client, so tree UI updates.
Needs actual alarm logic (latch, acknowledge, ..).

PR #156, opened 2018-05-09
Master project explicitly named to "phoebus".

claudio-rosati:
In NetBeans (and possibly in other IDEs), when the master pom is opened, the corresponding project will be named "parent", and this is very confusing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,6 +5,7 @@
   <artifactId>parent</artifactId>
   <version>0.0.1-SNAPSHOT</version>
   <packaging>pom</packaging>
+  <name>phoebus</name>
 
   <properties>
     <maven.repo.local>${project.build.directory}/.m2</maven.repo.local>


kasemir:
I agree that 'phoebus' looks better than 'parent' as the top-level project name.
At the same time, it's good to have the underlying name of the pom match the display name, i.e. just stay with what's in the
```
<artifactId>parent</artifactId>
```
instead of overriding that in this one case with
```
<artifactId>parent</artifactId>
<name>phoebus</name>
```

==> Would it be better to just rename the `artifactId` into "phoebus"?

claudio-rosati:
I think yes, probably it is better, unless you want to keep the semantic meaning of being the parent project, made by many sub-projects.
In any case, Maven always wants the name being explicitly given in order to publish it on nexus (also because name can contain spaces, thing not allowed in the id).

kasemir:
Well, Kunal created the parent pom, so I'll leave it to him to decide if the `<name>` should be Phoebus and the ID stays, or if the name is also updated. I'm just saying that I agree with "Phoebus" being a better top-level name than "parent".

claudio-rosati:
@shroffk 
Can you take the decision?

shroffk:
Like mentioned above...the reason behind the module being named "parent" is cause it is meant to be the parent of all other modules.

I agree we can set the <name> as phoebus

PR #157, opened 2018-05-09
Drag and drop feedback image

claudio-rosati:
It seems thereis a bug pixelReader.getArgb(...) when on Mac when screen is scaled to have more resolution (no retina scaling is used).

So I've decided to draw only the widget outline (this time correctly aligned to the cursor), without the type icon. All is done using pure JavaFX code, no more Swing/AWT one.


claudio-rosati:
@kasemir 
I'll port this back on the CS-Studio version of DB.

Issue #150, opened 2018-05-07
PhoebusApplication cleanup


kasemir:
* Remove code that's duplicated in ApplicationLauncherService
 * Sort entries in "Applications" menu
 * Cleanup toolbar (remove??)

kasemir:
Toolbar can be hidden; Applications menu is sorted

PR #163, opened 2018-05-16
Fixes

kasemir:
Fix thread error in ToolTip;
Sort "Applications" entries;
Keep autocomplete menu from clobbering entered text with first item


kasemir:
This sorted the menu entries, part of #150.

Fixes #164

PR #166, opened 2018-05-20
Split

kasemir:
Adds "Split Horizontally" and ".. Vertically" to the tab's context menu.
Saves & restores resulting arrangement in memento.

**Note:** This does change the memento content by adding `<pane>` resp. `<split>` elements, so existing memento won't load.

split

Still to add: Logic that removes a split when closing all its content.


shroffk:
yay!!
This is great.

PR #180, opened 2018-06-07
Alarm move

kasemir:
Adds 'move' to alarm tree context menu


kasemir:
No idea to trigger another Travis build, so added commit for context menu entry that copies PV name to clipboard, which is useful in general

PR #186, opened 2018-06-15
Fixes

kasemir:
* Option to hide/show toolbar
 * Sort "PV" context menu alphabetically.. which puts "Copy PV name on top". Would be good to additionally somehow keep it on top..
 * Imports from other display builder


kasemir:
* Option to show/hide the splash screen

PR #191, opened 2018-06-19
Update version of javax.mail dependency.

shroffk:
This adds email attachment support.

@kasemir can you test if this works in the IDE and your setup?

### DIFF
--- a/applications/email/pom.xml
+++ b/applications/email/pom.xml
@@ -21,7 +21,7 @@
     <dependency>
       <groupId>com.sun.mail</groupId>
       <artifactId>javax.mail</artifactId>
-      <version>1.6.0</version>
+      <version>1.6.1</version>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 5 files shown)


kasemir:
Updated the .classpath files for dependencies and product.

Builds and runs for me with maven, ant, Eclipse!

When switching between text/plain and text/html, the message body is different.
It's not lost: Can switch back to the 'other' mode, copy the text, switch mode, paste the text.
Still, I guess users would prefer if it kept the text through mode changes.
I even think it would be OK to loose formatting when going html -> plain -> html.

What's a real problem on Linux:
In text/html mode, I cannot enter a new line. The "Enter" key doesn't do anything within the body of the email dialog. Tried Ctrl-Return, Alt-Return, Shift-Return, all are simply ignored.

But I think these are all issues seen before. The update certainly fixes the Eclipse build issues for me.

shroffk:
> It's not lost: Can switch back to the 'other' mode, copy the text, switch mode, paste the text.
> Still, I guess users would prefer if it kept the text through mode changes.

This would mean us handling the parsing from text to html, which is non trivial...maybe once we have a general markup support solution this would be address via that.

Issue #200, opened 2018-06-26
channelfinder based autocomplete


shroffk:
Add autocomplete for pv names using the channelfinder directory service.
Ensure proper thread management to prevent long queries from blocking the UI.

shroffk:
TODO: 
 - [ ] reconfigure the properties to match the default preference settings.
 - [ ] when the client connection fails the provider should deactivate.

Issue #136, opened 2018-03-20
ChannelFinder


shroffk:
Move the channelfinder applications from cs-studio to Phoebus.

- [x] Move the channelfinder rest client library and all its dependencies to phoebus  
- [x] Move the channel view basic table viewer application
    - [x] Convert the jface/swt table to javaFX tableView
    - [x] Create SearchJob using the phoebus job framework
- [x] Move the channel tree by property view application
- [x] Move the channel line viewer
- [x] Move the channel waterfall plot
- [x] add support for adding and removing tags
- [x] add support for adding and removing properties
- [x] add support for creating new channels
- [x] replace properties usage with the java preferences

PR #201, opened 2018-06-26
Issue 136

shroffk:
Issue #136 and Issue #200

### DIFF
--- a/applications/channel/utility/.classpath
+++ b/applications/channel/utility/.classpath
@@ -1,5 +1,32 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/"/>
+	<classpathentry kind="src" output="target/classes" path="src/main/java">
+		<attributes>
+			<attribute name="optional" value="true"/>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources">
+		<attributes>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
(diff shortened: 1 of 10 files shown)


shroffk:
@kasemir 
In the current channelfinder autocomplete, I directly query the service.
I tried to do this via an existing Job however the result of the query and the autocomplete table started getting out of order.

kasemir:
That's OK. The completion implementation is already called on a background thread.
It should then do what it needs to do, and return the information when done.
See https://github.com/kasemir/phoebus-sns/blob/master/product-sns/src/main/java/org/phoebus/sns/completion/SNSPVProposals.java, it also connects to RDB etc.

kasemir:
`channel/utility/.classpath` listed `src/test/resources`, which doesn't exist.
19c2c6120fba1d6467d446836cf179ef00e0c3a2 removed that in the ongoing alarm_actions branch.

The channel finder preference file, `channelfinder/src/main/resources/config/channelfinder.properties`, doesn't match the expected name for preference files (`*_preferences.properties`), so it's not included in the online help section that describes available settings (http://phoebus-doc.readthedocs.io/en/latest/preference_properties.html).
At this time, `core/mail` is a good example for handling preferences in the suggested way: Resource `*_preferences.properties` that's copied into the online help, with comment to indicate the package name,  and a `*Preferences` class that loads defaults via the `PreferencesReader`.

Since the `CFProposalProvider` is included in the default project, a default configuration for https://localhost:8181/ChannelFinder is fine, but would be good to have a way to disable it, instead of continuing to run into connection errors for a missing channel finder setup.
For example, setting `channelfinder.serviceURL` to nothing could be a way to disable channel finder lookups, but that results in NPE during application startup (which until 70c925c70c700ea4758d6e91cdafb703d1706a15 prevented phoebus from starting up at all).

shroffk:
#200 has been updated

PR #202, opened 2018-06-26
Update .travis.yml

shroffk:
Creating a matrix build for jdk9 and jdk10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 sudo: required
 language: java
 dist: trusty
-jdk: oraclejdk9
+jdk:
+  - oraclejdk9
+  - openjdk10
+  - oraclejdk10
+
 
 script:
   - mvn clean install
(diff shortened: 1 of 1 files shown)


kasemir:
Works for Oracle 9 & 10, but not openjdk 10.

I assume that's because for openjdk, you need to separately add JFX.
.. which is fundamentally possible for openjdk11, https://github.com/shroffk/phoebus/issues/143, but that was a manual download. No idea how to do that for travis and openjdk10.

shroffk:
> I assume that's because for openjdk, you need to separately add JFX.

yes... that is exactly it. I am not sure if it is worth updating the matric to install openjfx.
Maybe we could remove openjdk 10 and wait for openjdk11

kasemir:
I'm OK with for now only using oraclejdk10.
Drop oraclejdk9, and skip openjdk10.

PR #205, opened 2018-06-27
Alarm actions

kasemir:
Adds 'cmd:...' to automated actions.
Moved PVTree and Probe to "Display" menu, both with icon.
Fixes.


kasemir:
Increased the tolerance of time-sensitive automated action test twice because saw it fail in local builds. No failure observed with travis, just triggered it once more via close/re-open of pull request.

Issue #204, opened 2018-06-27
Cleanup Logbook Interfaces


shroffk:
@es-sns121 
Cleanup the logbook interfaces. 
- [x] Remove the extra methods, like `getlogbooknames()`, which were added for convenience before jdk8 streams
- [x] refactor org.phoebus.logging to org.phoebus.logbook, rename the core-logging module into core-logbook
- [x] some of the core interfaces have static factory method and some have builders...a common design pattern is needed.

shroffk:
@es-sns121 @kasemir 
I have simplified the core interfaces of org.phoebus.logging
 - removed necessary methods no longer needed in java8
 - removed builder for all default implementations, except logEntry, and replaced them with static factory methods

kasemir:
Pull request?

Issue #64, opened 2017-10-04
In memory logbook with markup support


shroffk:
Missing requirements when creating entries:

- Markup support (Bold, Italics, Color)
- Inline attachments
- Auto correct
- Clickable links in clients
- Tables and Lists

Very nice to have
- cropping
- rotate
- annotations

shroffk:
Markdown Writer FX (https://github.com/JFormDesigner/markdown-writer-fx)

mvitorovic:
Java generic markup parsing library 
https://github.com/atlassian/commonmark-java

mvitorovic:
We need to make some progress on this. IIRC Caludio was said he will do the initial work?

claudio-rosati:
Despite my intentions, phoebus is now a very low priority task for me, being we are preparing for commissioning having a better Display BUilder (meaning also courses etc.) requires me on it almost at 100%

shroffk:
The markdown-writer-fx is not supported with java9.

shroffk:
Creating some prototype controls for displaying logentries

Single log entry:
![image](https://user-images.githubusercontent.com/2111304/42107202-7a10cd20-7ba4-11e8-93ed-5c29be855c30.png)

Log Entry Table:

![image](https://user-images.githubusercontent.com/2111304/42107229-8d0ad3c6-7ba4-11e8-9e6a-ef44800c2f7e.png)

shroffk:
Ok
The time and Title can be in the same line, the body remains as is.
We can also set the Title to bold.

shroffk:
commonmark support has been added to the olog ui clients.
currently I don't see a way to add it to the common ui without effecting other users - elog/sns logbook
closing issue

PR #208, opened 2018-07-02
Issue 204

shroffk:
#204 #64


shroffk:
Todo:

- [ ] add title in ui widget.

PR #212, opened 2018-07-11
Issue 204

shroffk:


kasemir:
The `AuthorizationService` was only used to enable/disable UI elements, so it started in core-ui.
I'll move it into your new core-security before too long, looks like a better place.

For secure storage, maybe we can use something based on `java.security.KeyStore`

Issue #135, opened 2018-03-19
File Browser


kasemir:
Desired functionality (still open to discussion) and progress on the file browser introduced in #134:

 * [x] Browse files
 * [x] Context menu to "Open" file in appropriate app
 * [x] Allow selecting app if there's more than one
 * [x] Preference for setting the initial root directory for new file browsers
 * [x] Show error when file can't be opened in dialog (in addition to the current log message)
 * [x] Context menu to "Open With" to pick app
 * [x] Double-click to "Open"
 * [x] Allow opening in external tool
 * [x] Icon of the (default) application
 * [x] Allow opening file in text editor (is that just one of the external tools?)
 * [x] Save/restore state. AppResourceDescriptor with create(URL) so that it allows opening navigator for directory, keeping that as the file:// resource which is saved/restored on restart? Or use memento for the URL? Or use AppResourceDescriptor for URL, plus memento for the currently selected file, expanding file tree to that on restart?
 * [x] Hide 'hidden' files based on `File.isHidden()`
 * [x] Context menu: Rename files
 * [x] Context menu: Delete files
 * [x] Context menu: Create new directory
 * [x] Context menu: Copy/paste/duplicate files
 * [x] Context menu: Refresh
 * [x] Allow 'dropping' files from outside to copy or move them into the target directory
 * [x] Context menu: Show/edit properties (executable, read-only, limited to what's supported by NIO package for all operating systems)?
 * [x] Copy full path of file to clipboard (so can then use 'chmod' etc. in terminal)
 * [x] Watch file system and refresh display as files are added/removed/renamed

kasemir:
Already works quite well, so far no performance issues when browsing files!

Small layout issue in the top row on Linux, where label and button are too narrow:
![fb](https://user-images.githubusercontent.com/1932421/37605135-f599f89a-2b68-11e8-97bf-164b1a22a5b2.png)

shroffk:
Ok...I can clean that up

The double click open is also implemented...it opens using the default app or prompts the user to select an app when multiple apps can handle the selected file.

kasemir:
Regarding double-click, I think it should be like File/Open: If there are multiple apps, it uses the preference to pick the default. (See use of `Preferences.default_apps` in `PhoebusApplication`.(

Right now double-click is like File/Open With:  If there are multiple apps, it prompts. You can already to that via the context menu, Open With. The double click should be the fastest possible operation, using the Preferences.default_apps, or the first one found, never prompting.

[...]

PR #217, opened 2018-07-13
Add support for 'application' based on external tools

kasemir:
@shroffk , this adds support for registering external tools as applications.

For example, on Linux it can start the text editor `gedit` for file extensions txt, py, ini, and the image viewer `eog` for png, jpg files.

This is done by extending the `ApplicationService` to not only register `App*Descriptors` via SPI but also add app entries based on a new preference setting for external tools.

Both the `ApplicationService` and the `ResourceHandlerService` used to run over the SPI for `AppResourceDescriptor`. Both the `PhoebusApplication` and the `ResourceHandlerService` used to `start()` apps with resource. That has been consolidated by removing the `ResourceHandlerService`. So `ApplicationService` now has the map of all apps and all file extension mappings, only checking the SPI once. Only `PhoebusApplication` "starts" all applications.


kasemir:
Example preference settings:
```
# ---------------------------------------
# Package org.phoebus.framework.workbench
# ---------------------------------------

# External applications
#
# Defines applications to use for specific file extensions
#
# Format:
#
# Each definition consists of name, file extensions, command.
#
# Name is the name of the definition, used to register the application.
# File extensions is a '|'-separated list of file extensions (not including the 'dot').
# Command is the path to the command.
#
# Multiple definitions are separated by ';'
#
# Example:
#
# Start 'gedit' for text files, and 'eog' for images:
#
# external_apps=text,txt|py|ini,gedit;image,png|jpg,eog
#
# The command will be invoked with the full path to the file.
external_apps=

# Directory where external applications are started
# May use system properties
external_apps_directory=$(user.home)
```

kasemir:
Part of #135 : Start text editor on files

kasemir:
Example setting for initial SNS tries to allow opening IOC *.db, startup files, PDF docs, images
```
org.phoebus.framework.workbench/external_apps=text,txt|py|ini|db|cmd|sh|st,gedit; image,png|jpg|gif|jpeg,eog; pdf,pdf,firefox
```

PR #220, opened 2018-07-17
Log entry view

es-sns121:
This is the mostly complete implementation of the log entry dialog/view. It currently treats all attachments as images since the AttachmentImpl doen't allow the content type to be set, but otherwise is almost entirely functional. The final bits of functionality include setting default logs and saving user credentials. They will be done on another branch.


kasemir:
It tries to download http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/jdk-10.0.1_linux-x64_bin.tar.gz which fails.

kasemir:
Pull request #221 has the same symptom: Previous builds worked, but since last commit the JDK10 based build fails to obtain JDK.

PR #234, opened 2018-07-25
Issue 64

shroffk:
added a simple viewer for log entries...
One minor change to Attachment. It turns out I need a 'getName' since olog returns on the name and I have to fetch the file latter.
In SNS case the can simple implement 
```
    @Override
    public String getName() {
        return file.getName();
    }
```


kasemir:
Ok, added the getName() to SNS implementation.

PR #235, opened 2018-07-26
#64 accidentally added the pane twice.

shroffk:
Fixing the LogEntryCalender bug found by @es-sns121


es-sns121:
On line 96 of LogEntryCalendar.java you get the Agenda.css file, however it fails to find the resource for me unless I change "Agenda.css" to "/Agenda.css". Would that still work on windows with that change?

shroffk:
Go ahead and make that change...I am currently not using that css file anyway and it is simply there as a placeholder for future work

PR #237, opened 2018-07-27
EMail Dialog: images, ..

kasemir:
Extracting the image list for viewing image attachments, adding files, adding screenshot, .. from the SendLogbookDialog, so same image panel used for logbook and email submission.

Then doing same for general attachment (files other than images), so again email and logbook dialog are the same for attachments.


shroffk:
+1 on having a common "image gallery"

why not use javafx.scene.layout.TilePane, it does a pretty decent job of figuring out how rows and columns to best fill the space and fitting as many images in that space.

kasemir:
Yes, as long as the selection is possible (to select an item, and that enables a 'remove' button), that would be better. I'm adding the "File" section for non-image attachments, and then look into updating the now-common ImageList with a TilePane.

kasemir:
This is what it looks like now:

nontile

What I like:

1) Can see a "big" preview of a selected item
2) Can select an image and then delete

The TilePane is a layout, not like a ListView that's aware of selection.
I think what it would look like: A row of same-sized images. No way to per se "select" an image, but each image could have an "X" in the upper right corner to remove that image.

I'll try that and post screenshot.

kasemir:
Actually, I give up on the TilePane.
I makes every tile the same width. Problem is that the tiles, i.e. ImageViews, will on their own assume the full size of the image, which tends to be way too large.
So need to update the ImageView.fitWidth and fitHeight yourself, for example set to total width divided by number of images. Means you're doing the layout yourself.
In addition, the TilePane doesn't scroll, so you soon end up with may tiny images, while the original approach allows for many images, all having a reasonably sized preview in the scrollable list, plus the larger preview section.

I may be wrong, and if somebody can do better, the `ImageList` is only 200 lines of code, so easy enough to try other approaches, which will then be used in both the email and logbook dialog.

kasemir:
Here's one of my sorry attempts to use TilePane:

tiles

PR #245, opened 2018-08-02
Issue 218

shroffk:
A new alarm config logging service


kasemir:
Not sure what's broken, seems different from the frequent cannot-get-jdk travis error

shroffk:
I tried to do some acrobatics with the branches...it seems to be trying to get a deleted branch

kasemir:
Travis indeed tries to clone an `Issue_alarm_config_service` branch instead of `Issue_218`.
Did you rename the issue or branch? Close and create new pull request?

PR #246, opened 2018-08-02
Issue 64

shroffk:
A new log entry table view

image


kasemir:
Can the timestamp use `org.phoebus.util.time.TimestampFormats..._FORMAT`?
"2018-08-02 13:52" looks better than "2018-08-02T18:52Z"

kasemir:
Otherwise great, in fact should merge right away.

PR #247, opened 2018-08-02
Archive engine

kasemir:
Creates services/archive-engine that builds the RDB-based archive engine.


kasemir:
Similar to this, I'll move the alarm-server-product and scan-server-product into services

PR #264, opened 2018-08-13
Alarm logger readme

es-sns121:
Tweaks to properites and classpath files.

The .classpath files contained references to non-existent "test" directories.
 
The alarm_logging_service.properties file had an IP address set in it. Commenting this out let the address default to localhost.

Updates to the readme to flesh out setup instructions.


shroffk:
Thanks

PR #266, opened 2018-08-14
Alarm state logger fixes

es-sns121:
-  AlarmStateLogger: Extract PV using AlarmTreePath.splitPath()

   Using the regex method resulted in a config of
   "/Accelerator/sim://sine(0, 10, 1)" having a PV field of
   "sim://sine(0,".

   Using splitPath, the PV field is now "sine(0, 10, 1)".

- Make config, severity, current_severity keyword.

   This allows searches and aggregations to be performed upon these fields.


shroffk:
> - Make config, severity, current_severity keyword.
> 
> This allows searches and aggregations to be performed upon these fields.

Text fields are also searchable, the only thing is the a different query would have to be created, instead of a term query we would have to use a match one.

es-sns121:
Ah, ok. Would text also be able to be aggregated? I wasn't able to perform any aggregations in Kibana on fields that were text, but when I changed them to keyword I could.

shroffk:
hmm... aggregation should be possible
however while reading the documentation I found 
` hostnames, status codes, or tags, it is likely that you should rather use a keyword field`
So I guess your changing the field type is good

PR #268, opened 2018-08-15
Build alarm logger

kasemir:
Fixes parent/child links in pom.

Adds `ant dist` support to build a distributable `alarm-logger-0.0.1.zip` that one can download, unzip, and run (with alarm-logger.sh for Linux).

Adds command line arguments for site-specific settings:
```
-help                       - This text
-server  localhost:9092     - Kafka server for reading alarm updates
-config  Accelerator        - Alarm configuration to monitor (more than one separated by ':')
-es      localhost:9200     - Elastic Search for writing alarm updates
-logging logging.properties - Load log settings
```

Adds a simple console so when running, "shutdown" will, well, shut it down. This allows running it in `procServ`:
```
./alarm-logger.sh 
2018-08-15 10:50:34 INFO [org.phoebus.alarm.logging] Alarm Logging Service (PID 8209)
2018-08-15 10:50:34 INFO [org.phoebus.alarm.logging] Starting logger for '..State' of Accelerator
...
> help
Commands:
	help             - Show help.
	shutdown         - Shut alarm logger down and exit.

 > shutdown
 > 
Done.
```


shroffk:
hangout?

kasemir:
Updated for #269 , keeping the AlarmStateLogger, ElasticClientHelper and alarm_logging_service.properties as found there except for trailing spaces.
Adds command shell, ant build, and option to set java.util.logging.

kasemir:
Command line args:
```
-help                                    - This text
-topics   Accelerator                    - Alarm topics to be logged, they can be defined as a comma separated list
-es_host  localhost                      - elastic server host
-es_port  9200                           - elastic server port
-bootstrap.servers localhost:9092        - Kafka server address
-properties /opt/alarm_logger.propertier - Properties file to be used (instead of command line arguments)
-logging logging.properties              - Load log settings
```

shroffk:
you must have noticed that the arguments have some inconsistencies.

es_host  & es_port have an underscore
boothstrap.server has a "."

I am trying to match the properties as described in the respective clients to elastic and kafka...maybe this is not great and it might be better to have consistent cmd line arguemtents

es             localhost:9200
kafka        localhost:9092

What do you think?

kasemir:
`-es localhost:9200` and `-kafka localhost:9092` is what I had for command line args. But you seem to like the property file, adding a command line option to select a different property file. Having the same name everywhere is best. Since we cannot change what the other libs expect to find in the property file, the command line args thus have to use the ugly property names. I think that's OK for this type of service that no end user will ever see.

PR #269, opened 2018-08-15
Issue 218

shroffk:
should be merged with #268

### DIFF
--- a/services/alarm-logger/alarm-logger.sh
+++ b/services/alarm-logger/alarm-logger.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+#
+# Alarm Server launcher for Linux or Mac OS X
+
+# When deploying, change "TOP"
+# to the absolute installation path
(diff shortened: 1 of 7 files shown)


kasemir:
OK, looking at the dates, it seems you had started to fix some important broken things (bad pom, ignoring topic, no start script, no way to configure, ..) a week ago, but there was no pull request.
So I fixed the same things today, and added the build config.
==> We need to pull/merge more often to prevent duplicate work.

kasemir:
I'll merge yours first, then fix the merge issues with my additions.

shroffk:
logging.properties is good...we can surely have that.

the alarm_logging_properties allows the user to give a properties file instead on listing everything on the command line

`java -jar -properties /my/config/file.properties`

I have not been very regular sorry...

shroffk:
@kasemir I do not have the shell commands which would be great to have too.

PR #272, opened 2018-08-16
Make dockpane tab name more readable

es-sns121:
The dockpane tab names were harder to read as "applicationName" than "Application Name".


shroffk:
@es-sns121 the application names are used to launch from command line and integrated actions in the old cs-studio...avoiding spaces in NAME is preferable.

The DISPLAYNAME should be the one with spaces.

For the dockpane, uses the getDisplayName...so we can add that method in the appropriate applications.
```
public DockItem(final AppInstance applicationInstance, final Node content)
    {
        this(applicationInstance.getAppDescriptor().getDisplayName());
        getProperties().put(KEY_APPLICATION, applicationInstance);
        setContent(content);
    }
```

es-sns121:
The NAME's are back to applicationName, and I added getDisplayName methods for the three logbook applications.

shroffk:
Cool,
Can I request one more thing that will make Kay extremely angry...

Since the rest of the class uses 
```
getDiplayName() {
}
```

Could we preserve this format for this method too...pretty please

es-sns121:
Since you said pretty please

Issue #275, opened 2018-08-17
Alarm Server: 'Passive' model


kasemir:
Look into a 'passive' option or maybe split the code so that the alarm server model can be used to simply capture the alarm configuration without creating PVs, sending state updates.

kasemir:
Slight reorg in @279 allows either reading model from Kafka, then export as XML, or build in-memory model detached from Kafka, then export as XML.
So there are classes to handle alarm model without PV, Kafka, alarm server.

PR #279, opened 2018-08-17
Alarm config xml 275

kasemir:
For #275..

@shroffk On one hand, this moves some code from the alarm server to the alarm model, a generally useful reorg. In addition, it adds an example.

With this, all the code for keeping an in-memory model of the alarm configuration, and formatting that as XML is in app-alarm-model.

`AlarmModelWriterTest`: Shows how to programmatically create an alarm tree with root, subnodes, leaf nodes in memory and then write that as XML. The in-memory model uses `AlarmClient*` classes, but it's not connected to an actual Kafka client, doesn't read from Kafka, doesn't write to Kafka.
So the same type of code could read from ElasticSearch to build an alarm tree, then export as XML.

`AlarmModelSnapshotDemo`: Uses the `AlarmClient` to fetch alarm configuration from Kafka, then dumps as XML. When the alarm client builds the model from the Kafka stream of config updates, there is the technical issue of knowing when we're "done". In principle, the alarm config stream never ends, a user could at any time change the configuration, creating another update. This code shows how to wait for a pause in updates, considering that as a stable configuration.
With ElasticSearch, that would not be necessary because it reads the configuration updates until reaching a certain time stamp, and then it's "done" and can export as XML.

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/ResettableTimeout.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/ResettableTimeout.java
@@ -5,7 +5,7 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  *******************************************************************************/
-package org.phoebus.applications.alarm.server;
+package org.phoebus.applications.alarm;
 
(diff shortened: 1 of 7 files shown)


shroffk:
This is great.
+1

PR #284, opened 2018-08-22
Add PV name Drag to display runtime, alarm table, and accept in data browser

kasemir:
Idea started in https://github.com/ControlSystemStudio/cs-studio/issues/2500, was implemented in RCP display builder.


pavel-ch:
When dragging sim://sine(-100, 100, 50, 1) the following dialogue appears (needs to find braces before splitting PV1, PV2, ... PVn string)
image

kasemir:
You're right, this could be better.
#289

PR #293, opened 2018-08-24
Alarm index name helper

es-sns121:
- [x] Integrate the IndexNameHelper into the AlarmLoggerService and AlarmLogger.

- [x] Command line arguments for index_span_units and index_span_value.

- [x] Add scripts to create and delete templates from elastic search.

- [x] Make it so the AlarmLoggerService can only create a single non time based index if desired.


kasemir:
Yes, setting the value to 0 would be a good way to say you want just one index

kasemir:
Looking into the unit test failure. There is a race condition inside the alarm server 'filter'.

Issue #297, opened 2018-08-28
"Save File" dialog too small via remote-ssh


kasemir:
* `ssh` to a Linux host
 * `ssh` once more to a VM on that linux host
 * start phoebus inside the VM, displaying via the two-hop remote-X connection
 * open data browser, add PV, close the data browser

--> The "Save File" prompt for the dirty input is zero-sized and cannot be used.
Only happens when doing that strange double-hop.

kasemir:
Same with "Save Layout As .." dialog.

PR #298, opened 2018-08-28
Fixes

kasemir:
.. for the alarm filter test and #297


kasemir:
Just stopped an Alarm filter test that had been running since August 29. Before the fix, it would experience a hangup within a day, so fix indeed seems to work.

PR #302, opened 2018-08-31
NetBeans integration

claudio-rosati:
Added 2 goals for the NetBeans integration allowing to build with and without tests.

### DIFF
--- a/nbactions.xml
+++ b/nbactions.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<actions>
+        <action>
+            <actionName>CUSTOM-Skip Tests</actionName>
+            <displayName>Skip Tests</displayName>
+            <goals>
+                <goal>clean</goal>
+                <goal>verify</goal>
+            </goals>
+            <properties>
+                <skipTests>true</skipTests>
+            </properties>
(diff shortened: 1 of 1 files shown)


claudio-rosati:
I'm sorry if, after using NetBeans since it was called [Xelphi](https://nbguru.wordpress.com/2008/05/14/dotnet-history-of-the-ides-netbeans/) and having contributed to its graph library and RCP codebase, it just happens to be part of only the 8% of the developers population using NetBeans. The fact is that with NetBeans I can just be much much more productive than with Eclipse or any other tool for which I need to learn how to use it, its keystrokes, etc.

At ESS we java developers are evaluating the possibility to move to IntelliJ IDEA, probably the best development tool ever... it ha a single disadvantage: it's not free. But for multi-million facilities few hundreds euros should not be a problem to improve the life of who is maintaining them running.

On paper the Visual Studio Code, far to be a complete IDE, seems a good editor, and I'll try to give it a run (mostly to give our Linux development machines a lighter editor). Anyway what I think important for Phoebus is being able to say "we support Eclipse, Netbeans" and any other editor we are now free to use thanks to having abandoned Eclipse RCP (as far as including IDE/editor specific project files will not collide with others).

Issue #303, opened 2018-09-03
Test failed in core-util trying to build the whole project.


claudio-rosati:
```
------------------------------------------------------------------------
Building core-util 0.0.1-SNAPSHOT
------------------------------------------------------------------------

--- maven-clean-plugin:2.5:clean (default-clean) @ core-util ---
Deleting /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/target

--- maven-resources-plugin:2.6:resources (default-resources) @ core-util ---
Using 'UTF-8' encoding to copy filtered resources.
skip non existing resourceDirectory /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/main/resources
Copying 0 resource

--- maven-compiler-plugin:3.6.2:compile (default-compile) @ core-util ---
Changes detected - recompiling the module!
Compiling 62 source files to /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/target/classes

--- maven-resources-plugin:2.6:testResources (default-testResources) @ core-util ---
Using 'UTF-8' encoding to copy filtered resources.
Copying 9 resources
Copying 0 resource

--- maven-compiler-plugin:3.6.2:testCompile (default-testCompile) @ core-util ---
Changes detected - recompiling the module!
Compiling 41 source files to /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/target/test-classes
/Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/time/TimeParserTest.java: /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/time/TimeParserTest.java uses or overrides a deprecated API.
/Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/time/TimeParserTest.java: Recompile with -Xlint:deprecation for details.
/Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/text/CsvParserDebug.java: /Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/text/CsvParserDebug.java uses unchecked or unsafe operations.
/Users/claudiorosati/Projects/GitHub4/phoebus/core/util/src/test/java/org/phoebus/util/text/CsvParserDebug.java: Recompile with -Xlint:unchecked for details.

[...]

PR #304, opened 2018-09-03
Set Locale and Timezone in test for better reproducibility

kasemir:
#303


claudio-rosati:
It works fine now.
Thank you

PR #307, opened 2018-09-03
Added "Developing with NetBeans" instructions to the README file.

claudio-rosati:


claudio-rosati:
Don't merge yet.
If #307 is approved, I'll add another paragraph of how opening the product project and setting it as the default one to be run.

kasemir:
This is 307. When you write "If 307 is approved", do you mean 306?
I'm fine with merging both.

claudio-rosati:
Oh yes, sorry.
OK I'll add the further paragraph so you can them merge both.

PR #308, opened 2018-09-04
Fixes

kasemir:
Import of Data Browser rule fix.

Preserve last "To:" value for email dialog.

Fix layout reload


shroffk:
@kasemir in preparation for the appliance archiver plugin I did a little bit of code restructuring for the databrowser.
I am still working on it so it is not ready for PR but I thought I would let you know.

kasemir:
OK, so let's merge this now to avoid too many open branches that might diverge.

kasemir:
I'm unclear about Claudio's pull requests since one mentions "Don't merge yet" but then refers waiting for 307, which is itselve...

kasemir:
Otherwise we should merge those as well.

PR #312, opened 2018-09-06
Issue 305

shroffk:
Needs reviews and input 

TODO:
system dependencies can be removed


kasemir:
Let's merge this so I can fix the ant files and my beamline build & distribution setup.

shroffk:
I tested the ant build also...added changes to the ant_settings

kasemir:
Good, thanks.
I guess the basic build instructions remain:
Can either use `mvn -DskipTests clean install` from the top, or do that only for dependencies and then proceed with the `dependencies/phoebus-target/target/lib/`. Only difference is the added `phoebus-target` directory level.

shroffk:
> I guess the basic build instructions remain

yes...I created the new modules to make sure that old build instructions work

> Only difference is the added phoebus-target directory level.

yes, we can put the output of phoebus-target into /dependencies...if we really want this...

kasemir:
Some ant files still needed the `phoebus-target` added, now OK.

kasemir:
Removed all systemPath entries from the pom.xml, only remaining one is for the Oracle JDBC jar.
Will move the profile that's enabled when `env.ORACLE_JDBC_JAR` is defined to install-jars, so oracle JDBC jar is installed in similar way, but instead of including a copy in git as with pbraw (which we cannot do for legal reasons) you must already have it & and point an environment var to it.

Issue #310, opened 2018-09-06
Problems launching phoebus product


claudio-rosati:
When I start phoebus-product I have plenty of the following channel finder-related exception (both launching from NetBeans and from terminal):

```
CI0011766:target claudiorosati$ java -jar product-0.0.1-SNAPSHOT.jar 
2018-09-06 14:03:32 INFO [org.phoebus.product.Launcher] Phoebus (PID 31949)
2018-09-06 14:03:34 WARNING [org.phoebus.framework.workbench] external_apps missing name, ext, command for '
2018-09-06 14:03:34 CONFIG [com.sun.jersey.spi.service.ServiceFinder] Running in a non-OSGi environment
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices] Searching for providers that implement: interface com.sun.jersey.client.proxy.ViewProxyProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices] Searching for providers that implement: interface com.sun.jersey.spi.inject.InjectableProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices]     Provider found: class com.sun.jersey.core.impl.provider.xml.SAXParserContextProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices]     Provider found: class com.sun.jersey.core.impl.provider.xml.XMLStreamReaderContextProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices]     Provider found: class com.sun.jersey.core.impl.provider.xml.DocumentBuilderFactoryProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices]     Provider found: class com.sun.jersey.core.impl.provider.xml.TransformerFactoryProvider
2018-09-06 14:03:34 CONFIG [com.sun.jersey.core.spi.component.ProviderServices]     Provider found: class com.sun.jersey.multipart.impl.MultiPartConfigProvider
2018-09-06 14:03:34 FINE [com.sun.jersey.core.reflection.ReflectionHelper] Unable to load class javax.annotation.PostConstruct using the supplied class loader jdk.internal.loader.ClassLoaders$AppClassLoader.
java.lang.ClassNotFoundException: javax.annotation.PostConstruct
[...]

PR #317, opened 2018-09-07
#310 fixing the logging.properties

shroffk:


kasemir:
Thanks. Can't see how this will break the build, so not even waiting for travis..

PR #326, opened 2018-09-18
Fixes

kasemir:


kasemir:
@shroffk Turns out the install-jars pom was broken.

It had an install-jar-lib element to install pbrawclient-0.0.8.jar.
If there was an ojdbc8-12.2.0.1.jar, that enabled a profile to install it as well.

.. but it was done in a way where it updated the plugin configuration, so the plugin ran once and did this:
```
maven-install-plugin:2.5.2:install-file (install-jar-lib) @ install-jars 
 Installing dependencies/install-jars/lib/ojdbc/ojdbc8-12.2.0.1.jar to .m2/repository/com/oracle/jdbc/ojdbc8/12.2.0.1/ojdbc8-12.2.0.1.jar
 Installing dependencies/install-jars/lib/pbraw/pom.xml to .m2/repository/com/oracle/jdbc/ojdbc8/12.2.0.1/ojdbc8-12.2.0.1.pom
```

So it installed the ojdbc jar with the pbraw/pom.xml.

Moved the 'configuration' section into the 'execution', and use one execution per jar to install, so now both jars are installed correctly.

shroffk:
> Moved the 'configuration' section into the 'execution', and use one execution per jar to install, so now both jars are installed correctly.

sorry, you are right..it should have been in the execution section.

shroffk:
@kasemir 
Do you want me to move the pbraw inside a separate profile too?

kasemir:
I think the configure was OK for the usual case that we execute some plugin once.

pbraw is fine to always install, don't change that.

I wish we could also always install ojdbc, but alas..

Issue #218, opened 2018-07-16
Alarm services


shroffk:
A project issue ticket for the various pieces of the alarm environment.

- [x] Alarm state message logger
      Tracks changes in alarm severity for PVs and alarm hierarchy up to the root.

- [x] Alarm command message logger
      Tracks alarms being acknowledged or un-acknowledged

- [x] Alarm config message logger
      Tracks current configuration and history of changes (PV added, guidance updated, related display changed, PV disabled, .., PV removed)

- [x] Alarm talk message logger
      Tracks alarms being annunciated

kasemir:
Will these reports be in a web client, using some existing tools for elastic search where all that's left to do is define some query-type configuration, or will this be Java code & UI running in Phoebus?

shroffk:
> Will these reports be in a web client, using some existing tools for elastic search where all that's left to do is define some query-type configuration, or will this be Java code & UI running in Phoebus?

Both,
I will add a simple phoebus UI. I would like a context menu action which brings up the table with the alarm history of a pv.

shroffk:
The alarm services have been implemented.
new feature requests should be opened as new issues.

PR #333, opened 2018-09-20
#218 fix NPE in the properties helper for reading default properties

shroffk:


kasemir:
I've seen the jdk10 build fail because travis could not install java 10.
After pushing another bogus whitespace commit to re-trigger travis, it succeeded.

shroffk:
Ok, Thanks

Issue #341, opened 2018-09-23
Try to compile phoebus on Ubuntu 16.04.5 LTS


hjunkes:
Can't compile on Ubuntu 16.04.5 LTS with java-9-openjdk-amd64 or jdk-10.0.2. 

[h1@earth phoebus (master)]$ ant clean dist
Buildfile: /home/h1/PHOEBUS/phoebus/build.xml

clean:

clean:
   [delete] Deleting directory /home/h1/PHOEBUS/phoebus/core/framework/target

clean:

...

clean:

core:

core-framework:
    [mkdir] Created dir: /home/h1/PHOEBUS/phoebus/core/framework/target/classes
    [javac] Compiling 62 source files to /home/h1/PHOEBUS/phoebus/core/framework/target/classes
    [javac] /home/h1/PHOEBUS/phoebus/core/framework/src/main/java/org/phoebus/framework/spi/MenuEntry.java:7: error: package javafx.scene.image does not exist
    [javac] import javafx.scene.image.Image;
    [javac]                          ^
    [javac] /home/h1/PHOEBUS/phoebus/core/framework/src/main/java/org/phoebus/framework/spi/MenuEntry.java:43: error: cannot find symbol
    [javac]     public default Image getIcon()
    [javac]                    ^
    [javac]   symbol:   class Image
    [javac]   location: interface MenuEntry
    [javac] /home/h1/PHOEBUS/phoebus/core/framework/src/main/java/org/phoebus/framework/spi/ToolbarEntry.java:3: error: package javafx.scene.image does not exist
    [javac] import javafx.scene.image.Image;
    [javac]                          ^
    [javac] /home/h1/PHOEBUS/phoebus/core/framework/src/main/java/org/phoebus/framework/spi/ToolbarEntry.java:24: error: cannot find symbol
    [javac]     public default Image getIcon()
    [javac]                    ^
    [javac]   symbol:   class Image
    [javac]   location: interface ToolbarEntry
    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] Note: Some input files use unchecked or unsafe operations.
    [javac] Note: Recompile with -Xlint:unchecked for details.
    [javac] 4 errors

BUILD FAILED
/home/h1/PHOEBUS/phoebus/build.xml:65: The following error occurred while executing this line:
/home/h1/PHOEBUS/phoebus/core/framework/build.xml:6: Compile failed; see the compiler error output for details.

[...]

PR #338, opened 2018-09-21
#366

shroffk:
simplify the running of the product via maven using
mvn exec:java

PR #342, opened 2018-09-24
Clarify need for JFX, JDK 9 or later. Build before trying to run.

kasemir:
#341 #338


kasemir:
The build failure appears to be a timing-sensitive test of the alarm actions.
Will fix that on the other pending pull request.

PR #343, opened 2018-09-24
ActionButton: Apply 'disabled' style to button, not pane

kasemir:
Import of https://github.com/kasemir/org.csstudio.display.builder/issues/447


kasemir:
Added more tolerant timing to test that failed in #342

PR #350, opened 2018-10-03
3d viewer

es-sns121:
Further updates to the 3d viewer.
 - [x] Save resource to dockitem so reopening app uses saved file.
 - [x] Add a reset view button
 - [x] Add a shape file Lithium Niobate (LiNbO3)
 - [x] Add a clear view button
 - [x] Define default directory for file chooser
 - [x] Change viewer background
 - [x] Opening a .shp file opens the viewer
 - [x] Online help
 - [x] Only allow one open viewer per resource
 - [x] Create pom file
 - [x] Create ant build file
 - [ ] Make into display builder widget

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -98,6 +98,7 @@
     <ant dir="app/alarm/model"/>
     <ant dir="app/alarm/ui"/>
     <ant dir="app/update"/>
+    <ant dir="app/3dViewer"/>
   </target>
 
   <target name="product" depends="app" description="Assemble product">


es-sns121:
Which directory would make the most sense to be the default?

This could be set by a preference such as with the file browser.

My first thought was to use the user's home directory. 

     `new File(System.getProperty("user.home"))`

However, this is likely not reliable outside of *nix based operating systems. 
Windows has a poorly defined idea of a home directory.

kasemir:
Yes, a preference similar to the file browser makes the most sense for all the reasons you mention, so similar to https://github.com/shroffk/phoebus/blob/master/app/filebrowser/src/main/resources/filebrowser_preferences.properties

In practice, I think we'll point the file browser and the 3d viewer to similar yet slightly different locations, so separate preferences make sense, for example:
```
org.phoebus.applications.filebrowser/default_root=$(user.home)/phoebus_files
org.phoebus.applications.viewer3d/default_dir=$(user.home)/phoebus_files/molecules
```

es-sns121:
I made some changes to the app/display/editor/build.xml because every time I tried to build with ant, it would fail due to the javadocs failing. This was due to the html/generated directory having been deleted and recreated during `ant -f javadoc.xml clean all`. Due to this I've added a line recreating this directory during the javadoc target. 

However, I'm not sure if this is correct.

PR #351, opened 2018-10-05
Slide button

kasemir:
Import of https://github.com/kasemir/org.csstudio.display.builder/pull/455


shroffk:
In the past when I have seen this issue it has usually resolved itself in a few mins.
I have restarted the build but with the branch deleted not sure it this will have any impact.

kasemir:
As you thought, "Could not find remote branch slide_button to clone"

Issue #143, opened 2018-04-19
Java 10, Java 11


kasemir:
Phoebus runs OK with Java 10.

The one new thing in Java 10 is 'var', but I don't see a desperate need to use that and break Java 9 compatibility.

Creating this ticket to track progress with Java 11:

 * Announcement that JDK 11 will no longer include JFX, https://blogs.oracle.com/java-platform-group/the-future-of-javafx-and-other-java-client-roadmap-updates

 * Oracle white paper "Java Client Roadmap" http://www.oracle.com/technetwork/java/javase/javaclientroadmapupdate2018mar-4414431.pdf

 * OpenJDK "making sure .. can easily create and run JavaFX applications [in] Java 11", 
https://github.com/javafxports/openjdk-jfx/issues/52

claudio-rosati:
I don’t think I’ll never use ‘var’: from a maintenance point of view is one of the worst thing.

kasemir:
Java 11 Update (http://mail.openjdk.java.net/pipermail/openjfx-dev/2018-September/022475.html):

 * New OpenJFX community site http://openjfx.io/
 * "final release of JavaFX 11" available as download from that site or maven central as openjfx:javafx-COMPONENT:11 with COMPONENT=base, graphics, controls, ...
 * Runs with preview of JDK 11 or JDK 10 (OpenJDK build only)

I take this to mean:
Once we switch to Java 11, we add  openjfx:javafx-*:11 to the dependencies, and expect that to work with both Oracle and OpenJDK 11, neither of which will bundle JFX.
Java 9 and 10 will then need to be abandoned, because Oracle9/10 include JFX and that would clash with the JFX11 from the dependency.

willrogers:
I have someone working on building and running CS-Studio with Java 11 - OpenJDK plus OpenJFX.

claudio-rosati:
Next week I'll start looking to extend the Integration plugin to open Display Builder and Databrowser 3 in Phoebus from CS-Studio.

shroffk:
> Oracle9/10 include JFX and that would clash with the JFX11 from the dependency.

I am having trouble with travis and openjdk + javafx

The trouble is that openjfx is no longer available for trusty. It is available for xenial and bionic distributions of ubuntu...but travis still does not have support for those distributions.

So when we move forward to 11, we might have to rewrite the travis build files.

shroffk:
[...]

PR #352, opened 2018-10-08
Java11

kasemir:
Supports building with Java 11, #143

This adds the dependencies for the javafx jars that are now separate org.openjfx packages.
It builds and runs with openjdk11.

Since the javafx jars differ with operating systems, `ant dist` now creates a separate `phoebus-0.0.1-linux.zip`, `phoebus-0.0.1-mac.zip` or `phoebus-0.0.1-win.zip`.

Good news:

 * It is possible to cross-build, i.e. to build the `phoebus-0.0.1-mac.zip` on Linux.

 * It still works with Oracle JDK 9 and 10.
   Not 100% sure why, but what happens for them is that the build adds the new org.openjfx packages
   to the dependencies/phoebus-target, yet in the end the java runtime uses the JavaFX that's included
   in Oracle JDK9 or 10, ignoring the javafx jars that we placed on the classpath.

Bad news:

 * pvAccess PVs no longer work because of the unfortunate corba dependency,
   and corba is no longer included in JDK11.
   Don't know if the latest pvAccess still needs that.
   Should be as simple as replacing the corba BooleanHolder and ShortHolder
   with AtomicBoolean and AtomicInt.

Screenshot running with Oracle JDK10:
JRE and JFX have the exact same version, since they're from the same JDK

jdk10

.. and OpenJDK11:
JFX is from the maven central module, slightly different version than the OpenJDK JRE.
(Yes, it's openjdk, still reports 'Oracle')
jdk11


kasemir:
Merging since it adds J11 support while J9/10 still work.

Note that pva:// pvs will only function under J9/10 and not J11 until the pvAccess library has been updated to the latest from master, see  https://github.com/epics-base/epicsCoreJava/blob/master/pvAccessJava/src/org/epics/pvaccess/impl/remote/IntrospectionRegistry.java

PR #359, opened 2018-10-17
removing jdk10 from CI

shroffk:


shroffk:
I think we need to prepare to move to 11.
I propose removing jdk10 from the travis CI for the time being

shroffk:
@kasemir @claudio-rosati 
I see that openjdk11 listed in the avaiable jdks...

kasemir:
It builds with openjdk11 for me locally, so maybe we should enable that.
Only issue is running pvAccess, but that'll be fixed as we update the pvAccess lib, and doesn't affect the build.

shroffk:
Ok... I will set this up for us to try

PR #361, opened 2018-10-17
Issue logbook

shroffk:
Not complete but creating a PR in keeping with the phoebus philosophy of merging often


shroffk:
I just pushed the merged / updated travis
The travis update had been done on a seperate branch so that it could be independently tested and merged...

PR #366, opened 2018-10-18
Issue 218

shroffk:


shroffk:
@es-sns121 I made a few changes in the logbook api
Like the search job takes a map instead of a string.

I also feel like we should be integrate the write and view bits more...do you want me to do a demo of the recent changes?

PR #372, opened 2018-10-23
testing openjdk11 - WIP

shroffk:
Adding jdk 11


shroffk:
@claudio-rosati @kasemir should we add the jdk 11 to the CI

kasemir:
Good!

What's the outlook for updated pvAccess, pvData, vtype jars so I can try those and make pva:// PVs work with JDK11?

PR #400, opened 2018-10-31
kafka-config-path

shroffk:
it works with jdk 10 and 11

https://issues.apache.org/jira/browse/KAFKA-6855


kasemir:
I'm just in the process up updating the Readme to kafka_2.11-2.0.0, the latest release, where this is no longer necessary. It works with Java 10.

kasemir:
I'll merge, but the next update will remove that because no longer necessary

shroffk:
Any plans on how/when we will upgrade? I am getting update fatigue :)

kasemir:
For the USPAS training setup, I'm going through the notes and noticed that a new version is out, so I installed that and updated the Readme (kafka_update branch).
Works fine with out current client, which is using kafka-clients 1.1.0, but I'll look at updating to 2.0.0 tomorrow.

PR #422, opened 2018-11-12
Fix #415

### LINKED ISSUE #415: Alarm Logger query to Remote Elastic Search server fails
When alarm-logger jar is run with '-es_host' option set to a remote elastic-search server, it errors out complaining it can't find the 'accelerator_alarms' index. This issue is not seen when elastic search server is run on localhost, but only seen when a remote ES server is used. I have confirmed that 'accelerator_alarms' index do exist on remote ES server but the alarm-logger still complains it can't find it.

tanviash:
Alarm-Logger : This is the fix for the problem where ElasticClientHelper wasn't using the properties file to get es_host and es_port values. Hence, it was always trying to make elastic search connection to default server which is localhost:9200.

### DIFF
--- a/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
+++ b/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
@@ -7,6 +7,7 @@ import static org.phoebus.alarm.logging.AlarmLoggingService.logger;
 
 import java.io.IOException;
 import java.util.logging.Level;
+import java.util.Properties;
 
 import org.apache.http.HttpHost;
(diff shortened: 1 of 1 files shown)


shroffk:
@tanviash the PreferencesReader is a phoebus service for managing and storing preferences with their defaults.
It is great that you have discovered the source of the issue here...maybe we can debug why the PreferencesReader is not getting the host and port values.

tanviash:
@shroffk When PreferenceReader is used by other modules, they have an option "-settings" to pass these preferences. We do not have any such '-settings' option for alarm logger. Also, i9t'll be redundant to have '-properties' and '-settings' to set the same params.

PR #426, opened 2018-11-14
hide the attachment accordion when no there are no attachments

shroffk:


kasemir:
Good idea!

I'll add the same to the LogEntryDialog's attachment view.

Issue #448, opened 2018-11-28
Font popover localization issue


pavel-ch:
Trying to make localization of Display Builder Editor, I have found that Font popover (and expecting the same for Color popover) is not in the right locale, although the rest of texts in the Editor is localized as expected.
I suspect WidgetFontPopover.java line 46 being the problem - loads always  the default locale texts:
`final InputStream iStream = WidgetFontPopOver.class.getResourceAsStream("messages.properties");`
Should I try to resolve the issue or you have an idea right now?
By the way, some applications like Probe or File Browser have no `messages.properties` so far. Is there a plan to make it localizable?

kasemir:
You're correct, it should use https://github.com/shroffk/phoebus/blob/master/core/framework/src/main/java/org/phoebus/framework/nls/NLS.java to fetch the localized messages, instead of loading a fixed filename.
Ideally, everything would be localized, it's just a matter of time.
If you can contribute some localization, that would be great.

pavel-ch:
Ok, idea with filename is working, but I do not know if it suits to you.
Advantage: using fxml %keyword construct for localization.
WidgetFontPopover.java line 46 replaced by:
```
String fname = "messages_" + Locale.getDefault().getLanguage() + ".properties";
InputStream iStream = WidgetFontPopOver.class.getResourceAsStream(fname);
if (iStream == null)
{
    iStream = WidgetFontPopOver.class.getResourceAsStream("messages.properties");
}
```
How to combine fxml localization with your NLS class is unclear for me.

kasemir:
How about you extract something like this into a new NLS method:
```
class NLS
{
  ..
  /** Get stream for messages
    *  Tries to open "messages_{LOCALE}.properties",
    *  falling back to generic "messages.properties" 
    *  @param clazz Class relative to which message resources are located
    *  @returns Stream for messages or null
    */
 public static InputStream getMessages(Class<?> clazz)
 {
            // First try "messages_de.properties", "messages_fr.properties", "messages_zh.properties", ...
            // based on locale
            String filename = "messages_" + Locale.getDefault().getLanguage() + ".properties";
            InputStream msg_props = clazz.getResourceAsStream(filename);

            // Fall back to default file
            if (msg_props == null)
            {
                filename = "messages.properties";
                msg_props = clazz.getResourceAsStream(filename);
            }

            if (msg_props == null)
                getLogger().log(Level.SEVERE, "Cannot open '" + filename  + "' for " + clazz.getName());
[...]

PR #449, opened 2018-11-29
Font popover localization issue - part 1

pavel-ch:
Solving issue #448 
NLS.java - helper for internationalization fxml dialogs
Color and Font PopOvers - using NLS instead of fixed filename for messages.properties


kasemir:
Thanks!

PR #451, opened 2018-11-29
Font popover localization issue - part 2 Filebrowser internationalization

pavel-ch:
Filebrowser + Probe internationalization continuing #448


kasemir:
Thanks!

PR #455, opened 2018-11-30
Issue 218

shroffk:
- update the indices for state and command logging
- add remote git push support

### DIFF
--- a/services/alarm-config-logger/.classpath
+++ b/services/alarm-config-logger/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-    <classpathentry kind="src" output="target/classes" path="src/main/java"/>
+  <classpathentry kind="src" output="target/classes" path="src/main/java"/>
   <classpathentry kind="src" output="target/classes" path="src/main/resources"/>
   <classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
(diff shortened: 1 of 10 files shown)


shroffk:
@tanviash some changes to the logger

tanviash:
What does alarm commands mean?

kasemir:
The XXXCommand topic, used by the client to send commands to the alarm server.
Commands are used to acknowledge or un-acknowledge alarms.

shroffk:
So this will not record both the State and the Command messages...
https://github.com/shroffk/phoebus/tree/master/app/alarm

So we will have a record for both 
- when the user performs and action (acknowledges)
- when the alarm server sends out an associated state update (alarm state changes to Acknowledged)

kasemir:
If all fails, read the documentation, https://github.com/shroffk/phoebus/blob/master/app/alarm/Readme.md, "Message Formats", "Command Topic"

kasemir:
The messages are sent via the config, state and alarm topics because those 3 streams have different direction (who's sending?) and traffic volume. That might allow optimizing the Kafka config, like "Command" topic can simply delete old messages, while config and state can compact.

At the same time, the messages are such that you can tell from the content what type of message it is. In case a client simply reads from all topics, it doesn't have to check what topic delivered the message, the content tells all there is.

tanviash:
Thank you Kunal & Kei.

kasemir:
I'm OK with merging that.

PR #462, opened 2018-12-05
 Filebrowser: few more texts externalized

pavel-ch:
#448 continue externalization


kasemir:
For what it's worth, I would not localize the text of Exceptions.

Exceptions are not meant for the end user.
When an exception occurs, the text is meant for the developer.
It's then helpful to have a known text which you can locate in the source code, and just like the rest of the source code (comments, variable names, ..) that should be in English.
A localized exception means you first need to figure out what the foreign-language exception means...

pavel-ch:
That is right. But these exceptions should be announced to user somehow, otherwise he/she will be confused that nothing happens. Should not there be a msgBox with the error instead?
I see the situation analogical to CreateDirectoryAction.java:52, for example.

kasemir:
I think a good general rule is to throw exceptions way down in the code, and report them to the user up high.

The exception carries some detail that mostly makes sense to a developer (cannot locate file, cannot read file, URL retrieval timeout, not valid XML, missing certain XML tag, value out of range, ...), and the exception message is in English, unaffected by localization.

In some cases the error is only logged. In other cases, on a higher level, the error might be reported to the user, for example via org.phoebus.ui.dialog.ExceptionDetailsErrorDialog, which can use a localized title and message ("Error opening ...", "Fehler beim Öffnen von ..."). It offers a "Details" pane for the exception message and stack trace.

kasemir:
Sentence structure can differ between languages.
Instead of constructing messages from multiple pieces like this:

```
Messages.PasteAlert1 + original + " " + Messages.PasteAlert2 + directory
``` 

I would use a 'format' string like "Bla bla {0} bla {1} bla" and then use that like this:

```
MessageFormat.format(Massages.PasteAlertFormat, original, directory)
```

This way, the localization would be any of
```
PasteAlertFormat=Dude, the folder {1} doesn't want to get {0} pasted into it
PasteAlertFormat=Failed to paste {0} into {1}
```

Issue #89, opened 2017-11-20
Externalize Strings, NLS in core


kasemir:
`org.phoebus.framework.nls.NLS` supports internationalization similar to the `Messages.java` and messages.properties` files in RCP to ease porting existing RCP-based code, but the core module itself still has many non-externalized strings.

Need to go through all the files and externalize texts.
Translations are also appreciated.

PR #487, opened 2018-12-12
 Text externalizations on core part 1

pavel-ch:
Continuing #448, partially #89


kasemir:
Thanks!

Are you doing this all by hand, or have you found a way to configure the Eclipse Source->Externalize Strings... tool to help with this?

pavel-ch:
Picking only some texts, looking for forgotten etc. Seems to be faster by hand. Thanks for the hint.

PR #516, opened 2018-12-27
Issue 218

shroffk:


kasemir:
Is the create_alarm_index.sh still needed? Doesn't the create_alarm_template.sh cover it?
In the latter, can the argument be removed, because just one "alarms_config_template" works for all, no need to create "accelerator_ alarms_config_template", "other_alarms_config_template", "xyz_alarms_config_template".

shroffk:
> Is the create_alarm_index.sh still needed? Doesn't the create_alarm_template.sh cover it?

If you disable auto creation of indexes then you would need this script to manually create the indexes...we could maybe say that auto creation of indexes must be enabled.

> In the latter, can the argument be removed, because just one "alarms_config_template" works for all, no need to create "accelerator_ alarms_config_template", "other_alarms_config_template", "xyz_alarms_config_template".

yes, I think you are right, this should work. I will remove the arguments.

kasemir:
OK, let's merge

kasemir:
@shroffk after merging this, the alarm logger fails to add messages like this:
```
Rejecting mapping update to [ihc_alarms_config_2018-12-14] as the final mapping would have more than 1 type: [alarm_cmd, alarm_config]]]
	at org.elasticsearch.rest.BytesRestResponse.errorFromXContent(BytesRestResponse.java:177)
	at org.elasticsearch.client.RestHighLevelClient.parseEntity(RestHighLevelClient.java:653)
	at org.elasticsearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:628)
	at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:535)
	at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:508)
	at org.elasticsearch.client.RestHighLevelClient.index(RestHighLevelClient.java:348)
	at org.phoebus.alarm.logging.ElasticClientHelper.indexAlarmConfigDocument(ElasticClientHelper.java:114)

```

I did run the new `create_alarm_template.sh`, and `GET /_template/*alarm*` confirms that elastic knows only about those three new index_patterns "*_alarms_cmd*", "*_alarms_state*" and  "*_alarms_config*".
I see that one index is auto-created, ihc_alarms_config_2018-12-14.

I think the problem is in `create_alarm_template.sh`.
Under both `"index_patterns":["*_alarms_cmd*"]` and `"index_patterns":["*_alarms_config*"]` is lists

```
  "mappings" : {  
    "alarm_cmd"
```

In the latter it should use a different mapping name, for example
```
  "mappings" : {  
    "alarm_cfg"
```

kasemir:
I'll fix that on the alarm_log_latch branch where I'm also adding a new "latch" field that allows finding those messages that indicate a new alarm, i.e. detecting when an alarm severity latched for the first time.

shroffk:
Crap...copy paste
So, in the create_alarm_index, I have correctly named the mapping type as "alarm_cmd" and "alarm_config"
Do  you mind if we keep the templates in sync with the create index and call in "alarm_config" instead of "alarm_cfg"

also, the client code also uses "alarm_config"

Sorry, I should have double checked the template script.

kasemir:
Yes, must be "alarm_config"  because that's used in the java code that writes. Works now with #519

PR #523, opened 2019-01-09
 XYPlot: Use update throttle to avoid out-of-order handling of X vs. Y

kasemir:
Import of RCP version, for kasemir/org.csstudio.display.builder#247


kasemir:
.. plus fix for https://github.com/kasemir/org.csstudio.display.builder/issues/486

PR #527, opened 2019-01-10
Alarm History Table to display all types of messages

tanviash:


kasemir:
I see no harm in merging this...

PR #533, opened 2019-01-15
AlarmConfigLogger update with unique stream name

shroffk:
@tanviash I suspect that the name space overlap in the KStreams might have resulted in the messages being randomly consumed by either of the two alarm micro services when run on the same machine.
Hopefully the following patch should address that.


kasemir:
Not sure how the streams work.
On the lower level, each 'consumer' must be in a group.
Messages to a topic are sent to consumers in the group.
You don't know with consumer will get the message, it'll be one in the group.

The client code used by the alarm table, tree, panel, .. use a UUID for the group to assert that each one gets all the messages:

https://github.com/shroffk/phoebus/blob/master/app/alarm/model/src/main/java/org/phoebus/applications/alarm/client/KafkaHelper.java#L60

Is the StreamsConfig.APPLICATION_ID_CONFIG the same thing as the "group.id" in the underlying KafkaConsumer?

shroffk:
Is the StreamsConfig.APPLICATION_ID_CONFIG the same thing as the "group.id" in the underlying KafkaConsumer?

I assume it should be similar

PR #534, opened 2019-01-16
#218 adding a readme for the alarm server

shroffk:


shroffk:
This PR needs to be squashed

Issue #542, opened 2019-01-23
Preference to suppress menu entries


kasemir:
Each site needs at least one site-specific product.
Maybe more: One for the accelerator control room, one for beam lines, one for office users.

When creating products, it's pretty easy to remove the complete functionality of a jar file, but a little harder to add a jar, since it needs to be listed in the classpath inside the manifest

-> Best to build a site-specific product with everything included, then delete the jars that are not required for the respective use case.

This, however, only allows deleting a complete jar file.

Propose to add a new preference option that's checked by the MenuEntryService to suppress SPI-provided menu entries:
```
# Hide the PV List and 'freeze' test
org.phoebus.ui/hide_spi_menu=org.phoebus.ui.pv.PVListMenuEntry,org.phoebus.ui.monitoring.FreezeUI
```

PR #543, opened 2019-01-23
Preference to hide menu entries

kasemir:
#542


kasemir:
Minor fix: PV Table 'info' dialog wasn't positioned

PR #554, opened 2019-02-06
Issue 136

shroffk:
Converting the greeting app to a diagnostic app, the diagnostic app shows the current selection/s

Updating channelfinder views with context menu support


kasemir:
Don't see a reason to not merge this, so going ahead.

Issue #555, opened 2019-02-07
Display Buider: Rule script not using alpha transparency channel


pavel-ch:
In my use case, I need to use colors including the alpha channel in rules. Current implementation generates the rule script without alpha transparency channel.

PR #556, opened 2019-02-07
Display Builder: Use color alpha channel in RuleToScript

pavel-ch:
Use color alpha transparency channel in RuleToScript. See #555


kasemir:
Thanks!

Issue #573, opened 2019-03-14
alarm config logger bugs


shroffk:
1. It fails to properly handle pvnames with special chars  which cannot be converted easily to file names

2. The alarm config logger only creates a restore script upon initialization, this should be changed so that each change in the git repo structure is also represented with an associated restore script.

PR #574, opened 2019-03-14
#573 fix regex for splchar in pvName & dump restore script for ea change

shroffk:
#573 @tanviash @berryma4


shroffk:
> like ':' -> '3A', '/' -> '2F', ..?

This would make the git repo / change history less readable.  

for example: 
SR3ARF{DEV3A3}

but... I agree with your point, we might have to find a way to distinguish these

Issue #577, opened 2019-03-18
ChannelFinder Client - logging


pavel-ch:
When ChannelFinder server is not available, the ChannelFinder client generates some console outputs with exceptions every time the Autocomplete is used. These outputs are generated with PrintStackTrace and System.out.println (ChannelFinderClientImpl:1113, 1115, 1120)
I suggest to use the logger for these outputs, to be able to control printing of outputs.
Suggestion follows in a pull request.

PR #578, opened 2019-03-18
 Logging in ChannelFinderClientImpl

pavel-ch:
Logging in ChannelFinderClientImpl see #577


shroffk:
I am sorry about this.
I started the exception handling and logger here but did not complete it.
https://github.com/shroffk/phoebus/commit/e452fe3d6ca2b5bbaa4a46c7bbfdd2f185b10c0f

also added a flag to deactivate CFProposalProvider when the service is not found
#586

Issue #606, opened 2019-03-29
Ctrl+Drag and Drop does not copy widget in Display Editor


pavel-ch:
There is already a DragDrop support in editor. What should from my perspctive work, too, is a Ctrl-Drag and Drop gesture for the widget copy operation in editor.
At the moment, the drag from a widget returns always the same string:
```
<?xml version="1.0" encoding="UTF-8"?>
<display version="2.0.0">
</display>
```
And thus the copied widget cannot be instantiated based on this string.
I can try to find the issue but tracking all the DnD support takes too much time to me. Can I get an advice where to start?

kasemir:
When you select one or more widgets, the `Tracker` is what's showing the blue selection rectangle with little handles to move and resize.

Check `org.csstudio.display.builder.editor.tracker.SelectedWidgetUITracker`, line 124, `setOnDragDetected`, which checks for the start of a drag operation.

Normally, when you just press-left-button-and-drag, we intend to move the selected widgets, which is why `setOnDragDetected` bails out right away unless `event.isShortcutDown()`, meaning you must be holding Ctrl (or Command on the Mac) to start a drag-and-drop instead of just moving the tracked widgets around.

If you raise the log level or change this source line to see what widgets are selected, you'll find that none are when the ctrl-drag is started: 
```
         // Allow 'dragging' selected widgets
        setOnDragDetected(event ->
        {
            if (! event.isShortcutDown())
                return;

            logger.log(Level.FINE, "Starting to drag {0}", widgets);
```

I think that's because Ctrl-click is also used to toggle selection of individual widgets.
Say you select a bunch of widgets, you can then Ctrl-click to add/remove widgets to/from the selection.

So select a widget, Ctrl-hold-down to start dragging
==> The Ctrl-click de-selects the widget, and you drag nothing.

kasemir:
The handling of Ctrl-Click to toggle the selection of widgets starts in `org.csstudio.display.builder.editor.tracker.SelectedWidgetUITracker` line 117:
```
        // Pass control-click down to underlying widgets
        addEventFilter(MouseEvent.MOUSE_PRESSED, event ->
        {
            if (event.isShortcutDown())
                passClickToWidgets(event);
        });
```

and a stack trace from there on looks like this:
```
at org.csstudio.display.builder.editor.WidgetSelectionHandler.toggleSelection(WidgetSelectionHandler.java:76)
[...]

PR #610, opened 2019-04-02
 Display Editor: Enable Ctrl-Drag to copy widgets 

pavel-ch:
See #606


kasemir:
That looks like a very elegant fix for the issue!

I'll look into changing `if (start_x < 0 || event.isControlDown())` into `..isMetaDown()` so Mac can use its Command key instead of Control.

kasemir:
.. meant `isShortcutDown`.

claudio-rosati:
The problem that macOS wants Alt (Option) not Meta (Command) for copy operations.

PR #659, opened 2019-05-06
Display fixes

kasemir:
kasemir/org.csstudio.display.builder#529


kasemir:
I'm looking at it.

kasemir:
It was worse:
The error message was meant to list the remaining widgets, but didn't.
Plus opening example/01main, then a subscreen, then going back to previous resulted in stack traces about being unable to locate parent item.

PR #687, opened 2019-06-12
Alarm Server & UI: Merge config and state topic

### LINKED ISSUE #675: alarm server configuration import failure
The alarm server fails to correctly import the alarm configuration file exported from the beast server.
1. disabled alarms are not loaded at all

2. the latch, annunciation, delay parameters are ignored and replaced with default
image

image

kasemir:
Fixes #675

### DIFF
--- a/app/alarm/Readme.md
+++ b/app/alarm/Readme.md
@@ -209,12 +209,12 @@ Messages sent by the UI also include the user name and host name to help identif
 
 _______________
 
-- Type `C:`, Config Topic:
+- Type `config:`, Config Topic:
 
 The messages in the config topic consist of a path to the alarm tree item that is being configured along with a JSON of its configuration.
 Example key:
 
-    C:/Accelerator/Vacuum/SomePV
+    config:/Accelerator/Vacuum/SomePV
     
 The message always contains the user name and host name of who is changing the configuration. 
 
@@ -250,7 +250,7 @@ format will only contain the used elements.
 
 For example, a PV that has no guidance, displays, commands, actions will look like this:
 
-    C:/path/to/pv : {"user":"user name", "host":"host name", "description":"This is a PV. Believe it or not."}
+    config:/path/to/pv : {"user":"user name", "host":"host name", "description":"This is a PV. Believe it or not."}
 
 - Deletions in the Config Topic
 
@@ -258,7 +258,7 @@ Deleting an item consists of marking a path with a value of null. This "tombston
 
 For example:
 
-    C:/path/to/pv : null
+    config:/path/to/pv : null
     
 This process variable is now marked as deleted. However, there is an issue. We do not know why, or by whom it was deleted. To address this, a message including the missing relevant information is sent before the tombstone is set.
 This message consists of a user name, host name, and a delete message.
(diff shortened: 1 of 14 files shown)


kasemir:
@shroffk , this fixes the alarm import problem in my tests.
Do you have a way to check?

As described in the readme, the "...State" topic is gone.
The main topic now handles configuration and state messages.
All message keys now start with a prefix:

`C:/some/path` -  config message   
`S:/some/path` -  state message   
`A:/some/path` -  command (action) message   
`T:/some/path` -  talk message   

The loggers will need to be updated, somewhat like this:
```
// We used to have
//    String path = record.key();
// Now the key is type:path
if (record.key().length() < 2)
{
    logger.log(Level.WARNING, "Invalid key, expecting type:path, got " + record.key());
    continue;
}

final String type = record.key().substring(0, 2);
final String path = record.key().substring(3);
```

shroffk:
I am going to update my deployment and check

shroffk:
> C:/some/path - config message
> S:/some/path - state message
> A:/some/path - command (action) message
> T:/some/path - talk message

why not just call them 

config:/some/path
state:/some/path
action:/
talk:/

do we really save a lot by excluding the characters...with the above example the messages are quite self explanatory even to people/tools/apps who don't know the key. 
In elastic it also makes searches a bit more intuiative
search for configuration changes to mypv:  
```config*mypv  ```
vs  
```C*mypv  ```

**Comment by kasemir** [comment]
A fixed-length prefix is nice because you can simply use `record.key().substring(0, 4)` to get the prefix, compared to having to split on ':', which means looping over the first couple characters.
So going to some fixed length scheme like

cfg:/some/path
ste:/some/path
act:/
tlk:/

would have next to no impact on speed and be a little more readable.
Still, compared to all the json parsing looking for a ':' doesn't matter, so OK, I'll update that to the strings you suggested.

**Comment by kasemir** [comment]
>  I'll update that to the strings you suggested.

Actually, I'll use  config, state, __command__, talk.

**Comment by shroffk** [comment]
Ok...things are getting better now.

I have been able to successfully load and reload the same configuration file consistently. The state updates are happening correctly and I do not see the previous null pointer errors.

There are some lingering issues with loading a configuration -> loading empty configuration
On the server the configuration is correctly clean up but the client still sees some of the nodes from the previous configuration
image

after reloading the empty configuration a few times I can finally reach an empty configuration

**Comment by shroffk** [comment]
I am fine with merging this and keeping the issue ticket open

**Comment by kasemir** [comment]
Load-empty-doesn't-clear-nodes-in-client is new, I need to look at that (next week).
I'm all for merging, but this will break service-alarm-*logger which need to strip the "xxxx:" prefix from the path resp. split "type:/path".
Still, merge and then update the loggers in a new pull request?

**Comment by shroffk** [comment]
Yes, 
I know that the associated services and tools will have to be re-written.

If I can get to working on that today I will merge it this afternoon else merge it monday and start updating the other pieces.

**Comment by kasemir** [comment]
OK, so feel free to merge this one.

**Comment by kasemir** [comment]
> .. else merge it monday

Are you ready to merge this?
I'll then look at the issue with loading an empty config.

**Comment by kasemir** [comment]
I can reproduce the issue related to importing an empty config.
When the alarm server is running and then a config with PVs in alarm is deleted, this is what happens.
Import deletes the items:
```
config:/root/path/PV1=null
config:/root/path=null
```

The alarm server sees that PV1 is deleted. So it updates the parent 'path' to no longer be in alarm:
```
state:/root/path=OK
```

Client sees that and adds 'path' back in. Server thus needs to follow each item removal with a `null` state update for that item.

PR #688, opened 2019-06-12
Formula PV

kasemir:
As discussed at developers meeting, a skeleton for formula PVs.

PVs now support "eq://expression" with shortcut "=expression".

Expression is numeric based on core-formula, with \`SomePV\` to include the value of PVs.

Example shows formula PV 
```
=`sim://sine` + `sim://ramp` + 15
```

plotted in comparison to its input PVs

eq

### DIFF
--- a/app/scan/model/pom.xml
+++ b/app/scan/model/pom.xml
@@ -27,6 +27,11 @@
     </dependency>
     <dependency>
       <groupId>org.phoebus</groupId>
+      <artifactId>core-util</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>org.phoebus</groupId>
       <artifactId>core-pv</artifactId>
       <version>0.0.1-SNAPSHOT</version>
     </dependency>
--- a/build.xml
+++ b/build.xml
@@ -9,12 +9,12 @@
     <ant target="clean" dir="core/framework"/>
     <ant target="clean" dir="core/util"/>
     <ant target="clean" dir="core/email"/>
+    <ant target="clean" dir="core/formula"/>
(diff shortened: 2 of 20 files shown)


pavel-ch:
The `phoebus/core/pv/src/main/resources/pv_formula_preferences.properties`  is referenced as `pv_eq_preferences.properties` from the code `class FormulaPVPreferences` -> Exception

kasemir:
Good catch. Don't understand how I missed that.
#691

pavel-ch:
Extremely useful feature.
Would it be possible to have a public access to formula PV's list of underlying PVs (or minimally PV names)? Example use case: custom Tooltip with some debug info etc.

kasemir:
Well, you'd have to check `if (some_pv instanceof FormulaPV) ..`.
Are you then suggesting to add something like `String getExpression()` and `Collection<PV> getInputs()` to the `FormulaPV`?

pavel-ch:
Yes exactly. The check `if (some_pv instanceof FormulaPV) ...` is legal because formula PVs are a) fixed part of Phoebus and b) something extraordinary having more PV inputs.
From my understanding, the `Collection<PV> getInputs()` on `FormulaPV` is enough. The expression string itself is clearly visible from the formula PV name.
I think this can be useful for formula debugging ui in any form. Some form of ui (tooltip?) can eliminate lots of user questions in the future.

kasemir:
OK, added that.
But go slow, this formula implementation is a skeleton for what we've discussed at the recent developer's meeting. A big missing part is an SPI mechanism for adding functions, based on requirements for handling the existing formula use of BNL NSLS-II BOY displays.

pavel-ch:
Thank you. And I understand, there is no hurry, this was for a long run. No direct activity atm.

PR #693, opened 2019-06-17
Scatter plot widget: First demo

kasemir:
@shroffk This is a basic scatter plot demo/preview/discussion point.

Not meant to be merged right away.

It's very similar to the XYPlot, but expects scalar X, Y and maybe error PV.
Adds their values to a circular bufffer, and then plots those buffer waveforms.

There is, however, no X/Y correlation logic. It simply adds each received PV update to the buffer.
In practice, there are several approaches to merge X and Y updates.
1) Have 'Y' update trigger a data point, using the then-current value of X
2) Have 'X' update ... of Y
3) Have a third 'Trigger' PV
4) Pick values where the time stamp matches with 0.01 seconds
5) Something entirely different
Trying to implement this in the scatter plot widget runtime will always end up missing some desired option.

My leaning would be to forget this, and instead move the correlation to an IOC. Whoever generates the samples would know how they correlate. That person can add a 'compress' record, configured as a circular buffer, for both the X and Y waveform to an IOC, updating them 'in sync' as required.
The area detector also has circular buffer support.
The point is that an IOC, which might be an intermediate one, generates the correlation waveforms.Those waveforms can be displayed with the existing XY plot.


kasemir:
scatter

PR #711, opened 2019-06-26
Strip chart: Add to examples

kasemir:


kasemir:
Similar to yesterday, maven still has trouble downloading itself:
```
[ERROR] Plugin org.apache.maven.plugins:maven-failsafe-plugin:2.22.0
or one of its dependencies could not be resolved:
Failed to read artifact descriptor for org.apache.maven.plugins:maven-failsafe-plugin:jar:2.22.0:
Could not transfer artifact org.apache.maven.plugins:maven-failsafe-plugin:pom:2.22.0
from/to central (https://repo.maven.apache.org/maven2):
Access denied to:
https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-failsafe-plugin/2.22.0/maven-failsafe-plugin-2.22.0.pom
```

shroffk:
It looks like the trouble is with travis containers n/w settings

shroffk:
Local maven builds continue to work fine...same with local jenkins maven builds

shroffk:
I think we can chalk this up to the cost of a free service.

kasemir:
The failures are rather random; you might be right about network settings.

Issue #712, opened 2019-06-26
phoebus jdk 9 support


shroffk:
Do we have a time line for when we plan to remove jdk 9 support for phoebus and thus updated travis build to only build again jdk 11.0.2

kasemir:
I'm OK with dropping J9 from the travis build now.

georgweiss:
No objection

PR #713, opened 2019-06-26
#712 removing oracle jdk 9 from CI

shroffk:
Considering we are moving eclipse based cs-studio to jdk11 it makes sense to remove the oracle jdk9 from CI build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: required
 language: java
 dist: trusty
 jdk:
-  - oraclejdk9
   - openjdk11
 
 before_install:


kasemir:
I don't think we're "done" moving Eclipse to Jdk 11, but the issue there is moving off Jdk 8. Going to Jdk 9, 10, 11 or 12 makes no difference, and we should not be stuck on Jdk 9, which is no longer available for download. Sorry if I merged this too quickly, though.

georgweiss:
Are you saying that Phoebus development in Eclipse does not work with JDK 11? I'm using IntelliJ...

kasemir:
Phoebus is great.
CS-Studio/RCP hasn't 100% moved to Java 9/10/11/12. There are still people who can't build it.

The reason for keeping Java 9 build tests in Phoebus was that maybe combining CS-Studio/RCP and phoebus with both on J9 is easier than with J11.
But for phoebus any J9-12 is fine,
and for CS-Studio/RCP any 9-12 is still problematic.

georgweiss:
But since 9-12 is problematic for CSS/RCP I see no advantage in keeping 9 for Phoebus.

willrogers:
Yes, ditch Java 9.

PR #716, opened 2019-06-27
Issue 715

shroffk:
@kasemir this might need a bit of reviewing :)

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/pva/PVAStructureHelper.java
+++ b/core/pv/src/main/java/org/phoebus/pv/pva/PVAStructureHelper.java
@@ -50,18 +50,25 @@ public class PVAStructureHelper
             final PVAData field = struct.get(name_helper.getField());
             if (field instanceof PVAStructure)
                 actual = (PVAStructure) field;
-            else if (field instanceof PVAStructureArray)
-            {
-                if(elementIndex.isPresent()) {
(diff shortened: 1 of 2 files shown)


kasemir:
Looks good!
We'll fix it if we find problems.

kasemir:
Doesn't handle this type of PV:
```
# softIocPVA -m N='' -d demo.db
record(calc, "ramp$(N)")
{
  field(SCAN, "1 second")
  field(INPA, "ramp$(N)")
  field(CALC, "A+1")
  field(LOW,  "1")
  field(HIGH, "100")
  field(LSV,  "MINOR")
  field(HSV,  "MINOR")
  field(FLNK, "waveform$(N)")
}
record(compress, "waveform$(N)")
{
  field(INP, "ramp$(N)")
  field(NSAM, "100")
  field(ALG, "Circular Buffer")
}
```

`pva://waveform` shows the array.
`pva://waveform[1]` gives this error:
```
Expected struct array for field value, got epics:nt/NTScalarArray:1.0 
    double[] value [55.0, 56.0...
	at org.phoebus.pv.pva.PVAStructureHelper.getVType(PVAStructureHelper.java:69)
	at org.phoebus.pv.pva.PVA_PV.handleMonitor(PVA_PV.java:69)
```

Fix would be to handle PVADoubleArray.

PR #727, opened 2019-07-04
Issue 720, Phoebus versions

georgweiss:
As discussed, changed from 0.0.1-SNAPSHOT to 4.6.0-SNAPSHOT. Actually 4.6.0 might not be the best choice... My understanding is that we start with 4.6.0 for all modules, but as modules evolve they will increment the version as needed.

Ant build works, but dependencies/ant_settings is left unchanged.

Using mvn release:prepare will require the user to:

- Force SNAPSHOT versions for epics and vtype
- Confirm release and next version for all artifacts, unless -DautoVersionSubmodules=true is specified on the command line. However, this is not an option when versions start to differ between modules.

Updated two pom files with version spring-boot-maven-plugin in order to suppress warnings from maven.

In the top level pom the maven local repository property has been commented out as that caused a problem with the maven release plugin. Maybe a bug with the plugin...

### DIFF
--- a/app/3dViewer/pom.xml
+++ b/app/3dViewer/pom.xml
@@ -4,19 +4,19 @@
   <parent>
     <groupId>org.phoebus</groupId>
     <artifactId>app</artifactId>
-    <version>0.0.1-SNAPSHOT</version>
+    <version>4.6.0-SNAPSHOT</version>
   </parent>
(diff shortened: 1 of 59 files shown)


shroffk:
This looks promising...I will try this out myself next week.

kasemir:
I'm OK with merging this.

Issue #726, opened 2019-07-04
Cannot save *.bcf file


MaticPogacnik:
Hi,

I am trying to create bcf file with predefined classes.
But when I try to save it i get an error that a file does not have suggested file extension. And I am unable to save it.

![bcferror](https://user-images.githubusercontent.com/5747624/60650202-f7dd8c80-9e43-11e9-9f6e-859b21ca9431.png)

It doesn't matter if I click Cancel or OK, either way I get dialog to select a name again and file is not saved.

Thanks,
Matic

kasemir:
I can confirm that for now you'll need to save it as a *.bob file and then rename.
Will fix this next week.

kasemir:
You can also try to fix this and submit a pull request.
At first glance, the required changes should be limited to the code around this section in https://github.com/shroffk/phoebus/blob/master/app/display/editor/src/main/java/org/csstudio/display/builder/editor/app/DisplayEditorInstance.java#L315:
```
 final File proper = ModelResourceUtil.enforceFileExtension(file, DisplayModel.FILE_EXTENSION);
```

Right now it always uses the *.bob `DisplayModel.FILE_EXTENSION`.
Should check `model.isClassModel()` to determine if maybe the bcf extension from 
`WidgetClassSupport.FILE_EXTENSION` is required.

kasemir:
Thanks for finding this and then fixing it.
Please continue ;-)

PR #728, opened 2019-07-05
#726 Allow to save class files

MaticPogacnik:
Great, that was nice advice, it works for me.
Is it done properly?

Maybe we should add an option to create new class file: Applications -> Display -> New class file ?


kasemir:
Can you check if you have tabs in there? Should all be spaces.

MaticPogacnik:
True, there were three tabs.
Sorry about that.

Issue #704, opened 2019-06-21
PhoebusApplication start and/or help issue with JFX11 and high log levels


kasemir:
OpenJFX11 has a bug in the WebView that causes it to fail when the log level is high,
https://stackoverflow.com/questions/53908982/new-webview-causes-java-lang-unsupportedoperationexception-not-implemented/53909677#53909677

When phoebus starts up and cannot find an existing layout to restore, it will open the "Welcome" page, which is based on a web view. So the overall application startup is then impacted, the saved layouts are not listed etc.

The 'help' pages won't open as well since they also use the web view.

Workaround is to reduce logging to WARNING.

kasemir:
Update to OpenJFX12 should also fix this.

According to https://github.com/javafxports/openjdk-jfx/blob/jfx-12/doc-files/release-notes-12.md#release-notes-for-javafx-12, `JavaFX 12 requires JDK 11`, so should be OK for anybody using JDK11 or 12.

PR #738, opened 2019-07-16
JavaFX12

kasemir:
Updates openjfx from 11 to 12.

@shroffk there is at this time no desperate need to update JFX.
There is no new API on which we depend.
JFX11 has a bug, #704, but there's a workaround.
https://github.com/javafxports/openjdk-jfx/blob/jfx-12/doc-files/release-notes-12.md#release-notes-for-javafx-12 mentions a fix for blurry fonts on Ubuntu and Debian which I cannot check.

This pull request is mostly to verify that everything still builds and runs with JDK12 and JFX12, check that we don't get behind.
The JFX12 description states that it required JDK11 or 12, and I've verified that it runs with both on Linux.
I found that the code also still runs with Oracle JDK 10, but then it uses the JFX 10 included in the JDK, ignoring the JFX12 jars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,7 @@ http://phoebus-doc.readthedocs.io
 
 
 ## Requirements
- - [JDK11 or later, suggested is OpenJDK](http://jdk.java.net/11).
-   (For the time being, Phoebus still builds with Oracle JDK 9 and 10,
-    but it must be a JDK that includes JavaFX).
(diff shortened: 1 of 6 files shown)


shroffk:
I have actually encountered the blurry text before, glad to hear that it is fixed in 12 and more importantly that we can still use javafx12 with openjdk11.

Since I do not have a reproduce able way to produce the bug, I am not sure how to test it either.

I will leave this PR open for a bit if that is ok

PR #746, opened 2019-07-23
Rename split horiz, vert into left/right, top/bottom

kasemir:
SNS end users tend to complain about "Split Horizontally" creating a vertical separator and vice versa.
Using "Split Left/Right" respl. "Split Top/Bottom" as another attempt.


kasemir:
Similar discussion that indicates that emacs, vim, .. also cannot agree: https://english.stackexchange.com/questions/293520/split-horizontally-or-vertically-which-one-is-which

PR #751, opened 2019-07-29
Allow compilation of pva lib with Java 8

kasemir:
Hugo submitted these changes to allow compilation of the pva lib with Java 8, which in turn allows calling it from Matlab.


kasemir:
Same basic `List.of` => `Arrays.asList` change allows all of `core.pv` to run with Java 8, which simplifies its use in display builder web runtime.

PR #768, opened 2019-08-12
Patches

shroffk:


kasemir:
How do I best disable the channel finder to avoid these error messages:
```
2019-08-12 12:38:21 INFO [org.phoebus.channelfinder.autocomplete.CFProposalProvider] 
Failed to create Channel Finder PVProposalProvider
...
Connection refused
```

shroffk:
Well, we could not include it in the product

shroffk:
excluding the app-channel-utility jar should get rid of the error...

kasemir:
I'm OK with including it.
Also OK to have localhost as default URL.
But have a way to disable, for example by setting the URL to empty.

PR #774, opened 2019-08-13
Ensure that the alarm services use UTC when recording alarm events

shroffk:
fixes the alarm messages source map to always use UTC


kasemir:
Finally got around to update our production setup, and it works great.
In the Kibana 'Discover' section, the "Table" now shows a message with local time of 11:52:.. as just that, "11:52:..". In the "JSON" tab you see that the raw data from Elastic is in UTC (Z), "2019-08-15T15:52:02.967Z".
So all as it should be, thanks!

PR #785, opened 2019-08-16
Collect JaCoCo code coverage metrics

andersharrisson-ess:
At ESS we have our own CI set up for Phoebus and would like collect code coverage.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,6 +50,25 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.4</version>
+        <executions>
+          <execution>
+            <id>prepare-agent</id>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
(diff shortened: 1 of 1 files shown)


kasemir:
Good idea.
How much does it add to the build time?
Is there a way to enable/disable based on properties or maven profile?

andersharrisson-ess:
> How much does it add to the build time?
Completely non scientific measurements gave me 54 secs extra build time with JaCoCo enabled.
> Is there a way to enable/disable based on properties or maven profile?
You can set jacoco.skip to completely skip the jacoco goals. If you prefer I can move the plugin to a profile. If so, do you prefer that profile to be active by default or not?

kasemir:
1 minute is OK, certainly if it can be diabled via `-Djacoco.skip=true`.
Based on the RCP experience I'm just paranoid that we jump from a basic compile time of 30 seconds to 45 minutes.

andersharrisson-ess:
Thank you for the merge, Kay. 
Running coverage metrics on every build may be unnecessary. If so, feel free to move it to a non-active-by-default profile. We'd appreciate if you give us a heads-up if you do, so that we can activate the profile in our CI pipeline.

PR #810, opened 2019-09-03
Unify MenuEntry and ToolbarEntry to both use Callable<Void>

kasemir:
@shroffk, this example would make both MenuEntry and ToolbarEntry use `Callable<Void>`, instead of ToolbarEntry defining its own very similar method.

Downside: Every implementation needs to return a Void, i.e. null.
Upside: Using existing interface.

We could also switch to Runnable instead of Callable<Void>, but I think it's nice for the implementation to allow exceptions which are then logged by the caller.
Having to add 'return null' is a lot simpler than having to catch & log exceptions.


shroffk:
> We could also switch to Runnable instead of Callable, but I think it's nice for the implementation to allow exceptions which are then logged by the caller.
> Having to add 'return null' is a lot simpler than having to catch & log exceptions.

+1

PR #814, opened 2019-09-05
Display Editor UI settings persistence

pavel-ch:
Persistence of UI user settings in the editor, namely:
    Grid, Snap and Show Coordinates presets
    Divider positions between Tree, Edit and Property panes.


kasemir:
Good, thanks!

PR #827, opened 2019-09-23
Add Gitlab CI pipeline

andersharrisson-ess:
At ESS we would like to run continuous integration on Phoebus and deploy the binaries to artifactory.esss.lu.se. For this we require the pipeline configuration to be included in the source project.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,11 @@
+stages:
+  - build
+
+build:
+  stage: build
+  image: registry.esss.lu.se/ics-docker/maven:openjdk-11
+  tags:
+    - docker
+  script:
+    - mvn --batch-mode -DaltReleaseDeploymentRepository=${RELEASE_REPOSITORY} -DaltSnapshotDeploymentRepository=${SNAPSHOT_REPOSITORY} clean install org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy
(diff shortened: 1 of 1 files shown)


shroffk:
Is the docker file being used here publicly available?

Could it be possible to include the docker file also, this would mean that other can use the gitlab pipeline too...

I think we might also want to consider how to organize the growing list of CI configuration files...
phoebus/ci/.tarvis
phoebus/ci/appveyor.yml
phoebus/ci/.gitlab-ci.yml
...

andersharrisson-ess:
Thank you for the comment, Kunal.

The Docker registry is publicly available as well as the Dockerfile: https://gitlab.esss.lu.se/ics-docker/maven/blob/openjdk-11/Dockerfile

The pipeline is somewhat generic, but it has explicit configuration to publish the resulting builds to ESS Artifactory, which ofc needs correct credentials. Same thing with the SonarQube analysis configuration. This means that the pipeline, in its current state, is quite useless to anyone but ESS.

For the plethora of CI configuration files, I would not make a plan for it. I think the community should have only one pipeline (I would suggest Travis) that pushes the jars to a global repository (i.e. Maven Central). However, until this is in place, there must be a way for all stakeholders to build the jars themselves. As soon as the jars are available globally, I will gladly submit a PR to delete the Gitlab pipeline.

PR #828, opened 2019-09-24
Csstudio 1100

georgweiss:
Fixed dependency to vtype-json. Also removed some unused code.


kasemir:
I suggest we merge this.

PR #832, opened 2019-09-25
Gitlab pipeline: Reference external pipeline configuration

andersharrisson-ess:
I decided that it's better to keep the pipeline configuration in an ESS Gitlab repo and just reference it from the .gitlab-ci.yml in the Phoebus repo.
This is to avoid further PRs for changes on the Gitlab pipeline.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,11 +1,2 @@
-stages:
-  - build
-
-build:
-  stage: build
-  image: registry.esss.lu.se/ics-docker/maven:openjdk-11
-  tags:
-    - docker
-  script:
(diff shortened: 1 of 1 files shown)


kasemir:
Not sure if it's good to have site-specific links like ...esss.lu.se in the shared, common repo, but this PR doesn't fundamentally change the situation, just refactors it a bit, so I'd go ahead and merge.

shroffk:
Well, the first PR was referencing ess links which were publicly accessible.
In anycase... I will accept this PR with the hopes that at some time in the future we are able to provide a generic solution.

kasemir:
The 'include' in this one is still accessible, just refactors it a bit, so I'm fine with it. But agree that site-specific code shouldn't be in the shared repo.

andersharrisson-ess:
Thank you for the merge.

I fully agree that site specific content should not be in the community project, but until there is a generic CI pipeline set up, I appreciate that we can have our configuration in the repo.

PR #847, opened 2019-10-07
Support changing fonts in rules

krisztianloki:
Changing the font requires the usage of WidgetFont (just like WidgetColor for colors); this PR implements that.


kasemir:
Thanks!

PR #848, opened 2019-10-07
Warn if script PV has unresolved macros

krisztianloki:
Warn if a pv that is set to trigger a rule or script contains unresolved macros.


kasemir:
Thanks!

Issue #836, opened 2019-09-26
Recursive copy of directory in file browser


georgweiss:
If a directory in the file browser is dragged and dropped onto itself or on any other node (directory or file) in the same source directory, a recursive copy operation is  started. The pain that follows is worse on Windows than on Mac OS. 

Will fix it.

kasemir:
By default, the file browser copies in response to a drag/drop.
That's because early on I tended to often drag files from the browser to a terminal or text editor, just to get the filename. .. but the file browser then deleted the file, thinking it had been 'moved'.

Compared to other file browsers, the default should be move (which now requires pressing ctrl or shift, don't remember), and the file browser would notice if the drop target didn't do anything, so it would NOT delete files when you drag them into a text editor, terminal, ...

Maybe you can look at that as well.

georgweiss:
Sure I can take a look, but this will boost the scope a bit and I may need to push this to after ICALEPCS. If we agree to change the default behavior from copy to move, I would advocate it since - as you suggest - that is the default behavior on Mac, Win and Linux file browsers.

georgweiss:
Fixed.

PR #855, opened 2019-10-14
Issue #836

georgweiss:
Fixes recursive directory copy if drop target contains source directory.
Also: default drag-n-drop behavior is move, modifier key (alt/option on Mac, ctrl on all other) press changes to copy.


kasemir:
In `FileTreeCell` `isModifierKeyPressed()`, consider using `PlatformInfo.is_mac_os_x` to check if it should use Ctrl or Alt, because that's only evaluated once and also used in other places to check for the platform.

kasemir:
While the move/copy works, the determination might be too early.
Right now I need to hold Ctrl before starting a drag to achieve a file copy operation.
In other file browsers, you can start the drag operation and then press/release Ctrl to update the drag mode.

georgweiss:
I wasn't aware of the possibility to modify transfer mode in file browsers after drag-n-drop gesture has been started. However, I cannot find a way to capture key press events in JavaFX after drag-n-drop has been initiated. Looks like key events are suppressed, or delivered at a later point in time when they are no longer of interest for the drag-n-drop operation.
Tested on both Mac and Windows with same result.
If anyone knows how to capture key press events during a drag-n-drop operation, I'm all ears.
As for using PlatformInfo.is_mac_os_x, I will create a new pull request.

georgweiss:
I wasn't aware of the possibility to modify transfer mode in a file browser after drag-n-drop has been started. 
However, it seems this is not possible in JavaFX. I am unable to capture key press events after drag-n-drop has been initiated, regardless of where I register the key event listener (TreeView or TreeView's parent or scene). Verified this (odd) behavior on both Mac and Windows.
If anyone knows a way to capture key events during an ongoing drag-n-drop operation, I'm all ears.
As for using PlatformInfo.is_mac_os_x, I will create another pull request.

kasemir:
The windows, mac and Linux file browsers somehow support this, which I find helpful since I can never remember which combination of Ctrl, Shift, Alt, Command causes move, copy, link.
So I start a drag and then keep trying key combination, watching how the drag icon updates until it reflects what I want.

You're right that it's not obvious from the API how to do that. It needs to be done in the `setOnDragOver` handler, but it receives a DragEvent with coordinates and Nodes, no keyboard info.

Still, this seems to work:

1) In `setOnDragDetected`, don't lock the TransferMode to COPY or MOVE. Instead, use COPY_OR_MOVE to allow both.
2) In setOnDragOver, when you print the event.getTransferMode(), you'll find that it's either COPY or MOVE, depending on keys which are currently held. On Linux, I get COPY by default, and MOVE when I press/hold the shift key. Calling `event.acceptTransferModes(event.getTransferMode())` in the drag over handler causes the drag item to update.
3) In the drag done handler, one then needs to check what the transfer more was, as before.

georgweiss:
The default transfer mode in all file browsers is move. When testing your proposal the default transfer mode is instead copy, which to me introduces a confusion (JavaFX bug?). Further, on neither of my test boxes (Mac, Windows and CentOS) I am able to modify the mode with a key press and setting the mode calling event.acceptTransferModes consequently does not alter the mode.

Maybe I'm missing something as we have inconsistent test results.

georgweiss:
Correction: on Windows I can see a transfer mode change, it goes from move to copy (which is consistent with the file browser). To add to the confusion, on Mac I also can see a mode change, but it's from copy to move, and not for the alt/option button (which is the modifier button in Finder), but using the command button.

All in all I find i to difficult to replicate the exact file browser behavior on all platforms.

kasemir:
When I `startDragAndDrop(TransferMode.COPY_OR_MOVE)` in `setOnDragDetected` and then check the `event.getTransferMode()` in `setOnDragOver`, I get this:

Linux: Defaults to COPY. MOVE with 'shift'.

Mac: Defaults to COPY, then `null` with control or command.

So it only works as one would think on Windows.
On Linux it's swapping move/copy.
I tried calling `startDragAndDrop(TransferMode.MOVE, TransferMode.COPY)` to first request MOVE, then COPY, but the order doesn't seem to matter.

On Mac the default I see is also COPY, but when pressing modifier keys, I get `null`, not MOVE. Tried `event.acceptTransferModes(mode == null ? MOVE : mode)` but that has no effect, no 'drop' happens.

Sigh.

So locking the MOVE vs. COPY when starting the drag is the best option.
Good news is that the drag icon does show a '+' or not, so one can see what the started drag will do and abort if that's not desired. Sure would be nice to change the operation in flight as possible in other file browsers, but seems impossible without hacks.

georgweiss:
OK, good we arrive at the same conclusion, which is no surprise as we have basically the same tests.
Let's blame JavaFX and move on.

PR #856, opened 2019-10-14
ant: add core-launcher

kasemir:


kasemir:
@shroffk apart from the  travis errors related to installing openjdk11 I'm unsure about the EPICS libs.
Some files call for
jca-2.4.3.jar, epics-util-1.0.1.jar, epics-pvdata-6.1.3.jar, epics-pvaccess-5.1.3.jar, vtype-1.0.1.jar, others for the same with `-SNAPSHOT`.
Which one should it be right now?

kasemir:
@shroffk We also need some update on how to run the product from the IDE.

Opening the Launcher class and then calling "Run" no longer works because of this line:
```      LogManager.getLogManager().readConfiguration(Launcher.class.getResourceAsStream("/logging.properties"));
```

Those resources are in the product class.
Should I move the default logging.properties to the core-launcher module so it's possible to run from the IDE again?

kasemir:
I just can't get it to work in the IDE.
The product has no code, so you can't "Run" it in Eclipse.
When I move the logging.properties resources to core-launcher, it runs, but you get an empty product since only the phoebus-product has dependencies on all the core-* and app-* modules.

What works: Add this type of main class to the phoebus-product:
```import org.phoebus.product.Launcher;
public class Main
{
    public static void main(String[] args) throws Exception
    {
        Launcher.main(args);
    }
}
```

Now you can open that and "Run" in Eclipse, the dependencies get all the core- and app- modules on the classpath and it works.

kasemir:
Nevermind, no need for a product/Main.java. Added info to Readme on how to run the Launcher in Eclipse with the product dependencies.

kasemir:
Travis keeps failing at the JDK11 install, but otherwise mvn, ant and the Eclipse way of building/running work again:

* Cleaned up .classpath files to reflect deleted source or resource dirs
* Moved logging.properties to core-launcher because Launcher loads it
* Added core-launcher to ant build
* Updated readme to show how one runs the product now that its 'main' class moved to core-launcher

PR #863, opened 2019-10-17
Creating a profile which generates the javadocs for phoebus

shroffk:
Adding support for creating javadocs.

@kasemir @georgweiss I would like for this profile to be active by default but this will produce a lot of warnings and errors pointing out places where javadocs are missing or we are using wrong tags and links


shroffk:
image


image

kasemir:
It won't be done for a while, but I'm OK with merging this so we all see the warnings and can fix them as we touch a piece.

PR #874, opened 2019-10-18
More version updates

kasemir:
.. pulled in via maven, somehow


shroffk:
I don't think "somehow" is fair... The updated version of the elastic library had newer dependencies.

As much as you really hate maven...I hate manually having to maintain dependency trees. Finding out all the dependencies...adding them to some file...

The CI systems which use maven properly did not catch this..."problem".

Anyway....In the future we will manually keep updating the list.

kasemir:
Trying to describe the target platform idea in new dependencies/readme, #875
As mentioned in there, maven is indeed convenient for handling the deps for you.

PR #879, opened 2019-10-23
Updated save-and-restore pom to properly use resoure filtering, updat…

georgweiss:
…ed dependency pom to include vtype-json


kasemir:
Good, thanks!

Issue #679, opened 2019-05-23
Display Editor - enhancements


pavel-ch:
Working on the demo screen, some hints from the work with Display Editor on Windows OS:
 - [x] PolyLine Editor: line points placing should follow grid constraints when the Grid Alignment is On
 - [x] Group behavior
    a) It is necessary to be able to start "lasso" selection gesture from inside the group. Currently it is very impractical in the case of heavy Groups usage - having to find a piece of free display to start lasso.
    b) Possibility to add a new widget to the Group from the Widgets panel - currently not possible
 - [ ] Macroized Property: 
    a) especially for the PV Name, it would be nice to add some kind of interactive support for "macroizing" the string using current macros ("reverse" macro application)
    b) enhance the PV completion, allowing work with macroized PV Names, if possible
 - [x] Better SAVE function availability:
    a) sometimes Ctrl-S does not work from unknown reason (perhaps focus problem), also the context menu must be clicked twice before it appears
    b) Place SAVE button to Editor toolbar and/or Editor display context menu would help

Thank you for your opinions.

kasemir:
All good. Specifically, I can confirm the Save-As issue, have reopened associated ticket. Would be very helpful to get some reproducible recipe, since I can get into the situation where save-as is disabled, but then can't remember how I got there.
I won't get to do anything about any of these in the next 2 weeks, though.

pavel-ch:
Fine. Meanwhile, I can look into items and send usable results as PRs (if available)

DorianHas:
I would add to this list the ability to rotate widgets. This could be very useful when modelizing beam line.

It could be associated to Ctrl-R or changed in Position section in  widget properties.

kasemir:
Re Group behavior, starting a 'rubberband' (lasso) selection from inside the group, that's now possible by holding the Alt key. Ordinarily, trying to 'rubberband' within a group doesn't work because the initial click selects the group. Now, holding Alt, the initial click is not selecting the group but instead allows starting a rubberband selection of items within the group.

kasemir:
Re "add a new widget to the Group from the Widgets panel - currently not possible":
I tried dragging a new widget from the list of widgets into a group. Works fine. As widget is dragged out of the list of widgets and over a group, the group is highlighted. When then letting go, the widget is added to that group.

kasemir:
Re Group behavior, starting a 'rubberband' (lasso) selection from inside the group, 
that's now also possible for the Tabs widget:
[...]

PR #880, opened 2019-10-23
DisplayEditor Ctrl-D gesture duplicate Widgets

pavel-ch:
See #679. Still some issues with Groups, will be tuned gradually


kasemir:
> Still some issues with Groups, will be tuned gradually

Would it be easier to implement that by basically calling "copy", then "paste"?
Or does that also still have group-related problems?

pavel-ch:
I explicitly do not want to erase the content of the clipboard, this is the principle why this shortcut exists -> cannot use copy-paste. The problem with groups comes for exotic multiselections between inside and outside of the group. It is a marginal issue, I think the problem is the same for Copy-Paste.

PR #881, opened 2019-10-23
DisplayEditor Ctrl-D shortcut / Group issues compensated

pavel-ch:
Turned to be simple at the end... see #679


kasemir:
Thanks!

PR #890, opened 2019-10-30
Csstudio 1143

georgweiss:
This is a fix to allow a zero-length string as a macro value. This is possible in BOY, so suggestion is to  align in Display Builder.


kasemir:
Needs corresponding update to ResourceParserTest, where test expected null, now gets ""

Issue #896, opened 2019-11-01
Display Builder - small enhancements


pavel-ch:
Into this Issue, I will group small contributions in order to enhance Display Builder and its infrastructure. These suggestions appear when we slowly start working on displays.

First issue is about JFX dialogs: on Windows, they show a default icon which is ugly. Setting the Owner to dialogs the icon is inherited from the stage. PR follows.

kasemir:
On Windows, how do you launch the program? Something like https://github.com/shroffk/phoebus/blob/master/phoebus-product/phoebus.bat works, and seeing the console messages in the cmd.exe window can be helpful, but it has similar issues to the "ugly icon" that you just fixed for dialogs:
End users tend to not notice the *.bat file. The file ending is by default hidden, and there's a plain batch file icon, no CS-Studio icon. At runtime, you have that extra window, which can be useful to see the console messages, but it's also distracting and has some default icon.

pavel-ch:
At the moment, this is not solved. Running the product usually not from bat but from sh file using bash.exe from git installation, effect is the same.
We can create a shortcut .lnk file like this with javaw which does not open command window (tested now):
target: "C:\Program Files\Java\jdk-11.0.4\bin\javaw.exe" -jar product-4.6.0-SNAPSHOT.jar  --add-modules=ALL-SYSTEM
Start in: the installation directory
Icon: we must supply icon in a supported format .ico
![image](https://user-images.githubusercontent.com/771816/68061250-360c8f00-fd04-11e9-8cc9-7eed296fd5d6.png)

pavel-ch:
Shorcut can be generated by powershell on an installation step - to fix the absolute paths inside, see for example:
https://dotnet-helpers.com/powershell/create-shortcuts-on-desktops-using-powershell/

pavel-ch:
**Display Builder Open Display Action: New Tab into a named pane**
In my use case, I need a way how to ensure single tab in a named pane (closing all other tabs before opening a new one). I have an implementation in a site specific widget.
I can imagine that this feature can be interesting for others, too.
So what I suggest is to extend the `org.csstudio.display.builder.model.properties.OpenDisplayActionInfo.Target` on SINGLE_TAB enum member, extend the `org.csstudio.display.builder.representation.javafx.ActionsDialog` with this radio button and implement handling in `org.csstudio.display.builder.representation.javafx.JFXStageRepresentation`.
The handling looks principially like this (example from my code, will be simplified):
```
[...]

PR #897, opened 2019-11-01
Display Builder JFX dialogs: set dialog Owner to get the right dialog icons on Windows

pavel-ch:
See #896


kasemir:
Neat, small change that makes is overall look better, thanks

PR #901, opened 2019-11-06
Issue 900

shroffk:


kasemir:
I was just about to do this, then answered some other email, and now you did it. Thanks!

PR #905, opened 2019-11-06
ES sniff feature for logging-ui

tanviash:
The change is similar to the one I made on alarm logger side. This is to enable ES sniff feature for logging-ui client so it can retrieve data from other cluster ES nodes in case of failover.

### DIFF
--- a/app/alarm/logging-ui/pom.xml
+++ b/app/alarm/logging-ui/pom.xml
@@ -16,7 +16,12 @@
     <dependency>
       <groupId>org.elasticsearch.client</groupId>
       <artifactId>elasticsearch-rest-high-level-client</artifactId>
-      <version>6.3.1</version>
+      <version>6.4.2</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


tanviash:
@shroffk @kasemir Please review this code change.

kasemir:
Sorry, I don't see anything new.
It needs a basic
```
  if (sniffer != null)
    sniffer.close();
```

Also, instead of `prefs.get("es_sniff").equals("true")` you can use `prefs.getBoolean("es_sniff")`

kasemir:
Ah, I didn't know how to handle the 'review' feature. Had entered something, but failed to press some 'commit' button so I guess you never saw my original comment.

kasemir:
Check the PreferencesReader. Instead of `Integer.valueOf(prefs.get("es_port")))))` you could use `prefs.getInt("es_port")`, and for `prefs.get("es_sniff").equals("true")` there's `getBoolean("es_sniff")`.
I can't test the cluster failover, so no idea if the change does anything to support it, but otherwise it looks OK to me.

Issue #620, opened 2019-04-04
Scaled Slider increment ignored?


kingspride:
Hi,

is it possible to change the scaled slider in a way that it doesn't jump directly to the clicked location, but instead obeys the increment setting and only changes by increment size in the clicked direction?

I'm asking because my facility is currently converting old opi files and CSS into bob files and phoebus. from the old environment, they are used to clicking somewhere left or right from the sliders knob to change the value in (small) steps.

what I want:
![slider-old](https://user-images.githubusercontent.com/45198837/55548133-1b712980-56d3-11e9-8d6a-e78f94045176.gif)

what phoebus currently does:
![slider](https://user-images.githubusercontent.com/45198837/55547664-0f389c80-56d2-11e9-81bb-690193a0c309.gif)

shroffk:
Could this be a widget property so that the behaviour can be configured by the OPI designer.

kasemir:
This behavior is hardcoded inside the JavaFX Slider, i.e. in the graphics library that we use with the Display Builder, no matter if Eclipse/RCP or Phoebus.
It's this code inside `com.sun.javafx.scene.control.behavior.SliderBehavior`:
```
    /**
     * @param position The mouse position on track with 0.0 being beginning of
     *        track and 1.0 being the end
     */
    public void trackPress(MouseEvent e, double position) {
        // determine the percentage of the way between min and max
        // represented by this mouse event
        final Slider slider = getNode();
        // If not already focused, request focus
        if (!slider.isFocused()) slider.requestFocus();
        if (slider.getOrientation().equals(Orientation.HORIZONTAL)) {
            slider.adjustValue(position * (slider.getMax() - slider.getMin()) + slider.getMin());
        } else {
            slider.adjustValue((1-position) * (slider.getMax() - slider.getMin()) + slider.getMin());
        }
    }
```

I can even argue that's a good idea, because that way a click has a predictable response: The slider moves to the clicked position.
If a click-and-hold results in a sequence of updates, how fast the slider moves really depends on how busy the computer is at the time, plus you create one 'write' to the PV per update, and when each written value reaches the IOC and takes effect is yet another question.

I do understand the desire to adjust some value in "steps" while interactively monitoring the response.
With the slider, that's possible by pressing the left/right respectively up/down cursor keys. Each key press adjusts the slider by the 'increment' value which can be configured on the widget in edit mode or at runtime via the context menu.

kingspride:
[...]

PR #906, opened 2019-11-06
Slider: Click on track  to inc/dec instead of jumping to that value

kasemir:
Makes slider behave like the one in BOY
#620


georgweiss:
Is the behavior the default one?

kasemir:
The default behavior is now the same as BOY, as original ticket requested.
I think it would be misleading to have a widget property for this behavior, because then users end up with screens where identically looking sliders behave in different ways.
It could be a preference setting, i.e. each site can decide which behavior they want for their sliders on all their displays.

PR #910, opened 2019-11-07
Slider: Preference to select behavior

kasemir:
The inc/dec behavior that's compatible with EDM, BOY is the default, because it avoids large value jumps from accidental clicks on the slider, for example to activate it for keyboard tweaks.
Jump-to-clicked-value is available via a preference setting for sites which prefer that behavior.


kasemir:
#620

Issue #367, opened 2018-10-19
Drag/drop doesn't work via SSH (Dock tab, Display editor, ...) with GTK3 on Linux


kasemir:
Dragging tab panes around between windows doesn't work when working on a remote Linux system via ssh.
Tabs can be dragged, but instead of dropping in the desired location, a new window is created because the drag-over and drop events are not received.

kasemir:
Drag/drop works when using `-Djdk.gtk.version=2`, and fails with `-Djdk.gtk.version=3`.
Details may depend on setup of computers used on server and client.
In the tested case both are RHEL 7.5.

PR #919, opened 2019-11-08
JavaFX 13.0.1

kasemir:
Updates JavaFX libraries to the current one, 13.0.1.

Requires JDK 11 or later, i.e. no change from JFX 12.

Big news: On Linux, this fixes the GTK3 drag-and-drop problems, #367, where tabs could not be moved, widgets not be dragged in the display editor, unless one ran with `-Djdk.gtk.version=2`.
Now both GTK2 and GTK3 seem to support drag & drop.
https://bugs.openjdk.java.net/browse/JDK-8211302

Otherwise found no API changes, everything still compiles.

https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-13.md


georgweiss:
Go for it!

PR #923, opened 2019-11-08
Destination named pane URI hint into DisplayRuntimeApplication

pavel-ch:
See #896 I just tried and it works fine


kasemir:
Note that we already handle "target=pane_name" for command line arguments, https://github.com/shroffk/phoebus/blob/master/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java#L804

So this should also use "target=..." instead of "pane=..." to be consistent.

Issue #917, opened 2019-11-08
ContextMenuEntry <V> 


kasemir:
The ContextMenuEntry takes a `<V>` type parameter, which is not documented, and not used.
It's returned by callWithSelection(), but no caller uses the result.

Suggest to remove `<V>` and turn callWithSelection() into void method.

PR #942, opened 2019-11-14
ContextMenuEntry: Cleanup, document

kasemir:
Removes the unused `<V>` type parameter.
Adds `Class<?>` to avoid `rawtypes` warning.
Documents how to contribute and how to use context menu entries.

#917


kasemir:
@shroffk can you check?
This removes the unused `<V>` in `ContextMenuEntry<V>`, adds documentation, removes `rawtypes` warning, which I think is progress for such a key SPI item.
On the other hand, if you had certain plans for actually using the `<V>`, we shouldn't remove it but instead, well, use it.

PR #960, opened 2019-11-19
Do not use getLocalPath()

krisztianloki:
It is deprecated and not needed anymore


kasemir:
It was left in place so that the source code for the Eclipse/RCP-based display builder and the Phoebus-based display builder can remain similar.
Are you saying we should stop considering that?

kasemir:
Maybe we should, and if somebody asks, it was your idea via this PR ;-)

PR #963, opened 2019-11-22
CSSTUDIO-1166 Improvements to alarm guidance handling

georgweiss:
A few changes to how guidance texts are handled in alarm views (table, tree).

- A guidance context menu item is included only for a "leaf" alarm item, and only if one single item is selected.
- The text of context menu item for guidance is always the same, i.e. not taken from the title of the guidance.
- When the guidance context menu item is selected, the pop-up will show all guidance texts configured for that alarm.

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmContextMenuHelper.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmContextMenuHelper.java
@@ -93,12 +93,12 @@ public class AlarmContextMenuHelper
         //
         // Considered tool tip, but unclear how to attach TT to menu item.
         final AtomicInteger count = new AtomicInteger();
-        for (AlarmTreeItem<?> item : selection)
-        {
-            addGuidance(node, menu_items, item, count);
-            if (count.get() >= AlarmSystem.alarm_menu_max_items)
-                break;
+
+        // Add guidance menu item only if one single AlarmTreeItem is selected
+        if(selection.size() == 1 && selection.get(0).getGuidance() != null && !selection.get(0).getGuidance().isEmpty()){
+            addGuidance(node, menu_items, selection.get(0));
         }
+
         added.clear();
         count.set(0);
 
@@ -157,21 +157,9 @@ public class AlarmContextMenuHelper
 
     private void addGuidance(final Node node,
                              final List<MenuItem> menu_items,
-                             final AlarmTreeItem<?> item,
-                             final AtomicInteger count)
+                             final AlarmTreeItem<?> item)
     {
-        for (TitleDetail guidance : item.getGuidance())
-            if (added.add(guidance))
-                if (count.incrementAndGet() <= AlarmSystem.alarm_menu_max_items)
-                    menu_items.add(new ShowGuidanceAction(node, item, guidance));
-                else
-                {
-                    menu_items.add(createSkippedEntriesHint(node, "guidance messages"));
-                    return;
-                }
-
-        if (item.getParent() != null)
-            addGuidance(node, menu_items, item.getParent(), count);
+        menu_items.add(new ShowGuidanceAction(node, item));
     }
 
     private void addDisplays(final Node node,
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/ShowGuidanceAction.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/ShowGuidanceAction.java
@@ -7,6 +7,9 @@
  *******************************************************************************/
 package org.phoebus.applications.alarm.ui;
 
+import javafx.scene.text.Text;
+import javafx.scene.text.TextFlow;
+import javafx.scene.web.HTMLEditor;
 import org.phoebus.applications.alarm.model.AlarmTreeItem;
 import org.phoebus.applications.alarm.model.TitleDetail;
 import org.phoebus.ui.dialog.DialogHelper;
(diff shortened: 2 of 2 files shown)


kasemir:
Why do you hide the guidance when multiple elements are selected?
This was done on purpose: When you select N elements, you get the guidance for those N elements listed in the context menu. When you then open one of them, the title shows from which item it originates.

kasemir:
>  A guidance context menu item is included only for a "leaf" alarm item

No!  You can assign guidance to any element in the hierarchy, and it then needs to show up for that element.
Assign guidance "Common" to /Accelerator, and "Vacuum Info" to /Accelerator/Vacuum  -> Both show up in the context menu of  /Accelerator/Vacuum and any item below.

kasemir:
> The text of context menu item for guidance is always the same, i.e. not taken from the title of the guidance.

 Nooo!  Users configure that on purpose, like "Overview", "Procedure Detail", ...  Just calling everything "Guidance" means you end up with 10 entries all called "Guidance" and you don't know which one is the "Overview", which one is the "Procedure", ..

georgweiss:
Because I find it confusing. Suppose you have one guidance text for a number of alarms, all with the same title (which is possible as the titles are free text defined by others). You will then have X menu items with the same text:
Screen Shot 2019-11-22 at 08 50 14

georgweiss:
This is my proposal:
Screen Shot 2019-11-22 at 14 26 06

kasemir:
If the user selects items, they see the guidance for those items.
To prevent the context menu from exploding in case everything was selected, there already this:
```
if (count.incrementAndGet() <= AlarmSystem.alarm_menu_max_items)
...
else
    menu_items.add(createSkippedEntriesHint(node, "guidance messages"));
```

kasemir:
Well, OK, one menu entry, then join all the texts in there.
But we already have alarm configurations that have been used for a decade with separate guidance entries, and operators can pick "Procedure.." from the list to see just that, not one "Guidance" link and then you scroll through to find what you wanted.

kasemir:
Just like displays.
Do you want to combine related display links into just one "Displays" and then open all?
Or actions, just one "Action" link to execute all???

georgweiss:
Let's talk about alarms since that is the topic here.

kasemir:
I'm talking about the alarms context menu, which has entries for guidance, related displays, actions.
There all handled the same way: We list what's configured in the context menu.
You want to change that for the guidance.
I'm pointing out that it makes no sense, and would be even less useful for the display links and actions.

kasemir:
Here's where you configure guidance, displays, commands:

ctx

They are all listed in the context menu, based on the hierarchy, i.e. you see everything for the selected items above. You propose to hide guidance for intermediate items, and join it for the rest.

That clearly wouldn't make sense for displays and commands, and I don't think it does for guidance, either.

kasemir:
Note how the code already detects duplicates for related displays.
If you select N items, and they all inherit the same "Overview" display link, then you only see that display link once. You could do something similar for guidance: If the same guidance is already in the context menu, don't duplicate it.

georgweiss:
OK, I realize I did miss the point of inheritance. Let me rework this a bit. But I insist on not including guidance items in the context menu when multiple nodes/rows are selected.

kasemir:
The exact same inherited guidance is already only shown once in the context menu.
There might be room to merge more: If N items all contribute guidance with the same title, then only show that title once in the context menu, joining the text.
But when the titles differ, then just list them as now.
Hiding guidance of selected items is wrong.
Smartly merging them (same title) is certainly good and currently not implemented.

kasemir:
> But I insist on not including guidance

Do you also insist on not including displays? On not including commands?
When the user selects N items, who are we to decide that maybe they didn't mean to select N items?

georgweiss:
Take a look at the screen shot above, which is data from ESS, not just my dummy test data. User selects four items in the tree, context menu shows only three, all with the same title since that's how they are configured. When one of the items is selected, a guidance text is shown for...user doesn't know until the pop-up is rendered and needs to compare with the PV name. And

kasemir:
> all with the same title 

Right, that's why I suggested that you work on joining texts for guidance that has the same title, then presenting only one context menu entry for that title.

georgweiss:
@kasemir , can you please send me an alarm configuration and a screen shot of how that is visualized in the context menu?

kasemir:
Right, it was meant to be used in a certain way.
Here's an example configuration:
```
<?xml version="1.0" encoding="UTF-8"?>
<config name="Accelerator">
  <component name="Vacuum">
    <guidance>
      <title>Vacuum Contacts</title>
      <details>Fred: 123-456-7890</details>
    </guidance>
    <display>
      <title>Vac. Overview</title>
      <details>vacuum.bob</details>
    </display>
    <component name="Sector1">
      <guidance>
        <title>Sector 1 Info</title>
        <details>Contact normal conducting RF personnel for longer vacuum issues</details>
      </guidance>
      <display>
        <title>Sec.1 Vacuum</title>
        <details>vac_sec1.bob</details>
      </display>
      <pv name="loc://Pressure(0)">
        <description>Sector 1 Vacuum Pressure</description>
        <enabled>true</enabled>
        <latching>true</latching>
        <annunciating>true</annunciating>
        <guidance>
          <title>Pressure Guidance</title>
          <details>Check pressure history display for trend. If momentary vacuum excursion appears to settle, continue operation.
If pressure remains elevated, activate secondary pump.
If pressure does not go down within 5 minutes while secondary pump is on, call vac group leader.</details>
        </guidance>
        <guidance>
          <title>What to do</title>
          <details>Bad example 1</details>
        </guidance>
        <display>
          <title>Vac. Shutdown Procedure</title>
          <details>http://site.org/opdocs/VacShutdownProcedure.pdf</details>
        </display>
      </pv>
      <pv name="loc://Pump(0)">
        <description>Sector 1 Vacuum Pump</description>
        <enabled>true</enabled>
        <latching>true</latching>
        <annunciating>true</annunciating>
        <guidance>
          <title>Pump Guidance</title>
          <details>If pump is off because of overtemperature, check blower, let cool for 5 minutes, then turn back on.
If unsuccessful, follow vacuum shutdown procedure to secure system and call in vac group leader.</details>
        </guidance>
        <guidance>
          <title>What to do</title>
          <details>Bad example 2</details>
        </guidance>
        <display>
          <title>Vac Pump</title>
          <details>vac_sec1_pump.bob</details>
        </display>
        <display>
          <title>Vac. Shutdown Procedure</title>
          <details>http://site.org/opdocs/VacShutdownProcedure.pdf</details>
        </display>
      </pv>
    </component>
  </component>
</config>
```

The idea is to support hierarchic configuration: Vacuum has links to a vacuum overview, to telephone numbers for vacuum at large. Vacuum/Sector1 may then link to a more detailed display for vacuum in that sector. Finally, PVs have guidance and maybe links to what's specific to the PV.

Next I'll get a 'good' and a 'bad' screenshot of the context menu.

kasemir:
When selecting just one PV, the context menu looks like this:

alarm1

We first list the guidance infos, then the displays.
They are ordered from most specific (PV) to more generic (sector, finally vacuum at large).

Overall, I'd say that's very reasonable.
In the RCP-based implementation, we actually added some modifiers to the icons to show if this guidance or display is from the item itself, or inherited from the hierarchy. Not sure if that helped, or caused questions.

kasemir:
Now the example where multiple PVs are selected:

alarm2

Again we first list the guidance infos, then the displays.

Good:
We merge guidance and display.
Both PVs inherit the "Sector 1 Info", "Vacuum Contacts", "Vac. Overview", .. but those are only listed once.

Room for improvement:
We first list specific, then generic entries for PV1, then for PV2, ...
==> Should be overall ordered by specific - to - generic:
1) Specific entries for PV1, PV2, PV2, ...
2) Items inherited one level up for PV1, PV2, ..
N) Top level items

Both PVs have a guidance entry with title "What to do".
When you open them, you see for which PV that's meant, but in the context menu you simply see two entries of the same label.
==> Guidance with the same title could be merged into one longer entry.

Unclear what to do about displays.
If N entries have display links with the same title, but _different_ display file, should that be one menu entry which opens all displays? I think not. Take that as configuration error and keep handling it like now, with multiple menu entries of the same name?

georgweiss:
Thanks for the examples. I'm fine with the inheritance mechanism, it makes sense.
However, comparing with the local examples at ESS shows that users may configure their alarms to not work well with the "intended usage". How is this intention documented? The XML allows for an arbitrary number of guidance tags at any level of the structure, which is not properly accounted for (and which by the way triggered this issue).
For the time being I need to park this issue, I will revisit next week or week after that.

kasemir:
Close PR for now?

georgweiss:
Yes.

PR #967, opened 2019-11-24
Show an error dialog if model cannot be loaded

krisztianloki:
This is basically the same as what the editor does when the model cannot be loaded.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
@@ -31,6 +31,7 @@ import org.phoebus.framework.macros.Macros;
 import org.phoebus.framework.persistence.Memento;
 import org.phoebus.framework.spi.AppDescriptor;
 import org.phoebus.framework.spi.AppInstance;
+import org.phoebus.ui.dialog.ExceptionDetailsErrorDialog;
 import org.phoebus.ui.docking.DockItem;
 import org.phoebus.ui.docking.DockItemWithInput;
 import org.phoebus.ui.docking.DockPane;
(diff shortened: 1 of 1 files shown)


app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
```diff
@@ -284,13 +285,39 @@ public void loadDisplayFile(final DisplayInfo info)
...
+            catch (Exception ex)
+            {
+                logger.log(Level.SEVERE, "Cannot load model from " + info.getPath(), ex);
+
+                final String exception_message;
+
+                if (ex.getCause() != null)
+                    exception_message = ex.getCause().getLocalizedMessage();
+                else
+                    exception_message = "";
```
kasemir: if (ex.getCause() == null) set exception_message = ex.getMessage()?

krisztianloki: I used `exception_message = ""` because `ex.getMessage()` will be shown by default by the dialog itself and it looked odd that the message was duplicated

kasemir: Yes, the complete message will be shown in the detail section of the dialog as part of the stack trace, if the user opens it.
But your main section messages ends in a colon "...:" to suggest that there's more:

```
Cannot load model from .. path: TheNestedMessage
```

But when there's no nested message, you end with a colon and then there's nothing:
```
Cannot load model from .. path:   <nothing>
```

krisztianloki: I will create a new pull request removing the colon when there is no cause.

PR #971, opened 2019-11-25
Allow hiding the menu; save/load via memento

kasemir:
Allows hiding/showing the menu.
Not added as a menu option, because unclear how user would then show the menu after accidentally hiding it.

Meant to be used by editing the memento to create a kiosk-like setup:
Create a layout with locked panes, disable tags, disable toolbar, close product.
In the memento, change "show_menu" from "true" to "false".
When now running phoebus, it loads the memento and the layout is pretty bare, with all the user can do is mostly close the app.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
+++ b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
@@ -99,6 +99,7 @@ public class PhoebusApplication extends Application {
     public static final String LAST_OPENED_FILE = "last_opened_file",
                                DEFAULT_APPLICATION = "default_application",
                                SHOW_TABS = "show_tabs",
+                               SHOW_MENU = "show_menu",
                                SHOW_TOOLBAR = "show_toolbar";
 
     /** Menu item for top resources */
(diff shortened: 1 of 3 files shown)


nugatritter:
How about hiding the context menu with entries like "Open in editor" to avoid unwanted changes?

kasemir:
To make displays read-only, there are already two ways:

Simply make the files read-only to the user. That way they can open them, edit, .., but have to save them under a different name. That's useful to empower end-users to come up with improvements, but they cannot accidentally modify the production displays. They can still open, edit, save in different location, and then contact control system engineers to look at the updated display and move it into the deployed set of screens.

For that reason we also support "edit" for displays read via http:// links. You get a "do you want to download to edit" type of dialog.

If you think your end users cannot be of any help, you can delete the app-display-editor jar from the product or build a custom product without the editor, so you get a runtime only tool.

pavel-ch:
You can disable "Open in editor" menu entry in authorization.conf - there is a specific user right:
# Anybody can run Display Editor
edit_display = .*

just delete the .*

kasemir:
Indeed, you had added that 2 years ago. Thanks!

Issue #968, opened 2019-11-24
Silently ignoring widgets with errors


krisztianloki:
In `ModelReader.readWidgetsAllowingRetry()` every exception (except `ParseAgainException`) is ignored and reading continues with the next widget. The exceptions are logged (with `WARNING` level) but there are no visual indications to the user that something went wrong and what is displayed is not correct. Moreover, it is really easy to unintentionally delete these "problematic" widgets when the display is saved in the editor.

kasemir:
This is by design.
The operator in the control room it's important that the display loads & runs as well as possible under the circumstance. Errors are logged for the system maintainer to investigate, but we don't want to simply stop parsing the display, or prompt with a dialog for every single error.

krisztianloki:
What about a single error dialog notifying the operator that there were some errors and they should check the log for more information?
And maybe replacing the unparsed widgets with placeholders (just like the Embedded display and Picture widgets)?

kasemir:
Those are both good ideas.

Did you actually run into exceptions while parsing a widget? I don't remember a real-life scenario.
But in any case, assume there's a display with 1000 widgets, it worked OK when it was edited and initially used, but maybe now it's called with some macro that's meant to contain a number, but for some reason contains a string, so 20 of the widgets which use the macro cannot be parsed?!

What we need to avoid:

 * Can't abort loading the file. Still need to show the other 980 widgets, because they might in fact cover all that the operator needs to do right now.
 * Can't open 20 dialog boxes, let alone modal ones, that keep the operator from interacting with the display.

One dialog box would be acceptable, as you suggest with text like "Errors loading file, check log for details". Also OK to put placeholders there, but for a placeholder you do need position, size and ideally widget type, so you can use something like the `UnknownRepresentation` which shows the outline and the widget type.

krisztianloki:
It seems that the widgets Claudio created are only available in the ESS build of CS-Studio/Eclipse or were not ported to Phoebus so that's where I encountered those exceptions.
I will try to implement the dialog box first and then play with the placeholders. Thanks for the suggestion of `UnknownRepresentation`!

georgweiss:
Yes, there are widgets that are missing in Phoebus when comparing with Display Builder in CS Studio/Eclipse. Currently we are working on porting at least the thumbwheel widget, but there are for sure other candidates. However, porting is complicated by the fact that some of Claudio's widgets have dependencies to third party libraries that we would need to carefully analyze before adding them to Phoebus.
[...]

PR #975, opened 2019-11-26
Show error dialog if widgets could not be parsed

krisztianloki:
Might also be useful to include the number of errors?

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java
@@ -187,7 +187,22 @@ public class DisplayModel extends Widget
         if (clean == null)
             return true;
 
-        return clean.booleanValue();
+        if (clean.booleanValue() == false)
+            return false;
+
+        // Check embedded displays and navigation tabs too
+        for (Widget child: getChildren())
+        {
+            java.util.Optional<WidgetProperty<DisplayModel>> child_dm_prop = child.checkProperty("embedded_model");
+            if (child_dm_prop.isPresent())
+            {
+                final DisplayModel child_dm = child_dm_prop.get().getValue();
+                if (child_dm != null && child_dm.isClean() == false)
+                    return false;
+            }
+        }
+
+        return true;
     }
 
     @Override
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java
@@ -15,6 +15,7 @@ import static org.csstudio.display.builder.model.properties.CommonWidgetProperti
 
 import java.util.List;
 
+import org.csstudio.display.builder.model.persist.ModelReader;
 import org.csstudio.display.builder.model.persist.NamedWidgetColors;
 import org.csstudio.display.builder.model.persist.WidgetColorService;
 import org.csstudio.display.builder.model.properties.WidgetColor;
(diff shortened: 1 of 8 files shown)


krisztianloki:
This does not work if the error is in an embedded display.
I'm having a hard time figuring out how (and when) can I get to the DisplayModel of an embedded display...

kasemir:
BOY loaded the display and its embedded display on the UI thread, merging them into one display model, with obvious performance implications.

In the display builder, embedded displays are by design treated as a "black box", in background threads. They do not delay loading and representing the top-level display. Instead, embedded content appears asynchronously as it's obtained.

This loading of the embedded model is implemented in the EmbeddedDisplayRepresentation. That's a compromise. In principle, the 'model' and 'representation' should remain separate. But loading the embedded model in the representation of the container simplified showing a preview of the content in the editor, while also supporting the runtime use of the embedded displays.

You could thus add similar "There was an error" dialogs to the EmbeddedDisplayRepresentation, but care should be taken to minimize them. A display with 10 embedded displays, each of them failing, should not pop up 10 error dialogs. 

Maybe do something like this to set a flag in the top-level display model:
```
final DisplayModel top = model_widget.getDisplayModel().getTopDisplayModel();
boolean showed_errors = top.getPropertyValue("showed_errors");
if (! showed_errors)
{
            // Prevent another dialog
            top.setPropertyValue("showed_errors", true);
            
            // show dialog
}
```
.. and also set the "showed_errors" flag in the code you aleady prepared, so for a given display there will only be one dialog at most.

krisztianloki:
Thanks! I've just finished implementing another solution; using a Phaser to be able to wait for the representation of all child widgets.
This implements Idea Nr. 1 of #968

app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java
```diff
@@ -167,6 +171,40 @@ public final boolean isTopDisplayModel()
...
+    }
+
+    /** @return <code>true</code> if this display was loaded without errors,
+     *          <code>false</code> if there were widget errors
+     */
+    public final boolean isClean()
+    {
+        if (clean == null)
+            return true;
+
```
kasemir: Does 'clean' need to be volatile, then making a safe copy before checking?
```
final Boolean safe = clean;
if (safe == null)
  ..
else if (safe.booleanValue() ...
````

Comment (krisztianloki) [reply app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java]: Yes, you are right; `clean` needs to be volatile. But what would the safe copy achieve?

Comment (kasemir) [reply app/display/model/src/main/java/org/csstudio/display/builder/model/DisplayModel.java]: Get the current 'volatile' once, then check its value against several options.
Otherwise it's possible that 'clean' changes during the execution of this code:
```
if (clean == null)
  ...
else if (clean.booleanValue()) ..
```
If 'clean' is set from other threads, it's possible that it was non-null for the first `if`, then set to null before the second `if`. Not sure if it indeed goes through those values, but when checking a changing value against several options, it's best to get a safe copy of the value, which can then not change while you're testing it.

krisztianloki: I see. In this case `clean` should not become null once it is set but I will add the safe copy to be safe

kasemir: I think you're right. `clean` is first null, then non-null, without going back to null. Still, I think it's a useful habit to keep the check 'atomic' by checking one and the same value (the safe copy) and not a potentially changing value. That'll always we predictable without having to check other setters of the value across the codebase. Thanks!

Issue #944, opened 2019-11-15
Phoebus release


shroffk:
It seems like the product has reached a point where a release would make sense.
The release would also be a good way to co-ordinate with the eclipse based cs-studio release.

georgweiss:
Please clarify "coordination" with Eclipse-based CSS. Not in terms of features, I guess, and not as a long-term ambition...?

shroffk:
So the Issue_944 has a profile and the release plugin which provides the following functionality.

### Release

#### Step 1:

To perform a release  
```
mvn release:prepare -P release
```
- Check that there are no uncommitted changes in the sources
- Check that there are no SNAPSHOT dependencies
- Change the version in the POMs from x-SNAPSHOT to a new version (you will be prompted for the versions to use)
- Transform the SCM information in the POM to include the final destination of the tag
- Run the project tests against the modified POMs to confirm everything is in working order
- Commit the modified POMs
- Tag the code in the SCM with a version name (this will be prompted for)
- Bump the version in the POMs to a new value y-SNAPSHOT (these values will also be prompted for)
- Commit the modified POMs

Additionally:
- Before commiting the changes, there is a script in the target platform 'release_classpath.py' which will be executed. This script can be modified to updated the .classpath and other files which need manual intervention during a release.

#### Step 2:
Currently, this is not used since we have to resolve where to deploy things to.  
```
mvn release:prepare -P release
```

- Checkout from an SCM URL with optional tag
- Run the predefined Maven goals to release the project (by default, deploy site-deploy)

shroffk:
@kasemir the `release_classpath.py` was added to specifically ensure that the commits tagged as release could contain the updated .classpath files which ensure that when a release tagged version is checked out both the maven and ant build work.

georgweiss:
ESS offers an Artifactory instance as an SCM deploy target for release builds.

shroffk:
@georgweiss @andersharrisson-ess 
You can test this release with 

```
 mvn -P release release:clean -DdryRun=true -DskipTests=true
```

it should produce the final poms and other files without actually changing or pushing

shroffk:
It seems like github also got into the deployment business

https://help.github.com/en/github/managing-packages-with-github-packages/using-github-packages-with-your-projects-ecosystem

andersharrisson-ess:
Interesting to read about Github Packages. I had no idea they had released a binary repository service. Thank you for the info, @shroffk.
[...]

PR #979, opened 2019-11-28
Issue 944

shroffk:
#944 
Here is the maven release plugin with documentation
@georgweiss @andersharrisson-ess could you have a look at this and let me know if this is good enough to merge.

TODO: test release on branch...
The maven release plugin supports this, but it is basically an action which we could include in the profile but we can also just simply call in via the release command

mvn release:branch -DbranchName=my-branch -DupdateBranchVersions=true


georgweiss:
Not sure I understand the release profile in phoebus-target/pom.xml. It specifies the release_classpath.py script, but that one only writes some information to a log. Plus there should be a newline char following os.getcwd(). Is the objective to also run mk_classpath.py?

shroffk:
The 'the release_classpath.py' is just a placeholder.

'mk_classpath.py' this is independent from this PR it was there before

kasemir:
I've added 'mk_classpath.py' for those using Eclipse as a helper for creating a .classpath file that lists all dependencies from the target/lib folder. I don't actually use it, not aware of anybody using it, by I keep getting questions about it  ==>  Just remove it.

georgweiss:
Running the build creates dependencies/phoebus-target/release.log. This should be added ti .gitignore

shroffk:
I have added the changes requested by @kasemir and @georgweiss

pom.xml
```diff
@@ -167,10 +170,45 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-release-plugin</artifactId>
+            <version>2.5.3</version>
+            <configuration>
+              <preparationGoals>exec:exec</preparationGoals>
+              <checkModificationExcludes>
+                <checkModificationExclude>**/.project</checkModificationExclude>
+                <checkModificationExclude>**/.classpath</checkModificationExclude>
+                <checkModificationExclude>pom.xml</checkModificationExclude>
+                <checkModificationExclude>**/phoebus-target/pom.xml</checkModificationExclude>
```
andersharrisson-ess: Out of curiosity, why do you need to exclude these files from the modification check?

kasemir:
>  Generally, I prefer .classpath files kept local

We agreed at the start to support multiple ways of building. As documented in https://github.com/shroffk/phoebus/blob/master/dependencies/README.md and https://github.com/shroffk/phoebus/blob/master/dependencies/README.md you can

a) Depend on the pom.xml for maven to assemble the dependencies from the internet
b) Unzip a target platform and then build with ant
c) ..target platform, import the projects into Eclipse and run

For a) and b), the .classpath files are ignore or overwritten. For c), we use the .classpath files.
If you delete them, we lose option c).

andersharrisson-ess:
> We agreed at the start to support multiple ways of building.

Thank you for the clarification, Kay.
I was not aware of the fact that there were official decisions to support not only Maven and Ant, but also Eclipse. Since that is the case, I have no further objections or comments to this PR.

kasemir:
We can reconsider if there are conflicts, if for example the nbactions.xml added by Claudio or the .gitlab-ci.yml added by you broke the build process for another tool. But so far I think these files can happily coexist.

PR #981, opened 2019-11-29
Csstudio 1126 Close empty windows

georgweiss:
This PR will make sure windows without any remaining DockPanes are automatically closed after a drag-n-drop gesture. Does not apply to main window, though.

Verified on all platforms.

This does NOT close windows where all tabs have been closed. That will be a another PR.


kasemir:
How do you decide:
a) User removed the last tab and would like to close the window
b) User is re-arranging tabs, moved some out, is ready to move tabs into the window, when, oops, the window vanishes

georgweiss:
Suggest you try it out.

georgweiss:
I realize I did not consider split panes, so I will fix that.

However, I do think empty windows are a distraction and can be perceived as an application flaw. If you disagree, I will make the behavior conditioned by a preference.

kasemir:
I agree that empty windows tend to be superfluous. Except when you're re-arranging tabs. Then again, the (firefox) web browser simply closes the window upon removing the last tab, so maybe that's now commonly accepted.

georgweiss:
Updated such that empty windows will be closed if org.phoebus.ui/close_empty_windows=true. 
Close of tabs using close button considered, and split panes are also considered.

kasemir:
If you're done for now, then go ahead an merge.
If it "works", I think we should then make it the default, no preference necessary.
This is pretty essential, so the behavior should be consistent, not based on a pref.

PR #989, opened 2019-12-04
Issue 988

mhedges:
This allows for setting `org.phoebus.pv.ca/name_servers` in a `settings.ini` file when starting phoebus, addressing Issue 988

### DIFF
--- a/core/pv/src/main/resources/pv_ca_preferences.properties
+++ b/core/pv/src/main/resources/pv_ca_preferences.properties
@@ -32,3 +32,6 @@ dbe_property_supported=false
 # Mask to use for subscriptions
 # VALUE, ALARM, ARCHIVE
 monitor_mask=VALUE
+
+# Name server list
+name_servers=localhost
--- a/core/pv/src/main/resources/pv_ca_preferences.properties
+++ b/core/pv/src/main/resources/pv_ca_preferences.properties
@@ -3,7 +3,7 @@
 # -------------------------
 
 # Channel Access address list
-addr_list=10.0.16.88
+addr_list=
 
 auto_addr_list=true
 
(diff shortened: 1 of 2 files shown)


core/pv/src/main/resources/pv_ca_preferences.properties
```diff
@@ -32,3 +32,6 @@ dbe_property_supported=false
 # Mask to use for subscriptions
 # VALUE, ALARM, ARCHIVE
 monitor_mask=VALUE
+
+# Name server list
+name_servers=localhost
```
kasemir: Can we make the default empty, so there's no change for existing setups?
```
name_servers=
```
I'm not sure what otherwise happens to existing setups which would suddenly look for a name server.

kasemir: While at it, you can also make the default address list be empty, because the current setting is very specific:
```
addr_list=10.0.16.88
```

kasemir:
See minor comments, otherwise: Thanks!

Since I've never used a CA name server: Which one are you using? What happens if the name server is down (or set to the wrong host)? Will the CA client fall back to broadcasting (addr_list), or will it give up?

core/pv/src/main/resources/pv_ca_preferences.properties
mhedges: Done!

mhedges:
I have only ever set `EPICS_CA_NAME_SERVERS=localhost:port1 localhost:port2 ...` for whichever ports I have opened via `ssh`.  In the C/C++ implementation of Channel Access (I do not know about the java implementation), name_servers and addr_list do not conflict. I have been able to connect to local PVs on addr_list with `EPICS_CA_NAME_SERVERS=localhost:5069` regardless of whether I am using port `5069`.

kasemir:
Alight, thanks!

Issue #997, opened 2019-12-08
Display Runtime: ProbeDisplayContextMenuEntry distorts formula PV names


pavel-ch:
When transfering PV name to ProbeDisplay, the ProbeDisplayContextMenuEntry distorts PV names for formula PVs due to URI conversion. Some characters like '+' get lost.

kasemir:
Thanks for fixing this.

PR #998, opened 2019-12-08
Encoding some characters for URI transfer to Display Probe

pavel-ch:
See #997. Not clear if this is enough.


kasemir:
You're right, this covers `=` and `+`, but who knows what else could be in a formula or actually PV in general.

Check `ResourceParser.decode`, https://github.com/shroffk/phoebus/blob/master/core/framework/src/main/java/org/phoebus/framework/util/ResourceParser.java#L237.
I would add an `encode` routine, calling `URLEncoder.encode` with that same `UTF_8` constant for the encoding,and then use that new `ResourceParser.encode` to handle whatever strange characters are found in the PV.

kasemir:
Thanks!

PR #1002, opened 2019-12-10
Csstudio 1179

georgweiss:
Added a button in the XY plot toolbar that rests the axis ranges to the initial value. Use case is: user wishes to return back to the initial plot view after a series of zoom/pan actions.

The reset action is itself undoable so the user can continue from the last zoom/pan state.

The added button is to the left of the zoom-in button. The icon for the button is taken from the BOY XY plot toolbar implementing the same action. I can think of a better icon, but this one should at least be recognized by BOY users. The placement of the button is not carved in stone.


kasemir:
"Resets the X and Y axis ranges to the initial values. These initial valuesare found by looking for the first {@link ChangeAxisRanges} action in the undo stack"

The undo stack is size limited. What if you can't find the "first" axis range change?

georgweiss:
Then nothing happens. Limit is 50, right?

georgweiss:
Thanks

kasemir:
That button now also appears in the data browser, but it doesn't do anything.
Can you fix that?

kasemir:
Works OK in the Display Builder XYPlot.

kasemir:
In the data browser, this is what happens:
```
Exception in thread "JavaFX Application Thread" java.lang.IllegalArgumentException: 2 Y axes, but 1 new ranges
	at org.csstudio.javafx.rtplot.internal.undo.ChangeAxisRanges.<init>(ChangeAxisRanges.java:75)
	at org.csstudio.javafx.rtplot.internal.undo.ChangeAxisRanges.<init>(ChangeAxisRanges.java:109)
	at org.csstudio.javafx.rtplot.internal.Plot.resetAxisRanges(Plot.java:1381)
	at org.csstudio.javafx.rtplot.RTPlot.resetAxisRanges(RTPlot.java:587)
	at org.csstudio.javafx.rtplot.internal.ToolbarHandler.lambda$5(ToolbarHandler.java:192)
```

I think that's because it simply goes back to the first rescale entry, when there were no or fewer channels = traces = axes.

You need to look for the first rescale entry that has a matching number of axes.

kasemir:
In the XYPlot, a changing number of axes is less likely, but still possible since there can be a script which adds/removes traces and axes at runtime.

georgweiss:
But finding the rescale action with a matching number of axes might not be the one I'm looking for since a zoom or pan action might have occurred before axes were added.

georgweiss:
Sure, I can remove it in the Data Browser.

georgweiss:
Of course, finding matching number of axes might be better to avoid exceptions, user just needs to be aware of how the reset button works.

kasemir:
Well, the whole idea of finding the first one is a bit arbitrary.
This is basically a "reset it to what I think it should be" button without a clear definition what that might be, hoping that the last 50 entries contain it.

georgweiss:
The limit of 50 entries is a... well, limitation. So if the reset button is limited by the same rule, I think that is OK.
The alternative would be to implement a complete reset (with respect to axis ranges, number of traces etc) to the state when the plot was first rendered.

kasemir:
A complete set doesn't make sense.
People use the data browser, add a channel, remove a channel, add another channel. Resetting to the very start would be the same as close/open new one, loosing everything.
In XYPlot, you could add this to the XYPlotRepresentation: Preserve the original model settings for the axes, then restore to that.
Adding this code to the RTPlot is one level too deep, impacting all code that uses it.

kasemir:
Similar to the XYPlot resetting to  the values saved from the original model, the data browser could reset to the values from the *.plt file, if there is a file and this is not an interactive session. In any case, the actual values to reset to would depend on the use case.
When instead trying to base this on the undo history, your best bet is to find the first entry with the same number of axes.

georgweiss:
I'm a bit on thin ice here, but using/preserving the original model in the XYWidgetRepresentation is one thing. The ToolbarHandler operates on the plot and has no access to the widget representation, has it?

kasemir:
Correct.
The easiest way to just fix what you did so far would be to stay in the plot code, stay with the undo history, but look for the first change that uses the same number of Y axes.

georgweiss:
Well actually I would prefer to do it in the XYWidgetRepresentation, but then I would probably need to add another method in the RTPlotListener to handle a request for the plot to reset to the original model. 
How is a trace added from script?

georgweiss:
Same number of Y axes I guess could be ambiguous too if traces are added and removed dynamically.

kasemir:
You could call that another limitation. A "make it look like I think it should" button isn't easy.

georgweiss:
Trying to convince users to press the undo button multiple times is not easy either.

kasemir:
Well, then go ahead and implement a "make it look like I think it should" button that works.

georgweiss:
I think it would be better to require the current and "original plot" to have same number of axes, and if not, display message and not do anything. This would be more consistent, i.e. reset is not meaningful if the plot was augmented with more traces.

kasemir:
It's very normal for a data browser to contain more than one trace.
By default, each trace gets its own axis.
So in most practical cases, that button will open a dialog to tell the user that the button doesn't work.

It's OK for the XYPlot, though, so use the RTPlot.addToolItem to add your button in the XYPlotRepresentation, i.e. only in the XYPlot, and not in the underlying plot that's also used by the data browser, scan data monitor, .. where it doesn't apply.

georgweiss:
I will exclude the button from Data Browser.
However, in XYPlot I still think it is a good idea to show that the reset operation cannot be performed, even if the use case is rare.

kasemir:
> I will exclude the button from Data Browser.

OK.

Issue #800, opened 2019-08-27
Alarm Config Logger - Remote git not updating


tanviash:
@shroffk Hi Kunal, Remote git is not updating with config changes with the latest code. Config.xml updates but then repo itself doesn't create new nodes and PV hierarchy underneath.

shroffk:
Hi Tanvi, 
Were you able to get any error messages from phoebus?
Also, I remember adding a preference to enable/disable pushing to remote repo...can you check if the preference is correctly set.

tanviash:
@shroffk I can look for logs and send it to you. What is the new preference setting called for enable/disable pushing to remote? I can't find it.

tanviash:
Yes, we only use remote location to enable pushing to remote when a value is provided. Moreover, I see config.xml on remote getting updated but the directories inside the git repo do not update.

shroffk:
and... I remember you saying that you tried to wipe out the old repo and started from a fresh one and still had the problem

tanviash:
Yes, I manually deleted the old directories from the repo.

shroffk:
Ok, I will wipe out and try a fresh installation myself... if you can find error logs or messages that would be very helpful

PR #1005, opened 2019-12-10
Fix alarm-config-logger directory path

tanviash:
Code change is to fix the directory path in alarm config logger which was making it unable to push correct directories to local and remote repo.


kasemir:
Is this for #800?

tanviash:
@yes, for #800

kasemir:
Good, another one fixed.

shroffk:
I think we can merge this. I can test it later this week

PR #1023, opened 2019-12-13
Csstudio 1173 Support for SVG

georgweiss:
We need support for SVG, added it for the Symbol widget. Tested on Mac and Linux with a few different SVGs of varying complexity. Looks promising, though gradients are not always 100% correctly rendered.

There is a Dependency to Apache Batik, it inflates the size of the lib folder by ~3 MB.


dependencies/phoebus-target/pom.xml
```diff
@@ -10,6 +10,7 @@
   <properties>
     <spring.boot-version>2.0.5.RELEASE</spring.boot-version>
     <preReleaseScripts>${project.basedir}/release_classpath.py</preReleaseScripts>
+    <batik.version>1.12</batik.version>
```
kasemir: Suggest the batik.version should be in the top level (parent) pom, and then also be used in the app/display/representation-javafx/pom.xml

georgweiss: Makes sense. Will fix it.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/SVGHelper.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param svgFilePath Non-null input stream to SVG file.
+     * @return A {@link Image} object if all goes well.
+     */
+    public static Image loadSVG(InputStream svgFilePath) {
+        TranscoderInput input = new TranscoderInput(svgFilePath);
+        try {
+            bufferedImageTranscoder.transcode(input, null);
+            return SwingFXUtils.toFXImage(bufferedImageTranscoder.getBufferedImage(), null);
+        } catch (TranscoderException e) {
+            throw new RuntimeException(MessageFormat.format("Unable to transcode SVG file, cause: {0}", e.getMessage()));
```
kasemir: e.getMessage() might be empty, plus the original stack trace of the TranscoderException is lost.
Suggest `throw new RuntimeException("Unable to transcode SVG file", e)`.
Also, since calling code some levels up already catches Exception, throw plain Exception instead of RuntimeException.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SymbolRepresentation.java
```diff
@@ -783,7 +786,13 @@ public void setSize ( double w, double h ) {
...
-                        return new Image(ModelResourceUtil.openResourceStream(imageFileName));
+                        InputStream inputStream = ModelResourceUtil.openResourceStream(imageFileName);
+                        if(imageFileName.toLowerCase().endsWith("svg")){
+                            return SVGHelper.loadSVG(inputStream);
+                        }
+                        else{
+                            return new Image(inputStream);
+                        }
                     } catch ( Exception ex ) {
                         logger.log(Level.WARNING, "Failure loading image: ({0}) {1} [{2}].", new Object[] { fileName, imageFileName, ex.getMessage() });
```
kasemir: Similar: ex.getMessage might be empty, and stack trace is lost.
Use `logger.log(Level.WARNING, "....", ex)` to log complete exception with message, trace, cause.

Issue #1017, opened 2019-12-12
How to correctly set up .classpath


shroffk:
The app email has a dependency to javax.activation and javax.mail. These are inherited from core-email.

However in eclipse the imported project reports errors despite the .classpath being manually updated to include phoebus-target.

![image](https://user-images.githubusercontent.com/2111304/70723038-42521780-1cc6-11ea-9114-e1ab37bbefc1.png)

Once I configure this as a maven project, the error disappears.

![image](https://user-images.githubusercontent.com/2111304/70723251-98bf5600-1cc6-11ea-9586-759d59a66e0b.png)

What am I doing wrong here...

kasemir:
Not sure, but overall I think we're fine.

I simply import the projects via File, Import, General, Existing Projects into Workspace, using the .classpath file without changes, and see no error.

![email](https://user-images.githubusercontent.com/1932421/70726822-ad065180-1ccc-11ea-8fdb-c1d7594a179d.png)

Those classes where you get the error are resolved to the binary in dependencies/phoebus-target/target/lib/activation-1.1jar

==> I'm fine.

You prefer to import as maven projects, and the error disappears

==> You're fine


You might be able to fix the Eclipse problem by opening in an instance of Eclipse that doesn't have additional maven tools installed, and then might still have to 'Project/clean' just the affected project. I do get that type of error occasionally for the XML parser imports, and Project/clean' on just the affected project fixes it.

PR #1028, opened 2019-12-16
Build svg

kasemir:
#1023 and #1025 added SVG support, but didn't add the new test/resources folder to the .classpath, didn't list the new batik and XML libs in the dependencies .classpath, and set batik.version in both the parent pom and a lower pom.

### DIFF
--- a/app/display/representation-javafx/.classpath
+++ b/app/display/representation-javafx/.classpath
@@ -5,6 +5,7 @@
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
 	<classpathentry kind="src" path="src/main/resources"/>
 	<classpathentry kind="src" path="src/test/java"/>
+	<classpathentry kind="src" path="src/test/resources"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/phoebus-target"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-framework"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-ui"/>
--- a/dependencies/phoebus-target/.classpath
+++ b/dependencies/phoebus-target/.classpath
@@ -119,6 +119,25 @@
 	<classpathentry exported="true" kind="lib" path="target/lib/spring-web-5.0.9.RELEASE.jar"/>
 	<classpathentry exported="true" kind="lib" path="target/lib/spring-webmvc-5.0.9.RELEASE.jar"/>
 	<classpathentry exported="true" kind="lib" path="target/lib/hibernate-validator-6.0.12.Final.jar"/>
+        <classpathentry exported="true" kind="lib" path="target/lib/batik-anim-1.12.jar"/>
+        <classpathentry exported="true" kind="lib" path="target/lib/batik-awt-util-1.12.jar"/>
+        <classpathentry exported="true" kind="lib" path="target/lib/batik-bridge-1.12.jar"/>
+        <classpathentry exported="true" kind="lib" path="target/lib/batik-constants-1.12.jar"/>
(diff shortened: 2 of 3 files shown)


georgweiss:
Following the procedure in the README I am capable of building with both maven and ant without this commit. What am I missing? What do I need to do in order to detect build issues?

kasemir:
What IDE are you using?

georgweiss:
IntelliJ. And command line to test build, since that is what the CI is using.

kasemir:
I think we all check building with maven.
Kunal then imports the maven projects into Eclipse, which doesn't always work perfectly.
You import them into IntelliJ, which seems to work perfectly.
I also build with ant, and I appreciate that you checked that.
In addition, I import into Eclipse via plain Eclipse projects (.project, .classpath).

==> I check maven, ant, eclipse.
You checked maven and ant.
Kunal checks maven.

shroffk:
I have given up importing projects as maven in eclipse. With the presence of the classpath this always has problems with resolution etc.

I am sorry for my preference towards maven.
While I have ci tasks set up for maven which is why for open PR I usually only check maven I will start testing more religiously with ant too.
I really don't want to test with eclipse and the mangled .classpaths that I eventually land up with to just get things to work.

georgweiss:
Well, IntelliJ is not perfect, but I rarely have issues. Plus if I fix the issues, they stay fixed.
Once you cover some mileage with IntelliJ you will realize it does a lot of small things much better than Eclipse. When summed up, that clearly compensates for the small amount of pain points.

shroffk:
Ok I think I will start moving to intellij

I will continue to check the maven build and ant from command line.

kasemir:
> I have given up importing projects as maven in eclipse. With the presence of the classpath ...

The .classpath files are for the plain Eclipse import.
When using the maven import, Eclipse updates them, and the result seems to help nobody.
Does the maven import work when you delete all .classpath files?

Then we could do this:
a) Use "Import existing Eclipse project" to use the .classpath as they are
b) Run some 'delete all .classpath' script, then import as maven project, where Eclipse now creates proper .classpath files from scratch.

shroffk:
With other maven projects, when I import them as maven projects...eclipse generates the .classpath files and I don't touch them. This works.

> Then we could do this:
> a) Use "Import existing Eclipse project" to use the .classpath as they are
> b) Run some 'delete all .classpath' script, then import as maven project, where Eclipse now creates proper .classpath files from scratch.

Thanks, but I don't see the value in spending this effort. In eclipse we should advertise that we only support importing as eclipse projects. I have been doing the same and am fine with this.

I have run configurations defined for building using maven and ant which work very reliably...even when the workspace shows errors.

My bigger issue with eclipse is that even when importing **as eclipse projects** the workspace frequently goes into error states that make no sense.
See #1017, which disappear when you close or refresh or other random stuff...but I guess the answer to that lies in intellij

kasemir:
> import them as maven projects...eclipse generates the .classpath .. works
> I don't see the value in spending this effort.

For Mac and Linux, it's trivial:
`find . -name .classpath -exec rm {} \;`

If you have a way to delete all .classpath files on Windows, then we can support both using the .classpath files or re-creating them from a maven import.

kasemir:
based on https://superuser.com/questions/1142143/deleting-all-files-that-start-with-certain-pattern-windows-command-line, this might work for the Windows power shell:
```
$curDir = Split-Path -Parent $MyInvocation.MyCommand.Path
foreach ($file in Get-ChildItem -force $curDir -Recurse)
{
    if (($file.Extension -match '.classpath'))
    {
        Remove-Item $file.FullName -Force | Out-Null
    }
}
```

PR #1029, opened 2019-12-17
Issue 1026

shroffk:
@kasemir please add the additionally changes needed for fixing the ant and eclipse builds


kasemir:
Done, thanks!

Issue #894, opened 2019-10-30
Java code conventions


shroffk:
It seems that as a collaborative project we should have some code conventions to make it easier for others to read our code.

I am sure this will prompt some strong opinions...but the current solution of "each person uses their own" makes for a very messy code base.

shroffk:
One declaration per line is recommended since it encourages commenting. In other words,
```
int level; // indentation level
int size; // size of table
```
is preferred over
```
int level, size;
```

shroffk:
We prefer putting an opening curly brace "{" at the end of a line, and not on a line by itself.
```

if (x < y) {           if (x < y) {
    x= y;                  x= y;
    y= 0;                  y= 0;
}                         } else {
else {         (or)        x= 0;
    x= 0;                   y= y/2;
    y= y/2;            }
}
```


 A scarce resource is the number of lines that one can see on the monitor at one time, and this convention wastes that resource.

```
if (x < y)            if (x < y)
{                       {
    x= y;                  x= y;
    y= 0;                  y= 0;
}                       } else
else           (or)  {
{                          x= 0;
    x= 0;                y= y/2;
    y= y/2;         }
}          
```

kasemir:
Well, I do feel strongly about having balanced indentations for the opening and closing quotes.

georgweiss:
@kasemir , are you referring to constructs like:
```
String a = "first" +
                  "second" +
                  "third";

```
Regarding braces, I think
```
if(){
}
else{
}
```

is more readable than the @shroffk 's example, even though this "wastes" an additional line.

kasemir:
Sorry, braces not quotes.
I'm referring to https://en.wikipedia.org/wiki/Indentation_style#Allman_style, being in favor of balanced braces
```
int func()
{
    while (true)
    {
    }
}
```
over placing the opening and closing brace at different indentation levels:
```
int func() {
    while (true)  {
    }
}
```

willrogers:
I think it's a very good idea to have consistent formatting. My preferred method is to have automatic formatting on save and a CI check to make sure the code conforms.

I don't really mind what the convention is.

kasemir:
Not sure about automatic formatting. Just like code doesn't write itself, it doesn't necessarily format itself.
```
// For instance, sometimes it better to have a short
final int end = (xx > l) ? xx : l-4;
// while at other times it's better to align like this
final int end = some_larger_expression
              ? value_for_one_case
              : value_for_other_case;

[...]

PR #1031, opened 2019-12-17
Code conventions

kasemir:
Initial writeup for #894, also mentioning file locations, naming of tests etc. which had been discussed in hangouts but should be documented with the code base.

### DIFF
--- a/CodeConventions.md
+++ b/CodeConventions.md
@@ -105,7 +105,7 @@ File Locations
 .. follow general Maven practice:
 
  * `src/main/java` - Java code for release
- * `src/main/resources` - Resources (*.fxml, *.png, ...)
+ * `src/main/resources` - Resources (`*.fxml`, `*.png`, ...)
  * `src/test/java` - Test code
  * `src/test/resources` - Resources used in the test (example input files, ...)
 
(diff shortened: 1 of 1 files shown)


georgweiss:
Do not quite agree that we should label the following as discouraged:
`if(){
}
else{
}`

IDEs are capable of highlighting opening braces so finding them should not be a problem.

kasemir:
I thought in #894 and hangouts we had already agreed on balanced braces.

georgweiss:
Maybe, do not quite recall. But how are we to enforce it? Are we do change all source?

kasemir:
We don't enforce it by blindly applying some source code formatter, that's kind of the point, because it would break all the manual formatting that people already put it.
These are suggestions for humans to follow as they add code or update some existing code.
Step one is to document it.
Step two might be enabling IDE support for removing trailing white space, or 4-spaces-instead-of-tabs. But no blind overall reformatting.

georgweiss:
That's fine, but I still do not agree with the statement of discouraged use of braces.

kasemir:
OK, then we scratch this.

Issue #1032, opened 2019-12-17
PVA structure value change notifications


shroffk:
In pvaccess, when references a part of a larger structure there is a possibility that the sub structure will not have a timestamp. 

eg.
Complete Structure:

![image](https://user-images.githubusercontent.com/2111304/71022476-8e3dfb80-20ce-11ea-9914-2c0f333242bd.png)

Substructure being used in cs-studio/phoebus:

![image](https://user-images.githubusercontent.com/2111304/71022458-854d2a00-20ce-11ea-9fae-110be80e7aab.png)

When the substructure is used in Display builder...despite the fact that the value is changing, since the time stamp remains the same the widget does not update. (I am not sure if the event is not being delivered to the widget or being discarded due to the same timestamp)

Ideas:
- core-pv/pva should somehow extract the timestamp from the parent structure
- display builder, like probe should respond to value changes even when the time stamp is the same.
- we have a flag with just adds the current time stamp

kasemir:
I don't think it's a timestamp issue. It's a matter of getting or not getting events. 
Increase the log level of org.epics.pva to FINE and check if you get "Received monitor #.. update for <channel name>" messages in one case but not the other.

kasemir:
What are you using as a pva server? Is that something I can reproduce locally?

shroffk:
Ok, 
Ignore the above report. After some digging the problem is not to associated with the timestamp.

The problem lies in the value comparison in the PropertyChangeHandler

```
       // Any change at all?
       if (new_value != null  &&  old_value != null  &&  new_value.equals(old_value))
           return;
```

In my case, the value is a IVDoubleArray... where the equality is checked by checking the underlying raw data points to the same array instead of looping through the array values.

```
    @Override
    public boolean equals(Object obj) {
        if (obj == this)
            return true;

        if (obj instanceof ArrayDouble) {
            ArrayDouble other = (ArrayDouble) obj;
            
            if ((array == other.array) && startIndex == other.startIndex && size == other.size)
                return true;
        }

        return super.equals(obj);
    }

```

Despite having different array values the arrays in the old and new are the same. 

![image](https://user-images.githubusercontent.com/2111304/71097545-21823a00-217e-11ea-9d29-8d6aebec952a.png)

shroffk:
The server is available here:
https://github.com/shroffk/scope

kasemir:
Does the array data indeed not change, so that part would be correct?

[...]

PR #1036, opened 2019-12-18
#1032 decode pva arrays into a new array

shroffk:


core/pva/src/main/java/org/epics/pva/data/PVABoolArray.java
```diff
@@ -91,9 +91,7 @@ public void encodeType(ByteBuffer buffer, BitSet described) throws Exception
     public void decode(final PVATypeRegistry types, final ByteBuffer buffer) throws Exception
     {
         final int size = PVASize.decodeSize(buffer);
-        boolean[] new_value = value;
-        if (new_value == null  ||  new_value.length != size)
-            new_value = new boolean[size];
+        boolean[] new_value = new boolean[size];
```
kasemir: Can you make them `final` as in the double example?

shroffk: Yes... makes sense.

kasemir: Thanks!

PR #1038, opened 2019-12-19
Display Builder Runtime - Extending WidgetInfoDialog

pavel-ch:
See discussion in #896
Extending WidgetInfoDialog with a recursive underlying PVs information for widgets with Children


kasemir:
Thanks!

PR #1049, opened 2020-01-07
Display Editor: in main Editor panel, for inline editor input field (…

pavel-ch:
…PV name or label), commit changes when inline editor looses focus
Continuing in #1042 and  #1047 - inline editor in the graphic editor panel.
Will NOT do the Widget Tree inline editor / not that important & not that easy to hack to TextFieldTreeCell edit mode finish conditions.


kasemir:
Thanks!

PR #1056, opened 2020-01-09
Csstudio 819

georgweiss:
ComboWidget maintains an ArrayWidgetProperty that is by default empty. Other widgets (e.g. SymbolWidget or MultiStateLEDWidget) are initialized with a non-empty ArrayWidgetProperty. The side effect of the empty ArrayWidgetProperty is that when display is opened in editor mode and a ComboWidget (where no items have been added yet) is selected, the dirty flag is set.

This PR fixes this (admittedly small) issue.


kasemir:
Good, thanks!

PR #1064, opened 2020-01-13
no reason to request full data w/ DBR_CTRL_*

mdavidsaver:
Value won't be used anyway, and larger than expected requests may trip over max array bytes computed assuming `DBR_TIME_*`.

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/ca/JCA_PV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/ca/JCA_PV.java
@@ -208,8 +208,9 @@ public class JCA_PV extends PV implements ConnectionListener, MonitorListener, A
             // (i.e. fetching the string as a BYTE[])
             // crashes the IOC.
             // --> Using the same request count as for the subscription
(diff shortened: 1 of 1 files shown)


kasemir:
Well, note the comment in that code section:
```
// With R3.15.5, fetching just one element for a record.INP$
// (i.e. fetching the string as a BYTE[])
// crashes the IOC.
// --> Using the same request count as for the subscription
```

Have you tested that with `some_calc.INP$`?
Maybe the issue was only in R3.15.5? I need to use this with R3.14.{several} and 7.x, so if the issue was only in R3.15, I don't personally care.

mdavidsaver:
Right, well.  I managed not to internalize that point.  This issue has been fixed in 3.15.6 (https://bugs.launchpad.net/epics-base/+bug/1678494).

For my purposes, `1` isn't a magic number.  It's just less than `9000000`.  Does capping the `request_count` at say 100 seem a reasonable way to go?

kasemir:
Thanks for checking that it's been fixed.
I'll update the comment.

PR #1066, opened 2020-01-15
Csstudio 834

georgweiss:
Added a read-back PV property to the BoolButtonWidget. The idea is to set the state of the button (LED, background and image, as configured) based on a read-back PV rather than on the primary PV.  If the read-back PV name is blank, the behavior is as before.
The label of the button is however always set by the primary PV value.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -155,6 +155,7 @@ public class Messages
     public static String WidgetProperties_BarBackgroundColor;
     public static String WidgetProperties_BarColor;
     public static String WidgetProperties_Bit;
+    public static String WidgetProperties_ReadbackBit;
     public static String WidgetProperties_BorderAlarmSensitive;
     public static String WidgetProperties_BorderColor;
     public static String WidgetProperties_BorderWidth;
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/BoolButtonWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/BoolButtonWidget.java
@@ -7,22 +7,6 @@
  *******************************************************************************/
 package org.csstudio.display.builder.model.widgets;
 
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.newBooleanPropertyDescriptor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.newFilenamePropertyDescriptor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propBackgroundColor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propBit;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propConfirmDialogOptions;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propConfirmMessage;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propEnabled;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propFont;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propForegroundColor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propLabelsFromPV;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propOffColor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propOffLabel;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propOnColor;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propOnLabel;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propPassword;
-
 import java.util.Arrays;
 import java.util.List;
 
@@ -40,15 +24,16 @@ import org.csstudio.display.builder.model.persist.NamedWidgetColors;
 import org.csstudio.display.builder.model.persist.NamedWidgetFonts;
 import org.csstudio.display.builder.model.persist.WidgetColorService;
 import org.csstudio.display.builder.model.persist.WidgetFontService;
-import org.csstudio.display.builder.model.properties.ConfirmDialog;
-import org.csstudio.display.builder.model.properties.EnumWidgetProperty;
-import org.csstudio.display.builder.model.properties.WidgetColor;
-import org.csstudio.display.builder.model.properties.WidgetFont;
+import org.csstudio.display.builder.model.properties.*;
+import org.epics.vtype.VType;
 import org.phoebus.framework.persistence.XMLUtil;
 import org.w3c.dom.Element;
 
-/** Widget that provides button for making a binary change
- *  @author Megan Grodowitz
+import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.*;
+
+/**
+ * Widget that provides button for making a binary change.
+ * @author Megan Grodowitz
  */
 @SuppressWarnings("nls")
 public class BoolButtonWidget extends WritablePVWidget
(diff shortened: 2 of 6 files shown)


kasemir:
Please avoid wildcard imports like ` import org.csstudio.display.builder.model.properties.*` and instead import the specific classes.

Why does the label not also follow the readback?

georgweiss:
The reason the label state is bound to the primary PV only is to acknowledge that the button h as been pressed. If the label would follow the same logic as the LED, the user would not be able to determine if the button was pressed or not.

georgweiss:
As for the wildcard import I blame IntelliJ. Not sure why it auto formatted the imports, I will update.

kasemir:
The indications of a widget must be consistent. The text, color, border, .. must all reflect the same state. For the button, the text is the main indicator. The color is a nice addition which might help to see the current state from a distance, but we would discriminate against users with impaired color vision if we depended solely on the color to indicate something. Same with border: The color is nice for most people, but we also use the border style to assert that everybody can tell 'minor' from 'major' severity.

If you want to write to one PV, but indicate the current state via a separate readback PV: Fine, add the readback PV as you did and make all of label, color, image, border, ... reflect the readback PV's value.

If you want to indicate two different things, i.e. write to one PV and indicate that state, while at the same time indicating the state of a separate readback PV, that should be done via two widgets. Like an LED for the readback placed next to the button for the setpoint.

georgweiss:
The implemented behavior is what our users are asking for. The read-back PV is optional.

kasemir:
We cannot add inconsistent behavior to a widget, breaking compliance with support for users with disability, simply because one user at one site wants some feature.
If the user needs an indication for a readback separate from the written value: Place an LED next to the button.

georgweiss:
Firstly, I do not agree that the behavior is inconsistent. The optional PV is there to inform the user of the actual status, while the label acknowledges that the button has been pressed. I fail to see how disability is a factor, please clarify.
Our users are well aware of the workaround of using another widget to indicate status, but they do no like it and are asking for this additional feature of the boolean button.

kasemir:
Color vision impaired users may not see the color, so they cannot see your readback PV indication.

With a button writing to the 'setpoint', indicating the state of the setpoint PV via label, color, border, combined with an LED for the 'readback', indicating its state via label, color, border works for everybody.

georgweiss:
If a color vision impaired user cannot use the LED, it should not matter if the LED state is determined by the set point or the read-back. Further, for color vision impaired users LED and background colors can be customized, or the button may use images instead of the LED.
We think this is a design that mimics the behavior of physical switches equipped with a LED indicating the actual state; the position of the switch indicates the set value, while the LED state indicates the outcome of the action.

kasemir:
> If a color vision impaired user cannot use the LED

A color vision impaired user can use the LED widget because there are state labels.
Turning the button into a widget where only the color is used to indicate something is not acceptable for the general user community.

If you want to operate displays at your site that way, you can add a rule to the button that changes the background color (both on and off color) based on a readback PV. Or you can derive a site-specific bool button for your site's product. Or update the label as well, so label, color, border are all connected to the readback PV (if it is configured).

georgweiss:
I think we all agree that we should avoid site specific solutions. Further, supporting a feature natively is better than using rules. Again, the feature is optional, if you do not like, don't use it.
You have not suggested a way to indicate button press if we let the label follow the behavior of LED, background or image.

kasemir:
Even optional features shouldn't be broken.

You're correct, if the label, color, border all reflect the readback, nothing indicates the setpoint.
Using the label to indicate the setpoint, with the color indicating the readback is not generally acceptable because color on its own cannot be used as an indicator.
A separate LED widget for readback, with button for setpoint does it, but you don't like that.
So you would need to add something else to indicate a setpoint/readback discrepancy. Could be a new overlay/decorator on the widget. But that can lead to odd behavior. With a separate LED and button, you press the button, and a short time later the LED for the readback would update. If you add a discrepancy indicator, pressing the button will at least briefly flash that discrepancy indicator, suggesting to the user that the operating system is at least briefly broken whenever you push the button.

georgweiss:
Well, let's agree that we disagree on the status of the optional feature. One can also question the purpose of the LED if you compare the bool button with a real-world physical button. Is it there to acknowledge that the set point was written or to indicate an outcome?
It would be valuable to have a second opinion from the community.

kasemir:
The button indicates the current state of the PV.
Note that clicking the button doesn't change the button state at all.
Clicking the button writes the value to the PV.
The PV either accepts the written value, we receive that value from the PV and update the button label, color, alarm.
Or the PV refuse the value, so nothing happens.
Or somebody else uses 'caput' or another screen or a sequence or .. to update the PV, we receive those updates and reflect it in the button.
So in any case, the button indicates the current state of the PV. Pushing the button writes the 'other' state to the PV, i.e. tries to toggle it, and the button then shows if that had an effect or not on the PV.

pavel-ch:
I can't judge as we do not use r/w pvs (always ro and wo) / using another communication scheme. Generally, to support readibility,  we are trying to avoid to:
- combine a write widget like a push button and its readback (preferably using readback symbols with contex menu ON, OFF,... actions, or alternatively separate buttons & readbacks)
- have flipflop push buttons (we always separate On and OFF actions)

So we have no use case for the BoolButtonWidget and I can't help you with this particular dilemma.

georgweiss:
Thanks @pavel-ch for your feed-back.

shroffk:
Hello,
I have been trying to insist that user split write and read back to seperate widgets. 
Your above use case (as I understand it) would be split into an LED - to display the readback value and a button which would be ties to the write PV and the values that are to be written.

These widgets are grouped together or kept next to each other.
I know this would take up more space on the screen but the separation of the widgets would mean that what is being displayed on the screen would be easily understood by any users without causing confusion regarding which of the widgets visuals match which underlying pv.

@georgweiss I feel like there is a possibility I might have oversimplified your use case. Can you describe the underlying system. Maybe knowing more details of your example might help me better understand this.

georgweiss:
Originally this is a user request which to me makes perfect sense. A push button used to toggle a HW state should optionally have the possibility to show if the underlying HW not only has received the signal, but also to show if the signal has had the wanted effect. This is how physical on/off switches usually work, and this I think is reasonable to mimic in a CS Studio widget. 

The argument that the purpose of a widget is to show the state of the PV it is bound to I think is limiting, in particular for a widget like the boolean button: what is the LED there for? If the label can confirm that the PV has acknowledged the signal, the LED only duplicates that information. I would also assume that not all control room personnel are familiar to EPICS and the internals of IOCs and PVs, so binding the boolean button behavior to one PV is to only take the engineer's point of view.

I have solved the user request using rules, but I still think an optional PV controlling the LED, background and image would be both justified and better than rules.

kasemir:
A good reason to keep the widgets single-purpose is long term display handling. Most operational sites deal with a mix of display tools.  Whatever you use now, in 10 years you will need to support at least one other display tool. For us it's currently EDM, EDM web display, BOY, Display Builder desktop, Display Builder web runtime. Displays using single-purpose widgets run quite well on all these tools based on automated translators or compatibility handlers.

georgweiss:
I guess I basically find the LED on the boolean button confusing and superfluous. To me it fills no purpose since the label conveys the state of the PV.
What would make more sense to me would to have a boolean button with label only, and then another widget where a LED would be bound to read-back PV.

kasemir:
I agree that the ' Show LED' option of the BoolButton is not necessary. It doesn't indicate any additional information. Some users may like it because they can tell from a little further away if the button PV is in the on or off state, but users with reduced color vision couldn't care less. "Show LED" was added to the Display Builder for compatibility with imported BOY screens, but EDM for example doesn't have a colored circle on its on/off button, and the next tool into which we import the display builder screens might also not have one.

It's a good example for better being careful what you add. As the BoolButton stands now, it's possible to configure it without labels, depending only on the LED colors to reflect the PV state, which is a bad idea. But with 15+ years of using BOY displays and now already a few years of *.bob displays in operation, taking the "Show LED" feature out would be disruptive. Best we can do is be careful what we add.

shroffk:
> What would make more sense to me would to have a boolean button with label only, and then another widget where a LED would be bound to read-back PV.

This is broadly how I prefer to address the use case too.

> A good reason to keep the widgets single-purpose is long term display handling...

Yes, in addition to long term management...the old xygraph is an example of how over complicated widgets can be difficult not just to maintain, debug, and optimize... but difficult to use too.

georgweiss:
Long term management is of course a factor to consider. However, that should not be an obstacle for improvement. Sometimes one needs to throw things overboard, just as we're doing with Eclipse/RCP.

georgweiss:
Will propose a different solution based on discussions with @shroffk

PR #1071, opened 2020-01-21
Csstudio 1190

georgweiss:
Added support for SVG in picture widget as ESS has a legacy to consider.


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/SVGHelper.java
```diff
@@ -39,6 +44,32 @@
...
+     */
+    public static Image loadSVG(String imageFileName, double width, double height){
+        String cachedSVGFileName = imageFileName + "_" + width + "_" + height;
+        return ImageCache.cache(cachedSVGFileName, () ->
+        {
+            // Open the image from the stream created from the resource file.
+            try(InputStream inputStream = ModelResourceUtil.openResourceStream(imageFileName)){
+                return loadSVG(inputStream, width, height);
+            } catch ( Exception ex ) {
+                logger.log(Level.WARNING, "Failure loading image: {0} [{1}].", new Object[] { imageFileName, ex.getMessage() });
```
kasemir: Instead of
```
logger.log("Error: " + ex.getMessage());
```
use 
```
logger.log("Error", ex);
```

The former only logs the exception message, which might be null. The latter logs the message (if there is one), plus the stack trace, and the optional nested 'cause'.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
```diff
@@ -243,4 +265,73 @@ else if (h_prime < pic_h)
...
+            // Resolve new image file relative to the source widget model (not 'top'!).
+            // Get the display model from the widget tied to this representation.
+            final DisplayModel widgetModel = model_widget.getDisplayModel();
+
+            // Resolve the image path using the parent model file path.
+            return ModelResourceUtil.resolveResource(widgetModel, expandedFileName);
+
+        } catch ( Exception ex ) {
+
+            logger.log(Level.WARNING, "Failure resolving image path: {0} [{1}].", new Object[] { imageFileName, ex.getMessage() });
```
kasemir: Another `ex.getMessage()` that would be better if it logged the exception

kasemir:
Good, thanks

PR #1072, opened 2020-01-21
Olog

shroffk:
Updated client for the new olog-es service.
WIP: do not merge since some core interfaces have minor updates.


shroffk:
@kasemir 
I have made one change the core logbook api.
The client now asks the service for a list of supported levels. This is to allow different sites to allow different levels.

As with the in memory logbook, you can populate this list from a properties file packaged with your site specific client.

I hope this works with SNS client easily

shroffk:
Additionally, where user defined times are uses I am reverting to using MILLI_FORMAT

kasemir:
I saw the 'level'. We do have those, so I can implement it for our logbook. I didn't look into the olog detail, that's all yours.

shroffk:
ya... the olog stuff is about the olog specific client.
Only the level, the associated create logentry dialog, and the query which uses MILLI_FORMAT should affect you.

kasemir:
OK to use MILLI_FORMAT.

kasemir:
If there were a logbook implementation that doesn't support levels, then you could consider supporting a UI that detects when the list of options is empty and thus removes that UI element.
But we can leave that for a time when it's necessary, since for now all supported logbooks have a 'level' (or 'priority' or something like that).

PR #1076, opened 2020-01-23
Csstudio 1191

georgweiss:
ESS needs the Source Sans font family. In Eclipse-based CS Studio it was provided as plugin.


kasemir:
How was it contributed in the Eclipse-based CS Studio?

georgweiss:
Through an ESS specific plugin found in https://github.com/ESSICS/org.csstudio.ess.product.git

kasemir:
Makes sense. So please contribute them to phoebus through an ESS-specific module as well. Or maybe a module in the common repo, but in any case a module separate from the Display Builder.

The display builder contributes one set of fonts, the Liberation ones, so that its examples "work", so that the built-in default fonts work, so that everybody can get started.
Every site is of course free to use whatever fonts they want, and if those are open source fonts, I'm OK with adding them as modules in a shared repository so that more than one site can easily include them in their site-specific product. But we cannot have the display builder provide all the fonts for everybody.

kasemir:
Technically, you can contribute fonts from your module by registering an AppDescriptor, and in its start() routine you then call Font.loadFont similar to the code in the display builder's CommonFonts class.

georgweiss:
To me it makes sense to provide them in this manner. The font files are smaller than for Liberation. We think the community product is improved since the Source Sans font is very well suited for building OPIs.

kasemir:
So by all means add them as a new font module like app-fonts to the common repo, and also by default include them in the example phoebus-product.
But they don't belong into the display builder. It only comes with one set of fonts to support examples and defaults.

georgweiss:
Sure, I can move them, but what about putting them in core/ui rather than a separate module?

kasemir:
I vote for a separate module. Fonts are a matter of personal taste. Sites should be able to decide if they want to include it or not.

georgweiss:
If you mean a maven module under app, how would sites exclude it?

kasemir:
In your product pom.xml, you can decide to include app-pvtree or not. So then you'll be able to include app- font-sans or not.

kasemir:
This is the pom from the example product. Note how various core-* and app-* modules are listed: https://github.com/shroffk/phoebus/blob/master/phoebus-product/pom.xml In your product, you can include or omit what you want.

georgweiss:
So a site that wishes to exclude the fonts would need to fork the product pom, right?

kasemir:
No need to fork.
Each site can have a site specific product, where they decide what to bundle into their product.

Examples include https://github.com/shroffk/nsls2-phoebus, https://github.com/kasemir/phoebus-sns, ...

Yes, the phoebus sources also include an example product, https://github.com/shroffk/phoebus/tree/master/phoebus-product, which tends to include 'everything' as a starting point for sites to create their own product. Sites should not fork that create their own.

@shroffk , can you explain this better?

georgweiss:
We will solve this differently, i.e. in an app specific for ESS.

PR #1087, opened 2020-01-29
Data Browser: Allow limiting the number of concurrent archive requests

kasemir:
Default allows 1000 concurrent requests, i.e. no practical change, but can now configured to a few or only 1.


shroffk:
Was too many concurrent requests a problem with your RDB archiver?

kasemir:
Generally, it hasn't been. Reading N channels "in parallel" tends to work OK and is faster than reading them one by one. Using the same _connection_ was a problem, but each thread establishes its own connection, and especially when accessing an RDB cluster _and_ the channels happen to be in different partitions, reading them in parallel is good.

... except when we have too many connections, too many readers, then the RDB throttles them back, and it's slower than if you read sequentially. Of course user A doesn't know how many other users there are right now reading data, so limiting the concurrency within each instance is limited. Still, I'm adding that parameter so we have something else to try.

Issue #1093, opened 2020-01-31
Liberation font installation unpredictable 


georgweiss:
The CommonFonts.install method installs Liberation fonts, but this does not always give the wanted result. 
Example 1:
1) Launch Phoebus without any tabs besides file browser.
2) Add a label to an existing or new display.
3) Select the font picker to launch the dialog.
4) The Liberation font family may be missing in the list, probably due to race conditions (that I have not investigated in depth). This becomes confusing if a class is used that defines precisely that font since the font dialog cannot show the association.
Example 2:
1) Launch Phoebus without any tabs besides file browser.
2) Start the data browser.
3) The label (y axis) font is expected to be Liberation (as defined in Model.java), but is not as CommonFonts.install has not been called. A fallback font is used instead, but this will of course be different depending on OS and maybe also OS version.

All in all CommonFonts.install should be called once at some point early in the Phoebus application startup phase. We could do it the way ESS installs custom fonts, i.e. separate application with this sole purpose.

kasemir:
I'll extract the Liberation fonts into a separate 'application'.

PR #1095, opened 2020-01-31
 Extract fonts used by example display and font size calibration ..

kasemir:
.. into separate module.
Fonts are registered via AppDescriptor.start() instead of static class
initialization.
This makes the fonts available to all tools, instead of depending on
first opening a display.

In principle also allows building a site-specific product without the
Liberation fonts, but note that this will disable the font size
calibration used when reading *.opi files.

#1093


shroffk:
> Fonts are registered via AppDescriptor.start() instead of static class
> initialization.

+1

PR #1100, opened 2020-02-03
Issue 1097

shroffk:
Packages services and the common phoebus product as tar and zip files

### DIFF
--- a/dependencies/.gitignore
+++ b/dependencies/.gitignore
@@ -1,1 +1,3 @@
 /bin/
+/phoebus-target.tar
+/targetRepository/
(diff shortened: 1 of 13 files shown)


shroffk:
Currently, the packaging into a tar and zip is part of the default build lifecycle.
Should this be made optional?
It does take a few extra seconds and the output folder takes up more space... but I am currently inclined to keep this unless someone objects.

shroffk:
@georgweiss @kasemir can we merge this?

kasemir:
I'm OK. Stayed out of this because you had requested review by 2 others already.

PR #1105, opened 2020-02-05
Csstudio 1184

georgweiss:
Updated pull request due to release and version cahnges


kasemir:
You're changing the file settings of for example  app/update/mk_update_settings.sh, removing the 'execute' bit, so I can no longer invoke in on Linux or Mac without first updating it back to `chmod +x ...`.

kasemir:
Same for `phoebus.sh`, removed the 'x' bits, so only Windows users can now launch the UI.

georgweiss:
Sorry about this, it was caused by a restore procedure when I got a new computer.

PR #1107, opened 2020-02-05
Grid constraint applied when pasting to editor with grid on.

pavel-ch:
See #679 Grid constraint applied when pasting to editor with grid on.


kasemir:
Thanks!

PR #1111, opened 2020-02-05
Display Editor: Align selected widgets to Grid / menu item

pavel-ch:
See #679 This PR adds a menu item for "Align selected widgets to Grid".
It also enhances the zoom combo box function (the selected item is now cleared when zoom is requested by mouse, to be able to select any new level from items list).


kasemir:
Thanks!

PR #1119, opened 2020-02-06
Zoom action enhanced for DisplayRuntime

pavel-ch:
Clear zoom combo on mouse wheel zoom / the same like for DisplayEditor
See #1111 & #679


kasemir:
Thanks!

Issue #1126, opened 2020-02-06
Alarm Log Table: disable automatic updates


mark0n:
The Alarm Log Table looks like a great tool for troubleshooting issues. I had a hard time using it though: I had to sort (which works flawlessly). Due to the amount of detail in this table I need a couple of minutes to make sense of the data. Unfortunately, the table automatically updates every few seconds which can be quite distracting and might move the results I care about around. To make matters worse whenever the table updates it loses its search properties forcing me to resort the table.

Have you considered a feature allowing users to disable automatic updates (like a checkbox next to the search button)?

tanviash:
@shroffk Hi Kunal, I can work on getting this checkbox to disable auto updates of the table. Do you have any other suggestions that can work better than check-box?

shroffk:
> To make matters worse whenever the table updates it loses its search properties forcing me to resort the table.

I think we might want to address this... we should save the search and the sorting features of the alarm table.

> disable automatic updates

We can add a checkbox, however I think the above feature might be more important :)

PR #1128, opened 2020-02-07
Alarm Log Table: Fix #1125 and #1126

### LINKED ISSUE #1125: Alarm Log Table: pressing enter in query string field doesn't trigger search

tanviash:
This pull request is to fix #1125 and #1126


shroffk:
Haven't tested this but it looks good to me...
Do you have store the column or could we just store the index. In any case this can be done latter.

Issue #1129, opened 2020-02-11
Display Editor - cannot paste and duplicate widgets containing language specific characters


pavel-ch:
In Display Editor, widgets containing language specific characters (like "umlaut" etc) in any text or name cannot be copied, throwing an exception in reading XML.
The problem is in converting widget XML byte stream to and from String. Using default platform encoding now, which is wrong and must use UTF-8.
PR follows.

PR #1130, opened 2020-02-11
Display Editor - Enabling paste and duplicate widgets containing lang…

pavel-ch:
…uage specific characters
See #1129


kasemir:
Thanks!

Issue #1136, opened 2020-02-11
Display model PathTest.java fails in windows


DominicOram:
As of https://github.com/shroffk/phoebus/issues/991 the tests in PathTest.java fail on windows with:
```
[ERROR] Tests run: 4, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 0.002 s <<< FAILURE! - in org.csstudio.display.builder.PathTest
[ERROR] testRelative(org.csstudio.display.builder.PathTest)  Time elapsed: 0.001 s  <<< FAILURE!
java.lang.AssertionError:

Expected: "other.bob"
     but: was "..\other.bob"
        at org.csstudio.display.builder.PathTest.testRelative(PathTest.java:77)

[ERROR] testCombine(org.csstudio.display.builder.PathTest)  Time elapsed: 0 s  <<< FAILURE!
java.lang.AssertionError:

Expected: "/home/beamline/main.bob"
     but: was "https://webopi.sns.gov/webopi/opi//home/beamline/main.bob"
        at org.csstudio.display.builder.PathTest.testCombine(PathTest.java:67)

[INFO]
[INFO] Results:
[INFO]
```

DominicOram:
The linked PR allows the tests to run OK on both windows and linux without reducing code coverage but I'm not sure if it actually requires more thought as to how the functions behave?

kasemir:
It's definitely progress to have the test pass on Windows, even if that requires two different tests.

The requirements of the ModelResourceUtil are in no small part "Needs to work with deployed displays", where buttons, related displays etc. need to work relative to local files but also when used with "http://server/displays/some/display.opi". The same displays can be used both as files at a beam line, and via URLs when viewed from outside the beamline. So can't use file tools because the paths may refer to URLs, at the same time cannot use URL tools because the paths may not be valid URLs. For example, a space " " is OK in a file path. For  URL it needs to turn into "+" or some other URL escape, which then won't work when used as a file name.

Adding tests and having them pass is good, since in case we eventually decide to re-implement the ModelResourceUtil, those tests are the best description of how it needs to function. So thanks for fixing this on Windows.

DominicOram:
Ok, neither the " " or "+" cases are being tested in the unit tests so I will add them. Are you saying it's possible to have a mix of URL and path in the same string? Otherwise you could pass the tests by calling `isUrl` then running the appropriate URL relativize or Path relativize.

DominicOram:
[...]

PR #1140, opened 2020-02-12
Display Model - Use nio functions

DominicOram:
See https://github.com/ControlSystemStudio/phoebus/issues/1136

* Adds additional unit tests to catch edge cases
* Refactors `getRelativePaths` to make better use of pre-existing functions


kasemir:
This is a bit tricky, you never know if there's some installation out there that will break when you tweak the file resolution, but as long as it passes the unit tests, all that we're aware of should be fine, so I'm in favor of merging this.
If we then find an issue, we'll add it to the unit tests and fix it.

DominicOram:
I'm pretty sure this is leveraging `nio` as much as is easily possible. Also has improved coverage so happy for it to be merged if you like.

PR #1142, opened 2020-02-12
Add Github Actions build

willrogers:


shroffk:
@willrogers  @andersharrisson-ess could you help resolve the this PR
I see the current .github has 2 yml files which are quite similar. Should we close this PR since the build.yml.

andersharrisson-ess:
This is indeed very similar to the existing workflows. I notice a few differences:

- This uses windows environment to build instead of Ubuntu.
- This uses v2 of the checkout action: `actions/checkout@v2`
- This runs Maven verify instead of just package

I prefer to run on linux more than windows, but it's just my preference. The result should be the same. Using later versions of the actions is probably good and if mvn verify works, we should probably do that. Hence, I recommend that this PR change the existing workflow (build.yml) to use v2 of the checkout action and change mvn package to mvn verify.

shroffk:
Thanks @andersharrisson-ess could you make those changes.
I will close this PR and we can have a new PR with @willrogers changes incorporated into the build.yml

Issue #1143, opened 2020-02-12
CI on windows


DominicOram:
Following https://github.com/ControlSystemStudio/phoebus/issues/1136 we should make sure that phoebus tests are run on windows. It looks like you can do this in travis but requires a bit of messing around.

PR #1144, opened 2020-02-12
Run CI on windows as well

DominicOram:
See https://github.com/ControlSystemStudio/phoebus/issues/1143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 sudo: required
 language: java
 dist: trusty
+os: windows
+
 jdk:
   - openjdk11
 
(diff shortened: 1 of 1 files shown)


DominicOram:
This build works but it is failing intermediately on:
```
[ERROR]   AutomatedActionTest.testAutomatedEmailFollowup:264 
Expected: "Send Email"
     but: was null
```

I had a quick look but it's not obvious to me why this is failing. @kasemir do you have more of an idea before I dive deeper into it?

kasemir:
This might be a time-sensitive test that doesn't always work.

kasemir:
If this is what's holding you up, change the test into something that prints a warning but otherwise proceeds.

DominicOram:
Thanks @kasemir, now passes in windows.

shroffk:
@DominicOram thanks for adding this to CI

Issue #1141, opened 2020-02-12
Documentation [ code-a-thon ]


shroffk:
#3
 - [x] merge the phoebus-doc https://github.com/kasemir/phoebus-doc and the phoebus repos
 - [x] create scripts & maven profiles to generate the docs. We can use the maven spyhix plugin...an example can be found in olog-es
 - [x] update the .rst files as needed
 - [ ] run the javadoc and sources maven job and fix as many warnings as possible
 - [x] publish/continue to publish to read the docs. Ensure changes don’t break anything.
 - [x] ensure building of multi module documentation. Preferences, Locations for example are described in the ‘core’ section yet many ‘app’ or ‘service’ documentation might want to link to it.

shroffk:
The source and javadoc should be moved to another ticket

PR #1147, opened 2020-02-13
#1141 merging Kay's documentation repo

shroffk:
WIP: 
the cloning and searching for sources can be removed since it is no longer needed


shroffk:
we still are having issues with images.

shroffk:
@willrogers and I had limited time work on this issue. We were having trouble getting the images from /src/resources/icons to be included in the generated docs

Also, relying on generating symlinks does not work well on windows.

shroffk:
This PR is ready to merge, 
1. The documentation is now under docs
2. The documentation can be built using both command line sphinx and the maven sphinx module
3. the documentation can be built on windows correctly
4. it is also available on read the docs https://control-system-studio.readthedocs.io/en/latest/

I apologies for all the inconvenience the delay in the PR has caused the collaboration and the incredible frustration people have had to endure when being unable to correct typos.
I try my best to avoid such situations in the future and once again appreciate your patience.

shroffk:
TODO: fix the links pointing to documentation. I shall keep working on that on this branch and open new PR's as needed.

georgweiss:
So I am trying to follow the README instructions:

- On Mac - using Brew - could be `pip3` instead of `pip`. Not sure, but it probably depends on the Mac OS version. On newest version it's `pip3`.
- `make clean install `should be `mvn clean install`?
- `mvn clean html` fails, there is no html goal.
- Javadoc should be possible to generate with Maven, I think.
- Line 84 has a typo: henerated -> generated.

kasemir:
>  make clean install should be mvn clean install?

It really meant  "make" for the sphinx "Makefile", https://github.com/kasemir/phoebus-doc/blob/master/Makefile

georgweiss:
So we're missing a makefile. Or makefiles?

kasemir:
Yes. I would suggest either importing the sphinx-generated Makefile, or adding at least the commands from the `clean` and `html` makefile targets into an ant or maven file.

shroffk:
> So we're missing a makefile. Or makefiles?

Yes, these were removed as part of the previous PRs
I have added them back

PR #1148, opened 2020-02-13
Create GitHub Actions workflows

andersharrisson-ess:
Github Actions workflows that builds and publishes Phoebus onto ESS Artifactory.
This PR requires credentials for Artifactory to be set in Github Secrets.

### DIFF
--- a/.github/maven/settings.xml
+++ b/.github/maven/settings.xml
@@ -1,0 +1,12 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <servers>
+    <server>
+      <id>artifactory</id>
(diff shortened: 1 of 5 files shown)


shroffk:
@andersharrisson-ess I have given you permissions to manage this repo... you can add the necessary secrets

andersharrisson-ess:
Thank you,
I have added the required credentials now.
The Github Actions workflows, if you choose to merge them, triggers build and publish to Artifactory for all commits pushed to master branch as well as tags. One caveat is that you must not push (or tag) a commit with release versions twice. The second workflow will fail since overwriting versions in Artifactory is not allowed.
Pull-requests (to all branches) as well as pushed commits to non-master branches will just trigger a "mvn package" - not a publish to Artifactory.

andersharrisson-ess:
This PR introduces an alternative to travis-ci, so obviously there is overlapping functionality. If you decide to move onto Github Actions, the .travis.yml should be removed once the Github Actions workflows have been verified.

andersharrisson-ess:
Thank you for the merge, @shroffk 
I can see that artifacts are being correctly pushed from Github Actions to ESS Artifactory now. You can safely remove the .gitlab-ci.yml file, since we no longer need our own CI at ESS.

Issue #1134, opened 2020-02-11
Formula Functions


shroffk:
Add the formula functions from the eclipse cs-studio to the new core-pv.

[X] Extend ‘FormulaFunction’ to describe arguments, so that auto-completion can be supported. Idea: replace `getArgumentCount()` returning 2 with `getArguments()` returning `[ “base”, “exponent” ]`
[X] Implement auto-completion of function names, showing their arguments
[ ] re implement as many of the formula functions as possible. Contribute using the new formula function SPI in core
[ ] Create a utility view for all the formula functions. This is useful for users.
[ ] Add documentation

kasemir:
`getArguments()` returning `[ "arg1", "..." ]` means: Needs one argument, plus maybe more.
Just `[ "arg1", "..." ]` means: Zero or more args.

kasemir:
Please create PR for added category soon, since that impacts everybody who's using the SPI

shroffk:
I thought we should work on this branch and then make a PR

shroffk:
I will create a PR and then keep this branch open for us to co-ordinate our work on

kasemir:
Thanks.

shroffk:
Hello,
It seems like with the use of String... in concat and highestSeverity, we run into some issues with Scanner. 
The scanner does not like "..." it breaks out of the argument parsing and then throws an exception since it is finding a "." instead of the ")"

@kasemir could you advise on the best way to clean this up

georgweiss:
The parsing of a formula like rescale(`sim://const(1,2,3)`, 20, 0) will call the formula implementation multiple times. The first time the first argument is set to a default VDouble NaN value. Second time it is called with the correct type and value.  The consequence is that checking for argument types becomes tricky. If the first argument is not an array, the implementation can throw an exception, but that would be logged. Or the implementation can check type and return null if it is not the expected, but ultimately this results in the value not being computed without any further information (except possibly in the log).

Any thoughts on this?

georgweiss:
Another thing: formulas are registered by name only, which does not allow for "method overloading". We could maybe append argument list to the key?

kasemir:
Kunal: Do you mean scanner for parsing formula for evaluation, or parser for completion proposals? When I asked about `concat` in yesterday's hangout, you sounded as if it basically evaluates, so I assume the issue in in the proposal lookup? That was implemented when there was no actual var-arg function. Should have implemented the concat while working on proposals...

[...]

PR #1150, opened 2020-02-13
Formula: Var arg, unit test

kasemir:
Formula proposals (autocomplete) handle var args from 'concat' example.

#1134


kasemir:
Leaving open in case people still like to add function implementations on the branch...

PR #1151, opened 2020-02-14
Display Editor: Points dialog commit changes on losing focus

pavel-ch:
See  #1042 - continuing with points dialog


pavel-ch:
For Macro dialog, there is a kind of conflict between current implementation with automatic edit of the next cell through ENTER, and losing focus commit behavior. Postponing for the moment to wait for the right idea ...

georgweiss:
Well, I spotted the same conflict and decided it wasn't worth the effort to fix it.

kasemir:
In the macro editor, the ENTER support is a real asset since I usually need to set more than one macro, so name ENTER value ENTER name ENTER value .. is convenient.

Maybe the focus handling needs a flag to know if focus is lost as a result of ENTER.

Issue #1000, opened 2019-12-09
X11 middle mouse button 'selection' copy/paste


kasemir:
User commented that we don't support the middle mouse button 'selection' copy/paste on Linux.

True.

It's fundamentally not supported by JavaFX:
https://bugs.openjdk.java.net/browse/JDK-8088117]
https://stackoverflow.com/questions/30032290/javafx-mouse-clipboard-does-not-work-in-unix/31278321
https://stackoverflow.com/questions/21844255/how-can-i-read-the-x11-clipboard-from-a-javafx-application

It is supported by AWT, and adding this to `ContextMenuPvToClipboard` would publish PV names not only on the normal copy/paste clipboard but also the X11 selection buffer:
```
import java.awt.Toolkit;
import java.awt.datatransfer.ClipboardOwner;
import java.awt.datatransfer.StringSelection;

...
final java.awt.datatransfer.Clipboard x11_sel = Toolkit.getDefaultToolkit().getSystemSelection();
if (x11_sel != null)
    x11_sel.setContents(new StringSelection(text), (board, lost_content) -> {});
```

But while that "works", it's unclear if interfacing with both the AWT and the JavaFX UI this way might have negative side effects. Also doesn't cover our most interesting use case:
1) Operator middle-clicks on some PV in EDM or ... to get it into the X11 selection buffer
2) Operator would like to paste that PV into Data Browser or newly developed display

In those cases, JavaFX text fields which are used to enter the PV name would need to react to middle clicks, which they don't out of the box.

For now leaving this as a Linux-specific issue that's not generally supported, so users need to go via the normal copy/paste (Ctrl-C, Ctrl-V) mechanism.

kasemir:
Fixing the most common use case:
Site uses MEDM or EDM or ... where PV names can usually be copied into the X paste buffer via middle button click. PVs are _only_ copied into the X paste buffer, not the usual clipboard.
Pasting the PVs into Phoebus PV name text fields used to require first pasting them into e.g. an X text editor, then using Ctrl-C to copy into clipboard, finally being able to Ctrl-V paste into Phoebus PV name text field.

On X11/Linux, we now support a plain middle-button-click in Phoebus PV name text fields to paste text from the X11 paste buffer.

This helps when using both old and new tools on Linux, for example in a control room.
[...]

PR #1156, opened 2020-02-19
Connected to #1060 / On X11, support pasting PV names via middle button

pavel-ch:
See #1000. On some platforms (Windows), AWT API Toolkit.getDefaultToolkit().getSystemSelection() can be null.


kasemir:
Looks like nobody pushed the middle button on Windows since 10-2017 ..  Thanks!

pavel-ch:
Not that long, since January 10, 2020 only :-) ... I have seen it but did not care till today...

kasemir:
Jan 10 was the `paste` counterpart (XPasteBuffer, which does check for Mac, Linux vs. Windows). The `copy` where you saw the NPE goes back to 2017.

PR #1158, opened 2020-02-21
Display Editor: context menu actions

pavel-ch:
- Expand All Tree Items in WidgetTree
- Collapse All Tree Items in WidgetTree
- Duplicate Widgets action in Display Editor

See #679


kasemir:
You could simplify it a little by using `TreeHelper` to expand/collapse, recursing through the subtree items. Otherwise looks good.

kasemir:
Thanks for the TreeHelper update!

PR #1160, opened 2020-02-21
Issue 1146

georgweiss:
This eliminates code duplication and consolidates as much as possible to one single VTypeHelper. However, remaining unchanged code needs more analysis, preferably by developers possessing the knowledge on how the helper methods are used.


kasemir:
This shares at least one commit with PR #1147

georgweiss:
Sorry, missed that one. Suggest we scrap 1147.

kasemir:
Confused.
Why does a PR about VTypeHelper include one commit of the doc re-org?
This PR simply copied the external docs in, step 1 of #1147 .
But it didn't do the other steps of #1147, like updating links now that the docs are in a new place.
For starters, https://github.com/ControlSystemStudio/phoebus/blob/master/docs/README.md that explains how to build the documentation still refers to the old location.

georgweiss:
I share your confusion, my work was done on branch Issue_1146, created by @shroffk.

georgweiss:
What if I remove those items contributed by step1 of 1147?

shroffk:
> Why does a PR about VTypeHelper include one commit of the doc re-org?

Sorry, this branch originated from a branch which had the first commit about the documentation.
Once the #1147 is merged this should resolve itself.

> But it didn't do the other steps of #1147, like updating links now that the docs are in a new place.

This is a work in progress.

kasemir:
> What if I remove those items contributed by step1 of 1147?

Yes, please remove the partial import of the ongoing doc re-org.

georgweiss:
OK, will do.

georgweiss:
On second thoughts... If those files in the doc module are harmless, why should they be removed only to be added later?

kasemir:
Right now I would be unclear where one would add info, or even just fix a typo
(see https://github.com/ControlSystemStudio/phoebus/blob/master/docs/source/authorization.rst: 
"interating" needs to be "interacting"):

a) In the original copy?
b) In the new copy (which isn't included in anything at this point)?
c) Both?

georgweiss:
Not sure what you are trying to say. Please clarify.

pavel-ch:
With this PR, Eclipse build is broken mostly because of missing core-vtype in .classpath, for
app-databrowser
app-display-convert-edm
app-pace
app-pvtable
app-pvtree
core-formula
core-pv
service-alarm-server
service-archive-engine
service-scan-server

georgweiss:
I'd like to settle the question on whether to keep the doc contributions first.

pavel-ch:
Do whatever you need, from my side it is a reported issue only. I continue using checkout before this commit temporarily.

kasemir:
@pavel-ch , you're correct. The Eclipse build is broken, the ant build is broken, and we have an unused copy of the 'doc' setup from the still ongoing #1147/#1141 work.
I'll fix it.

georgweiss:
I would appreciate if we could come to a joint decision on what to do with the doc module.
As for Eclipse specific .classpath files... Why do we manage them in the repo? I know they have been there for a while, but in general they are not required by the project to build.

kasemir:
The doc folder and how documentation is built is handled on  #1147/#1141. This PR is about core-vtype, and should not touch the otherwise unrelated handling of docs.

kasemir:
Don't worry about the .classpath files. If you don't need them, you can simply ignore them. They're there for those who use Eclipse, just like we include nbactions.xml for those who use netbeans and Intellij hints in the Readme for those who prefer Intellij.

georgweiss:
Well, if there are complaints about Eclipse not being able to build due to incorrect .classpath files, then I cannot ignore them. I'd be happy to update them, but I still think IDE specific stuff should not be managed in the repo.

kasemir:
Apart from the build issues: When you build with maven, does it actually work?
For me, the data browser doesn't show any data:
mvn clean install, run, open data browser, add sim://sine -> no plot

georgweiss:
I'm seeing the same in data browser, investigating it.
But maven build works, I check it before PR, and Travis checks it.

kasemir:
The build works, but the product doesn't plot.
Data browser has data, but cannot plot it.

db

georgweiss:
Found the issue, will fix promptly.

PR #1167, opened 2020-02-25
Issue 1146 Fix for  VTypeHelper error detected i databrowser

georgweiss:


kasemir:
I still see the same issue: Data browser sample view shows that there is data, but plot does not show a trace

db

kasemir:
The new PlotSampleTest fails with expected 7.7 but was NaN:

db2

georgweiss:
Works fine at my end.

georgweiss:
Is VTypeHelper rebuilt?

kasemir:
Turns out yes, it was a rebuild type of issue. The test now passes, and the plot also works again with both 'raw' and 'optimized' samples.

db3

Thanks!

PR #1171, opened 2020-02-26
Issue 1141 2

shroffk:
A cleaned up version of the PR #1147


georgweiss:
Builds fine using both mvn and make (on Mac)

Issue #1045, opened 2020-01-05
Ant build broken for alarm logger?


MaticPogacnik:
I was using alarm logger maven build for few months now, and everything worked ok, with exception of few warrnings:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.springframework.cglib.core.ReflectUtils$1 (file:/home/mpogacnik/Phoebus/release4.6.0/phoebus-parent-4.6.0/services/alarm-logger/target/lib/spring-core-5.0.9.RELEASE.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of org.springframework.cglib.core.ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
2020-01-05 11:29:29.146  WARN 998 --- [pool-1-thread-2] o.a.k.clients.admin.AdminClientConfig    : The configuration 'es_port' was supplied but isn't a known config.
2020-01-05 11:29:29.142  WARN 998 --- [pool-1-thread-1] o.a.k.clients.admin.AdminClientConfig    : The configuration 'es_port' was supplied but isn't a known config.
2020-01-05 11:29:29.152  WARN 998 --- [pool-1-thread-1] o.a.k.clients.admin.AdminClientConfig    : The configuration 'date_span_value' was supplied but isn't a known config.
2020-01-05 11:29:29.152  WARN 998 --- [pool-1-thread-1] o.a.k.clients.admin.AdminClientConfig    : The configuration 'es_host' was supplied but isn't a known config.
2020-01-05 11:29:29.152  WARN 998 --- [pool-1-thread-2] o.a.k.clients.admin.AdminClientConfig    : The configuration 'date_span_value' was supplied but isn't a known config.
2020-01-05 11:29:29.153  WARN 998 --- [pool-1-thread-1] o.a.k.clients.admin.AdminClientConfig    : The configuration 'alarm_topics' was supplied but isn't a known config.
[...]

PR #1178, opened 2020-02-27
#1045 making all the spring dependency versions the same

shroffk:


georgweiss:
I take it the affected alarm services are verified? I did the same change at my end and noted that a Spring Boot application is bundled with Tomcat 9. Previously - with differing versions of Spring Boot - the build pulled in Tomcat 8.

shroffk:
@georgweiss Yes, I verified that the alarm logger and the alarm config logger continue to work with 2.1.5.
This should reduce the number of dependencies with difference version needed.


@geng1 
I have also added the show channel info back 
image

image

shroffk:
Additionally,
If you include the channel finder util to your product. The context menu has an action which allows you to query for that pv in channelfinder

PR #1179, opened 2020-02-28
Issue_1162 Update Windows bat file to eliminate hard coded version

georgweiss:
Decided to not use Maven filtering as that would also require a change in the assembly configuration. 
Modified the bat file to behave lite the *nix counterpart, i.e. set jar file identity based on dir (ls) command.

NOTE: Verified on Windows 10 only!


kasemir:
Good, thanks!

Issue #1182, opened 2020-03-02
Online help no longer includes plain html from doc sources


kasemir:
The documentation contains some plain html, like the javadoc for display builder scripts.

https://github.com/ControlSystemStudio/phoebus/blob/master/docs/README.md still describes that it's supported:

```
# Build the Javadoc, i.e. html files to be included in the manual
( cd phoebus/app/display/editor; ant -f javadoc.xml clean all )

# Building the documentation
( cd phoebus/doc; make clean html )
```

.. but in the generated manual, those links don't work.
It's not just that the path to html that's copied in changed, the content seems simply missing.

Same when calling `mvn verify -P sphinx -N`

kasemir:
#1141

shroffk:
Updated the conf.py to copy generated html

The generated html is copied into 
build/html/raw_html/<module>/<path>...

e.g.
![image](https://user-images.githubusercontent.com/2111304/75703460-87626600-5c85-11ea-8639-046ce645afdf.png)

There is a change in the way the `<a href="html/generated/index.html">Java Doc for scripts</a>` is resolved.

Previously it resolved to to:
`phoebus/app/display/editor/doc/html/generated/index.html`

now it resolves to:
`phoebus/docs/build/html/html/generated/index.html`

The reason behind this is because the use of applications_links.rst instead of the applications.rst

shroffk:
While the old way in which the applications.rst was used meant the git would always see that file as modified which each run. I still feel like we might want to revert to the old way of modifying the application.rst directly to simply the linking of raw html

kasemir:
Your pick. All I want is to have online help.

shroffk:
Noted. Thank you

kasemir:
> Previously it resolved to to:
> phoebus/app/display/editor/doc/html/generated/index.html
> now it resolves to:
> phoebus/docs/build/html/html/generated/index.html

What if both `display/editor/doc` had some plain html, and `other/whatever/doc` also has some plain html. Would they both be copied into the same `docs/build/html/html`?

shroffk:
>  Would they both be copied into the same docs/build/html/html?

No this would surely not work. The static html has to go to a module specific location so that multiple modules can contribute their own html and other static content.

shroffk:
The last commit does the following
Copy the doc folder mainiting the same structure as the phoebus src.
Creates the application.rst and services.rst 
[...]

PR #1185, opened 2020-03-02
Issue #1182

shroffk:


shroffk:
@kasemir Please verify if this works as described in the ticket.

kasemir:
Yes, works now, thanks!

PR #1187, opened 2020-03-06
CSSTUDIO-1216 Embedded display always initiated to height=200

georgweiss:
Removed setting size of embedded display widget as that overrides the default values.


kasemir:
Good.
The default size needs to be set in the widget constructor. Changing it in `defineProperties` will have widgets come up with that initial size, but it's not the 'default' and that causes the problems you saw.

Turns out Tabs and Nav. Tabs also adjust their size in `defineProperties`, I'll fix that.

PR #1191, opened 2020-03-09
Display Editor: mark editor Tab as [Edit]

pavel-ch:
Looks ugly but I do not see any better solution. Tried using app icon but a) having troubles with Tab not showing label and b) must be supported by (all) other apps.
See #679


kasemir:
Yes, the thing is that the text in the tabs is already using 'graphics' in form of a `Label`.
That's because we need something to which we can attach drag listeners to support moving the tabs around. To add an icon, we'd have to set the 'graphics' of the tab to for example a `Group` of `ImageView` and `Label`, then make the group react to drags.

PR #1196, opened 2020-03-10
Linux launcher shell script bug, and improvement

jeonghanlee:
@kasemir, 

  Please look at this pull request, which is a simple, and tiny update on phoebus.sh in order 
1) to remove the following like error 
```
./phoebus.sh --help
./phoebus.sh: 34: [: x--help: unexpected operator
```
2) to put `-help` as the foreground instead of background. 

In addition, I added two more lines to allow the script know where it is. So, one can run this script outside the directory. 

Best,
Han

### DIFF
--- a/phoebus-product/phoebus.sh
+++ b/phoebus-product/phoebus.sh
@@ -4,13 +4,15 @@
 
 # When deploying, change "TOP"
 # to the absolute installation path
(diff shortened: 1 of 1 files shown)


kasemir:
Good, thanks!

PR #1208, opened 2020-03-13
ctrl+mouse scroll to zoom in fix #1194

### LINKED ISSUE #1194: Display Runtime: Ctrl+Mousewheel to adjust zoom
Ctrl+ scroll mousewheel down on an opi file zooms out the element but Ctrl + scroll mousewheel up does not zoom in the element.

tanviash:
#1194 On some linux setups, an extra scrollwheel event is seen with delta = 0 which causes zoom to work in just one direction. The fix is to ignore such erroneous scrollwheel events.


kasemir:
Thanks for finding and fixing this!

On most linux setups, we receive mouse scroll wheel events with delta=40 or delta=-40.
On systems that see issue #1194 we receive an extra event with delta=0.
In the original code that resulted in a double-step-zoom in one direction and no effective zoom in the other direction (because of a quick zoom in/out).

PR #1223, opened 2020-03-19
ColorFontUtil and ConsoleUtil classes

tanviash:
The PR is to make ColorFontUtil class available to scripts to provide compatibility with legacy scripts. Also, it ports ConsoleUtil class in a cleaner way than done in the PR before which extended the patchScript method.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/ColorFontUtil.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/ColorFontUtil.java
@@ -1,0 +1,96 @@
+package org.csstudio.display.builder.runtime.script;
+
+import org.csstudio.display.builder.model.properties.WidgetColor;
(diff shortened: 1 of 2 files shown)


tanviash:
> There are a number of color definitions in org.csstudio.display.builder.model.persist.NamedWidgetColors that overlap with ColorFontUtil. Could we consider merging these somehow to avoid code duplication?

@georgweiss Hi Georg, the aim is to only provide legacy scripts that already use this class backwards compatibility but going forward, we want new scripts to use the class WidgetColor. Hence I don't think it's a good idea to provide more predefined colors as part of this class.

kasemir:
You added the warning to getFont, so it's printed every time.
The MacroHandler example placed the warning in a static initializer, so it's only printed once when first accessed. ConsoleUtil could use the same warning.

kasemir:
Thanks!

PR #1227, opened 2020-03-20
CSSTUDIO-1229 Fix copy-paste issues in alarm config logger

georgweiss:
Several program arguments were mapped to same property.


shroffk:
thank you very  much :)

Issue #1231, opened 2020-03-23
Display Navigation


shroffk:
At the developers meeting it was discussed that having a tool or application that created a list/graph/tree of all the opi files linked to a particular opi might be useful.

This was aimed at helping with the migration from boy to display builder.

I felt like such an app/tool would be useful in normal circumstances to, when developing and managing screens it is always useful to understand the links from that screen.

I have created an application called display navigation view which parses a given opi screen and produces 
1. a unique list of all linked files
2. a tree view of all linked files.

This is the first iteration and we can all more feature as we need them.

![list-link](https://user-images.githubusercontent.com/2111304/77336186-8d37ee00-6cfd-11ea-9970-4fa5a5142826.PNG)

![link-tree](https://user-images.githubusercontent.com/2111304/77336192-8f01b180-6cfd-11ea-8c4b-634c064a48a8.PNG)

shroffk:
@tanviash let me know if this work for you

mark0n:
I like the idea of generating a tree that you can navigate (hide subtrees). I'm wondering how you are visualizing if your graph is cyclic (remember we want to find cycles)? Have you tried throwing your data into [GraphViz](https://graphviz.org) to get a graphical representation? There seems to be a [Java interface](https://github.com/nidi3/graphviz-java) as well.

shroffk:
@mark0n @tanviash while the graph is fun the current app...though not pretty... fulfils the core requirements.
Can we close this ticket and open a new one for the graphing

shroffk:
oops closed by mistake... you can reopen if necessary

PR #1232, opened 2020-03-23
Issue #1231

shroffk:
Application to show display navigation


kasemir:
Neat!
Can it read the model and process it in a background job?

shroffk:
> Can it read the model and process it in a background job?

Currently it does not... I actually have a note in the javadoc's which mentions that it should be offloaded onto a different thread.

I was planning on playing with the javafx.concurrency to implement this.

kasemir:
Alright, good.

shroffk:
I moved the list view population off the UI thread.

For the tree view this is a little tricky, since the children are found lazily... i.e. this is when the user clicks on the node.

shroffk:
@kasemir if you have a better suggestion for the tree solution let me know.
Otherwise I am ready to merge this.

kasemir:
By all means go ahead and merge.
This is for people who edit/port displays.
It won't affect end users who run displays in the control room.

Not sure how to best handle the tree update on multiple threads.
For the file browser, I have 'getChildren' first return the old/empty data, since you need to return something right away. A background jobs is scheduled, which then updates the child list.

https://github.com/ControlSystemStudio/phoebus/blob/master/app/filebrowser/src/main/java/org/phoebus/applications/filebrowser/FileTreeItem.java#L64

georgweiss:
Haven't tried his out until now, looks useful!
Just two question... In the attached screen shot the parent directory of the bob file being analyzed is listed as a child node. Intentional?
Also wonder if we should consider adding nodes representing other resources, e.g. images.
Screenshot 2020-03-25 at 10 04 23

shroffk:
Just two question... In the attached screen shot the parent directory of the bob file being analyzed is listed as a child node. Intentional?

yes,
this is an example with cyclic links.
The root has a link to a child which has a link to the root again.

The screenshot is of a set of files I use for unit testing 🙂
________________________________

georgweiss:
Thanks, I trie to delete the comment, it was not complete... In any case, what about also including other resources in the tree view, e.g. image files.

shroffk:
> By all means go ahead and merge.
> This is for people who edit/port displays.
> It won't affect end users who run displays in the control room.
> 
> Not sure how to best handle the tree update on multiple threads.
> For the file browser, I have 'getChildren' first return the old/empty data, since you need to return something right away. A background jobs is scheduled, which then updates the child list.
> 
> https://github.com/ControlSystemStudio/phoebus/blob/master/app/filebrowser/src/main/java/org/phoebus/applications/filebrowser/FileTreeItem.java#L64

Thanks, 

I had tried something similar with Tasks... however in my case the 'isleaf' requires 'getChildren' to be processed. In the file browser you can check the item itself to determine 'isleaf' which is unfortunately not possible here.

I will look into other ideas.

kasemir:
If the `isLeaf` value changes, you could still start with `isLeaf()` returning `false` and empty `getChildren()`. When the data then arrives, force a refresh via `TreeHelper.triggerTreeItemRefresh()`

PR #1234, opened 2020-03-24
DataUtil class for legacy scripts 

tanviash:
DataUtil class and its APIs are used by legacy scripts hence this PR is provide backwards compatibility for these legacy scripts.


tanviash:
> To load this class into JavaScript, do you patch the 'import' to handle the differences between the rhino and nashorn JS implementations?

@kasemir I manually patched the scripts to fix importPackage.

PR #1240, opened 2020-03-30
Csstudio 1232 Added logbook service availability check

georgweiss:
When user initiates creation of a new logbook entry, the application does not verify that the logbook service is available and that it has any configured log books. This commit adds a check, and if it fails an error alert is shown and the log entry dialog is suppressed. There are some things to consider:
1) Since the availability check dispatches a request to the remote logbook service, there is a small delay incurred before the log entry dialog is shown.
2) In order to handle connect timeout when the availability check is performed, the property org.phoebus.olog.api/connectTimeout or org.phoebus.olog.es.api/connectTimeout (in ms) must be set to something that is >0. Note that this timeout period will then apply to all calls to the remote logbook service. It may be better to set the default value in the properties files to >0, say 2000 (ms).


kasemir:
Does this switch the logbook body editor to HTML, with a default content that refers to                 `"");` ?

georgweiss:
OK, my bad, that was only a test. Will fix.

PR #1247, opened 2020-04-01
Display 'pv_name' macros on graphics

kasemir:
Unlike 'Monitor' or 'Control' widgets, 'Graphics' widgets are meant to be static in nature, they don't have a PV. Still, legacy *.opi files sometimes contain a "pv_name" property which is then basically used as a macro in attached scripts or rules. At least that's how FRIB decided to use it. To allow import, adding macro to all 'Graphics', and importing legacy "pv_name" property into the macros.


tanviash:
@kasemir Thank you very much Kay.

PR #1275, opened 2020-04-21
CSSTUDIO-1264 Bug fix logbook list view

georgweiss:


shroffk:
Hi, 
What was the bug here...would like to know in case I have it in other places

georgweiss:
It's my bug introduced when I migrated the attachments view to fxml.

PR #1278, opened 2020-04-24
Csstudio 1268 Alarm logger: make thread pool size configurable

georgweiss:
Thread pool size was hard coded to four, which allowed logging of only two alarm configurations. Added possibility to specify size in either properties file or on command line.


georgweiss:
Thanks

PR #1279, opened 2020-04-24
CoreUI: JFX dialogs: set dialog Owner to get correct dialog icons on Windows

pavel-ch:
See #896 #897 for motivation, this PR is targeted to Core UI. Usage example: `org.csstudio.trends.databrowser3.ui.AddPVDialog`


kasemir:
Thanks!

Issue #1235, opened 2020-03-25
Save and Restore: PVs disconnected


dsoklic:
In the Save and Restore application sometimes the PVs don't want to connect. I can however always access the PVs in other places, like the PV Table. The error seems to occur randomly. I can run Phoebus and everything works fine and then at the next run without changing anything the PVs can not be accessed, but only from Save and Restore.

I have attached a screenshot where this can be seen and also a screenshot of the PV table where the same PV does connect.

![pvtable](https://user-images.githubusercontent.com/217755/77531790-cb671600-6e93-11ea-81dd-0b8f82376d4f.png)
![sr](https://user-images.githubusercontent.com/217755/77531793-cc984300-6e93-11ea-99b6-f2ad71c314d3.png)

georgweiss:
Please attach output of terminal.

dsoklic:
Thank you for your quick reply. Here is the entirety of today's log:

[phoebus_save_and_restore.log](https://github.com/ControlSystemStudio/phoebus/files/4380737/phoebus_save_and_restore.log)

georgweiss:
I'm sorry, need to clarify: need the output from the CS Studio terminal. The service does not connect to PVs.

dsoklic:
Sorry about that. Here is the output of the terminal: 
[phoebus.log](https://github.com/ControlSystemStudio/phoebus/files/4381062/phoebus.log)

georgweiss:
The log suggests two different problems:
1) You are trying to save a snapshot under the same name as an existing one. This is not allowed.
2) You are trying to restore a snapshot, but the PV cannot be write connected. Is the PV writable or read-only?

dsoklic:
I don't think this is connected with the disconnected PV issue. This issue already exists when I only create a saveset and open it up. The Live Setpoint is already disconnected at this point.

Here is a log where nothing else is pressed: 
[phoebus2.log](https://github.com/ControlSystemStudio/phoebus/files/4385722/phoebus2.log)

georgweiss:
You see this behaviour irregularly, correct? When this happens, are you able to connect to the PV through an epics tool like caget?

geniejhang:
I also saw VEnum type PVs are disconnected all the time while I'm testing.
It looks like the PV with issue is also VEnum type.

Try this one with softIoc.

record(bi, "STATUS") {       
  field(DTYP, "Soft Channel")
  field(ZNAM, "OFF")
  field(ONAM, "ON")
[...]

PR #1280, opened 2020-04-25
VEnum support on both application and service

geniejhang:
Resolving Issue #1235 
Confirmed with https://github.com/epics-base/epicsCoreJava/tree/release/7.0

### DIFF
--- a/app/save-and-restore/ui/pom.xml
+++ b/app/save-and-restore/ui/pom.xml
@@ -40,22 +40,22 @@
 		<dependency>
 			<groupId>org.epics</groupId>
 			<artifactId>gpclient-core</artifactId>
-			<version>1.0.1</version>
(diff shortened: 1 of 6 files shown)


georgweiss:
Thanks @geniejhang for your patience in this matter, but I'm afraid this PR is not sufficient to support VEnum for both CA and PVA in save-and-restore. I am working on the issue and will hopefully create PR soon, but there is a dependency to epcisCoreJava which must be released in a new version before we can do the required changes in CS Studio.

Suggest we reject this.

shroffk:
Hello,
I will try to get the release out next week. 
Sorry about the delay...

geniejhang:
Thank you, @georgweiss 
At least I can learn something from the comparison to the released version.

PR #1284, opened 2020-04-27
Separating snapshot list to ListView from the TreeView

geniejhang:
One of our users requested take apart snapshot lists from the TreeView because the number of snapshot increases as time.
So, it becomes like the attached.
I'm pretty sure this should happen to all users sooner or later.

### DIFF
--- a/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/BrowserListCell.java
+++ b/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/BrowserListCell.java
@@ -1,0 +1,87 @@
+/**
+ * Copyright (C) 2019 European Spallation Source ERIC.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+ */
+
+package org.phoebus.applications.saveandrestore.ui;
+
+import javafx.fxml.FXMLLoader;
+import javafx.scene.control.*;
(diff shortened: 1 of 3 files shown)


georgweiss:
Frankly I do not see how this solves the problem of having too many snapshots. Moving them out of the tree into a list only moves the problem from one part of the UI to another. The nice thing about the tree is the ability to collapse nodes you are not interested in. Also, if the issue is a large number of snapshots you should maybe consider removing obsolete items.

Having a large number of configuration poses the same question: should they be separated into a list as well?

Can you please reconsider?

geniejhang:
Honestly speaking, I don't know the use case of this application well.
However, I agree on that snapshot lists can be lengthier than saveset and folder lists.

Say, one needs to look through savesets and each of which the latest snapshot.
If the snapshot list is integrated in the tree, one needs to collapse and expand each tree
or expand and scroll to reach the next saveset.

This modification is just auto-collapsing each saveset when clicked others providing less scrolling.
Plus, I believe people tend not to remove obsolete stuff if there were to have a tiny chance to give benefit later.

Anyway, this pull request is always optional for mainstream package.
One other way is to make this feature optional by putting toggle button.
The choice is yours.

Thank you for your opinion.

georgweiss:
In general I do not think the tool (Phoebus) - or any other software - should attempt to solve issues that are actually workflow issues like creating a lot of data and not removing it when no longer needed.

If you want to use this in a dedicated branch I do not object to that. But please do not merge to master branch. Alternatively, add a preference that enables this behavior and that would default to "false".

geniejhang:
Okay. I should have taken the optional way.
I thought this one might be useful to the others as well because that was the view in CS-Studio version of save and restore app.
Anyway, I'll try to update the optional one later.
Thanks.

georgweiss:
Yes, I know that this type of layout was used and I decided - based on discussions with my colleagues at ESS -  to arrange data in a tree instead for better overview and usability. I found the "old" UI layout difficult to use as it rendered correlated data in three flat lists.

geniejhang:
I see. Yeah, I was also confused when I first saw three-panel view.
I like the tree view. It's more intuitive :)

Anyway, I'll make an update later for the option.
Do you have any other plan for S&R app that you can share with me?
I'm writing code for FRIB-side requirement on S&R app.
If there're some overlap, I can do or leave it.

georgweiss:
If you can share your email-address I can send you my back-log.

geniejhang:
Thanks. It's changj@frib.msu.edu

georgweiss:
@geniejhang, does this PR contain a non-default option to use the list instead of tree nodes? If not, I suggest we close this and you can create a new PR where you make the list optional.

geniejhang:
@georgweiss Yeah. My original plan was to push another commits here, but flu pushed everything delayed. I'll create another PR. Thanks.

PR #1285, opened 2020-04-27
Feature: Support to disable/re-enable alarm email notifications

tanviash:
This PR is to handle a feature request from FRIB users to provide a way through Alarm UI to temporarily disable/re-enable alarm email notifications for an alarm tree. The changes add a new button in "Alarm Table" view right next to the maintenance mode button which can be used to disable/re-enable alarm email notifications for that alarm tree. This button is made visible if the setting "org.phoebus.applications.alarm/disable_notify_visible=true" is added to the ini file. Otherwise, the new button is invisible by default.

The changes use the alarm_cmd kafka message to transmit the info about the disable/enable of the notifications to the alarm server. Alarm server updates the alarm_state message by updating a new field called "notify" to show the if the alarm email notifications are in disabled state.


app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
```diff
@@ -112,6 +112,9 @@
     /** Nag period in seconds */
     public static final long nag_period_ms;
 
+    /** Disable notify feature */
+    public static final String disable_notify_visible;
```
kasemir, line 116: disable_notify_visible should be a boolean flag, not a string with values 
true", "false".

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/table/AlarmTableUI.java
```diff
@@ -277,7 +282,12 @@ private ToolBar createToolbar()
         search.setTooltip(new Tooltip("Enter pattern ('vac', 'amp*trip')\nfor PV Name or Description,\npress RETURN to select"));
         search.textProperty().addListener(prop -> selectRows());
 
-        return new ToolBar(active_count,ToolbarHelper.createStrut(), ToolbarHelper.createSpring(), server_mode, acknowledge, unacknowledge, search);
+	if (AlarmSystem.disable_notify_visible.equals("true"))
```
kasemir, line 285: disable_notify_visible should be a boolean that can simply be compared as `if (...disable_notify_visible)`

kasemir:
For the alarm state update messages, I would like to avoid adding `notify: true` to every message. Maybe only add for the `notify: false` case, and otherwise default to true, so `notify: true` is not actually included in the message? That way, there is zero added traffic in the 'normal' state.

tanviash:
> For the alarm state update messages, I would like to avoid adding `notify: true` to every message. Maybe only add for the `notify: false` case, and otherwise default to true, so `notify: true` is not actually included in the message? That way, there is zero added traffic in the 'normal' state.

@kasemir I think this already happens since in in JsonModelWriter.toJsonBytes, I set the "notify" only when the notifications are disabled else nothing is written to it.

 if (disable_notify)
 {
                jg.writeStringField(JsonTags.NOTIFY, JsonTags.DISABLE);
  }

Let me know if you are meaning something else here. Thanks.

kasemir:
That's it exactly, thanks for confirming.

Can those variables that are effectively yes/no flags become consistently boolean, not string?

app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
tanviash: Field updated to boolean

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/table/AlarmTableUI.java
tanviash: Done

kasemir:
The alarm logger now has test failures because of the updated message content:
```
[ERROR]   MapperTest.AlarmStateMessageTest:43 Failed to map the AlarmStateMessage expected:<...nt_message":"NONE","[]latch":false}> but was:<...nt_message":"NONE","[notify":true,"]latch":false}>
```

kasemir:
Somewhat unclear why a test needs to be updated with "notify:true" when we generally assume to not see the 'true' case. Only "notify:false" should appear in messages....

kasemir:
Haven't tried it, but appears OK now.

tanviash:
> Haven't tried it, but appears OK now.

Thank you Kay for the code review & approval.

PR #1286, opened 2020-04-29
Csstudio 1264 Improved error handling for logbook entry submission

georgweiss:
The logbook entry dialog now shows errors if submission fails (e.g. due to bad credentials or connectivity issues). This is verified for olog-es, and to some extent olog.


core/logbook/src/main/java/org/phoebus/logbook/LogEntrySubmissionResult.java
```diff
@@ -0,0 +1,31 @@
...
+package org.phoebus.logbook;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Builder
+@Data
+public class LogEntrySubmissionResult {
+    private LogEntry logEntry;
+    private int httpStatus;
```
kasemir, line 29: `httpStatus`: We cannot assume that all logbook implementations are based on web services.
For example, the SNS logbook makes entries via direct RDB access, so there is no 'httpStatus'.
A generic `String error` would work for everybody, or just an `Exception` where whatever error you want to convey can be in the message. When using an `Exception`, the logbook submission can simply throw an exception, no need to _return_ an exception, though.

georgweiss:
There is olog and olog-es, using different back-end stacks, both exposing a "web service". The clients have adopted different strategies to send requests and process the result. In view of these differences - that I find a bit unfortunate - this is the least painful I could come up with. I assume any client in Phoebus would prefer to use the web service and not connect to a database, which by the way does not even exist in the case of olog-es.
The reason I chose to include an exception and http status in the result is that I wanted to avoid implementing the same logic in two clients: why did the call fail, what error message should the UI show? So I put the logic in the common component.

kasemir:
> There is olog and olog-es

.. and the SNS logbook, and sites are permitted to implement their own logbook glue code.
The `core.logbook` API was designed to allow wrapping any of these without prescribing a specific implementation.

georgweiss:
OK, I did not realize there is another implementation using the same UI. Will come up with something different.

PR #1291, opened 2020-05-04
Csstudio 1241 Update of epcisCoreJava dependencies

georgweiss:
Fixes issues with VEnum and NaN in save-and-restore.


core/pva/src/main/java/org/epics/pva/client/MonitorRequest.java
```diff
@@ -159,6 +159,11 @@ public void handleResponse(final ByteBuffer buffer) throws Exception
 
                 // Decode type description from INIT response
                 final PVAData type = channel.getTCP().getTypeRegistry().decodeType("", buffer);
+                byte[] b = new byte[buffer.limit()];
```
kasemir, line 162: This printout in the PVA MonitorRequest looks like debug code that shouldn't be merged

georgweiss: You are quite right, I will fix.

PR #1303, opened 2020-05-07
Add MIT license file

georgweiss:
Need a license file in order to fulfil requirements for a Intellij Ultimate edition license. Listed copyright lines for contributing sites. This will over time become incorrect, question is if omitting copyright statements is more feasible.

### DIFF
--- a/LICENSE.html
+++ b/LICENSE.html
@@ -1,0 +1,259 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<!-- saved from url=(0042)https://www.eclipse.org/legal/epl-v10.html -->
+<html xmlns="http://www.w3.org/1999/xhtml"><head><meta http-equiv="Content-Type" content="text/html; charset=windows-1252">
+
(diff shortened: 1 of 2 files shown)


shroffk:
It turns out that there are ORNL files with EPL license.
ORNL is a member of the Eclipse foundation.... and thus feels strongly that we stick to the EPL license.

Currently, we will approach jetbrains and try to use EPL.

georgweiss:
What about copyright information? Do we need it? Do we want it?

shroffk:
I will add this to the next meeting agenda so all parties can express their opinions and preferences.
Currently, we can leave this as they are.

georgweiss:
The license as pointed out by @kasemir is in HTML on the Eclipse domain. Should we copy the content to LICENSE.md (which then renders a bit differently) or just add LICENSE.html?

shroffk:
@kasemir I hope this is fine

kasemir:
https://github.com/ControlSystemStudio/cs-studio has just a LICENSE file (EPL).
You now have both LICENSE.html (EPL) and LICENSE.md (MIT)

kasemir:
@kunal No, I don't thinks it's fine to have two licenses, one of which omits ORNL from the copyrights.

PR #1311, opened 2020-05-08
Databrowser AA Reader

tanviash:
DataBrowser
This PR is to add numeric to string conversion for alarm status field in Archiver Appliance Reader. Currently, the status field in DataBrowser is displayed as a number which is not very user-friendly as most users don't know what numeric value stands for. Displaying the status as a string will be helpful to understand the alarm type.

### DIFF
--- a/app/databrowser/src/main/java/org/phoebus/archive/reader/appliance/ApplianceValueIterator.java
+++ b/app/databrowser/src/main/java/org/phoebus/archive/reader/appliance/ApplianceValueIterator.java
@@ -332,41 +332,41 @@ public abstract class ApplianceValueIterator implements ValueIterator {
             return "HIHI_ALARM";
         } else if (status == 4) {
             return "HIGH_ALARM";
-	} else if (status == 5) {
+        } else if (status == 5) {
             return "LOLO_ALARM";
-	} else if (status == 6) {
+        } else if (status == 6) {
             return "LOW_ALARM";
-	} else if (status == 7) {
+        } else if (status == 7) {
             return "STATE_ALARM";
-	} else if (status == 8) {
+        } else if (status == 8) {
             return "COS_ALARM";
-	} else if (status == 9) {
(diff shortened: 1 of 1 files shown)


app/databrowser/src/main/java/org/phoebus/archive/reader/appliance/ApplianceValueIterator.java
```diff
@@ -313,4 +313,61 @@ protected static AlarmSeverity getSeverity(int severity) {
             return AlarmSeverity.UNDEFINED;
         }
     }
+
```
kasemir, line 316: The alarm status text to index mappings are defined in EPICS base, the C code running on the IOC.
For Java, the closest we have is in the Channel Access client library https://github.com/epics-base/jca/blob/master/src/core/gov/aps/jca/dbr/Status.java.

I would call  `gov.aps.jca.dbr. Status. forValue(int status)` instead of basically duplicating that in here.

app/databrowser/src/main/java/org/phoebus/archive/reader/appliance/ApplianceValueIterator.java
```diff
@@ -322,52 +324,9 @@ protected static AlarmSeverity getSeverity(int severity) {
...
-        } else if (status == 19) {
-            return "SIMM_ALARM";
-        } else if (status == 20) {
-            return "READ_ACCESS_ALARM";
-        } else if (status == 21) {
-            return "WRITE_ACCESS_ALARM";
-        } else {
-            return "UNKNOWN_ALARM";
-        }
+	String stat = String.valueOf(Status.forValue(status));
```
kasemir: I think you can use just
```
return Status.forValue(status).getName()
```
instead patching the toString representation. 


 getName()

tanviash:
@kasemir Thank you for reviewing the fix.

PR #1320, opened 2020-05-13
Support alarm auth rules per alarm server

tanviash:
This PR is to support 2 tasks.
1. Alarm auth rules like alarm_config, alarm_ack etc. per alarm server. See the auth.conf changes to know how to set these rules per alarm server. This is an optional change which means the existing way of alarm rules continue to work as before.  
2. New authorization rules for maintenance mode (alarm_mode) and alarm email notify disabling mode (alarm_notify).

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmUI.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmUI.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.phoebus.applications.alarm.ui;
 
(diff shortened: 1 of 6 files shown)


kasemir:
I like the general idea and the update to the `authorization.conf` format.
The changes to the alarm UI are small, and overall this accomplishes what you want.

The naming and the Authorization API change could be more obvious.
"Rule" suggests something far more elaborate, especially when it gets confused with the display's rule support. But I'm not sure what to name it exactly, so go ahead and merge this, we can than rename the API soon afterwards, won't affect the content of the `authorization.conf` nor the functionality.

tanviash:
@kasemir Thanks Kay for reviewing and approving the changes.

Issue #1324, opened 2020-05-13
Macros dialog editing behaviour improvement


shroffk:
Depending on where you click and the number of clicks users quite often find themselves in the situation that multiple clicks had resulted in the <Enter Name> to not only be deselected but the cursor has moved to the end of the default text. After which an edit cancel results in parsing error.

![image](https://user-images.githubusercontent.com/2111304/81853861-0d731680-952b-11ea-9a72-bbc7372036c1.png)

Suggested solution.
When an edit action is started the default text should be removed.

kasemir:
Might be easier to just remove the `Messages.MacrosTable_*Hint`.
Those in the know can still click on the cell to edit.
Those who don't see the "Add" button.

PR #1325, opened 2020-05-13
#1324

shroffk:


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/MacrosTable.java
```diff
@@ -253,8 +296,8 @@ private void fixup(final int changed_row)
         // Assert one empty row at bottom
         final int len  = data.size();
         if (len <= 0  ||
-            (data.get(len-1).getName().trim().length() > 0  &&
-             data.get(len-1).getValue().trim().length() > 0) )
+                (data.get(len-1).getName().trim().length() > 0  &&
+                        data.get(len-1).getValue().trim().length() > 0) )
             data.add(new MacroItem("", ""));
```
kasemir, line 301: I was actually proud of this formatting, how the data.get(.. lined up for the getName and getValue, but whatever...

shroffk: Sorry, as you can see I have set the brackets to follow your format. I shall fix this.

kasemir: These lines are a good example for auto-format simply not understanding what belongs together.

As for all the rest, the whole "<enter name>" stuff could also be removed, but now that you already put in the time to clear it, go ahead.

shroffk: For many, directly clicking on the table is easier than the button... do you see any errors in the code

kasemir: Hah, web page, can't just enter 'less than'. ...
Meant to say the whole "enter name" stuff .. could be removed, would be less code overall.
But again, go ahead, I've also often ended up with that 'hint' text in the name or value.

shroffk: thanks

PR #1331, opened 2020-05-14
Update config_name when change_handler is invoked

tanviash:
Currently AlarmSystem.config_name is final string, this PR updates config_name when change_handler is invoked so auth functions mayAcknowledge(), mayConfigure() etc. gets the current config name when it is changed from the default value.


kasemir:
Added comment to config_name and made volatile since we might read it from several threads.

PR #1337, opened 2020-05-20
Create or add to a saveset from ProcessVariable context menu.

geniejhang:
By default, the navigation look&feel the same as original save&restore.

Please review. @georgweiss

### DIFF
--- a/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ContextMenuCreateSaveset.java
+++ b/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ContextMenuCreateSaveset.java
@@ -1,0 +1,128 @@
+package org.phoebus.applications.saveandrestore;
+
+import javafx.scene.Parent;
+import javafx.scene.Scene;
+import javafx.scene.control.Alert;
+import javafx.scene.image.Image;
+import javafx.stage.Modality;
+import javafx.stage.Stage;
+import org.phoebus.applications.saveandrestore.model.Node;
+import org.phoebus.applications.saveandrestore.model.NodeType;
+import org.phoebus.applications.saveandrestore.service.SaveAndRestoreService;
+import org.phoebus.applications.saveandrestore.ui.saveset.SaveSetFromSelectionController;
+import org.phoebus.core.types.ProcessVariable;
+import org.phoebus.framework.preferences.PreferencesReader;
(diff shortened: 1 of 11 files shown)


app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ContextMenuCreateSaveset.java
```diff
@@ -0,0 +1,128 @@
...
+import org.phoebus.applications.saveandrestore.service.SaveAndRestoreService;
+import org.phoebus.applications.saveandrestore.ui.saveset.SaveSetFromSelectionController;
+import org.phoebus.core.types.ProcessVariable;
+import org.phoebus.framework.preferences.PreferencesReader;
+import org.phoebus.framework.selection.Selection;
+import org.phoebus.framework.workbench.ApplicationService;
+import org.phoebus.ui.javafx.ImageCache;
+import org.phoebus.ui.spi.ContextMenuEntry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
kasemir, line 20: Can you use plain java.util.logging?
That's what we agreed to use for phoebus, https://control-system-studio.readthedocs.io/en/latest/intro.html.
It's also what can be configured via the `-logging` command line option and the Utility/Logging Config menu.

geniejhang: Sure, I can. I think I'm confused with the other project saying I can use either one.

kasemir: I am sorry that you are confused.
Yes, the target platform, https://github.com/ControlSystemStudio/phoebus/tree/master/dependencies#target-platform, by now contains commons-logging, jboss-logging, log4j, 
log4j-to-slf4j, logback, ...
But the original idea when transitioning from Eclipse and starting Phoebus was to use java.util.logging. Not because it's the very best of all, but because it's there and does the basic job. When using external libs like Kafka etc. it's sometimes impossible to avoid other logging libs, but in code that's explicitly written for phoebus like a menu SPI, it should be easy to stick with java.util.logging.

georgweiss:
I will merge this as our users are asking for it. But I will ask you to change some minor UI things. Is that OK?

geniejhang:
Sure

PR #1342, opened 2020-05-22
(Optional) separating snapshots from TreeView and show them in ListView

geniejhang:
As we discussed earlier, default setting leave UI the same.


georgweiss:
@geniejhang, please resolve merge conflict.

geniejhang:
@georgweiss It was conflict with my previous PR.

PR #1343, opened 2020-05-22
Filtering PVs from the list using regex.

geniejhang:
This included minor UI change: SnapshotTable fills vertical gap

### DIFF
--- a/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
+++ b/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
@@ -234,6 +234,8 @@ public class SnapshotController implements NodeChangedListener {
 
                         if (!preserveSelectionCheckBox.isSelected()) {
                             item.selectedProperty().setValue(matchEither);
+                        } else {
+                            matchEither |= item.selectedProperty().get();
                         }
                         return matchEither;
                     }).collect(Collectors.toList());
--- a/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
+++ b/app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
@@ -27,25 +27,27 @@ import javafx.fxml.FXML;
 import javafx.scene.control.*;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
-import javafx.scene.layout.BorderPane;
-import org.epics.gpclient.*;
-import org.epics.gpclient.PVEvent.Type;
-import org.epics.vtype.VEnum;
-import org.epics.vtype.VString;
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyEvent;
+import javafx.scene.layout.VBox;
+import org.epics.gpclient.GPClient;
+import org.epics.gpclient.PVConfiguration;
+import org.epics.gpclient.PVEvent;
+import org.epics.gpclient.PVReader;
 import org.epics.vtype.VType;
 import org.phoebus.applications.saveandrestore.Messages;
 import org.phoebus.applications.saveandrestore.Utilities;
 import org.phoebus.applications.saveandrestore.data.NodeChangedListener;
+import org.phoebus.applications.saveandrestore.model.ConfigPv;
+import org.phoebus.applications.saveandrestore.model.Node;
+import org.phoebus.applications.saveandrestore.model.NodeType;
+import org.phoebus.applications.saveandrestore.model.SnapshotItem;
 import org.phoebus.applications.saveandrestore.service.SaveAndRestoreService;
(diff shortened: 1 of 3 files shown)


geniejhang:
Oh, this works both in savesets and snapshots

georgweiss:
What does the checkbox "Preserve...." work?
Also: I'd prefer if the filter was disabled if the input field is empty. As it stands you need to input '*' and the press Enter.

geniejhang:
By default, when opening a saveset selects all the PVs in it.
Filtering also works the same way.
If you filter the list, filtered PVs are all selected by default.
Checking the checkbox allows users to keep their selected PVs.
For example, let's say you have 100 PVs and you want to select only two different PVs.
One way is to use & to filter both at once without checking the checkbox.
The other was is to filter one (selected automatically) and check the box to keep the selection and filtering the other to find out where it is (not selected automatically) and select it.

georgweiss:
I'm sorry, but I do not understand the use case for the check box. If check it and filter to sort out one PV and then do the same thing with another regexp, I cannot see the first one. The list then does not show all the selected PVs. If this is misunderstanding on my behalf, please create a video that demonstrates this.
Also: if the user types a regexp such that no PVs in the list match, using the delete key to remove the regexp should result in the list showing all PVs as the filter/regexp is then removed.

geniejhang:
So, here's the story.
When users filter the PV list, we can assume that they have the following thoughts:
1. There're many PVs in the list they cannot find the one they want quickly by eye.
2. They're filtering PVs to observe the stored and live values and/or to restore filtered PVs.
3. For example, Let's say you want to select PVs containing A and B. In this case, users can filter with *A*&*B* to show all the PVs containing A and B.
4. By doing so, filtered PVs are all selected to be restored.
5. What if users want to observe C values while keeping the selections of A and B?
6. This case, checking "Preserve selection after" checkbox keeps selection preserved.

**Also: if the user types a regexp such that no PVs in the list match, using the delete key to remove the regexp should result in the list showing all PVs as the filter/regexp is then removed.**
This is what it behaves now. The list reacts on every keystroke you press. If the field becomes empty, then it'll show all the PVs.

kasemir:
Similar idea in the data browser search panel, where an "Add.." or "Replace" option allows you to add to the existing results, or replace them:

Screen Shot 2020-05-28 at 9 02 56 AM

georgweiss:
OK, now it works to delete regexp and unfiltered list. However, I still have issues with the following scenario (see attached images):
1) The list contains 22 PVs, all named HZB-V20*. All are selected by default.
2) In the regexp field I enter "C", which reduces the list to 7 PVs. 
3) I remove the regexp, check the "Preserve..." box and type "C" for the regexp filter. Now I of course get the same list of 7 matching PVs. 

So my understanding is that with the checkbox ticked, all 22 PVs are selected even when the filter results in displaying a subset. This I think can become confusing as the user would need to remember which PVs were selected in the unfiltered list. It becomes even more difficult if the user starts by selecting a subset in the unfiltered list, then ticks the "Preserve.." checkbox and starts filtering. My concern is that once the user is done filtering and selecting, the effective list of selected PVs is not shown.

I'd prefer I you can consider the scheme proposed by @kasemir, which also has the benefit of making save/restore consistent with other parts of Phoebus.

geniejhang:
@kasemir @georgweiss 
Thanks for the nice comments.
I was thinking about exactly the same concern lying on the best last night.

It looks like we can't have the same scheme as Data Browser because users start from an existing list.
In that sense "Add" doesn't mean anything to S&R app.
"Replace" option is just the same as the unchecked behavior.

I'll think about a little bit on this and propose what I come up with.
Please give me any idea on this if you have any.

georgweiss:
To be honest I think the actual problem is about how easy it is to create a lot of data - be that large save sets or large number of snapshots - but how difficult it seems to be to exercise appropriate discipline when organizing the data. So my question back to users would be: can you organize your save sets differently so that filtering would not be needed?
In any case, if you want to keep selected PVs in the list even though they do not match the current regexp, maybe you could color code these PVs instead?

geniejhang:
@georgweiss 
Thanks for the idea.
Since I wasn't here at the designing stage, I couldn't say anything about the current design of the application.
However, it looks common to have large number of PVs in beam line controlling situation.

geniejhang:
@georgweiss 
So, my final idea is to leave the selected PVs when "preserve" checkbox is selected no matter what filter value is.
If a selected but out of filter entry is unselected, it will be still there until the next filter event.
Coloring entry might give some alert to user, so let's not put that in.
Checkbox label should serve the role anyway.
What do you think?

geniejhang:
Note that I added another small feature here d3d79b7.

georgweiss:
I think this might work if you add one more change: when ticking the "Preserve..." box all PVs should be unselected. Reason is that if they are selected, the regexp won't filter on anything.

And I still think the use case still points to a way of working that could be improved. Instead of creating one big save set and filter later, users should take the time to create one save set for each use case.

geniejhang:
Even with 20~30 PVs, it's hard to find the PV user want instantly because PV names tend to be long.
Maybe I should label it as search instead of filter.

Anyway, you're right. I should have thought of that when I add the feature.
I'll update as you said. In addition, I'll add Shortcut+F to move focus to filter TextField.

PR #1351, opened 2020-05-26
CSSTUDIO-1281 Customize position of Acknowledge menu item

georgweiss:
Users have asked to place the Acknowledge/Unacknowledge context menu items on top. This change supports that if preference org.phoebus.applications.alarm/menuitem_acknowledge_ontop is set to true (defaults to false).

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
@@ -120,6 +120,8 @@ public class AlarmSystem
     /** Disable notify feature */
     public static final boolean disable_notify_visible;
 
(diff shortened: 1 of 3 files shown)


kasemir:
Since ack/un-ack is indeed a frequent operation, the user can also double-click on an alarm in the table to ack/un-ack.

kasemir:
I'm fine with ack/unack on top, so making that the default and only option.
Note that this
```
menu_items.add(0, new UnAcknowledgeAction(model, active));
```
needs to be replaced by
```
menu_items.add(new UnAcknowledgeAction(model, acked));
```
since otherwise you can no longer un-acknowledge. Fixing that as well.

shroffk:
> I'm fine with ack/unack on top, so making that the default and only option.

+1

georgweiss:
OK, thanks.

PR #1358, opened 2020-05-27
Issue 1356

shroffk:
WIP:
This PR should wait till it is confirmed that the SNS logbook client continues to work properly.


kasemir:
Go ahead.

core/logbook/src/main/java/org/phoebus/logbook/LogClient.java
```diff
@@ -14,434 +15,458 @@
...
+     * @return collection of LogEntry objects
      */
-    public Collection<String> listAttributes(String propertyName);
+    public List<LogEntry> findLogs(Map<String, String> map);
 
     /**
      * Return all the logs. ***Warning can return a lot of data***
-     * 
+     *
      * @return Collection of all LogEntry entires
```
kasemir, line 52: Several comments use "entires" instead of "entries"

shroffk: sorry, other that this one the other 'logentries' and 'logentires' both appears as a spelling mistake so I landed up ignoring them.
Fixed and creating PR now

Issue #1361, opened 2020-05-28
Docking - closing a named pane from context menu


pavel-ch:
Currently, when trying to close an empty named pane from the context menu, nothing happens. The pane must be renamed to an empty name and then closed.
I have added a confirmation dialog to enable user to close the named pane directly.
PR follows.

kasemir:
If you create a layout with named panes, I assume you do that because you plan to display content there, so the pane needs to remain available. In practice, you would thus also "lock" that pane, and the issue you describe shouldn't arise. But OK, a dialog is certainly better than simply ignoring the close request, so go ahead.

pavel-ch:
Thanks.

PR #1362, opened 2020-05-28
Docking - close named pane from context menu with confirmation dialog

pavel-ch:
See #1361


core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
```diff
@@ -243,7 +244,23 @@ private void showContextMenu(final ContextMenuEvent event)
...
+                    close.setOnAction(evt -> 
+                    {
+                        if (!getName().isBlank())
+                        {
+                            Alert alert = new Alert(Alert.AlertType.CONFIRMATION);
+                            alert.initOwner(dock_parent.getScene().getWindow());
+                            alert.setTitle(Messages.DockCloseNamedPaneTitle);
+                            alert.setContentText(MessageFormat.format(Messages.DockCloseNamedPaneText, getName()));
+                            alert.getButtonTypes().setAll(ButtonType.YES, ButtonType.NO);
+                            alert.showAndWait().ifPresent(type -> {
```
kasemir, line 256: Maybe add `DialogHelper.positionDialog(...` so the dialog shows up relative to the dock_parent?
Otherwise, on setups with multiple monitors, the dialog could be way elsewhere.

pavel-ch: I tried relative to dock_parent, but relative to current DockPane seems better to me...

kasemir: Agree, dock pane is better. In any case, try to assert that the dialog comes up close to the "close" button, not on a different monitor.

kasemir: Good, thanks!

PR #1363, opened 2020-05-28
Embedded display: Suppress scroll events when there are no scrollbars

### LINKED ISSUE #1355: Resized embedded displays may go out of frame when scrolling with mouse
Apparently an old bug inherited from RCP CS Studio, as reported by a user:
If an embedded display is scaled down with "Size content to fit widget", in runtime the mouse scroll will scroll out the content from it's frame.
The embedded display is rendered fine to start with. But if the Operator puts the mouse over the embedded display and scrolls with mouse (wheel or gesture), the widget goes out of frame depending the magnitude of the scaling. The bigger the scaling down was the more it goes out of frame. In some cases the widget can completely disappear.
Attaching bob files demonstrating the behavior
[Archive.zip](https://github.com/ControlSystemStudio/phoebus/files/4686718/Archive.zip)
.

kasemir:
Fixes #1355


georgweiss:
Thanks @kasemir

PR #1368, opened 2020-06-02
minor issues on alarm-logger and alarm-config-logger (arguments)

jeonghanlee:
@kasemir and @shroffk , please look at this pull request. Alarm-logger and alarm-config-logger have the mixed programming logic between standard java and springboot. So, the workflow in main function is not quite well organized. 

I only touched  minimum source files to minimize any impact on these application, since I also have limited knowledge on JAVA and SpringBoot programming. 

Now, `-h` option and `-help` option work, it terminates the main applications well.  And I remove "arbitrary" option `-h*`, because it may create potential issues  in near future. 

For `alarm-config-logger`, now the consistent properties options and the input arguments, and rename the properties file also. In addition, add the sh script for `alarm-config-logger`. 

Please let me know what you think, and feel free to change or reject them according to what you would like to do.

### DIFF
--- a/services/alarm-config-logger/alarm-config-logger.sh
+++ b/services/alarm-config-logger/alarm-config-logger.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+# Alarm Config Logger launcher for Linux or Mac OS X
+
+# When deploying, change "TOP"
+# to the absolute installation path
+TOP="."
+
+# Ideally, assert that Java is found
+# export JAVA_HOME=/opt/jdk-9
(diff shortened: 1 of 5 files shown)


shroffk:
Thanks Han for the code cleanup...
In production FRIB, ESS, or NSLS2 rarely uses the other arguments so it must have been overlooked.

We had started this and logging as pure springboot service but then wanted to try to bring the 3 services together...While we use pure java for the arg parsing, spring boot is useful for embeding an http server, binding to ldap and other authentication mechanisms, integration testing of resources,...

So I will have to find a way of integrating the spring application startup better and document it for others.

PR #1370, opened 2020-06-03
Display hierarchy view of PV list of saveset and snapshot

geniejhang:
It's optional in the settings.
Facilities can customize how to create tree depending on their PV naming convention.
By default, I put FRIBHierarchyParser which split PV at : and then _ as shown in the screenshot below.


kasemir:
Instead of introducing a `FRIBHierarchyParser` which basically splits on the regular expression `[:_]`, could you call that `RegularExpressionHierarchyParser` with a preference setting for the reg ex? The same reg ex would actually work for SNS as well, and I think many sites would just need to configure the reg. ex.

geniejhang:
@kasemir Thanks for the suggestion. In fact, our previous goal was to use regex instead of for loop.
However, I found there're so many exceptions due to the PV names not following the naming convention.
@dylan171 suggested me to add some default one, but I thought it would be enough using FRIBHierarchyParser if any PV follows the same rule.
Anyway, I'll add one for sure.

kasemir:
If one regex doesn't suffice, could you use a list of regex?

A PV name is checked against each of the regex patterns until there is a match, and the 'groups' in the match provide the hierarchical pieces of the PV name?

geniejhang:
@kasemir Yes. That was the case.
I found around 10 regex patterns for our PV names and just gave up at that stage.
Moreover, it has order dependence.
For example, let's say we have A:B:C and A:B.
If I try to match with (\w+):(\w+):(.*) and then use (\w+):(.*), both can be separated [A, B, C] and [A, B].
However, if I do the other way around, they'll be [A, B:C] and [A, B].

So, I'll put the note on the setting file.

kasemir:
Well, you'd have to list the patterns in order, stop when the first one matches, and a "match all" like `(.+)` needs to be avoided, instead using something like `([^:]+)` to match anything-but-colon.
But yes, naming standards suck ;-)

geniejhang:
@kasemir Yeah. I can do that for RegexHierarchyParser, not for FRIB :)

geniejhang:
@georgweiss Forgot tagging you here.

georgweiss:
I like the feature, though not sure if there is any demand for it at ESS.

Issue #1374, opened 2020-06-04
Contributions to context menu


shroffk:
In the past we have had request to allow configuring the addition and removal of context menu entires. This was achieved in the issue ticket listed below by preferences.
#1094

The use of preferences still requires building and packaging of modules not really needed by the product... with the preference basically excluding a module that was included. 
This pattern of including plugins also results in other problems, like warnings from the channelfinder plugin on startup which are very important at a site which has channelfinder but can't access it and are completely irrelevant at a site which has no channelfinder.

To address both these the phoebus framework should enable the easy creation of products with just the right pieces to address the needs of a site without forcing the inclusion of superflusious peices.

shroffk:
In order to achive this here are a few things.

1. The selection, adapter, and contextmenu API's are services can be improved and simplifed.

By looking at existing context menu contributions I have updated the selection API, I have added a few more methods and made most of the call method default.
Additionally thanks to the improvement to the contenxtmenu service by @kasemir and updates to adapter and selection service a simpler design pattern for making site specifc contributions to the context menu is significantly reduced.

The trends module with 2 implementation for contributing the email action to the databrowser can be used as an example.

The default/simple email action
![image](https://user-images.githubusercontent.com/2111304/83760465-f9f72f00-a642-11ea-86a9-bf3b5fc65abf.png)

The rich email action, this includes the list of pvs, the timerange, and a copy of the .plt file
![email-rich](https://user-images.githubusercontent.com/2111304/83760474-fcf21f80-a642-11ea-86ee-5feda450ee1d.PNG)


Sites which like the default can stick to the simple email action.
Sites which like the rich email options can inlcude the app-trends-rich-context-menu in their product.
Sites which want no email action can not include anything.

shroffk:
fixing the logentry in the same way 

![image](https://user-images.githubusercontent.com/2111304/101363936-fb164c80-386f-11eb-9ec3-b547feb0cf54.png)

shroffk:
This project was complete
SelectionServices, AdapterService, and ContextMenuService together provide all the functionality required to create a pugable and extensible context menu

PR #1375, opened 2020-06-04
Issue 1374

shroffk:
WIP:
#1374


shroffk:
#1374 
Cleanup of the ContextMenuEntry, the previous one existed prior to a lot of improvements to selectionService, contextMenuService, and adapterService...so now the context menu entries don'e have to do adaption of selection etc...

app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java
```diff
@@ -196,11 +202,23 @@ private void createContextMenu()
...
+
+            SelectionService.getInstance().setSelection(this, Arrays.asList(DatabrowserSelection.of(model, plot)));
+            List<ContextMenuEntry> supported = ContextMenuService.getInstance().listSupportedContextMenuEntries();
+            supported.stream().forEach(action -> {
+                MenuItem menuItem = new MenuItem(action.getName(), new ImageView(action.getIcon()));
+                menuItem.setOnAction((e) -> {
+                    try {
+                        action.call(plot.getPlot(), SelectionService.getInstance().getSelection());
+                    } catch (Exception e1) {
+                        e1.printStackTrace();
```
kasemir, line 214: Instead of `e1.printStackTrace()`, I would suggest `logger.log(Level.WARNING, "XYZ Failed", e1)`.
The printout can easily get lost.
Using the logger means we can control the level and destination of the message, there's hopefully a somewhat useful specific message to describe the problem, plus you still get the stack trace.
Tried to document that in https://github.com/ControlSystemStudio/phoebus/pull/1031

app/email/src/main/java/org/phoebus/applications/email/actions/ContextCreateEmail.java
```diff
@@ -1,32 +1,110 @@
...
+                    if (parent != null)
+                    {
+                        controller.setSnapshotNode(parent.getScene().getRoot());
+                        stage.setX(parent.getScene().getWindow().getX() + 100);
+                        stage.setY(parent.getScene().getWindow().getY() + 50);
+                    }
+                    stage.show();
+                } catch (Exception ex)
+                {
+                    logger.log(Level.WARNING, "Email dialog failed", ex);
```
kasemir, line 102: This is a good example for how one should log an exception!

core/ui/src/main/java/org/phoebus/ui/spi/ContextMenuEntry.java
```diff
@@ -48,11 +57,48 @@ public default Image getIcon()
      */
     public Class<?> getSupportedType();
 
+    /**
+     * Invoke the action associated with this context menu contribution
+     * @throws Exception on error
+     */
+    public default void call() throws Exception
+    {
+        throw new UnsupportedOperationException(getName() + ".call() Is not implemented.");
```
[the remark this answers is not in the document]
kasemir: Just looking at the API, i.e. all those default `call` variants which all throw exceptions, I'm not sure:
a) Can I ignore them all, since there's a default for all of them, so the compiler is happy?
b) Do I actually need to implement all, otherwise the Exception will cause me to fail at runtime?
c) Do I just need to implement one, because whatever uses it will fall back to the "other" variant of call until one succeeds?

Maybe all it needs is another comment in the Javadoc of what one needs to do.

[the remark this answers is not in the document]
kasemir: >  second iteration

OK

app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java
```diff
@@ -209,7 +209,7 @@ private void createContextMenu()
                     try {
                         action.call(plot.getPlot(), SelectionService.getInstance().getSelection());
                     } catch (Exception e1) {
-                        e1.printStackTrace();
+                        logger.log(Level.WARNING, "Failed to exectute " + action.getName() + " from databrowser plot.");
```
kasemir, line 212: Sorry, should still pass the exception so we get the stack trace:

```
                        logger.log(Level.WARNING, "Failed to exectute " + action.getName() + " from databrowser plot.", e1);
```
Note ", e1" at the end

shroffk: I thought I had included it... was using a search replace and got distracted with the CI failure

shroffk:
I think I understand the slow github actions... it seems like the push build and PR build happen in a kind of single threaded fashing with switching.

app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java
shroffk: sorry about this overlook

shroffk:
@kasemir is this ok for merging now.

I think we can change the requirements of the call method once we have a clear set of requirements

PR #1376, opened 2020-06-04
fixing logging

shroffk:
Especially exception logging.

This does not fix all but fixes  channel, logbook, some display and core modules.


kasemir:
Good, thanks!

PR #1379, opened 2020-06-05
logging clean and release of temp file fixes

shroffk:


shroffk:
the build has succeeded, github is not updating

PR #1382, opened 2020-06-09
Fix the bug which cannot handle settings file. 

jeonghanlee:
@kasemir and @shroffk,

Please check this request, and let me know. 

The current version cannot set `server` and `config` properly if we define them in *.ini file.  This request allows users to use `ini` file as the Kafka server and alarm config as well. However, these values are overridden by command line arguments `-config`, `-server`, or both. 

Moreover, use the "shutdown" instead of "start with shut".

@jeonghanlee

### DIFF
--- a/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
+++ b/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
@@ -572,7 +572,7 @@ public class AlarmServerMain implements ServerModelListener
                         throw new Exception("Missing -server name");
                     iter.remove();
                     args_server = iter.next();
-		    use_args_server = true;
+                    use_args_server = true;
                     iter.remove();
                 }
                 else if (cmd.equals("-config"))
@@ -581,7 +581,7 @@ public class AlarmServerMain implements ServerModelListener
                         throw new Exception("Missing -config name");
                     iter.remove();
                     args_config = iter.next();
-		    use_args_config = true;
+                    use_args_config = true;
                     iter.remove();
                 }
                 else if (cmd.equals("-logging"))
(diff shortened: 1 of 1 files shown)


services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
```diff
@@ -135,7 +137,7 @@ private boolean handleShellCommands(final String... args) throws Throwable
             restart.offer(false);
         else if (args.length == 1)
         {
-            if (args[0].startsWith("shut"))
+            if (args[0].equals("shutdown"))
```
kasemir, line 140: I would prefer to be lenient with the shell commands.
Both "shutdown" or just "shut", "shutd", ... should be recognized since I've been gotten used to just typing "shut".
If there was another shell command that also started with "shut", then of course we'd have to check more of the text, but the other commands "cd", "ls", ... pose no conflict.

kasemir: OK with the rest

shroffk: > I would prefer to be lenient with the shell commands.
> Both "shutdown" or just "shut", "shutd", ... should be recognized since I've been gotten used to just typing "shut".
> If there was another shell command that also started with "shut", then of course we'd have to check more of the text, but the other commands "cd", "ls", ... pose no conflict.

I agree with Kay here, starts with is a lot more forgiving...

For the other stuff:
I appreciate the logging on the conflict between the settings file and the arguments...it should be very useful for debugging.

services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
```diff
@@ -137,8 +137,8 @@ private boolean handleShellCommands(final String... args) throws Throwable
             restart.offer(false);
         else if (args.length == 1)
         {
-            if (args[0].equals("shutdown"))
-                restart.offer(false);
+	    if (args[0].startsWith("shut"))
```
kasemir, line 140: For what it's worth, try to use N*4 spaces for indentation, not tabs, nor a mixture of tabs/spaces.
But that's minor, and there's no agreement on style anyway, so go ahead.

services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
jeonghanlee: I thought, both of you made these comments, yes, I've agreed with both of you. Please check the latest changes.  However, I changed `-h` argument in order to keep all options consistently within `Alarm Trio`.  :smiley:

services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java
jeonghanlee: I am very bad about styles, can you explain "N*4 spaces" ?

kasemir: That file intents by 0, 4, 8, .., multiples of 4 spaces.
The `if (args[0].startsWith("shut"))` should be intended twice (8 spaces) to align with the rest. Looks like it's using "tab" plus 3 spaces. The width of a tab could be 4 or 8, who knows.

jeonghanlee: I do my best to match what you described, please let me know what you think.

jeonghanlee: @kasemir , I have few more questions on styles.

-  How do you make *N*4 spaces"? Do you type them all with space key?
-  Do you prefer this style in Java?  I would like not to break your style while I change / update codes.

kasemir: I set Eclipse to use spaces for indentation, specifically 4.
Most IDEs have settings for that.
In `vi` I do type 4 spaces.

This is getting ridiculous, just merge it, I'm the only one who cares about this anyway.

jeonghanlee: @kasemir , Sorry for these commits, but I will do this later according to what you describe as much as I can. I can see clearly why you try to do this while looking at github filechanges, which gives me more clean view.  I like it.

Issue #1392, opened 2020-06-12
Cancellable Job


shroffk:
Update the existing jobs to support user triggered cancels

PR #1393, opened 2020-06-12
Issue 1392

shroffk:
#1392


app/channel/utility/src/main/java/org/phoebus/channelfinder/utility/JobRunnableWithCancel.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A helper class which provides the framework for running tasks in the background while still monitoring for cancel requests.
+ *
+ * @author Kunal Shroff
+ */
+public abstract class JobRunnableWithCancel implements JobRunnable {
+
+    private static final Logger logger = Logger.getLogger(JobRunnableWithCancel.class.getName());
+    private static final ExecutorService executorService = Executors.newCachedThreadPool();
```
kasemir, line 22: Consider using `org.phoebus.framework.jobs.NamedThreadFactory`:
 1) That way, the thread will have a recognizable name
 2) The thread will be a daemon thread, allowing the app to exit

shroffk: I have used the NamedThreadFactory, not sure I have the best name for the time being

shroffk:
@geniejhang : I think these are some examples of a simple cancelable task.

app/channel/utility/src/main/java/org/phoebus/channelfinder/utility/JobRunnableWithCancel.java
kasemir: Sure, that's fine. When you see a thread named "JobRunnableWithCancel-1", you can locate that in the sources. With "pool-05-thread-03" it's a lot harder.

geniejhang:
Wow. That was fast. Thanks!

kasemir:
On one hand, I'm leaning to a move of `JobRunnableWithCancel` into `org.phoebus.framework.jobs`.
It certainly is one example of how to do things, and it makes it easy to wrap `Runnables` which were written without support for progress reporting and cancellation.

On the other hand, cancelling/interrupting a task this way is comparably brute force.
The basic Java Thread API, https://docs.oracle.com/javase/10/docs/api/java/lang/Thread.html, suggests not using `stop()` etc. but instead setting some flag and then having the code in the thread check for this flag to exit gracefully. Interrupting a thread in the end also depends on the code in the thread checking `isInterrupted()`.

How about we move the `JobRunnableWithCancel` into `org.phoebus.framework.jobs`, but add some comments:

```
Ideally, a JobRunnable is written with cancelation and potentially progress updates in mind:
 
   JobManager.schedule("Demo", monitor ->
        {
            monitor.beginTask("What I do", 100);
            for (int step=0; step < 100; ++step)
            {
                // Allow cancellation
                if (monitor.isCanceled())
                    return;
                
                // .. do something
                monitor.worked(1);
            }
        });

If the task cannot be cancelled gracefully because it is for example stuck in network calls or inside a 3rd party library, the  `JobRunnableWithCancel` may be used to brute-force interrupt the task in response to a cancellation request.
```

shroffk:
I am ok with moving this to the jobs package with the above comments.

In general I feel:
Running operations in a Job (even one that does not cancel) is better than UI thread.

Exiting threads gracefully is complicated and defers on the exact operation being performed and its implementation...I was looking https://docs.oracle.com/javase/10/docs/api/java/util/concurrent/FutureTask.html with the hope that we could use this in conjunction with the Jobs API... In the ideal case checking `isInterrupted()` would be best/

Like you mentioned in the example of REST calls, there is very little to do in terms of a graceful exit other than to kill the thread waiting for a response. I think over time we might be able to expand the jobs package with more templates.

kasemir:
Using background threads is certainly better than the UI thread.
..and the Java 'concurrent' package has a ton of ideas.

'org.phoebus.framework.jobs' is for things that the user might need to see.

You opened a *.plt and all you see is a blank page while the Data browser fetches archived data. The "Jobs: 5" in the lower right corner indicates that stuff is happening. You can even open the Jobs monitor to see for which PVs it's fetching 
data.

You opened a display, which might take some time to open the actual file.

Channel finder UI fetching data would be similar: You requested something, and the "Jobs" show that we're working on it.

Counterexamples:
When the data browser updates a plot, there are threads to compute the data range for autoscaling, plus the actual plot is drawn in background threads. These are comparably short in duration and happening all the time. They should certainly be in background threads, but not use "Jobs", since there's no need for the end user to see them. The number of running jobs indicated by the status bar would flicker all the time if we showed these.

PR #1394, opened 2020-06-12
VEnum index and choices need to be stored in DB

geniejhang:
@georgweiss 
So, master version only stores selected item among other choices.
It makes unable to compose ComboBox items unless live PV is read to retrieve those.

This changes the behavior and stores serialized index plus choices into DB.
so as to enable to edit cell value of VEnum PV.

### DIFF
--- a/services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/persistence/dao/SnapshotDataConverter.java
+++ b/services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/persistence/dao/SnapshotDataConverter.java
@@ -276,14 +276,27 @@ public class SnapshotDataConverter {
 				case ENUM:{
 					Object[] values = objectMapper.readValue(snapshotPv.getValue(), Object[].class);
 
-					int index = (int) values[0];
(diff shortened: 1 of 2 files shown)


georgweiss:
@geniejhang , I'm afraid this is not backwards compatible. Existing enum records in the  database will trigger exception when deserialized. You would need a migration script to deal with that, converting existing records to a form compatible with your changes. On the other hand, if this is not an issue (i.e. existing records containing enums causing exception is OK, or if you do not have any enum records), then fine with me. At ESS I do not believe we have not yet used enum PVs in save&restore.

geniejhang:
@georgweiss Yeah. This one is indeed not backward compatible. If anyone has used VEnum between your last gpclient update and the point this one is merged, those snapshots will cause exception.
I would just write some if statement for deserializer for backward compatibility. It should be quicker.

georgweiss:
Thanks, I think handling both formats is the least amount of pain.

geniejhang:
@georgweiss The updated code doesn't throw an exception, but also shows only one option stored in DB making it unable to switch to the other. To do it, user needs to create a new snapshot.

PR #1396, opened 2020-06-15
Fix alarm config tool

krisztianloki:
After the latest changes the Import and export of alarm configurations ignored the '-config' parameter.


kasemir:
Thanks for also cleaning up the 'tabs'

Issue #1242, opened 2020-03-31
Save And Restore: Stored Setpoint cell editing


martinapav:
We want to change »Stored Setpoint« in Save And Restore table. By clicking on the Stored Setpoint cell we can edit it. After changing stored set point to 5 and pressing enter also »Δ Live Setpoint« changes.
 
![image](https://user-images.githubusercontent.com/62931048/78021053-cbaf5780-7352-11ea-816b-520543fc7fd5.png)

If we click again into the cell and only press enter without entering any different value the value goes back to previous setting but »Δ Live Setpoint« does not change.
 
![image](https://user-images.githubusercontent.com/62931048/78021082-e08beb00-7352-11ea-8f80-5fbe8ae35134.png)

If I only edit the value and click Save snapshot, Live Setpoint seems to be saved. Is this field intended to be editable at all?

georgweiss:
No, not initially. I have other requests for making the stored setpoint editable (and possible to persist), but that has not been implemented.
Your observation points to an inconsistency. What I can do short term is to make the cell read-only.

geniejhang:
I'm working on this for our application. Just FYI.

georgweiss:
Thanks @geniejhang. 
There is one twist to this that I'd like us to agree upon: if the cell value is changed before restore, it must be clear to the user that the changed value is not saved in the snapshot by default. Doing so would in my view be very confusing and in part defeat the purpose of a snapshot. To persist the changed value in the snapshot should require the user to press the save button. Also, if the snapshot tab is closed without saving the value, the user should be prompted to choose to exit without save.

geniejhang:
@georgweiss 
Yes. I was thinking about that last night lying on the bed.
FRIB version of CS-Studio SaveSetRestore has pink highlight when the saveset/snapshot becomes modified state. I'm thinking about putting similar stuff in as well.
I agree on the behavior with you.
Thanks for the comments.

geniejhang:
@martinapav Since it's now merged to master, please test it 😄

PR #1400, opened 2020-06-24
Manipulating stored value cell - Issue #1242

geniejhang:
It's done in two ways:
1. Changing the multiplier at the right top on the screenshot.
2. Directly editing stored cell value.

So, brief explanation.
1. Changing the multiplier multiplies the number to all the non-read-only, stored values of VNumber and VNumberArray.
   - Delta values are updated accordingly as the attached.
      
    - You can also see "Save Snapshot" button is enabled. If the spinner value is not 1 or any cell has been changed from its original value by user, "(Modified)" will be added to snapshot name and tab title. In this case, by clicking "Save Snapshot" button created a new snapshot with modified values.
      
   - If you put non-number character in the spinner, it will become red and Restore button is disabled.
      

2. Changing one stored value cell behaves the same.
   - You see delta value is updated, "Save Snapshot" button is enabled, and Snapshot name and tab title have "(Modified)" at the end of them.
      

- Some comments:
   - Utilities class is updated to support unsigned type values. I don't know the frequency of those types, but anyway I added them.
   - SafeMultiply class is added to ensure multiplying number in spinner to VType value doesn't exceed its range.

Any comment is welcomed.

@georgweiss


georgweiss:
Good job @geniejhang , you did what I have not had the time to realize...

One small thing: we're trying to avoid wildcard imports. Can you please update to import classes?

geniejhang:
@georgweiss 
Sure thing. IntelliJ automatically changed all those import stuff.
I reverted those and disabled the option.

geniejhang:
Here's the update preview.

georgweiss:
@geniejhang , sorry for being focused on details...
We would like to support localization, so every piece of text in the UI should be taken from the messages.properties in the s&r UI module. You can leave the hardcoded stuff in the fxml files, but the text values should be set in runtime and read from the properties file.

Messages for the log file need not be localized.

Also, I think "Restoring values multiplied by" should be "Restore values multiplied by". But you would need to confirm with a native English speaker, which I am not.

dxmaxwell:
I like the new location. In terms of the label, I've been trying to think of a shorter (and more clear if possible), how about "Restore with scale"? I am a native speaker, but I wouldn't put too much faith in that.

georgweiss:
Or maybe "Restore with factor"?

dxmaxwell:
To me "scale" better implies multiplication (as "offset" implies addition). And I think users at FRIB want to use this feature to scale the beamline configuration for different types of beams.  But, I can certainly live with "factor", and perhaps it would be preferred by the broader community. @georgweiss  as the primary maintainer I think it is up to you, unless we get some more opinions.

georgweiss:
Since I'm not a native English speaker: I fold.

georgweiss:
I'm sorry, but "Restore with scale" is not in messages.properties. You can leave the hard coded value in fxml (makes it easier for SceneBuilder), but when UI is launched the label text should be taken from messages.properties.

geniejhang:
Oh. I didn't realize that you're mentioning that one since I don't see any other labels are in messages.properties.
Just updated.

georgweiss:
Yeah, I might have committed a sin or two myself :-). I will accept once CI passes.

FYI: I will be leaving for four weeks of vacation starting Monday. You should maybe ask for permission to merge, check with @kasemir or @shroffk .

PR #1407, opened 2020-07-01
Embedded display: $(LCID)

kasemir:
#1405 didn't quite work out.
It would always define $(LCID), but then all macros are replaced from the parsed XML.

Now $(LCID) is added to the 'effective' macros:
It doesn't show in the editor, you cannot assign a value.

Screen Shot 2020-07-01 at 12 22 43 PM
At runtime, however, it's available.

Screen Shot 2020-07-01 at 12 23 05 PM


kasemir:
Travis: 
```The command "~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts" failed and exited with 52 during .```

Issue #1410, opened 2020-07-01
opi "menu" widget with no label (i.e. label is an empty string) is translated to .bob "ActionButton" widget with label "No PV"


oksanagit:
Hi,
Please take a look at the attached pic. 
I actually made a fix for it and will create a PR in a few. 

![PhoebusNoPV](https://user-images.githubusercontent.com/21367956/86285873-6ec27800-bbb3-11ea-94fd-5c4eb55ca3d4.jpg)

kasemir:
How do you get that display?
Are you converting from MEDM to *.bob?
Or are you opening a *.opi display (which had previously been converted from MEDM)?

oksanagit:
I am opening an .opi display, which was previously converted from MEDM.  We would like to move to the new product as smooth as possible, and  it is my understanding that same code is being used in both scenarios: opening an .opi screen and translating from .opi to .bob and opening bob. Thank you.

kasemir:
There are 2 routes:

1) MEDM -> *.opi (using the converter in BOY) -> open *.opi in Display Builder
2) MEDM -> *.bob (using the converter in the display builder)

You chose 1), which is fine.
Can you attach the *.opi file?

oksanagit:
[pointGrey.opi.txt](https://github.com/ControlSystemStudio/phoebus/files/4864790/pointGrey.opi.txt)
I had to add .txt. .opi extension was not accepted as legitimate.

kasemir:
Thanks.
Can you add ADPlugins.opi, since that's the one with the actual menu buttons in question:

![Screen Shot 2020-07-02 at 10 58 02 AM](https://user-images.githubusercontent.com/1932421/86374781-f27f7180-bc52-11ea-8ac1-d68d491bb377.png)

oksanagit:
[ADPlugins.opi.txt](https://github.com/ControlSystemStudio/phoebus/files/4865580/ADPlugins.opi.txt)
Yes, sorry, this is an included display, the main part of the story.

oksanagit:
Let me know it anything else is needed. Thanks

kasemir:
OK thanks, will look at it.

kasemir:
Try with the latest. Thanks for finding this.

oksanagit:
I have tested the master branch, looks good to me. Thank you Kay!

PR #1411, opened 2020-07-01
 Now when Phoebus opens .opi screen with menu widget which has empty …

oksanagit:
…string as a label, the new Active Button widget will show empty string as well.


kasemir:
Let me look into it because while it would fix #1410 it could break #1293.

kasemir:
Have slightly different solution that fixes  #1410 and still allows #1293.

PR #1417, opened 2020-07-09
Ack/unack buttons follow auth

tanviash:
Currently, ack and unack buttons in alarm table's toolbar do not follow authorization rule set as alarm_ack in authorization.conf file. This PR is to fix this bug.


tanviash:
@kasemir Thank you for reviewing the changes.

Issue #1420, opened 2020-07-15
Databrowser - passing / using macros in databrowser


pavel-ch:
1) I was not able to pass macros to databrowser using standard DB action syntax:
![image](https://user-images.githubusercontent.com/771816/87546220-7e5cb900-c6a9-11ea-9cc7-a27a53ce3a9c.png)
What I get (on Windows) is always an exception:
![image](https://user-images.githubusercontent.com/771816/87546380-c380eb00-c6a9-11ea-9f22-09a040bb5263.png)
Because URI was treated as a filename including macros.
I would suggest, in `org.csstudio.trends.databrowser3.DataBrowserInstance.loadResource(URI)`, use:
`XMLPersistence.load(new_model, ResourceParser.getContent(new URI(input.getScheme() + "://" + input.getPath())));`
instead of original:
`XMLPersistence.load(new_model, ResourceParser.getContent(input));`
This works for me. Can you approve @kasemir ?

2) In databrowser, using macro for PV name generally works, but on several places shows macro name instead of expected resolved text like here:
![image](https://user-images.githubusercontent.com/771816/87547703-c67cdb00-c6ab-11ea-8533-95f5ecddf3a7.png)

3) I would really appreciate to have a possibility to macroize start / end time. This is suitable for loading one PLT file with many time ranges. I can do it, can you approve @kasemir ?

Thanks

kasemir:
Sounds good, thanks

kasemir:
I think this has been handled, re-open if not.

PR #1422, opened 2020-07-15
Databrowser - passing / using macros in databrowser

pavel-ch:
See #1420 
* Macros can be passed from URI
* Few basic UI changes to see resolved PV names (properties panel, samples panel)
* Macroized start/end time. At the moment ONLY when reading from PLT file -> suitable for fixed read-only PLT files (edited by hand).


kasemir:
Good, thanks!

PR #1424, opened 2020-07-22
Multiple tagging on a snapshot and tag searching

geniejhang:
Let me explain what this update is for by putting an instruction I wrote for internal use.
FYI, at the last part the instruction shows snapshots separated from the TreeView.
However, since we decided to implement both one TreeView and TreeView with ListView, the features apply to both UI.

@georgweiss


georgweiss:
Sorry for late reply, I was on vacation.
In my view the visual communication is a bit inconsistent due to the use of different icons (color, shape) for tags. I'd like to propose:
* In the top level context menu, use the yellowish tilted icon with no + sign, as seen in the sub-menu.
* Use the yellowish titled icon with magnifying glass for the search button.

geniejhang:
What do you think?

georgweiss:
Yes, this looks better.

geniejhang:
Wow. I haven't merged yet, though...
So quick!

georgweiss:
Then please create another PR when done.

PR #1426, opened 2020-07-27
CSSTUDIO-1062 Spinner widget PV name not updated in edit mode

georgweiss:
The spinner widget ha been updated to listen to changes of the PV name when in edit mode, i.e. same behavior as for instance text update.


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SpinnerRepresentation.java
```diff
@@ -352,6 +354,8 @@ protected void registerListeners()
         model_widget.propShowUnits().addUntypedPropertyListener(contentListener);
         model_widget.runtimePropValue().addUntypedPropertyListener(contentListener);
 
+        model_widget.propPVName().addPropertyListener(pvNameListener);
```
kasemir, line 357: call to addPropertyListener() should have corresponding removePropertyListener in `unregisterListeners` below.

georgweiss:
You're right, added unregister.

kasemir:
Thanks!

PR #1430, opened 2020-07-28
Display Builder Editor: New Selected Widget Distribution actions

pavel-ch:
... with predefined Gap. See #679


kasemir:
When the gap is zero, would the result be the same as before?
I'd like to avoid adding too many options, so maybe remove/replace the original gap options, i.e. always prompt for a gap.

pavel-ch:
No, this moves all widgets including the last one, and sets an equal gap between all. If the gap is 0, then widgets are back-to-back, if negative, they can overlap. The function does not work like the existing one, that is why I did not modify it. I think both variants have their use cases, and I would propose to distinguish between variants at the Distribution menu level.

pavel-ch:
This new functions are specifically good when all widgets should fit into the display grid, which is hard to achieve in practice with original functions.

kasemir:
OK, thanks!

PR #1433, opened 2020-07-28
Display Buider Editor: Widget Tracker: refined show/hide location Lab…

pavel-ch:
…el (X,Y)
At the moment with fixed size limits. Should be better to measure actual Label sizes, but not worth of invested time?
See #679


kasemir:
I agree this is good enough.

PR #1436, opened 2020-07-29
Slidebutton: Fix size

kasemir:
As a result of #1434, the slide button would always show but not properly respond to resizes.

Note the label that's larger than the widget, which simply overflows:
Screen Shot 2020-07-29 at 8 51 39 AM

With this fix, the widget honors the assigned space:
Screen Shot 2020-07-29 at 8 50 35 AM

Screen Shot 2020-07-29 at 8 50 41 AM

Of course one can still enable 'auto-size':
Screen Shot 2020-07-29 at 8 50 48 AM


georgweiss:
Well, I did test various scenarios with and without autosize but apparently I was not looking for the correct things.

Issue #1435, opened 2020-07-29
Spaces in paths&names


DorianHas:
Hello there,

We encounter an issue with Phoebus recent version. Spaces in names and directories aren't took into account. It's creating the following error message : 

![errorFileBrowser](https://user-images.githubusercontent.com/51746537/88774591-ec15e400-d183-11ea-83cd-379084a2e16a.png)

Dorian

kasemir:
Long story short: You best avoid spaces as well as characters like `(`, `)`, `:` in file names, or come up with a perfect rewrite of everything that works for everybody.

In the end, the resource names must be URIs. We need to support access to the same `display.bob` or `plot.plt` no matter if it's a local `file://path/to/plot.plt`, which is typically the case in a control room, or if it's accessed as `http://my.site.org/display_files/path/to/plot.plt`, which is often the case when users run the tool in their office. Inside a URL, spaces need to be escaped as `%20`. The technical difficulty is that when we write files, like from a display editor or the data browser, that's of course only possible with local files, not via URIs, which are read-only. So we do need to do a certain amount of conversion between URI and local files. Our code has undergone several simplifications to move from string manipulations to basic Java URI/Path/.. API, but it's tricky because for example Windows uses different file name rules than Mac OS and Linux.

kasemir:
The problem was introduced in #1422, and I think I've found a work around that'll again allow spaces in names.

kasemir:
Fixed

PR #1437, opened 2020-07-29
Data browser filename fix

kasemir:
#1422 added support for macros, but because it used plain string
manipulations to remove the "..?MACRO=value" query from the input URL,
it created problems with spaces in filenames, as reported in #1435.

This rewrites the URI using API that will properly escape spaces in the
file name.

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/DataBrowserInstance.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/DataBrowserInstance.java
@@ -194,7 +194,13 @@ public class DataBrowserInstance implements AppInstance
                 new_model.setMacros(macros);
 
                 // Load model from file in background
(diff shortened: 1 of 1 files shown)


pavel-ch:
URI rewrite is okay except that for me on Windows, plt files cannot be open and throw exception File not found. Reason is that at the end of filename, the new URI(....) constructor adds ? and #. This is caused by using empty strings instead of null for unused parameters (see URI docu). PR follows.

kasemir:
I've just received a similar complaint from a local user but couldn't look into more detail since I don't have a development setup on Windows. Would appreciate a lot if you can look into this!

pavel-ch:
With this patch, it works on Windows. Please test on other platforms, but should work since this is in compliance with documentation.

kasemir:
Looks OK on Mac OS, thanks!

PR #1439, opened 2020-07-31
Initial widget state incorrect if formula PV references disconnected PV

georgweiss:
Given how formula PVs are constructed and evaluated, this is the least intrusive solution I could find. Root cause for the issue is that even if input PVs in a formula are off-line, the evaluated PV will not be treated as disconnected when OPI is launched. This fixes the issue such that widget is rendered correctly if any of the input PVs is off-line.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/pv/RuntimePV.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/pv/RuntimePV.java
@@ -20,6 +20,7 @@ import org.phoebus.pv.PV;
 import org.phoebus.pv.PVPool;
 
 import io.reactivex.disposables.Disposable;
+import org.phoebus.pv.formula.FormulaPV;
 
 /** Process Variable, API for accessing life control system data.
  *
@@ -70,7 +71,8 @@ public class RuntimePV // TODO (Almost) remove. Use vtype.pv, only add setValue 
     {
         // If there is a known value, perform initial update
         final VType value = pv.read();
-        if (value != null)
+        // FormulaPVs should not trigger value change as PV inputs may be disconnected
+        if (value != null && !(pv instanceof FormulaPV))
             listener.valueChanged(this, value);
         listeners.add(listener);
     }
--- a/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
@@ -18,7 +18,9 @@ import java.util.logging.Level;
 import org.csstudio.apputil.formula.Formula;
 import org.csstudio.apputil.formula.VariableNode;
 import org.epics.vtype.Alarm;
+import org.epics.vtype.Display;
 import org.epics.vtype.Time;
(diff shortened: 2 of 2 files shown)


kasemir:
I'll look at it. The fix should be possible solely in the FormulaPV's initial value, without requiring  `instanceof FormulaPV` checks in the display runtime or other clients.

georgweiss:
I tried an approach giving the input PVs an initial value, but that had substantial side-effects.

kasemir:
I appreciate that. Still, adding`instanceof FormulaPV` checks to the display builder, data browser, probe, ... would lead down the wrong path. Whatever's wrong needs to be fixed in the formula PV.

georgweiss:
I guess one could add to FormulaPV (and skip instanceof check)

`@Override
    public VType read(){
        if(inputsConnected())
            return formula.eval();
        return null;
    }`

kasemir:
The variable node simply needs to start out 'disconnected'. Plain 'NaN' as currently the case doesn't  work because it misses the INVALID alarm state.
This one-liner (plus imports) seems to fix it as far as I can tell:

```
diff --git a/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java b/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
index 346c69f19..68d96b7ad 100644
--- a/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
@@ -18,7 +18,9 @@ import java.util.logging.Level;
 import org.csstudio.apputil.formula.Formula;
 import org.csstudio.apputil.formula.VariableNode;
 import org.epics.vtype.Alarm;
+import org.epics.vtype.Display;
 import org.epics.vtype.Time;
+import org.epics.vtype.VDouble;
 import org.epics.vtype.VString;
 import org.epics.vtype.VType;
 import org.phoebus.pv.PV;
@@ -60,7 +62,10 @@ public class FormulaPV extends PV
             final VariableNode vars[] = formula.getVariables();
             inputs = new FormulaInput[vars.length];
             for (int i=0; i<inputs.length; ++i)
+            {
+                vars[i].setValue(VDouble.of(Double.NaN, Alarm.disconnected(), Time.now(), Display.none()));
                 inputs[i] = new FormulaInput(this, vars[i]);
+            }
 
             // Set initial value
             doUpdate();
```

kasemir:
Will add a unit test and then create PR.

kasemir:
Code based on `read(){ if(inputsConnected()) return formula.eval(); return null; }` doesn't work.

Take the formula
```
=1 ? 42 : `missing_pv`
```
 If checking all inputs, it would blindly return 'null'. When actually evaluating the expression, it should return 42.

This type of formula can be used to select between two devices:
```
= `DeviceSelectorPV` == 1 ? `DeviceAReadout` : `DeviceBReadout`
```

When device A is selected, we want to see its value even though device B might be offline. That's the whole point of the `DeviceSelectorPV`. So the result of the formula must be based on actually evaluating it, no shortcut via `inputsConnected`

PR #1440, opened 2020-07-31
Picture widget not showing placeholder when SVG resource is incorrect

georgweiss:
Also removed unused code.


krisztianloki:
This will scale the (converted) image instead of re-transcoding the original SVG when the size of the widget changes which can result in ugly images if the change is big enough.

Issue #1387, opened 2020-06-10
Formula PV: formula PV appears disconnected though its input PVs are connected


pavel-ch:
We are extensively using formula PVs with 2 and more input PVs. Irregulary happens, that input PVs are connected but Fomula PV itself is disconnected (=has a value with DISCONNECTED alarm) after display startup. Proven by Probe, too.
My guess is that this is caused by initialization of PVs on display startup together with a threading issue:
The method `org.phoebus.pv.formula.FormulaInput.handleUpdate(VType)`:

```
    private void handleUpdate(final VType value)
    {
        logger.log(Level.FINE, () -> formula_pv.getName() + " updated by " + pv);
        variable.setValue(value);
        formula_pv.update();
    }

```
is called from more threads. If interrupted between `setValue `and `update ` calls, it can happen that the old initially disconnected input PV value wins over the new connected value.
I tried to make this method `synchronized` and the effect has gone.
However I do not know further consequences of making this method synchronized.
Can you please evaluate and let me know if the change is OK @kasemir ?
Thank you.

pavel-ch:
Well it appears that
> ... and the effect has gone.

is not really true. Effect is still there.
So please take this Issue as an misbehavior notification.
I must dig a bit deeper to find the real reason.
At the moment seems that setting PV logging level from WARNING to FINE really switches the effect off, perhaps because changing thread timing on my system.

pavel-ch:
I tested your code, the issue is not observed so far. Needs a longer run. I would propose to merge to master.

kasemir:
Thanks for testing. Just found one more thing: It's possible that the evaluation on thread X used an older value from an input updated on thread Y because of an omitted `volatile`.

pavel-ch:
Right this could be the originating issue. For me, problem did not show since the decoupling introduction.

kasemir:
Both are certainly useful improvements. Have to see if the latter was in fact the original problem.

Anything special in how you get into that state?
Is it formulas with many inputs? Or did you see it with just 2 inputs? I assume never with just 1 input?
I assume these inputs are using your MQTT-type PVs? Or do you see it with plain local and simulated PVs?
Does it only happen with the "initial" state after opening a display? When later changing one of the inputs, does the formula then properly update? Or does it also happen with a long-running formula that keeps updating, but for some reason it remains disconnected even though its inputs are changing?

pavel-ch:
[...]

PR #1442, opened 2020-07-31
Formula: Start PVs 'disconnected' until first value arrives

kasemir:
#1387 #1439

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/formula/FormulaPV.java
@@ -18,7 +18,9 @@ import java.util.logging.Level;
 import org.csstudio.apputil.formula.Formula;
 import org.csstudio.apputil.formula.VariableNode;
 import org.epics.vtype.Alarm;
+import org.epics.vtype.Display;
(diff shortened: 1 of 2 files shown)


georgweiss:
This does in my view not fix the issue in all aspects.
When a widget references a disconnected PV it shows the appropriate alarm border and content. With this change the widget will show something different if an input PV is offline. In the case of the TextUpdate widget it will show the border associated with INVALID and NaN, while I would expect the UNDEFINED border and (for isntance) <=2*`offlinepv`>:

Screenshot 2020-07-31 at 16 25 44

kasemir:
Yes, there is a subtle difference between `INVALID` and `UNDEFINED`. You see that when you compare
```
=`dummy`
=`dummy` + `sim://sine`
=`dummy` + `sim://intermittent`
```
That's a result of how the formula transforms the values.
You might see either `<the formula>` or `NaN`.
But key is that you always get either `INVALID` and `UNDEFINED`, you know that something's not quite right with the data, whereas before you'd see `NaN` without any alarm.

georgweiss:
Well, I was not aware that formulas should be possible to evaluate even if input PVs are not available...

PR #1443, opened 2020-07-31
Display Builder Editor: Open in external editor actions

pavel-ch:
Display Builder Editor: Open in external editor actions added for Script dialog ("Select" Menu Button) and Editor Pane (Context menu).
Opens the Script file, if an external application is registered for handling .py or .js files, if script is configured as file (not embedded).
Opens the Display file, if an external application is registered for handling .xml files.
See #679


kasemir:
Thanks!

PR #1450, opened 2020-08-03
Some updates on Save&restore application and service

geniejhang:
@georgweiss Please give some comments if any. Thanks.

Update on migration tool
- An option to migrate tags using multiple tagging instead of "Golden" tag
- Keep saveset/snapshot creation date and time

- Comments
  : Service API updated to support keep the date and time
  : Unit test part is updated in accordance with API changes

Update on SAR
- PV Case-sensitivity update (Test and test are now regarded as different PV)
- Possible to add same PV with different readonly setting
- value column in property table became VARCHAR(2000) from VARCHAR(100)
- foreign key added for snapshot_id in tag table to make tags deleted on cascade


geniejhang:
Please don't merge even if you don't have any comment on this.
I found two issues that
- Missing snapshot after migration
- Duplicate snapshots

Thanks.

georgweiss:
I have no objections.

geniejhang:
@georgweiss Did you write this tool? If it is and it works fine with ESS, I don't think I should push further update on the migration tool.
I found missing snapshots because the number of PVs in snapshotItems is larger than the number of PVs in a saveset.
I'm not sure how this situation happened, but it worked fine with us with CS-Studio somehow.

Duplicate snapshots part may need to be pushed if I fix it.

geniejhang:
@georgweiss Duplicate snapshots turned out to be real.
There were two commits with the same name somehow.

You can just merge these. There's no bug in the tool.

georgweiss:
Yes I did put together the migration code, but we do not need it any longer. Feel free to adopt it to your needs.

Issue #1441, opened 2020-07-31
Polyline drawing behavior a bit confusing.


georgweiss:
Not sure if this is an issue, but the following was reported on the polyline widget:
After placing the widget in the OPI user wishes to change line width. However, the polyline is still in edit mode, so clicking in the line width input field will create another point in the display area. This point is of course not visible as it obscured by the edit tools. Further, when the mouse is moved outside the visible display area to the line width input field, the cursor changes from the edit cursor (green plus sign) to an input cursor. Not possible to take screen shots, though.

If the above is not the intended behavior the I guess editing of the line should be "paused" or terminated when mouse is moved outside the visible display area.

A user also reports that occasionally it is in this scenario not possible to terminate polyline editing (using ESC). The following stack trace is dumped:

`2020-07-30 16:33:49 SEVERE [org.csstudio.display.builder.model] Unknown listener org.csstudio.display.builder.editor.poly.PointsBinding@1b5751d4
java.lang.Exception: Unknown listener
at org.csstudio.display.builder.model.properties.PropertyChangeHandler.removePropertyListener(PropertyChangeHandler.java:119)
at org.csstudio.display.builder.editor.poly.PointsBinding.disposeEditor(PointsBinding.java:126)
at org.csstudio.display.builder.editor.poly.PointsBinding.done(PointsBinding.java:228)
at org.csstudio.display.builder.editor.poly.PointsEditor.lambda$new$0(PointsEditor.java:101)
at com.sun.javafx.event.CompositeEventHandler$NormalEventFilterRecord.handleCapturingEvent(CompositeEventHandler.java:321)
at com.sun.javafx.event.CompositeEventHandler.dispatchCapturingEvent(CompositeEventHandler.java:98)
at com.sun.javafx.event.EventHandlerManager.dispatchCapturingEvent(EventHandlerManager.java:221)
at com.sun.javafx.event.EventHandlerManager.dispatchCapturingEvent(EventHandlerManager.java:180)
at com.sun.javafx.event.CompositeEventDispatcher.dispatchCapturingEvent(CompositeEventDispatcher.java:43)
at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:52)
at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
[...]

PR #1451, opened 2020-08-03
Polygon/line editor: Leave 'append' mode when mouse exits the view

kasemir:
As described in #1441, the 'append' mode is only useful within the view pane and interferes with for example setting properties.


georgweiss:
Thanks.
I'm not that familiar with the PolyLine, cannot find any means to add segments once the append mode has been abandoned. Is this correct?

kasemir:
Kindly try the poly line example.
**"Pressing the 'Space' key toggles between EDIT and APPEND mode."**

Screen Shot 2020-08-04 at 8 33 25 AM

georgweiss:
So I learn something new today too.

PR #1453, opened 2020-08-05
Picture widget fixes

krisztianloki:
Fix a race condition during initialization and correctly determine the native aspect ratio of SVGs


georgweiss:
Can you please elaborate a bit on the aspect ratio issue?

krisztianloki:
When the image is loaded in `contentChanged()` the aspect ratio is calculated and stored in  `native_ratio`. This value is later used in `updateChanges()` to determine the actual size of the `ImageView`. If the image in question was an SVG the aspect ratio was incorrect:
- before your changes the SVG was already converted to the size of the widget, so the aspect ratio of the widget was used
- with your changes the aspect ratio of the placeholder image was used

georgweiss:
OK, thanks for fixing. However, when testing I could not see any side effects.

PR #1455, opened 2020-08-06
CSSTUDIO-1316 Some widgets initially disabled when PV disconnected

georgweiss:
Fixes issue where some widgets are initially disabled if associated PV is disconnected. With this change the behavior is more consistent across all widgets.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/BoolButtonRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/BoolButtonRepresentation.java
@@ -353,8 +353,7 @@ public class BoolButtonRepresentation extends RegionBaseRepresentation<ButtonBas
         }
         if (dirty_enablement.checkAndClear())
         {
-            final boolean enabled = model_widget.propEnabled().getValue()  &&
-                                    model_widget.runtimePropPVWritable().getValue();
+            final boolean enabled = model_widget.propEnabled().getValue();
             jfx_node.setDisable(! enabled);
             Styles.update(jfx_node, Styles.NOT_ENABLED, !enabled);
         }
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/CheckBoxRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/CheckBoxRepresentation.java
@@ -227,8 +227,7 @@ public class CheckBoxRepresentation extends RegionBaseRepresentation<CheckBox, C
             // Don't disable the widget, because that would also remove the
             // context menu etc.
             // Just apply a style that matches the disabled look.
-            enabled = model_widget.propEnabled().getValue() &&
-                      model_widget.runtimePropPVWritable().getValue();
+            enabled = model_widget.propEnabled().getValue();
             Styles.update(jfx_node, Styles.NOT_ENABLED, !enabled);
             if (model_widget.propAutoSize().getValue())
                 sizeChanged(null, null, null);
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ComboRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ComboRepresentation.java
@@ -274,12 +274,7 @@ public class ComboRepresentation extends RegionBaseRepresentation<ComboBox<Strin
         }
         if (dirty_enable.checkAndClear()  &&  !toolkit.isEditMode())
         {
-            final boolean enabled = model_widget.propEnabled().getValue()  &&
-                                    model_widget.runtimePropPVWritable().getValue();
-            // When truly disabled, the widget no longer reacts to context menu,
-            // and the cursor will be ignored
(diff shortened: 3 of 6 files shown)


kasemir:
Will this enable the widgets even when you cannot write?
EPICS display tools tend to disable widgets when the PVs are read-only.

kasemir:
The `enablement.bob` example can be used to test.

kasemir:
With this change, applied to the BoolButton, the button will appear 'clickable' (enabled), even though it cannot operate since the PV is read-only:

Screen Shot 2020-08-06 at 10 00 27 AM

georgweiss:
I see your point. However, the Spinner widget shows the same behavior, i.e. when attached to a read only PV it is not disabled and silently rejects write operations. My point is that in order to make the initial state consistent across all widgets we should maybe consider making all widgets that may write enabled/editable until the point where the underlying PV connects and sets the PV writable property to false.

ralphlange:
Technically I would see "disconnected" as "not writable". (As you can't write when it's not connected.)
But consistency clearly has highest priority.

(Following my line, everything - also the reading widgets - would have to start disabled, and flip after the PVs connect. That would be technically correct but lead to a lot of flickering that very few users would appreciate.)

georgweiss:
Agree with @ralphlange , consistency is in my view imperative. We cannot expect all users to understand the fine-grained details of Pheobus or EPICS, so inconsistent behavior confuses, creates frustration and support tickets.

georgweiss:
Closing this unmerged as this issue requires an agreement on how to consistently display the initial state of a widget with PV.

kasemir:
> the Spinner widget ..  when attached to a read only PV it is not disabled 

That needs to be fixed in the Spinner widget. It should be 'disabled' like the Text entry, bool button, checkbox, combo, slider and others when it can't operate. 'Disabled' with the caveat that it might only appear disabled, since fully disabling in the JavaFX sense would also remove the context menu, and then you could no longer check which PV is disconnected resp. read-only. So in practice it might only apply a style that makes it look disabled.

kasemir:
> (Following my line, everything - also the reading widgets - would have to start disabled, and flip after the PVs connect.

Reading widgets appear disconnected when they are disconnected.
Since you can't 'do' anything with a text update, i.e. you cannot enter a text or click it to write to a PV, there is no need to 'disable' reading widgets. So reading widgets are either disconnected or connected.
Writing widgets on the other hand can be disconnected, connected to read-only PV, or connected to writable PV. In the former two states they are disabled since you can't operate them.

ralphlange:
Completely agree.

However, at ITER we had a _big_ discussion and applied a "fix" so that a freshly opened panel (BOY) starts with everything looking connected. Users did not like the impression that no matter what panel they opened, everything looked bad (disconnected) and only slowly and part-by-part changed appearance to good (connected). "That pink flashing has to disappear."

georgweiss:
So maybe this all boils down to the following:
1) Spinner widget needs fix to handle read-only PV.
2) BoolButton needs fix to handle context menu. Put button into Pane like in ActionButton?

kasemir:
> everything looked bad (disconnected)

Yes, I've been going back on forth on that as well.
It's a little unnerving when each newly opened display shows all pink/disconnected until the PVs connect, usually within a second. So should the PV layer delay the 'disconnected' state for say 5 seconds? But then what value are you supposed to show when you don't have a value? For displays, I don't want to make up bogus values just to avoid 'disconnected' when in fact the PV is disconnected.

shroffk:
> That needs to be fixed in the Spinner widget
+1 I agree with Kay here... Fixing the spinner would bring the consistency we all agree we need.

@ralphlange... I think everything connected was a wrong solution to a valid issue. With phoebus the screens load faster and the original issue of the users is addressed.

I really find the context menu with pv name very useful. I have been trying to find ways in which a disabled widget could be made to listen to the creatcontextevent...however no luck. This usecase seems very uncommon.

kasemir:
> Spinner widget needs fix to handle read-only PV.
Yes

> BoolButton needs fix to handle context menu. Put button into Pane like in ActionButton?
Try just removing the `jfx_node.setDisable(! enabled);`. It does already apply the 'disabled' looking style.

ralphlange:
Our "fix" was exactly what Kay mentioned: delay showing the unconnected state for the short time span that would solve 90% of the cases.

As an engineer, I actually prefer the flickering: I see it as a visual acknowledgement of connections happening successfully.
The main stream impression was: Look! The control system is slow and bad.

georgweiss:
removing `jfx_node.setDisable(! enabled);` fixes the initial state, but will not disable the button once the connection to a read-only PV is done. 
Also, write widgets that are initially disabled if PV is not connected would imply that PV is read-only, but since PV is not connected, that information is missing.
I'm fine if this is the default behavior, but do all write widgets exhibit the same behavior (Spinner excluded).

kasemir:
> removing jfx_node.setDisable(! enabled); fixes the initial state, but will not disable the button once the connection to a read-only PV is done.

Yes, because this is only evaluated for `dirty_enablement`:
```
final boolean enabled = model_widget.propEnabled().getValue()  &&
                                    model_widget.runtimePropPVWritable().getValue();
```

You need to update the handling of `dirty_enablement` vs `update_value`. Maybe merge them so that a value update will also trigger the style correction.

georgweiss:
Not sure that will do any difference... setting the native JavaFX button to disabled has a side effect on right click: it's the "background" (i.e. display) that will receive the mouse click event. That's why I suggested putting the button inside a Pane.

PR #1458, opened 2020-08-07
Use PV.isDisconnected() instead of pv.read()==null

krisztianloki:
Formula PVs are not necessarily return `null` after initialization if the input PVs are disconnected. PV.isDisconnected() correctly checks for the disconnected state.
#1387


kasemir:
Good, thanks!

PR #1460, opened 2020-08-07
Run else branch of rule when PV disconnects

krisztianloki:
When a PV that is used in a rule disconnects (after the rule has run at least once) run the 'else' branch of the rule to restore the original state of the widget.
This is useful when a widget's visibility is controlled by a rule; when the PV disconnects the widget becomes visible again if that was the initial state

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
@@ -234,6 +234,7 @@ public class RuleToScript
         final StringBuilder script = new StringBuilder();
         script.append("## Script for Rule: ").append(rule.getName()).append("\n\n");
         script.append("from org.csstudio.display.builder.runtime.script import PVUtil\n");
+        script.append("from java import lang\n");
         if (pform == PropFormat.COLOR)
             script.append("from ").append(WidgetColor.class.getPackageName()).append(" import WidgetColor\n");
         else if (pform == PropFormat.FONT)
@@ -383,8 +384,12 @@ public class RuleToScript
         else
             script.append(try_indent).append(setPropStr).append(formatPropVal(prop, -1, pform)).append(")\n");
 
-        script.append("\nexcept PVUtil.PVHasNoValueException:\n");
+        script.append("\nexcept (Exception, lang.Exception) as e:\n");
         script.append(try_indent).append(setPropStr).append(formatPropVal(prop, -1, pform)).append(")\n");
+        script.append(try_indent).append("from org.csstudio.display.builder.runtime.script import ScriptUtil\n");
+        script.append(try_indent).append("ScriptUtil.getLogger().info('Restored default value because of exception')\n");
+        script.append(try_indent).append("import traceback\n");
+        script.append(try_indent).append("ScriptUtil.getLogger().warning(traceback.format_exc())\n");
 
         return script.toString();
     }
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
@@ -278,7 +278,9 @@ public class RuleToScript
             {
                 final String varname = entry.getKey();
                 /*
-                 * PVUtil.getDouble() (used for pv0, pv1, etc) does not throw an exception when the PV is disconnected
+                 * PVUtil.getDouble() (used for pv0, pv1, etc),
+                 * getSeverity() and getLegacySeverity() do not throw
+                 * an exception when the PV is disconnected:
                  * Let's make sure that we have a pvInt for every PV
                  */
                 if (expr_to_check.contains(varname) || varname.contains("pvInt"))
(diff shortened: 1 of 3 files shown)


app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
```diff
@@ -276,13 +276,14 @@ else if (pform == PropFormat.FONT)
             for (Map.Entry<String, String> entry : pvm.entrySet())
             {
                 final String varname = entry.getKey();
-                if (expr_to_check.contains(varname))
+                /*
+                 * PVUtil.getDouble() (used for pv0, pv1, etc) does not throw an exception when the PV is disconnected
+                 * Let's make sure that we have a pvInt for every PV
+                 */
+                if (expr_to_check.contains(varname) || varname.contains("pvInt"))
```
kasemir, line 283: If the rule only uses `pvStr0` because the data is a string, will adding a `pvInt0` cause problems, since the string cannot be parsed as an integer?

app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
```diff
@@ -370,11 +377,14 @@ else if (pform == PropFormat.FONT)
...
         {
-            script.append("else:\n");
+            script.append(try_indent).append("else:\n");
             script.append(indent).append(setPropStr).append(formatPropVal(prop, -1, pform)).append(")\n");
         }
         else
-            script.append(setPropStr).append(formatPropVal(prop, -1, pform)).append(")\n");
+            script.append(try_indent).append(setPropStr).append(formatPropVal(prop, -1, pform)).append(")\n");
+
+        script.append("\nexcept PVUtil.PVHasNoValueException:\n");
```
kasemir, line 386: Instead of falling back to the original value when the PV has no value, one could argue that it makes sense to fall back to the original value for any error, so just `except:`, no specialization for `PVHasNoValueException`

app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
krisztianloki: I've just tested it and it did not complain. By the way, I was wondering why getDouble() returns NaN when the PV is not connected instead of throwing an exception?

app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
krisztianloki: Makes sense. But with a simple `except:` we will lose the stacktrace in the log. Is it possible to catch both Python and Java exceptions **and** print the stacktrace?

app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
kasemir: OK, thanks for checking.
For a double, NaN is a common representation of "I don't have a number". For an integer, there is no way to express "I don't have a number". Values like -1 or Integer.MIN_VALUE are all potentially valid numbers.

krisztianloki: Wouldn't it make sense to add a `getDoubleWithException()` variant that would throw an exception instead of returning NaN? Originally rules were not executed when the PV was disconnected so it was impossible to take advantage of checking NaN... and we could get rid of the "mandatory" pvInt

kasemir: getDouble() is meant as a quick way to get a double. Any problem turns into NaN.
If you care about the nature of a specific problem, like PV has no connection, you can already check via PV.isDisconnected.

So maybe instead of changing the PVUtil in all those places, just add a PV.isDisconnected check to the generated rule code.

app/display/model/src/main/java/org/csstudio/display/builder/model/rules/RuleToScript.java
kasemir: Similar to issue above: If you consider disconnected as a key state, then just check for that via PV.isDisconnected instead of updating the PVUtil to throw exceptions, adding a `pvInt` just to provoke the exception, and then having to catch that exception.

krisztianloki: If it is okay to fall back to the original value in case of any exception that is fine by me; I'll revert PVUtil.java
I will also print the stacktrace of the exception we caught.

app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/internal/RuntimeScriptHandler.java
```diff
@@ -45,7 +45,8 @@
     private final Widget widget;
     private final List<ScriptPV> infos;
     private final Script script;
-    private final boolean check_connections;
+    private final boolean is_rule;
+    private boolean check_connections;
```
kasemir, line 49: `check_connections` was `final`. Now it's changed at runtime, on different threads, so needs to be `volatile`.

krisztianloki: Oops, you are right...

krisztianloki:
I've also reintroduced PVUtil.PVHasNoValueException without changing the signature of the PVUtil functions to not print error messages every time the rule is run with a disconnected PV

kasemir:
Yes, this looks like it could work, let's go with it.

PR #1462, opened 2020-08-10
CSSTUDIO-1316 Boolean Button behavior when PV disconnected or read-only

georgweiss:
Boolean Button now behaves like Action Button.
Screenshot 2020-08-10 at 09 13 45
Screenshot 2020-08-10 at 09 15 07


kasemir:
Totally unrelated: In your screenshots, the fonts appear to have a spacing problem.

What should look like this..
Screen Shot 2020-08-10 at 1 50 41 PM

instead looks like this with overlapping characters:

Screen Shot 2020-08-10 at 1 50 50 PM

Is that just a side effect of the screenshot being resized, or is that really happening?

georgweiss:
That is a MacOS issue since an OS update a while ago. Haven't had time to analyze it.

PR #1464, opened 2020-08-12
Support changing points in rules

krisztianloki:
Also simplified RuleToScript a bit

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/properties/Points.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/properties/Points.java
@@ -18,6 +18,21 @@ public class Points implements Iterable<Point>, Cloneable
 {
     private final List<Double> points = new ArrayList<>();
 
+    public Points()
+    {
+    }
+
+    public Points(double... points)
+    {
+        if ((points.length % 2) == 1)
+            throw new IllegalArgumentException("Number of points must be even");
+
+        for (int i = 0; i < points.length; i += 2)
+        {
+            add(points[i], points[i + 1]);
+        }
+    }
+
     /** @return Number of points */
     public int size()
     {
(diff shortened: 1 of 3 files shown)


kasemir:
Do you have an example for changing points in a rule?

krisztianloki:
Do you mean like this one?
[polyline.bob.gz](https://github.com/ControlSystemStudio/phoebus/files/5064817/polyline.bob.gz)

kasemir:
Fascinating. The example can actually be opened in the original editor, the rule appears fine, can see & edit the points, just won't actually create a functional script: 

Screen Shot 2020-08-13 at 8 25 23 AM

I'm always worried about turning the display into a video game. To remain portable, displays should concentrate on add-widget, set-pv, and not rules which totally change what you see.
But whatever, when users complain in 5 years that their 95% rule-based displays no longer work, we'll say "I told you so"

bsobhani:
Is there a way to assign the value of a pv to a point's x or y value, so that as you change a PV a point moves up and down? I tried checking "Value as Expression", entering true for "Boolean Expression", and then entering [0,pvInt0,0,0] for "Value". But when I changed the PV for pv0 nothing seemed to happen.

I can make a rule to move the entire widget up/down by using the "y" property, but I cannot seem to move an individual point.

kasemir:
To allow `[0,pvInt0,0,0]` for the value, you'll have to update `PointsWidgetProperty.setValueFromObject`. For now, `setValueFromObject` only allows `Points`, not `List<Double>` or whatever `[...]` becomes when jython passes that in.

bsobhani:
I see. Thanks.

krisztianloki:
Actually `Points` is already imported; it is needed to restore the default state. So all you have to do is change `[0,pvInt0,0,0]` to `Points([0,pvInt0,0,0])`.

As a general rule, it is always a good idea to click on `Show script` to see how the initial / default state is set

PR #1466, opened 2020-08-13
Issue 1134 array formula functions

georgweiss:
Previously not integrated implementation of array formula functions.


kasemir:
For these array formula functions, their arguments show up in the auto-completion suggestions, but their function names are missing:

Screen Shot 2020-08-13 at 2 33 16 PM

georgweiss:
Will fix it.

shroffk:
The array functions override the default `getSignature()`
While we should surely add the function name.... I also see some value in having the argument type.

georgweiss:
...some value in having the argument type.

Being explicit specifying the V* types may be overkill and may not be of help to the user. I will update to make the types a bit clearer, e.g. String instead of VString, Number instead of VNumber etc.

kasemir:
Yes, the `V*` type is not too useful. With 'index', 'exponent', .. it should be obvious that it's a number, not a string.
See existing examples which are terse but helpful:

 * `sqrt(x)`: Any number x
 * `pow(base, exponent)`: First the base, then the exponent
 * `hypot(x, y)`: First the 'x' coordinate, then the 'y' coordinate
 * `atan2(y, x)`: First the 'y' coordinate, then the 'x' coordinate
 * `concat(String, ...)`: Needs strings, one or more

If more detail is deemed necessary, that can be provided in the online help, as has been done for the array functions.

PR #1467, opened 2020-08-14
Handle runtime width and height change in Poly* widgets

krisztianloki:
Polygone and Polyline widgets didn't handle width and height changes during runtime. With this PR they do.
I've decided not to override `defineProperties` in `PolyBaseWidget` but to add a separate function to keep the order of properties when saving the widgets.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/PolyBaseWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/PolyBaseWidget.java
@@ -15,8 +15,6 @@ import org.csstudio.display.builder.model.WidgetProperty;
 import org.csstudio.display.builder.model.properties.Points;
 
 /** Base for widgets that display points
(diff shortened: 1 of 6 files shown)


kasemir:
I'd be OK with changing the order, i.e. basically points 'first', since they're common to Polygon & line, then specializations later. Pick what's cleaner in the code.

krisztianloki:
Overriding `defineProperties` is definitely cleaner. I'll update the code.

pavel-ch:
Trying to create polyline or polygon in Editor: points are outside the widget area and inconsistent with mouse clicks in editor. Also after widget resize, the points are out of sync with handles in editor, see:
image
Reproducible on Windows.
I use a site specific copy of the original implementation of polyline widget due to some added features (two-color line) and this original code works just fine in this regard.
Please look at it or revert, thank you.

krisztianloki:
You are right @pavel-ch, I've just opened a PR to fix this: #1473

PR #1474, opened 2020-08-19
Save and restore: Bug fix and some updates

geniejhang:
Main changes are following

1. VEnum choices shown correct list.
2. Limit user input of node name to DB limit.

### DIFF
--- a/app/save-and-restore/model/src/main/java/org/phoebus/applications/saveandrestore/model/Node.java
+++ b/app/save-and-restore/model/src/main/java/org/phoebus/applications/saveandrestore/model/Node.java
@@ -52,6 +52,7 @@ import lombok.ToString;
  *
  */
 public class Node implements Comparable<Node>{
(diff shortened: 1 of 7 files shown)


georgweiss:
Two comments:
1) There is some code duplication in SnapshotTable and SnapshotTreeTable. Can this be reduced?
2) Instead of limiting the length of the Node name, can you instead adda migration script to increase the length of the table column (e.g. 512). That way we can remove some code, and also get rid of the coupling between the column definition and the hard coded value in Java.

geniejhang:
@georgweiss 
1. Yes. But let's leave it for now. That'll be ultimate goal for the other part of the code. I created some duplicate codes for supporting default UI and FRIB-preferred UI. Since the feature implementation is finishing here, the next step will be optimization for it :)
2. Also, yes. That also works. I'll just update it for 512 chars in an hour.

Issue #1463, opened 2020-08-12
XY graph autoscaling


shroffk:
We are trying to use the xygraph for both horizontal and vertical plots.

I have 2 waveform/array pv's. One pv holds a list of indices while the other holds the values

When I use the indices for the x axis and the value for the y axis then the autoscaling of both works fine.

When I reverse this the autoscaling of the y axis does not work... I am not sure I understand what is going wrong here. 
Sometimes simply adding a few extra elements to either of the arrays trigger the scale to be calculated correctly.

I added breakpoints to the rtplot and verified that it has all the correct data points (all 9 of them)...with the correct x and y value.

![image](https://user-images.githubusercontent.com/2111304/90022279-59477000-dc80-11ea-9c8a-789641ed1dfe.png)

shroffk:
here is the example screen:

[xf-grpah.zip](https://github.com/ControlSystemStudio/phoebus/files/5063055/xf-grpah.zip)

kasemir:
In PlotProcessor.autoscale, the value range is only checked for samples within the current X range.
Maybe print out the result of `determinePositionRange` and see if that explains what's happening.

shroffk:
Thanks, 
You are kind of right, the main issue is in the following
```
// Last sample is the one just inside end of range.
int stop = search.findSampleLessOrEqual(data, position_range.getHigh());
```
in the determineValueRange

The way the stop index is determined is the cause of the above behaviour.

The position range is being calculated correctly.... in the above case it is 0-100. The stop index is calculated at "4", since that is a index where the y value is 100.

kasemir:
For tests you can replace that start..stop determination with
```
start = 0;
stop = data.size()-1;
```

I guess we need to skip that start..stop determination and just use 0.. data.size()-1  if the 'position' data is not ordered (because in this case it's basically swapping the 'position' and 'value').

shroffk:
I feel like I am clearly missing some assumption about the values being used as index...cause moving some values results in the complete range being used 

![image](https://user-images.githubusercontent.com/2111304/90036569-34a7c400-dc91-11ea-9a51-f94f79106e88.png)

kasemir:
The find-sample stuff is a binary search. For ordered data, that's the fastest.
For non-ordered data, moving one value can get different results.

shroffk:
> For non-ordered data, moving one value can get different results.

yes,
[...]

PR #1475, opened 2020-08-19
XYPlot: Fix 'autoscale' of X axis when data is not ordered

kasemir:
#1463


oksanagit:
Looks fantastic ! Thank you both.
image

PR #1480, opened 2020-08-20
(Optional after #1474) Save and restore: CSV import/export of saveset/snapshot (backward compatible)

geniejhang:
@georgweiss This was requested by FRIB users.
If you don't think think is not that useful for public, please just reject it.

If this commit is accepted, it'll show the followings.
For the context menu of
Folder: shows importing saveset to the selected folder
Saveset: shows importing snapshot to the saveset, exporting the saveset to CSV file.
Snapshot: shows exporting snapshot to CSV file.

The exported CSV files are compatible to Git repository files and vice versa.

In case there's duplicate snapshot name, it'll show the prompt to change the importing name.
In case snapshot is not compatible to the saveset selected, it asks if user wants to create a saveset from snapshot.


georgweiss:
At ESS we for sure would not need it, but other sites may. Suggest you add a preference setting that would enable this, default=false.

PR #1483, opened 2020-08-21
CSSTUDIO-1124 Add create display item in file browser for folder nodes

georgweiss:
Users are requesting possibility to create a display from file browser.
Screenshot 2020-08-21 at 16 38 14

Also a small fix in create folder prompt.


shroffk:
Hi,
I will create a template action which adds to the file browser context menu using SPI and the ContextMenuService, SelectionService

You can refine it further.

georgweiss:
Do not merge!
Need to change a few things.
Sorry.

georgweiss:
This PR also include a fix for silent overwrite of a display file when launched from the application menu.

Support for new plot file from file browser will follow in separate PR.

PR #1484, opened 2020-08-21
Save & Restore: Setting threshold for highlighting PVs by their delta values

geniejhang:
@georgweiss I found threshold setting method is commented out.
So, I uncommented and implemented corresponding UI as below.

Delta cells are highlighted only when they're above delta threshold.
I also put the percentage number next to the absolute value.


georgweiss:
I do not quite understand the screen shots. Why is the delta live setpoint not highlighted in the second screen shot?

geniejhang:
Because it didn’t pass the Delta threshold. 21.66% is within 22%

georgweiss:
Of course, my bad. Sorry.
Appreciate you are working on this, the code was commented out as I did not consider it essential for an initial release of S&R. 
In any case, did the commented out code include the percentage indication next to the value? I'm asking since I think the percentage indication impacts the readability. After all, calculating the difference in percent is not that hard...

geniejhang:
No. Commented out part only sets the threshold values. Percentage part I added. I can make it optional if you like.

georgweiss:
Yes, please make it optional, e.g. a checkbox or other UI item to toggle between show/hide.

geniejhang:
Got it. Toggle button sounds good to me. 
I’ll do it tomorrow since it’s 3 am now.
I was fighting with a bat intruded in my house.

geniejhang:
@georgweiss Now it has toggle button at the top.

ralphlange:
Looking at the pretty screenshot...
Is there a Phoebus policy for labeling toggle buttons?
I.e.: Does the label show the current state, or does it show what happens when you press the button?
(Personally, I am afraid of clicking on any toggle button exactly because of that question.)

kasemir:
When a button is both 'indicator', it can indeed be confusing. Does it show what you'll get, or what you have now?
For the alarm table's "maintenance mode" button, the tool tip explains what you have and what you'll get, and I think in general a tool tip can be a good option to handle this.

geniejhang:
@georgweiss Any comments?

georgweiss:
On top of the tooltip, the pressed state of the button should show percentage. Otherwise no comments.

geniejhang:
Changed to "If pressed, show \u0394 percentage when PV has non-zero \u0394"

PR #1488, opened 2020-08-24
Fix IndexOutOfBoundsException when loading symbol widget

krisztianloki:


shroffk:
Shouldn't we log the exception when it occurs?
Maybe at the FINE or DEBUG level

kasemir:
Yes.
We decided to not have code conventions, #1031, but if we had:

Exceptions
----------

The best way ... often ... is to log it...
...
```
     catch (Exception ex)
     {
...
         // Better:
         // Uses logging, shows complete exception detail
         // with message (if there is one),
         // stack trace, nested exception ('cause')
         logger.log(Level.WARNING, "Failed to solve big problem", ex);
```

shroffk:
I think the scope of the ticket/s was too broad... 90%+ agreement was lost over 10% of disagreement.
I will try to put together a plan were we can start to reinforce our agreements...

Regarding exception handling I think there is universal agreement that we should log it as shown by Kay.

@krisztianloki thanks for finding this issue and fixing it :).. could you add the logging of the exception

krisztianloki:
The exception is caused by the `getElement()` call introduced by my previous PR: #1471 when the `SymbolWidget` is being parsed (with more than 2 symbols) and the `symbols` property is being populated, so it is kind of an "expected exception" that is why I do not log it.

kasemir:
>  so it is kind of an "expected exception"

Then I agree it's fine to not log it, but I would put that into a comment:
```
catch (Exception e)
{
   // Expected, thus ignored
}
```

PR #1489, opened 2020-08-24
#1483 Contributions to file broser context menu using SPI

shroffk:
WIP - do not merge
an alternative to the #1483 PR


kasemir:
This could work!
Does it rely on the gentlemens' agreement between the file browser and the display builder that `File` will always refer to a directory?
Technically, the receiver of a file can always call `getParent()` when a directory is needed. To the end user it should be OK to invoke "New Display" on a file and then the new display is created in that file's folder. It should not replace the original file on which the context menu happened to be invoked.

So does it not matter if `File` is a plain file or folder?
If it does, we could just use the exising `File` for a plain file and `Path` for a folder, and put that into the Javadoc of the `ContextMenuEntry`:
```
 * Generally supported types:
 * Process Variable - Represents a PV name
 * File: Should refer to a plain file
 * Path: Should refer to a directory.
```

Alternatively, you could also define a simple wrapper class to identify a folder:
```
  class Directory extends File
    {
        public Directory(URI uri)
        {
            super(uri);
            if (! isDirectory())
                throw new IllegalArgumentException(uri + " is not a directory");
        }
    }
```
That would assert that `Directory` is just that. But one could still use `File` for a folder, so also define `PlainFile extends File` for a plain file?? Maybe that's just too much effort?

shroffk:
> Alternatively, you could also define a simple wrapper class to identify a folder:

I thought of this, but it starts looking like eclipses IPath and IFile...

I much rather prefer the other solution.

> Technically, the receiver of a file can always call getParent() when a directory is needed.

We can add support for both file and folder selections...in the ContextMenuAction, if file then use parent...else use the folder.

kasemir:
>  ...IPath and IFile...

Right, would be silly to introduce our own wrappers when they're not 100% necessary.
.. which for now they aren't. The Display Builder's ContextMenuAction can handle folder (just use) as well as file (get its parent folder), and we can worry about other tools that need a 'File' context menu entry later.

georgweiss:
>Does it rely on the gentlemens' agreement between the file browser and the display builder that File will always refer to a directory
This is news to me. How is this manifested in code and APIs?

shroffk:
>  How is this manifested in code and APIs?

Currently, the context menu items contributed via SPI are added to the context menu only when there is a single selection and the selection is a folder

https://github.com/ControlSystemStudio/phoebus/blob/8dbd0e5bf5eb2aa94c12b77aa5163edb1d75e364/app/filebrowser/src/main/java/org/phoebus/applications/filebrowser/FileBrowserController.java#L428

> Should not 1) also show an error dialog?

yes, you are correct... I just wanted to put together the example for the context menu with SPI which did not need direct dependencies. @georgweiss it would be great if you could help add the error/exception handling for this action :). We could also check if the folder is writeable and either show an error message or add a warning to the dialog.

> "create new display" when a regular file is selected in the file browser. To me it is counter intuitive: what does it mean to create a new display in a regular file? User would need to be familiar with the underlying logic.

I see the logic...however we usually have dozens of files in a folder which fill the whole screen... so going to the parent folder would mean a lot of scrolling up. I think just using the parent would be nice once people learn about it, after all the action still opens a dialog which needs user interaction.

image

georgweiss:
As for supporting "create new display" when regular file is selected: I see your point with folders holding large amount of files. Plus it is currently also possible to create a new folder when a regular file is selected. I dislike it nevertheless...

I will add error handling, but my point is that the current action from the application menu overwrites an existing file, so there is a need to add error handling there too. So: the FileChooser will not show any prompt if user specifies a file name without extension. We hence need to check if a file with the full file name (i.e. including .bob) exists and fail on that.

georgweiss:
I'm afraid there is a detail we need to agree upon in the case user specifies a file name corresponding to an existing file. This happens in either of these two cases:
1) The native FileChooser will prompt the user if the specified file name matches an existing file. User may cancel or select to overwrite (at least on Mac and Windows).
2) User selects a file name that is the same as an existing file, but without extension. Next ModelResourceUtil.enforceFileExtension will add the extension so the java.io.File objects corresponds to an existing file.

How do we make the two cases consistent? If user in 1) chooses to overwrite, case 2) must also prompt for overwrite. On the other hand, if 2) should never overwrite, then case 1) will not do what the user has asked for (i.e. to overwrite).

shroffk:
> How do we make the two cases consistent? If user in 1) chooses to overwrite, case 2) must also prompt for overwrite. On the other hand, if we should never overwrite, then case 1) will not do what the user has asked for (i.e. to overwrite).

Ok, 
I think we can have a confirmation dialog after File object is created. If the File already exists ask the user for confirmation to overwrite.

I know this might be annoying since in case 1. the user has already asked to override the file... however this is an uncommon activity and a second warning might be ok. The annoyance results in consistency... when you are going to erase a bob file you will be asked "at least once" to confirm your action.

I hope the infrequency of this situation justifies the above action.

georgweiss:
To clarify: in use case 1) user will be prompted only once, by the native file chooser. In case 2) the prompt will be a custom confirmation dialog. Case 1) and 2) are mutually exclusive.

shroffk:
It seems like this PR has conflicts now

georgweiss:
I have already merge the (modified) code in the original PR. This should be closed without merge.

Issue #1494, opened 2020-08-25
Fall back to `messages.properties` on a per message level


tmijail:
Currently we only fall back to `messages.properties` if `messages_xx.properties` doesn't exist.
If we have an incomplete `messages_xx.properties`, we fall back to the name of the field in `Messages.java`.

For example, if `messages.properties` has the following content
```
Hello=Hello
HowAreYou=How are you?
```
`messages_de.properties`, the following
```
Hello=Moin
```
and our locale is `Locale.GERMANY`, then the value of `HowAreYou` will be `<HowAreYou>`.

I think this is an issue because it means that if a localization is not completely up to date, it will offer a worse user experience than having no localization at all (apart from the default one). A better solution would be to first fall back to the default (english) localization (`Hello=Moin`, `HowAreYou=How are you?`).

We can achieve this by taking advantage of Java's `ResourceBundle`s. I have a possible implementation in PR [#1495](https://github.com/ControlSystemStudio/phoebus/pull/1495).

PR #1495, opened 2020-08-25
Fall back to messages.properties on a per message level

tmijail:
Currently we only fall back to `messages.properties` if `messages_xx.properties` doesn't exist.
If we have an incomplete `messages_xx.properties`, we fall back to the name of the field in `Messages.java`.

This provides a worse experience than having no localization at all unless localizations are completely up to date.

More information on issue #1494

With the changes in this PR, an incomplete localization will first fall back to english for the messages missing from `messages_xx.properties` and then to the field names for the messages missing from both `messages_xx.properties` and `messages.properties`.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetColorPopOver.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetColorPopOver.java
@@ -51,8 +51,7 @@ public class WidgetColorPopOver extends PopOver
 	    try
         {
             final URL fxml = WidgetColorPopOver.class.getResource("WidgetColorPopOver.fxml");
-            final InputStream iStream = NLS.getMessages(WidgetColorPopOver.class);
-            final ResourceBundle bundle = new PropertyResourceBundle(iStream);
+            final ResourceBundle bundle = NLS.getMessages(WidgetColorPopOver.class);
             final FXMLLoader fxmlLoader = new FXMLLoader(fxml, bundle);
             final Node content = (Node) fxmlLoader.load();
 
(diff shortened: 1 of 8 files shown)


kasemir:
This is great, since missing entries in localized texts can indeed be a problem.
@pavel-ch , I think you have the most localizations. Maybe check if those still work, i.e. if the original filename `"messages_" + Locale.getDefault().getLanguage() + ".properties"` is the same as what this update will use.

pavel-ch:
I can test on Monday, but this must work. What we are about to lose is an easy way to see (in the form of log warnings) what keys were added into basic English message files. That was really good help for translation maintenance. Since this commit, I have to observe all commits for message files additions, because in the runtime, missing localized strings will be silently (if I understand well) replaced by English strings. If there is any better way to find additions than watching commits, let me know.

kasemir:
You're right, in our own NLS implementation you would get warnings for missing as well as superflous messages, both useful for maintaining localizations.

With this update, the end user is shielded from "<NameOfField>" and instead falls back to the default (English) text.

I think best would be both:
Instead of replacing our NLS implementation with ResourceBundle calls, add the fallback to our code, so we get all: Warnings for maintainers and fallback for end user.

kasemir:
The `ResourceBundle` supports a property `resource.bundle.debug=true`, but that just seems to show some file info, no detail on missing/extraneous values in the file.

I'll work on an NLS update that reads both the local and the generic file and logs when there are missing/extra values, since I agree with Pavel that it's otherwise hard to keep the message files up to date.

tmijail:
There's also the possibility of generating a report of ResourceBundle inconsistencies during build. Something like this https://maven.apache.org/plugins/maven-pmd-plugin/l10n-status.html

I was checking that out just now. If you're open to that kind of solution, I could keep investigating.

kasemir:
Pavel, you decide: The new PR that logs missing/extra values, or you guys look into maven reporting on this.

kasemir:
Well, looks like I'd have to update the FXML-based code as well. Forgetaboutit, I'm out. You merge this and then have maven tell you what's missing, or leave as is, I don't care.

pavel-ch:
Sorry for the late answer, not in the office today. The new code looks good. A kind of maven report (a file?) on missing/extra values would be perhaps even better than the old log warnings, being at one time on one place. Thanks on investigating that!

kasemir:
OK, I suggest we merge this, and @tmijail  you can then look into a maven report as a separate effort.

tmijail:
Great. I'll look at that over the weekend and get back to you.

Issue #1468, opened 2020-08-14
Paste properties does not update the Properties view


krisztianloki:
When using the copy-paste properties function the Properties view is not updated. Have to select a new widget and go back to see the changes and the display needs to be reloaded for the changes to take effect.

kasemir:
Cannot duplicate.
Tried for example:
1) Create Label and Rectangle.
2) De-select 'transparent' on the label
2) Select Rect, copy properties 'background color'
3) Select Label, paste properties -> Label background updates, as does the color indicated in the properties.
4) Create another label
5) Select first label, edit text, copy properties 'text'
6) Select second label, paste properties -> Label text updates, as does the text shown in the properties view

krisztianloki:
Then it will be a bug in the `Text Symbol` widget; copying the `Symbols` property produced the result I described. The `Symbol` widget even threw an exception.

krisztianloki:
So, it seems that neither `ArrayWidgetProperty` nor `StructuredWidgetProperty` calls `firePropertyChange()` in `readFromXML()`.
In `ArrayWidgetProperty.readFromXML()` it is possible to call `setValue(elements)` instead of `value = elements` and it solves the issue but `StructuredWidgetProperty` overrides `setValue()` and throws an `IllegalAccessError` so `firePropertyChange()` needs to be explicitly called.
However, I'm not sure if there are side effects of adding these new notifications...

krisztianloki:
Fixed by #1497

PR #1497, opened 2020-08-26
Inform listeners of value changes

krisztianloki:
(Possible) fix for #1468 ; it does fix the issue but I don't know if it causes side effects.
Not sure about calling `setValue()` in `ArrayWidgetProperty.readFromXML()` because `setValueFromObject()` calls `firePropertyChange()` explicitly... which one is correct?


kasemir:
`setValue` vs. direct call to `firePropertyChange` should be nearly the same.
We've seen the side effect of _not_ sending change notifications, so let's see how this works out.

PR #1498, opened 2020-08-26
NLS: Fall back from messages_{Locale}.properties to default ...

kasemir:
... and log when texts are missing or extraneous in the default file or
the optional localized file.

Functionally like #1494 but will log warnings about missing/extra texts, while implementation in #1495 would be silent and thus complicate the maintenance of the message files.


kasemir:
Arg, would have to update the code that uses FXML as well. I give up.

PR #1501, opened 2020-08-27
TextEntry Selector for CSS Compatibility

TheDuz:
Adds Selector Property (DateTime File None) to TextEntryWidget.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -198,6 +198,7 @@ public class Messages
     public static String WidgetProperties_Font;
     public static String WidgetProperties_ForegroundColor;
     public static String WidgetProperties_Format;
+    public static String WidgetProperties_Selector;
     public static String WidgetProperties_GridColor;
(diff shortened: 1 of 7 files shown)


kasemir:
In BOY, the text entry field had an option to change from a plain text field into a file or date selector.
In the Display Builder, the text field is a plain text field. For file selection, there is a separate file selection widget.
The motivation is that just like right now we might be transitioning from *DM or BOY to the Display Builder, in 5..10 years we'll have to transition from that to the next tool, so keeping the widgets focused on a specific purpose is best for long term maintainability. See also https://github.com/ControlSystemStudio/phoebus/wiki/Display-Builder-Compatibility.

When importing legacy *.opi files, the old text-with-file-selector-option widget should be converted into a pair of text field and file selector widgets, see `TextEntryWidget` `configureFromXML`.
There is already code/comments to treat text-with-date-selector-option the same way, alas we have no Date/Time widget.

So if you want to support date/time entry, please add a standalone date/time widget, and then maybe update the legacy file support to translate text-with-date-selector-option into a pair of TextEntry and that to-be-implemented Date/Time widget.

PR #1504, opened 2020-08-27
Advanced 'Replace with...'

krisztianloki:
Morphs properties not just widgets: for example the `file` property of PictureWidget can be morphed to `symbols[0]` of SymbolWidget and vice versa
Plus, it only copies the tooltip if it is different from `getInitialTooltip()`

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/app/MorphWidgetsMenu.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/app/MorphWidgetsMenu.java
@@ -22,6 +22,7 @@ import org.csstudio.display.builder.editor.WidgetSelectionHandler;
 import org.csstudio.display.builder.editor.undo.AddWidgetAction;
 import org.csstudio.display.builder.editor.undo.RemoveWidgetsAction;
 import org.csstudio.display.builder.editor.util.WidgetIcons;
+import org.csstudio.display.builder.model.ArrayWidgetProperty;
 import org.csstudio.display.builder.model.ChildrenProperty;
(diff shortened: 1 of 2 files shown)


kasemir:
Well, this is somewhat tricky business.
When you have a text update field that's pretty much all default, just a PV name, and morph that into a text entry because you decided that you want to allow changing the PV after all, should morphing create the text update with the background, foreground, ... copied from the text update? That results in a text entry which looks like a text update, lacking the somewhat distinctive background color which suggests that this is a field where you can enter text. So should morphing only copy properties that are not at their default value?
Also, while being able to "morph" widgets is cute, it would be hard to support custom mappings of all widget types and all their properties. Like think about morphing X to Y and then Y back to X and trying to make it go back to the original state.

So I think it's simply not well defined how all the widgets should morph, and frankly not worth investing too much time into this very special operation. But OK to tweak it a little, like the tooltip check, which is actually like the idea of only copying properties that have been changed from their default.
I like the `doMorphProperty` implementation: It allows adding some specialized mappings in one place. The widgets themselves remain unaware, it's avoiding for example code inside the SymbolWidget and PictureWidged to make them depend on each other.

Do you want to look at updating it a little more so that only non-default values are copied over?

krisztianloki:
I've used 'Replace with...' *a lot* lately, that is why I dove into it.
I can update to skip non-default values, but I would like to unconditionally copy all properties in the Position category and also Name

kasemir:
> I've used 'Replace with...' a lot lately, that is why I dove into it.

Maybe you need to create your displays correctly in the first place ;-)

OK, let's merge this, and if you think it's worthwhile to do more about the handling/skipping of default values, you do that as time permits. Thanks!

PR #1506, opened 2020-08-28
Do not steal DEL and copy-paste keypresses

krisztianloki:
When a property is being edited while the mouse was inside the widget editor area pressing the DEL key deleted the widget instead of the character. I'm not sure if this PR fixes all corner cases but it definitely fixes the one that has bitten me several times.


kasemir:
Maybe the initial approach was wrong. Instead of making the `key_handler` an event _filter_ that receives every key click, and then having to sort out which key strokes to ignore, there might be a better way by adding that key handler to the tracker, i.e. that rectangle which appears on top of selected items, and naturally only receives key clicks when it has the focus.
Meanwhile, I can see how your update helps to tame the existing implementation.

PR #1507, opened 2020-08-28
Added forward compatibility

krisztianloki:
Do not load displays / widgets if their major version is higher than supported.
Also, disconnect the editor from the file if the model could not be loaded instead of creating a new model with the same filename.

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
@@ -477,21 +477,23 @@ public class EditorGUI
             {
                 canon_path = file.getCanonicalPath();
                 model = ModelLoader.loadModel(new FileInputStream(file), canon_path);
+                this.file = file;
             }
             catch (final Exception ex)
             {
                 canon_path = null;
                 logger.log(Level.SEVERE, "Cannot load model from " + file, ex);
                 ExceptionDetailsErrorDialog.openError("Creating empty file",
-                        "Cannot load model from\n" + file + "\n\nCreating new, empty file", ex);
+                        "Cannot load model from\n" + file + "\n\nCreating new, empty model", ex);
                 model = new DisplayModel();
                 model.propName().setValue("Empty");
+                // Don't associate this editor with the file we've failed to load
+                this.file = null;
             }
 
             if (! file.canWrite())
                 model.setUserData(DisplayModel.USER_DATA_READONLY, Boolean.TRUE.toString());
             setModel(model);
-            this.file = file;
 
             try
             {
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
@@ -514,7 +514,9 @@ public class EditorGUI
             if (canon_path != null && model.isClean() == false)
             {
                 ExceptionDetailsErrorDialog.openError("Errors while loading model",
-                        "There were some errors while loading model from " + file + "\nNot all widgets are displayed correctly; saving the display in this state might lead to losing those widgets. Please check the log for details.", null);
+                        "There were some errors while loading model from " + file + "\nNot all widgets are displayed correctly; " +
+                        "saving the display in this state might lead to losing those widgets or some of their properties." +
+                        "\nPlease check the log for details.", null);
             }
 
         });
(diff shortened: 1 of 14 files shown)


shroffk:
> It's OK to log a warning "Widget X has version 4, expected version 3", but throwing an exception and not showing that widget at all would severely impact operation when operators would otherwise have been able to use the display without any problems.

+1

krisztianloki:
Would it be okay to run the display as if there were no warnings (but logging the warnings and showing a messagebox to the operator) _but_ treating the widgets as unknown ones in the editor? The rationale is that saving an opi in an older editor will silently discard the unknown properties.

My idea with this PR was that compatible changes could be introduced by incrementing the minor or patch number while incompatible changes would increment the major version. A _compatible_ change would be for example adding a border or the `interactive` property to the `TextSymbol` widget. An _incompatible_ change would be for example changing the `symbols` property of `TextSymbol` and/or `Symbol` widget from `ArrayWidgetProperty` to `StructuredWidgetProperty` (similar to the `MultiStateLED` widget). Such a change would definitely impact the user experience as an older Phoebus version would execute the widget completely differently. I am planning to implement that transition and while it can be done by creating two new widgets I would rather not add new widgets that behave 90% similar to existing ones. In fact I would like to see the `LED` and `MultiStateLED` widgets merged too.

kasemir:
As for LED and MultiStateLED, they are separate on purpose.
See https://github.com/ControlSystemStudio/phoebus/wiki/Display-Builder-Compatibility#led-multistateled or pretty much any presentation that has ever been given on the display builder.
BOY had those widgets as one, and changing the number of states modified the list of available properties ?! In 'LED' mode, the value is a boolean or a bit in a number, whereas ' MultiStateLED' mode always treats the value as a number, you can't access a certain bit. That's just very different behavior, so they're on purpose two separate widgets.

If you want to update your text symbol to have totally different properties, that would also become a different widget.

krisztianloki:
> If you want to update your text symbol to have totally different properties, that would also become a different widget.

Even if it would be implemented in a backward compatible way? If the symbol - PV value association is zero based and sequential (like it is today) then the widget would be saved with the current version number and an ArrayWidgetProperty type `symbols` but if the association is different then it would be saved with version `3.0.0` and a StructuredWidgetProperty type `symbols`.

kasemir:
Based on what we learned from BOY, examples of what I want to avoid.

LED widget has on/off labels and connects to a boolean PV, or a bit in a numeric PV. Then you change the number of states from 2 to 3, and the on/off labels suddenly change into label 0, label 1, there's a new label 2 property, and the bit property vanishes.
Instead, we now have a plain LED with on, off, bit, and a multi-state LED with label 0, label 1, label 2, ...

An XYGraph widget with buffer size, append-or-replace, that can display a scalar over time or a waveform over index or two waveforms against each other. Seemed like a powerful idea at the time, but in reality it won't show anything if you don't match the buffer size to the waveform size?! If you want to show values over time, make sure that you configure the X axis to use a time format. At runtime, users can accidentally  change that back to numbers and then email you that EPICS is broken. Or they change a waveform-based plot to show time stamps on the X axis, again wondering what that's about.
Instead, we now have an XYPlot for waveforms, buffer size is automatically fetched from the waveform. And a separate strip chart widget for scalar over time, automatically using a time axis.

As a result, I've had much fewer "How do I use this?" questions because one widget performs one function, it doesn't drastically change functionality based on one property (number of states), and there are no properties that will break its functionality (time axis). Also makes it easier to handle questions, because the functionality and properties of each widget are fixed. No "well, it depends, ...". A new tool like the web runtime or whatever we'll have to do in 10 years will be easier to implement as well.

I hope this explains the specific design choices for the LED widgets and puts them into context.

As for the Symbol widget, it for now acts like the MultiState LED. There are N symbols, and the PV should be value with N states. There is no special case for boolean PVs, a symbol widget with 2 states doesn't change "symbol 0" into "Off Symbol" and "Symbol 1" into "On Symbol", nor does the 2-state scenario support checking a bit in a number.
If you want to change the symbol widget to use a structure for each state, fine, but try to minimize the impact.

 * Naturally increment the version number of the symbol widget
 * New code should read older version, i.e. still decode the list of symbol files
 * New code will naturally persist in the new format, so older runtimes won't be able to find any symbol files.
    If loaded into old runtime, the old code will read a symbol without any list of symbol files, so it shows `examples:/icons/default_symbol.png`

Screen Shot 2020-08-31 at 9 39 04 AM

So if you change the new version that way, there will be a loss of backwards compatibility, and it will be silent.
Going forward, a generic version warning would be useful, so next time around loading a newer version we get warnings in the log, but again that should not be an exception, it must continue to load and operate as much as possible.

You could have the new version still write the old `<symbols>` array in addition to the new structure, just to support loading the new format in older runtimes. You could actually write it with a comment
```
   <!-- Deprecated, left for compatibility with version 2.0.0 runtimes -->
   <symbols>
     ...
```

krisztianloki:
Actually what the Symbol and TextSymbol widgets lack compared to MultiStateLED is the ability to specify the numeric value for a symbol. That is why I would like to use a StructuredWidgetProperty. To keep backwards compatibility if the numeric values for the symbols is a zero-based sequence incrementing by one then the current on-disk format would be used.

PR #1509, opened 2020-08-31
Spinner widget: Support disable state, context menu and font property.

georgweiss:
Spinner widget now consistent with TextEntry: disabled when PV is off-line, JavaFX default edit context menu suppressed in favor of toolkit menu, and font property also supported.


georgweiss:
Need to add some changes to handle font property. Do not merge.

georgweiss:
Considered incrementing the widget version but decided not to as the font property is silently ignored when not supported. Is this the correct approach?

kasemir:
I would leave it. New tool loading old files will use the default font and look just like before. Old tool loading new files will ignore the font; might look a little different but still "work".
We just merged the update that would log a warning when old tools load new files. So going forward, one can increment the version number whenever one would like to get a warning.

PR #1512, opened 2020-09-02
Csstudio 1038 Set size of stand-alone OPI window

georgweiss:
When launched as stand-alone window the size is calculated to consider the width of the toolbar. I have adjusted the numbers to eliminate the need of scroll bars. Unfortunately the result is not the same on all platforms, see screen shots. Setting different numbers based on platform may give consistent result, but then again it could fail in future releases of the OS or JavaFX. For the sake of completeness, these are the numbers giving consistent results:
Windows 10: extra width: 18, extra height: 105
MacOS 10.15.5: extra width: 3, extra height: 96
CentOS 7.6/xfce4: extra width: 13, extra height: 105
To cover all bases I decided to of or the Windows numbers, which will result in some padding on Mac OS and Linux.
Screenshot 2020-08-31 at 09 34 18 (1)
Skärmklipp (1)
Screenshot 2020-08-06 at 16 50 03 (1)


kasemir:
You could check  `org.phoebus.ui.javafx.PlatformInfo.isWindows()` etc. to pick different values, but on Linux there is likely an additional difference for the various desktop environments (Qt, Gnome, ..), so this might be good enough.

PR #1516, opened 2020-09-03
Add default save path

CC-g2ca:
Small modification we made: add a default path property for saving. You can define a default saving directory. Not much but quite useful.

* Add property in ui properties
* Add condition in SaveAsDialog.java
* Add variable in SaveAsDialog.java
* Add variable in Preferences.java
* Add instanciation in Preferences.java constructor


kasemir:
I can see at least three use cases:

 * Always start in the same place as you now implemented, so end users who might be unfamiliar with the overall file system don't have to look for a place to save stuff.

 * Start in the location of the currently executed or edited display. You likely have a file tree for these, and starting in the location of the currently active display is likely a good starting point.

 * Start in the location of the last file loaded or saved.

I don't see how to magically figure out which one is best right now, and it can also change over the runtime of the tool.

PR #1524, opened 2020-09-09
ChannelTable: Show Active PVs only Checkbox

tanviash:
This PR adds an optional checkbox to ChannelTable UI to easily filter active PVs in channelfinder. The new checkbox can be turned on by adding a new setting "org.phoebus.channel.views.ui/show_active_cb=true" to settings.ini file.
image


shroffk:
I think this might be useful...
I was hoping that the preference controlled the visibility of the checkbox. If a site does not have the "active" property the the checkbox can be confusing. So the preference could be used to control the visibility of the checkbos itself to avoid it being shown in products where the associated property does not exist.

shroffk:
Oh wait...
I see you are adding and removing the checkbox from the grid using the preference.

This is exactly the feature my previous comment was requesting.

tanviash:
> Oh wait...
> I see you are adding and removing the checkbox from the grid using the preference.
> 
> This is exactly the feature my previous comment was requesting.

Correct, in case the needed setting is not provided in settings.ini, checkbox is not visible on ChannelTable UI.

shroffk (approved):

shroffk:
> Correct, in case the needed setting is not provided in settings.ini, checkbox is not visible on ChannelTable UI.

Is there a reason you used the adding and the removing of the control instead of changing the visibility of the widget... I would assume it might be easier.

tanviash:
> > Correct, in case the needed setting is not provided in settings.ini, checkbox is not visible on ChannelTable UI.
> 
> Is there a reason you used the adding and the removing of the control instead of changing the visibility of the widget... I would assume it might be easier.

Either way, it had the same effect on UI.

PR #1525, opened 2020-09-09
Display Builder: Adding Line Style to Shape Widgets

pavel-ch:
See #896 Thank you in advance for approval @kasemir


krisztianloki:
Not a big deal but `line_width` and `line_style` from PolylineWidget and `line_width` from PolygonWidget could be moved to PolyBaseWidget

pavel-ch:
Yes I considered that, but decided not to do it, just because these properties have nothing to do with being PolySomething, so PolyBase is by my understanding not a systematic place for that.

PR #1527, opened 2020-09-10
Save&Restore: Bunch of bug fixes

geniejhang:
@georgweiss  One main change is that I disabled modifying stored value in comparison view.
So, when you comparing two or more snapshots, base snapshot values can be changed, but compared snapshots' values cannot be changed.


georgweiss:
Thanks for the bug fixes!

Issue #1380, opened 2020-06-05
Release Guildeline and Procedure


shroffk:
Discussion:

It is great that over the last few months we have been able to have multiple releases of phoebus.
The release process has been quick and simple...this should help us push out changes and bug fixes quickly to our users and improve the general user experience.

With phoebus moving into production at multiple sites (SNS and ESS have had it in the control room for awhile now) I think this would be a good time to formalize and document the release procedure. The practices at each site and the good practices associated with putting out a release that would keep everyone happy.

Initial thoughts.
Release:

1. pre release:
 - [ ] Notification or PR (the release plugin can be configured to push to a branch which we can merge via a PR with approval... the NSLS2 product, run-scripts, and ansible based deployment process  needs to be [updated](https://github.com/shroffk/nsls2-phoebus/commit/547f88efc21c07e7ff1c6219370ac04016a9c252) with new releases, so such a notification would be helpful)

2. release process:
 - [ ] Releases should be tagged. (the release plugin does this already)
 - [ ] Source and Common product binaries hosted on the github project. (instead of working hard on automating this we could simply as a community upload win, mac, and linux platform tars)
 - [ ] Change log and Release Notes
 - [ ] prepare for next release, update version etc. (we will have to co-ordinate and find a solution on how the pom and build.xml are both updated)
 - [ ] tech-talk announcement?

ralphlange:
Is your argumentation only about the maintenance/bugfix digit? I would generally agree.
Or are you arguing against using numbered versions at all? I would strongly disagree.

ralphlange:
fwiw: ITER has strict requirements on traceability. That's why for "external" products a released version is preferred over a git hash - under the assumption that a released version has a release note, complete and matching documentation etc etc.
A date is not unique and does not provide an acceptable level of traceability.

shroffk:
I think we could add our use cases and requirements first. It might be better to sort out the details once they are all collected.

At NSLS2

I have 2 different products for the accl and beamline.
A ansible based deployment solution which build the latest phoebus product and the alarm services on the target machines.
Currently, we checkout the master and build that. In the accelerator control room there are talks about wanting a more rigidly controlled version/tag/release.

kasemir:
Hi Ralph, great Friday discussion, right?

I'm not specifically arguing for a certain numbering scheme, but trying to point to the issue:

[...]

PR #1529, opened 2020-09-15
#1380 a changelog for documenting important aspects of each release

shroffk:
While not perfect, the goal of this changelog is to start documenting changes with each release.

The changelog is open to modification by ALL developers to add notes about API changes, features, apps, bug fixes...developers should keep appending the "current development" section.

Just before/after a release we should create a new  "current development" segment.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,8 @@ Date:
 
 * Alarm Datasource
 * Simplification of the API's for the ContextMenuService, SelectionService, and AdapterService
+* File browser context menu items to create new display or data browser plot.
+* Array operations in formula functions.
 
 Release 4.6.3
 -------------
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,0 +1,36 @@
+Change Log
+==========
+
+A list of API changes, major features, and bug fixes included in each release.
+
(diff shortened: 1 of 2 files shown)


georgweiss:
@kasemir , @krisztianloki, @pavel-ch , @geniejhang , @tanviash 
Can you please consider if there is any change that you think should go into the change log?
@shroffk, @geniejhang , where are we on the Epics core update?

krisztianloki:
Maybe the behavior change of rules when a trigger PV is disconnected, something like:

Restore the property values of rules if a trigger PV is disconnected

georgweiss:
@krisztianloki , IMHO that might qualify as a "minor" fix.

krisztianloki:
These maybe:
The symbol widget now shows a semi-transparent rectangle (in front of the symbol) with the color "INVALID" when the PV is disconnected.
The LED widget shows the label when the PV is disconnected and both the "On" and "Off" labels are the same.

georgweiss:
@krisztianloki , please submit your items into the change log.

shroffk:
Sorry @krisztianloki beat you to it...

Just a note, 
For the time being I want to follow the system where additions to the changelog should be made by developers as they see fit without requiring approval :)...we keep the system informal and open until we find a reason to be more restrictive.

PR #1532, opened 2020-09-16
Csstudio 213 - Statistics tab in data browser

georgweiss:
As per user request, added statistics tab in data browser. Data is calculated (by Apache Commons math3) on user request only, i.e. no real time update.


kasemir:
There might be an error in the start/end handling of the time range.
In this screenshot, the red trace has a value range of about -2.9 to 4.8, yet min..max value are shown as ~0 to 2.9.
Sample count inside the plot is 3, plus the one "before" the start time valued ~4.8 that determines the initial value, yet the statistics suggest a sample count of 2.

Screen Shot 2020-09-17 at 4 43 04 PM

Implementation wise, it's good that pressing the Refresh button starts the computation on background threads. But those background threads then update SimpleStringProperty count, mean, .. etc., which are used in setCellValueFactory, i.e. the UI is bound to them and will update via listeners, directly in the background threads.
In many cases that "works", but isn't guaranteed. UI access should only be performed in the UI thread, so this could have side effects later on where the UI suddenly misbehaves.

georgweiss:
When trying to reproduce this I get the correct number of samples. You should see a sample count of 3, corresponding to the "flanks". Did you refresh the data after changing the time range?
The plot draws a line to the left using the last value outside the start of the time range to the first sample within the time range. So it looks as the first value is not considered, but it is not apparent that it is outside the time range and does not contribute. This is a bit deceptive when that value is greater or smaller than any of the other samples since it is not considered in determining the max or min. Question is if including that point is the correct thing to do.

I will update such that the observables are updated in the UI thread.

kasemir:
Yes, I did press 'Refresh' several times.

Your screenshot shows Sample Count 3 for both traces. While the Green trace clearly has 3 'steps', the red trace suggests only 2 samples.

As for the first sample, a plot for a start...end time range needs to consider the first sample at-or-before the start time, i.e. the sample with a value that's still valid at the start time, until we then receive the next sample within the start...end time range.
"Setpoint" type PVs that only change when the operator enters a new value can remain unchanged for months.
In the plot they show up as a flat line, the actual sample is way to the "left" of the plot's start time. Statistics should show Sample Count = 1 for that one sample.

`PlotProcessor` includes this example of locating the sample indices:
```
final PlotDataSearch<XTYPE> search = new PlotDataSearch<>();
...
// Find start..stop indices from ordered positions to match axis range.
// Consider first sample at-or-before start
start = search.findSampleLessOrEqual(data, position_range.getLow());
...
// Last sample is the one just inside end of range.
stop = search.findSampleLessOrEqual(data, position_range.getHigh());
```

georgweiss:
In my screen shot the red plot does in fact contain three samples, but two adjacent samples have equal value.
I would argue that samples outside the plotted time range should not be considered. Granted, it may look confusing to see a straight line and then not being able to compute statistics as there are no samples within the time range. This should also be a corner case and not very interesting in terms of statistics calculations.

Cannot really understand why your screen shot shows a sample count of 2. Are you able to reproduce?

kasemir:
> Are you able to reproduce?

Just look at `sim://sine` and `..ramp` for a while and then zoom into some region.
In this example, the green trace has a minimum based on the last sample around -3.5, but stats suggest min value -2.86.
Green trace also has at least 4 samples.

Screen Shot 2020-09-18 at 12 37 30 PM

I think this is because
a) You ignore the initial value
b) you perform your own search just based on the seconds,
```
start = 1000 * timeRelativeInterval.getAbsoluteStart().get().getEpochSecond();
```
instead of using the full `Instant`.
If you used the `PlotDataSearch<Instant>` as everywhere else it would honor the complete time stamp, plus have consistent handling of the initial sample.

Issue #1508, opened 2020-08-31
Ways to check the status of localizations


tmijail:
On PR #1495, @pavel-ch pointed out that silently falling back to the english localization if a message wasn't present on `messages_xx.properties` would make maintaining a localization a pain because there would be no easy way to check whether there were messages missing from `messages_xx.properties`. 
He would previously watch the log for `{CLASS} is missing value for '{MESSAGE}'`, which now only appears if `{MESSAGE}` is missing from both `messages.properties` and `messages_xx.properties`.

I found two ways to make this easy again.

### 1. Localization Tools Maven Plugin
We generate a [report ](https://web.archive.org/web/20090105015431/http://mojo.codehaus.org/l10n-maven-plugin/report-mojo.html) that looks something like this.  

![L10n Status report for phoebus](https://user-images.githubusercontent.com/6920322/91719630-10425780-eb6c-11ea-8014-ca86ae731508.png)
([live report from another project](https://maven.apache.org/plugins/maven-pmd-plugin/l10n-status.html))

We'd just need to add [this profile](https://github.com/tmijail/phoebus/blob/7a4ab7c8c87230e9fbf262f217298016956bf4da/pom.xml#L207) to the pom, and then it's as easy as running `mvn site -P l10n-status`. The report will be located at `phoebus/target/site/l10n-status.html`.

### 2. Checkstyle
We can use Checkstyle's [translation](https://checkstyle.sourceforge.io/config_misc.html#Translation) rule and one of the [available plugins](https://checkstyle.sourceforge.io/#Related_Tools) to integrate Checkstyle with the IDE (or just use the Checkstyle CLI).
I've tried the [IDEA plugin](https://github.com/jshiell/checkstyle-idea) and it's pretty nice. 

![Checkstyle IDEA plugin scan result](https://user-images.githubusercontent.com/6920322/91718200-50eca180-eb69-11ea-8630-b44093d6f786.png)

You can double-click on the warnings to take you to the correct file and then use IDEA's Resource Bundle view to quickly find the missing messages. 

![Resource Bundle view](https://user-images.githubusercontent.com/6920322/91718531-f0aa2f80-eb69-11ea-96c4-f9ea3f54c976.png)

To support checkstyle, we'd just need to add a [checkstyle.xml](https://github.com/tmijail/phoebus/blob/checkstyle/checkstyle.xml) file somewhere.

---

[...]

PR #1533, opened 2020-09-17
L10n status

tmijail:
Implements a report to check the status of localizations as discussed in #1508 .


tmijail:
(sorry about the delay)

kasemir:
Very good to have this documented, and thanks for making it an optional/profile option.

PR #1541, opened 2020-09-22
Add logbook client for PSI Elog

ffeldbauer:
I developed a rather simple LogClient for the [PSI Elog](https://elog.psi.ch/elog/) based on their [python library](https://github.com/paulscherrerinstitute/py_elog).

The API supports the Attributes Author, Subject, Type, and Category.
In Phoebus Type and Category are named Logbook and Tag.

As this was my first attempt of programming with Java/Maven I oriented myself a bit on the OlogClient

### DIFF
--- a/app/logbook/elog/.project
+++ b/app/logbook/elog/.project
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>app-logbook-elog</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+</projectDescription>
--- a/app/logbook/elog/build.xml
+++ b/app/logbook/elog/build.xml
@@ -1,0 +1,9 @@
+<project default="app-elog">
+  <import file="../../../dependencies/ant_settings.xml"/>
+
+  <target name="app-elog" depends="compile-app">
+    <jar destfile="${build}/elog-${version}.jar">
+      <fileset dir="${classes}"/>
+    </jar>
+  </target>
+</project>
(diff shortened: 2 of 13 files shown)


georgweiss:
Interesting. Trying to connect to our Elog server, but this fails with "Remote host terminated the handshake", despite valid SSL certificate. Any clues? Will unhashed password work, if not, how should it be hashed (sha256?)?

ffeldbauer:
Hey @georgweiss ,
the URL for the getTypes() is `<ELOG-URL>/?cmd=new`. The page should contain a block like this:

```html
</tr><tr><td nowrap class="attribname">Type<font color=red>*</font>:</td>
<td class="attribvalue">
<select name="Type" onChange="mod();">
<option value="">- please select -
<option value="Meetings">Meetings
<option value="Notes">Notes
</select>
<input type=submit name="extend_1" value="Add Type" onClick="return mark_submitted();">
</td>
```

The password in the properties need to be hashed (did not want to have a clear text password there). To get the correct hash, I'm using the python script:
```python
#!/usr/bin/python3
import getpass
from passlib.hash import sha256_crypt
upwd = getpass.getpass()
print( sha256_crypt.hash( upwd, salt='', rounds=5000 )[4:] );
```

ffeldbauer:
Just to make sure, it's not a problem of the Elog configuration. We use the following for defining the available attributes:
```
Attributes = Author, Type, Category, Subject
Options Type = Meetings, Notes
Options Category = Mechanics, Sensor Modules, DAQ, Sensors, Electronics, Hardware, Vacuum
Extendable Options = Category, Type
Required Attributes = Author, Type
```

ffeldbauer:
@georgweiss  ok....sorry, I tested the two methods ElogAPI#getTypes() and ElogAPI#getCategories() standalone with the `jshell`...
The HTML body from the getTypes request contains `form name=form1`. For all other HTML requests, this string should only occure, if the authentication failed, therefore ElogAPI#validateResponse() throws an execption here....

I will fix this bug!

georgweiss:
@ffeldbauer, what does <ELOG_URL> point to? Is that a page prompting the user to select a logbook? In our Elog deployment the selection of logbook looks differently, so the HTML parsing in getTypes() fails.

ffeldbauer:
<ELOG_URL> is the URL of a Logbook. E.g. `https://localhost:8080/demo` (The page, where you see a list of available logbooks would be `https://localhost:8080` in that case)

georgweiss:
Sure, but how is such a page supposed to look like for getTypes to return the wanted data? Can you share a screen shot?

ffeldbauer:
screen

ffeldbauer:
screen01

georgweiss:
Thanks @ffeldbauer . I'm afraid we've changed a bit too much in our setup for this to work.

ffeldbauer:
That's a pitty...I actually talked with Kunal and Kay while working on this, and it turned out even the unmodified Elog version has a lot of customization options which makes it hard to implement such a client.
So I have now implemented the client for a rather basic configuration as shown above.

ffeldbauer:
@georgweiss I modified the ElogClient. Now Types and Categories can be specified in the preference file. If they are not set there, the HTTP request is used.
Maybe this works for your version of the Elog.

georgweiss:
Thanks for the update @ffeldbauer.

shroffk:
@georgweiss can we merge this

ffeldbauer:
@shroffk from my side, there are no objections.
Not sure if @georgweiss tested the latest version and if it also works for him, but as I understood he operates a modified version of the Elog.

georgweiss:
I have no objections. We will not use this due to how Elog was modified at ESS. We will migrate to olog-es, I tested this to see if this would be a viable alternative.

PR #1542, opened 2020-09-22
Jython Interface to Save and Restore

domensoklic:
This pull request adds methods that can be called from Jython to create and restore a snapshot. This adds flexibility to user scripts.

Here is an example script for saving:
```python
from org.csstudio.display.builder.runtime.script import PVUtil, ScriptUtil
from org.phoebus.applications.saveandrestore.ui.snapshot import SnapshotController
from org.phoebus.applications.saveandrestore import AppConfig
 
logger = ScriptUtil.getLogger()
children = widget.getDisplayModel().runtimeChildren()
 
setname_widget = children.getChildByName("text_setname")
setcomment_widget = children.getChildByName("text_setcomment")
 
setname_pv = ScriptUtil.getPrimaryPV(setname_widget)
setname = PVUtil.getString(setname_pv)
setcomment_pv = ScriptUtil.getPrimaryPV(setcomment_widget)
setcomment = PVUtil.getString(setcomment_pv)
 
PVset_list = ['System:SubSystem02:GasCorrectionFactor_SP', 'System:SubSystem01:GasCorrectionFactor_SP']
 
logger.info('Saving setname ' + setname + ' with comment ' + setcomment)
 
snapcont = AppConfig.snapshotControllerInstance()
 
snapcont.generateSaveSet("Test", "Comment", PVset_list)
 
logger.info('Saving script finished')
```

Example for restoring:
```python
from org.csstudio.display.builder.runtime.script import PVUtil, ScriptUtil
from org.phoebus.applications.saveandrestore.ui.snapshot import SnapshotController
from org.phoebus.applications.saveandrestore import AppConfig
 
logger = ScriptUtil.getLogger()
children = widget.getDisplayModel().runtimeChildren()
 
setname_widget = children.getChildByName("text_setname")
setcomment_widget = children.getChildByName("text_setcomment")
 
setname_pv = ScriptUtil.getPrimaryPV(setname_widget)
setname = PVUtil.getString(setname_pv)
setcomment_pv = ScriptUtil.getPrimaryPV(setcomment_widget)
setcomment = PVUtil.getString(setcomment_pv)
 
logger.info('Restoring setname ' + setname)
 
snapcont = AppConfig.snapshotControllerInstance()
 
snapcont.restoreSnapshotAndValidate(setname)
 
logger.info('Restoring script finished')
```


app/save-and-restore/ui/pom.xml
```diff
@@ -120,6 +120,16 @@
 			<version>5.1.7.RELEASE</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
kasemir, line 123: See https://github.com/ControlSystemStudio/phoebus/blob/master/dependencies/README.md
In short, when you think you have to add another dependency, also add it to dependencies/phoebus-target/pom.xml

app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
```diff
@@ -533,6 +539,92 @@ public void loadSaveSet(Node node){
...
+                    readers.add(readerPv);
+                }
+
+                try {
+                    await().atMost(2, SECONDS).until(() -> readers.stream().allMatch(x -> x.isConnected()));
+                } catch (ConditionTimeoutException ex) {
+                    LOGGER.log(Level.WARNING, "Some PVs could not be saved because they are disconnected");
+                }
+
+                countdownLatch.await(1, SECONDS);
```
kasemir, line 616: Any fixed timeouts like atMost(2, SECONDS), 1 second, .. are likely wrong.
Also, CountDownLatch already offers `await(2, TimeUnit.SECOND)`. Do you really need to drag in another dependency to duplicate standard Java API?

georgweiss:
Seems like all save sets created through the Jython API will be placed in folder directly in the top-level root folder. With a large number of folders on the top level the tree will difficult to navigate.
I'd prefer if the client could also specify a different location in the tree for new save sets.

app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
domensoklic: Is there a better way to determine the timeout? It takes a while for the PVs to connect and for data to be received. A timeout is definitely needed, but to what value should it be set?

We did have an issue with the PVs not connecting, and the code quietly continuing with some PVs not being saved. I will check if increasing the timeout is enough and that this library can be removed.

app/save-and-restore/ui/pom.xml
domensoklic: Thank you for your feedback. I didn't know there are guidelines for dependencies.

domensoklic:
> 
> 
> Seems like all save sets created through the Jython API will be placed in folder directly in the top-level root folder. With a large number of folders on the top level the tree will difficult to navigate.
> I'd prefer if the client could also specify a different location in the tree for new save sets.

Would specifying a full path be a good solution since there can be multiple layers of folders, and if I remember correctly their names are not unique?

georgweiss:
Folder names must be unique within same parent folder. A full path is fine, though you need to query the back-end in order to find the unique id of the parent folder in which the save set is to be persisted. Currently there is no back-end API to do that, but you are free to add it, should be useful. Or let me know if you prefer me to do it.

app/save-and-restore/ui/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
kasemir: The point is exactly that there's not one good universal value for the timeout. It depends on your site. The EPICS `caget` command defaults to a 1 second timeout, but if you have a larger network with several gateways, a 10 second timeout might be more appropriate. A tool that connects to 10000 PVs might take 30 seconds until all PVs connect.

That's why `caget` can use timeouts that you pass on the command line. Likewise, your code could use a preference setting to set the timeout, or pass that in, as done with some of the Displays Builder's https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/PVUtil.java

domensoklic:
I'm not familiar with the internal workings of the backend, so your help would be much appreciated.

georgweiss:
@domensoklic, I can do that, but probably not within the next two weeks or so. Can your PR wait?

domensoklic:
@georgweiss, that is fine, thank you.

georgweiss:
Please check PR #1558, it adds utility end points that makes it easier to work with "file paths" in a client. Note: the new end points do not replace any existing, i.e. to create nodes one still needs to know the unique node id of the folder node in which to create the new object.

domensoklic:
The PR looks good. I can then retrieve the Unique ID from the `Node`.

georgweiss:
Yes, the unique id is a property of a `Node`, but the "file path" is not. The additions facilitate working with file paths, but client will still need to manage such file path strings. For instance, a to create a `Node` in the path /topLevelFolder/subfoder1/subfolder2 you will need to call the rest end point to find the unique id corresponding to subfolder2.

app/save-and-restore/ui/pom.xml
domensoklic: I'm not sure if I understand the instructions correctly. A dependency needs to be added to both `dependencies/phoebus-target/pom.xml` and a submodule `pom.xml`?

kasemir: Yes. To dependencies/phoebus-target/pom.xml so we have a list in one place of all external dependencies. The result is also a 'lib' folder that can be used to build with ant or Eclipse or ...  For maven, you also need to add it to the submodule pom.

georgweiss:
Need to repeat @kasemir's concern about adding a dependency (awaitility) duplicating standard Java APIs.

domensoklic:
> Need to repeat @kasemir's concern about adding a dependency (awaitility) duplicating standard Java APIs.

Yes, I wanted to discuss that. I have kept a call to awaitility, and of course, still depend on that library. I need to wait for all PVs to connect, and using awaitility is the most elegant solution. I don't know if there is a built in functionallity to wait for the PVs to connect. An alternative to awaitility would be to create a Thread and have it check the PVs. What are your thoughts on this?

georgweiss:
@domensoklic ,can you please describe your use case in more detail?

domensoklic:
@georgweiss, we would like to retrieve snapshot data from MASAR and write the data to the corresponding PVs. During testing we have found that we need to wait for the `PV` instances to connect to be able to write successfully. Since there doesn't seem to be a method that we can call to wait for this to happen, the most elegant way was to use the awaitility library. This library hides away the complexity of creating a thread, periodically checking a condition, blocking until the condition is fulfulled or a timeout occurs.

georgweiss:
Thanks for clarifying. But what is more specifically the use case, on a higher level? How does scripting enter into the picture and how is it invoked, how many PVs, is this calibration or fine tuning or something else?

kasemir:
The `PV` used by everything else in Phoebus is primarily meant for tools that connect to a PV and keep running, handling updates, disconnects, reconnects over time based on a 'reactive' behavior.
The idea of waiting for a PV to 'connect' and then read a value is intrinsically flawed because the PV might disconnect just before you try to read it. The only dependable way to get one or more values is to subscribe to updates and then handle what arrives. Again most components of phoebus subscribe "forever" or until they're closed, but it's still possible to use it for a fixed number of updates like this:

```
        final CountDownLatch done = new CountDownLatch(3);

        System.out.println("Awaiting " + done.getCount() + " updates...");
        final PV pv = PVPool.getPV("sim://sine");
        final Disposable flow = pv.onValueEvent().subscribe(value ->
        {
            System.out.println("Received update " + value);
            done.countDown();
        });
        done.await();
        flow.dispose();
        PVPool.releasePV(pv);
```

This example will still wait "forever", because it's very hard to predict how long it'll take until the first value arrives.
You can replace `await()` with the version that takes a timeout, but that timeout should then probably be configurable via a preference setting in the final application code.

Example to read just one value for several PVs:
```
        final PV pv1 = PVPool.getPV("loc://x(42)");
        final PV pv2 = PVPool.getPV("sim://sine");

        try
        {
            final CompletableFuture<VType> value1 = new CompletableFuture<>();
            final CompletableFuture<VType> value2 = new CompletableFuture<>();
    
            final Disposable sub1 = pv1.onValueEvent().subscribe(value -> value1.complete(value));
            final Disposable sub2 = pv2.onValueEvent().subscribe(value -> value2.complete(value));
    
            CompletableFuture.allOf(value1, value2).get(5, TimeUnit.SECONDS);
            sub2.dispose();
            sub1.dispose();
    
            System.out.println(pv1.getName() + " reads " + value1.get());
            System.out.println(pv2.getName() + " reads " + value2.get());
        }
        catch (Exception ex)
        {
            ex.printStackTrace();
        }

        PVPool.releasePV(pv2);
        PVPool.releasePV(pv1);
```

It uses a fixed timeout of 5 seconds, which again is just for the demo, in application code that likely needs to be configurable.

georgweiss:
I'd like to point out that save & restore was not designed to be a "backup tool" handling very large number of PVs in a save set. The use cases for which it is intended is procedures like calibration or fine tuning of a sub-system where the list of PVs is reasonably small to be viewed and potentially manipulated in the save & restore UI. 

If there is a use case to extend save & restore to handle a headless backup scenario that should be treated as a separate issue ticket where more detailed requirements can be discussed and agreed upon.

kasemir:
For what order of magnitude in PV counts is it designed?

georgweiss:
If I remember correctly: "typically less than 100".

kasemir:
SNS detector settings on a beamline tend to have around 4000 PVs.
The PV Table acts as a simple save/compare/restore tool.

kasemir:
https://github.com/ControlSystemStudio/phoebus/tree/master/app/pvtable, https://control-system-studio.readthedocs.io/en/latest/app/pvtable/doc/index.html

kasemir:
Compared to save/restore, it's just a file with PVs and values.  It only opens one file at a time and shows you the PV names, their current value, and the optional saved value. It's left to the end user to save a snapshot in a specific file name, there's no management of the files, no comparison between files.

georgweiss:
To be honest, I wasn't aware of PV Table and its features. Seems there is quite some overlap with save & restore. I think it would be valuable to consider a merge.

kasemir:
The PV Table was one of the earliest tools in CS-Studio. When MAZAR and the Eclipse save/restore UI started, the PV table was considered.
While both have a list of PVs, get their current value, and maybe a saved value for comparison, the big difference is that the PV table is just that, anybody can use it without any additional infrastructure. It can store snapshots in its own XML format or in a format compatible with the IOC 'autosave' files. You can name the files whatever and place them wherever.
MAZAR/save/restore on the other hand requires a database or similar infrastructure. You first need to define save sets etc. That certainly has a purpose, but requires some preparations.
Any 'merge' should still allow both a simple standalone use and the use as basically the 'save set display' within a larger framework. In principle, that should be possible by for example having a PV Table with both standalone "file" data source and a more elaborate "database" data source. But it would require some planning.

georgweiss:
I agree, a merge would certainly not be a walk in the park. But still worth considering to save users from confusion, and developers from code duplication.

domensoklic:
We would like to do linac reconfigurations for different particles and energies.
We have an industrial application, so it's important to us that it's automatic, fast and reliable. That is why we would like to run saving and restoring this via scripts.

At this point in time, we don't have all the details, but we know that we will have different configurations that we want to apply and be sure that they applied to all the PVs.
A technician will run the scripts manually. The number of PVs is up to 100.

georgweiss:
Thanks @domensoklic for the clarification. 
I need to give this some more thoughts before I respond.

georgweiss:
As for general comments:
I'd prefer that `SnapshotController` would disallow automatic creation of nodes in the root folder, i.e. when client specifies `null` for the parent's path of the new node.
Further, I'd also like the scripting support to be isolated to a dedicated class instead of putting it in the `SnapshotController`. You can still user Spring's injection mechanism to do the wiring of dependencies.

In regards to the issue on the awaitility dependency, I'd need to agree with @kasemir . However, the solution is maybe not to code around (e.g. introduce a CountDownLatch to handle only write connection), but rather to switch the PV library altogether. This has been discussed previously, but I have not had the possibility to do the switch. If you can wait for an update of the save & restore implementation I can try to prioritize such an effort, but I cannot commit to anything better than "within two months".

domensoklic:
Thank you for your comment, @georgweiss.

I agree with your opinion, but we will need an intermediate solution to cover our use cases. I think it is best if we use a fork of the Phoebus for now, and then switch to the original version when the new PV library is supported.

georgweiss:
OK.
Just et me add that the PV library I am referring to is not new, it's already there in the Phoebus code base. 
I'll let you know when I'm done with the switch, but - as hinted previously - this will take some time due to other priorities.

shroffk:
@domensoklic @georgweiss What is the status of this PR

It looks like there is a project idea in the comments too.
@georgweiss - FYI, typically some of the masar configuration files do go into the thousands of PV's

georgweiss:
@domensoklic, it's on my backlog, but unfortunately pushed down due to more urgent tasks. I expect and hope to be able to return to this during Q2.

domensoklic:
> @domensoklic, it's on my backlog, but unfortunately pushed down due to more urgent tasks. I expect and hope to be able to return to this during Q2.

Hi @georgweiss, were you able to get any work done on this PR? We're making plans for our use case, and it would help us a lot if you could have this implemented in the near future.

georgweiss:
@domensoklic , I regret to say that I do currently not have any possibility to handle this. Please note that some changes have been done in save & restore, e.g. Spring dependencies have been removed.

I need to reiterate: save & restore was not designed with headless usage (from script) in mind. Supporting that use case well implies a considerable refactoring effort, which I would like to do anyway for other reasons. But unfortunately I cannot do this in the short term perspective.

georgweiss:
@domensoklic, I've tried to recall what was originally discussed and proposed in your PR.
In my view using the SnapshotController to read PVs and send the result to the remote save & restore service is a bit problematic. The SnapshotController is designed to be a JavaFX UI controller, and using it in a context like a script would in my view be a bit odd. Those parts of the SnapshotController that deal with the PV layer should be refactored to the service layer in order to support save & restore from script. 

Again, I did plan to refactor portions of save & restore and the scripting use case would be considered. But with my current duties I will not be able to do this until late 2021, or beginning 2022.

domensoklic:
@georgweiss, I understand your thoughts that the SnapshotController was meant to be a UI controller and it is weird for it to also be handling S&R business logic.

You are more familiar with the architecture of the Phoebus than we are, and you wanted to refactor the code anyways, so I would follow your guidance here. I think it would be best if we stick to our fork for now, until you are able to refactor the code. We can then simply switch to the scripting interface that you will provide.

georgweiss:
@domensoklic , to clarify: I'd say most of the S&R business logic is in SaveAndRestoreService. The pieces that I'd like to move out of the SnapshotController is reading and writing from/to PVs when doing save and restore operations. However, the SnapshotController still needs to connect to PVs as the UI displays the values.

kasemir:
Save/restore has a GUI.  Calling it from a script within a display is somewhat of an intermediate idea; why can't you just use the GUI? There could, however, be a use case for a complete non-GUI approach, some command line tool for restoring settings.

There are similar examples of offering command line access to functionality that's already in CS-Studio:
The EDM converter can be invoked via `phoebus -main org.csstudio.display.converter.edm.Converter ...`, the archive data export can be invoked via `phoebus -main org.phoebus.archive.Export ...`.
You could similarly offer a way to call for example `phoebus org.phoebus.applications.saveandrestore.CmdLineTool -restore -saveset XXX ....`

georgweiss:
Thanks @kasemir, I will certainly consider your input when refactoring save&restore.

georgweiss:
@domensoklic, some updates in this matter:
I have started working on save&restore script support, but this work thread is not highest priority so I do not expect this to be completed for at least a month or so. In any case, any scripting support should in my view be quite limited in the functionality it provides. For instance, manipulating nodes (create/delete/copy/move folders or save sets) should in my view not be scripted. In fact, the use cases I think would be useful is to restore a snapshot, and maybe to also create new snapshots for a given save set.

Would it be OK to close this PR at this point?

shroffk:
@domensoklic can you review this PR (https://github.com/ControlSystemStudio/phoebus/pull/2122)

georgweiss:
Closing due to lack of input.

PR #1543, opened 2020-09-23
CSSTUDIO-1023 Added PVList column to show PV value

georgweiss:
As requested by our users.
Also tweaked the width of the columns in the table.
Screenshot 2020-09-23 at 15 47 50


kasemir:
For the value column, please check that it doesn't crash when trying to show a 1MSample array value.

georgweiss:
I'd say the same concern should apply to WidgetInfoDialog, right?

kasemir:
Exactly, that's why it includes code to only show the type, not the full value, see
https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java#L179

georgweiss:
To be frank I cannot see the difference in the implementations, save for PVList not calling VTypeUtil.

Issue #1544, opened 2020-09-24
Failed save operations not clearly communicated to user


georgweiss:
If a display file (and other resources?) cannot be saved because (for instance) a remote disk is unavailable, Phoebus will log the problem but leave the display editor unchanged, i.e. asterisk in tab title is still indicating dirty file. There is no alert to indicate the issue. 
In case user tries to terminate Phoebus with dirty display file, the subsequent prompt will give the user the opportunity to save. If at this point save fails, the editor will remove the asterisk giving the impression that the file has been saved successfully.

I would maybe challenge the need to to the save operations in the background, even if they do take time (and most of the times save operations should be quick). It should be easy to communicate failures, and handle the outcome in a synchronuous manner, e.g. when file save is successful, the application may be terminated. If a user wants to close the application, it will prompt before close if a file is dirty, but then not actually close once the file has been saved.

kasemir:
> I would maybe challenge the need to to the save operations in the background

Absolute necessity.
In the control room, operators run an instance of CS-Studio with many displays, ideally for months.
When they now perform file access, be it opening a new display or saving a *.plt file, that can should be quick but it's possible that for example NFS issues block it for minutes. This must not result in the complete UI being blocked.

georgweiss:
NFS issues or not, the problem remains: if file save fails user is left with dirty resource or - even worse, as described previously - left with the impression that file has been saved successfully.

shroffk:
At nsls2 we struggled with opi screens of an ill configured NFS, significant freezing... especially when we have screens with hundreds or embedded screens each with embedded screens.
These screens would have auto resizing...
Overall opening files sometimes tooks minutes and cs-studio blocking for all that time would be terrible.

So... loading files off the UI thread is **very** useful.

Now that I have given my strong opinion on this topic... @georgweiss the problem you present is a completely valid one.

> if file save fails user is left with dirty resource or - even worse, as described previously - left with the impression that file has been saved successfully.

I think we can have the file save operation expanded... so that if the user cannot save the file then we should pop up a dialog to help them save the file somewhere else where they have the appropriate authority.

kasemir:
Details of how to save are necessarily left to the application, i.e. the `DockItemWithInput` `save_handler`, because saving could mean writing a file, submitting a transaction to a database, invoking some web service etc.
That save handler is a `JobRunnable` which can report progress, support cancellation or throw an exception to indicate failure.

[...]

PR #1545, opened 2020-09-24
Display editor: Pass exception from "Save As" up to trigger error dlg

kasemir:
#1544

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/EditorGUI.java
@@ -528,8 +528,9 @@ public class EditorGUI
 
     /** Save model to file
      *  @param file File into which to save the model
+     *  @throws Exception on error
      */
-    public void saveModelAs(final File file)
+    public void saveModelAs(final File file) throws Exception
     {
         logger.log(Level.FINE, "Save as {0}", file);
         try
@@ -545,7 +546,7 @@ public class EditorGUI
         }
         catch (Exception ex)
         {
-            logger.log(Level.SEVERE, "Cannot save as " + file, ex);
+            throw new Exception("Cannot save as " + file, ex);
         }
     }
 
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/Messages.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/Messages.java
@@ -80,6 +80,7 @@ public class Messages
     public static String RuleCountFMT;
     public static String Run;
     public static String SaveDisplay;
+    public static String SaveDisplayErrorFMT;
     public static String SaveDisplay_TT;
     public static String ScriptCountFMT;
     public static String SearchTextField;
--- a/app/display/editor/src/main/resources/org/csstudio/display/builder/editor/messages.properties
+++ b/app/display/editor/src/main/resources/org/csstudio/display/builder/editor/messages.properties
@@ -63,6 +63,7 @@ ReplaceWith=Replace with
 RuleCountFMT={0} rules
 Run=Execute Display
 SaveDisplay=Save display
+SaveDisplayErrorFMT=Error saving display as {0}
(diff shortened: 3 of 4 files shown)


georgweiss:
Thanks, this fixes the issue.
However, in the case  when user closes the application and is prompted to save if a dirty file is open. Currently the applications does not close after the file has been saved, even though that was the user's intention. Should this behavior be maintained?

kasemir:
I'll look at that.

pavel-ch:
Yes I also vote for changing this. Now, if a changed file is somewhere in the middle of your open tabs, the first bunch of tabs closes, then Phoebus asks for saving and stops closing. After you close the Phoebus then, and start again, you get your workplace back with the half of originally open tabs. This is not a bug, but it is really annoying.

kasemir:
I'm working on it.
The technical problem is that we only had a `boolean` is-ok-to-close check.
For a 'clean' tab it can return 'true' and you close the tab.
For a 'dirty' tab it had to return 'false' right now to keep the tab open and then start the process of saving.
The result was that all the clean tabs close, the dirty ones stay open, and the app remains open after all is saved.

In the update, it uses a `Future<Boolean>` is-ok-to-close check. That way, "close" can start a background thread for checking all the tabs, await their answer (which could be 'ok' after some delay for saving the file, or 'no' after some delay if saving failed), and then process to close once all are 'ok'.

Overall, this looks much better: "Close" will first prompt "Do you want to save?" etc., and then close everything down.
But it will take another day to test all the scenarios.

pavel-ch:
In practice, this change is working quite fine, but:
In DisplayRuntime, looks like with this change, I am getting exceptions when closing display via script ScriptUtil.closeDisplay(widget):
```
2020-10-09 12:18:14 SEVERE [org.phoebus.ui.docking] Failed to call prepareToClose
java.lang.Exception: Stack Trace
	at org.phoebus.ui.docking.DockItem.close(DockItem.java:651)
	at org.csstudio.display.builder.runtime.app.DisplayRuntimeInstance.close(DisplayRuntimeInstance.java:258)
	at org.csstudio.display.builder.runtime.app.DockItemRepresentation.lambda$0(DockItemRepresentation.java:143)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
	at com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:174)
	at java.base/java.lang.Thread.run(Thread.java:834)
```
because DockItem prepareToClose() has not been called before. In the runtime, we usually do not need a dirty state check, so the question is whether there should be an option to close intentionally without the "prepareToClose()" call. This is done anyway now, but with the side effect of the above Exception logged.

kasemir:
While the message looks SEVERE, the closing does continue just fine. End users won't see it.
Since the script is calling `ScriptUtil.closeDisplay`, it should be easy enough to update the `ScriptUtil` to perform the complete `prepareToClose`-then-`close` cycle, which will silence that message.

pavel-ch:
That is not that easy due to threading issues, both`ScriptUtil.closeDisplay` which is calling `DockItemRepresentation.closeWindow(DisplayModel)`do not know DockItem. Then DisplayRuntimeInstance.close() gets called in UI thread, but prepareToClose cannot be used there being on UI thread. Any idea?

kasemir:
I don't see a fundamental issue. Yes, ScriptUtil, i.e. the runtime, doesn't know the dock item, but the toolkit which is asked to actually perform the 'close' does. I'll work on it.

pavel-ch:
Yes principally that must be trivial, I only hesitated where to place the `prepareToClose` call. Thanks.

PR #1552, opened 2020-09-29
WidgetTree: optimized moving widgets

krisztianloki:
Instead of removing and re-adding the widgets simply move them in the list; removing and re-adding is expensive because all the listeners have to be removed and then re-added, the representation has to be discarded and re-created.

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/tree/WidgetTree.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/tree/WidgetTree.java
@@ -10,6 +10,7 @@ package org.csstudio.display.builder.editor.tree;
 import static org.csstudio.display.builder.editor.Plugin.logger;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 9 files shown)


kasemir:
Have you observed noticeable delays with the original remove/add approach?

pavel-ch:
In this context, would be good (nice-to-have) a drag and drop in widget tree for changing widget order...

krisztianloki:
> Have you observed noticeable delays with the original remove/add approach?

Yes, especially when using pictures, and when I pressed Alt+B/Alt+F several times and quickly

kasemir:
OK. Since a clean remove & add with associated update of the representation is needed anyway, simply using that was an easy initial choice. I agree that a dedicated 'move' is better, and if it makes a noticeable difference, go ahead.
Still, pressing Alt+B/Alt+F several times is likely a result of not being able to simply drag/drop to the desired location right away, so that would indeed be useful. Just need to make sure that you drag something that's specific to the tree.
A drag in the PV tree, then accidentally dropping it outside shouldn't result in widget removal. Should also probably not accept a widget dragged from the palette, so again best drag some object that's only recognized by the tree view.

pavel-ch:
@krisztianloki could you continue some time in the future with D&D or should I look at that?
There are some other pitfalls besides that described by @kasemir, like D&D from-to a group or Tab etc. Perhaps the D&D can be limited to the actual Parent. Also, a preview of the change in editor pane during drag phase would be nice (but maybe implementation difficult, no clue atm).

krisztianloki:
@pavel-ch if you have time to do it then please take a look at it. I have a bunch of things on my TODO list so I cannot promise anything...

pavel-ch:
Ok...

pavel-ch:
I have implemented DnD for WidgetTree (limited to single selection), see pavel-ch/phoebus widgettree_dnd branch). I do not want to PR before this #1552 is merged, why it is waiting?

kasemir:
No reason, let's merge

PR #1555, opened 2020-10-02
Display Buider Editor: Drag and Drop support in Widget Tree

pavel-ch:
See #1552 discussion. Initial implementation.
Single selection and drop to the same level (root/Group/Tab) only supported atm.


kasemir:
Looks good to me.
As for 'same level (root/Group/Tab) only supported' I'd say that's a feature.
When dragging widgets in the editor in or out of groups/tabs, it's quite obvious how you're changing both the location on the screen and the group/tab membership.
If the widget tree allowed moving widgets out of tabs/groups, without changing their location on the screen, they would easily end up behind a parent, i.e. hidden, or in front of, which for the most part looks like they're inside the parent, but actually aren't. Overall bound to cause more trouble than good, so specifically not allowing that via DnD in the widget tree is actually helpful.

PR #1563, opened 2020-10-09
ScriptUtil.close: 'prepareToClose' before closing

kasemir:
The proper cleanup that allows all panes to prompt "Do you want to save" etc. requires calling `prepareToClose` before `close`.
This makes that happen for `bothScriptUtil.closeDisplay`

#1545

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DockItemRepresentation.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DockItemRepresentation.java
@@ -1,3 +1,10 @@
+/*******************************************************************************
+ * Copyright (c) 2016-2020 Oak Ridge National Laboratory.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


pavel-ch:
After fetching this commit (and some before, do not know exactly how many) my Eclipse shows errors in some projects. I had to modify some .classpath files. I do not know whether that is my local issue or a common one. Is it something known? Changes are visible in my fork pavel-ch/phoebus master branch (compare to this repo).

kasemir:
It's quite possible that the .classpath files need updates to add the new 'core-vtype'.
That of course only affects those who use Eclipse to build, so such changes tend to follow a little later.
It's curious that I don't see any errors here, only need to edit dependencies/phoebus-target/.classpath to un-comment the javafx libs, but your changes actually make sense, so feel free to create a PR. There are a few different ways to do it, for example if the new 'core-type' included vtype-1.0.3.jar with `exported="true"` in its .classpath, that might reduce the changed line count a little, but not sure it's worth it.

pavel-ch:
Ok thank you I will PR as-it-is for now

Issue #1564, opened 2020-10-09
Conversion of shapes and their borders


shroffk:
BOY allowed for shaped to define borders. 
![image](https://user-images.githubusercontent.com/2111304/95625648-cf7d0e80-0a46-11eb-9d4b-ff2540c0959f.png)

While all the 16 border options are excessive and it does not make sense to reproduce them in Display Builder, supporting a few borders which map directly to the supported `line_styles` could be very useful.

shroffk:
While exploring this issue I think I appreciate how DisplayBuilder does not have overlapping and confusing properties.
In BOY the shapes had "borders" AND "outlines"... with borders being potentially alarm sensitive.

With the conversion, I hope to stick to the property lists already established in display builder shapes. i.e. outlines.
I think it is best to preserve the broadly passive nature of the shape widgets.

kasemir:
I appreciate that you appreciate it ;-)
The borders in BOY are partially a side effect of what SWT Draw2D figures happened to offer. A 'Group' or 'Title Bar' type of border on a circle doesn't really make any sense.
Importing a 'border' of solid, dash, dash-dotted style into the line style the new rectangle makes perfect sense.

pavel-ch:
Did Boy allow setting some parameters for dash or dash-dot segment lengths? In DB, it is a fixed ratio to line (or border)  width and thus sometimes unusable (try dash-dot with line width 1):
![image](https://user-images.githubusercontent.com/771816/95630557-59969880-0a82-11eb-8794-1c63b4cc0d71.png)

shroffk:
@pavel-ch as far as I can see there was no such parameters in BOY.

First iteration,
![image](https://user-images.githubusercontent.com/2111304/95632477-b9297f80-0a53-11eb-8613-20b8b840d386.png)

TODO: 
code cleanup, move handling to support class
add same functionality to arc, ellipse, and polygon

kasemir:
I'd like to avoid too many widget-level parameters, i.e. no per-widget "Dash Length", "Dot Length", "Gap Length" parameters.
But I guess your point is not to allow arbitrary patterns or detailed customization of the dash length anyway. You're concern is simply that the dashes become invisible for certain line width.
Maybe you can find some way to scale based on the line width? Like `gap = line_width`, `dash = 3*line_width`?

pavel-ch:
[...]

PR #1591, opened 2020-10-13
Display Builder shape widgets representation - scaling dash segment l…

pavel-ch:
…engths for line styles. Centralized in JFXUtil and limited for line widths under 4.
See #1564


kasemir:
One of the travis builds ends in "cancelled". Restarted, same thing. Ignoring it.

PR #1593, opened 2020-10-14
Update to Open JavaFX 15

kasemir:
Updates to the current version of Open JavaFX.

Release notes, https://github.com/openjdk/jfx/blob/jfx15/doc-files/release-notes-15.md, don't seem to include anything desperately needed for phoebus or to address known problems, but probably best to stay up-to-date.

Tried on Mac with Open JDK 15.
On Linux with Open JDK 14, and both `-Djdk.gtk.version=2` as well as `3`.


kasemir:
Are there concerns? Otherwise suggest we update to the current JavaFX release.

PR #1595, opened 2020-10-16
CSSTUDIO-1351 Make undo stack size configurable

georgweiss:
Users request ability to increase undo stack size. Still defaults to 50, but can be configured:
`org.csstudio.display.builder.editor/undo_stack.size=<new size>`

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/Preferences.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/Preferences.java
@@ -9,6 +9,7 @@ package org.csstudio.display.builder.editor;
 
 import java.util.HashSet;
 import java.util.Set;
+import java.util.logging.Logger;
 
 import org.phoebus.framework.preferences.PreferencesReader;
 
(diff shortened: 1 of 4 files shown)


app/display/editor/src/main/java/org/csstudio/display/builder/editor/Preferences.java
```diff
@@ -20,9 +21,11 @@
 {
     public static final String HIDDEN_WIDGETS = "hidden_widget_types";
     public static final String NEW_DISPLAY_TEMPLATE = "new_display_template";
+    public static final String UNDO_STACK_SIZE = "undo_stack.size";
```
kasemir, line 24: So far all the preference settings follow the pattern `some.package.name/lower_case` with dots in the package name and underscores in the variable name. 

https://control-system-studio.readthedocs.io/en/latest/preferences.html
https://control-system-studio.readthedocs.io/en/latest/preference_properties.html.

app/display/editor/src/main/resources/display_editor_preferences.properties
```diff
@@ -16,3 +16,6 @@ hidden_widget_types=linear-meter,knob,gauge,clock,digital_clock
 #
 # GUI Menu action Applications / Display / New Display opens the following template
 new_display_template=examples:/initial.bob
+
+# Size of undo stack. Defaults to 50 if not set.
+# undo_stack.size=
```
kasemir, line 21: All the other preferences use this pattern to describe the value and define it with its default value:
```
# Description
the_variable=default_value
```

So I'd make this
```
# Size of undo stack
undo_stack_size=50
```

https://control-system-studio.readthedocs.io/en/latest/preferences.html
https://control-system-studio.readthedocs.io/en/latest/preference_properties.html

kasemir (approved):
Looks good to me

PR #1604, opened 2020-10-22
CSSTUDIO-1350 File not dirty when undo stack is empty

georgweiss:
This is a fix for a corner case: when user has used up all undo operations (defined by the undo_stack_size limit) the file is not treated as dirty, even if there are additional changes. User may close file without being prompted to save -> changes lost.
The fix affects display editor as well as scan editor.


kasemir (approved):

georgweiss:
Thanks

PR #1607, opened 2020-10-25
Symbolrepresentation fixes

krisztianloki:
These changes fix some flickering issues with the symbol widget and forcibly shows the index value if the symbol cannot be found.


kasemir:
What caused flickering?

krisztianloki:
When all the images were loaded the correct image was displayed right away with `toolkit.execute()` and an update was then scheduled to show it again and set the width and height of the imageview. Sometimes it happened that the image was shown in its original size for a short amount of time.
Another flickering is with the X; the X was displayed immediately. Now it is only displayed if the correct image is not available.

kasemir:
Thanks for the explanation. I was wondering if the flickering was based on our code changing the picture, of if we ran into some JavaFX issue. Looks like it was simply code changing the picture, briefly, the wrong way, and it's been fixed.

PR #1611, opened 2020-10-28
Replace Jackson to Gson for de/serializing Json

geniejhang:
- For save&restore client
- Update epicsCoreJava package version accordingly

### DIFF
--- a/app/save-and-restore/model/pom.xml
+++ b/app/save-and-restore/model/pom.xml
@@ -36,6 +36,12 @@
 			<version>${vtype.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.epics</groupId>
+			<artifactId>vtype-gson</artifactId>
+			<version>${vtype.version}</version>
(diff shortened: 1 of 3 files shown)


geniejhang:
@georgweiss Please review the change.

georgweiss:
It looks as this replaces Jackson only on the client side. Save & restore service will still use Jackson do deserialize. If there is a performance gain in switching to Gson, I'd prefer if we could make that switch also in the service.

shroffk:
@georgweiss @geniejhang 
Should we merge this PR and open another one for the service

geniejhang:
@georgweiss @shroffk 
The main focus of the change was to enhance the performance for saving information to DB.
bottle neck was the client part and service performance is always good no matter what number of PVs we try to store.
Amount of work needed to update the service to Gson support exceeds the gain we get.
I'll do it in the end, but the priority is quite low.

I suggest merge this one first which is significant amount of improvement.

PR #1614, opened 2020-10-30
Actionbutton fixes

krisztianloki:
Changes:
- Do not actually disable the widget just apply the disabled style and ignore clicks
- Separate enabled and writable state of ActionButton; opening a display or running a script should be possible even if the ActionButton has a PV and that PV happens to be read-only or disconnected
- Show 'n actions' as label if there are more than one actions executed as one

Questions:
- Right now it is possible to circumvent confirmation by using the context menu and selecting the action there. Is it considered a bug?
- Is the 'Execute as one' feature a convenience or a constraint? Because right now it is possible to select any action using the context menu so it defeats the purpose of 'Execute as one' as a constraint.
- (Not specifically related to ActionButton) Shouldn't disabled widgets have their actions in the context menu also disabled?

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/Messages.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/Messages.java
@@ -15,6 +15,7 @@ import org.phoebus.framework.nls.NLS;
 public class Messages
 {
     // Keep in alphabetical order, synchronized with messages.properties
+    public static String ActionButton_N_ActionsAsOneFmt;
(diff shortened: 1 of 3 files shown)


kasemir:
Answers:

 * Yes, I think that's an oversight
 * Convenience. Simple "click" runs all actions instead of opening drop-down for selecting just one. OK for context menu to still offer the individual actions to invoke just one of them at a time.
 * Yes, just like 'confirmation' the 'disable' should also apply to the context menu.

Do you want to fix that within this branch, or merge what you have and then attack the rest later?

krisztianloki:
Thanks!
Please merge this one and I will create a new branch and PR for the rest.

PR #1616, opened 2020-11-04
Fontpopover fixes

krisztianloki:
- Select the matching named font when selecting family, style, or size. If there is no match, clear the named font selection
- Always reinitialize font popover so that `setInitialConditions()` is called every time
- Do not call `setFont()` from listeners if those listeners were called because of `setFont()`

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/WidgetFontPropertyBinding.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/WidgetFontPropertyBinding.java
@@ -39,25 +39,31 @@ public class WidgetFontPropertyBinding
     /** Update model from user input */
     private final EventHandler<ActionEvent> action_handler = event ->
     {
-        if (popover == null)
-            popover = new WidgetFontPopOver(widget_property, font ->
+        final WidgetFontPopOver previous = popover;
+        popover = null;
+        if (previous != null)
+        {
+            if (previous.isShowing())
             {
-                undo.execute(new SetWidgetPropertyAction<>(widget_property, font));
-                if (! other.isEmpty())
+                previous.hide();
+                return;
+            }
+        }
+        popover = new WidgetFontPopOver(widget_property, font ->
+        {
+            undo.execute(new SetWidgetPropertyAction<>(widget_property, font));
+            if (! other.isEmpty())
+            {
+                final String path = widget_property.getPath();
+                for (final Widget w : other)
                 {
-                    final String path = widget_property.getPath();
-                    for (final Widget w : other)
-                    {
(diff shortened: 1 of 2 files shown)


kasemir (approved):
I agree that updating some selected font/size/style into a matching 'named' font is good.

Now in principle it's possible that somebody really wants to use italics, a little smaller. Might happen to match "Comment" right now, but users really wants italics and a little smaller, without changing that if the definition of "Comment" is later changed.

Also possible that you would have defined fonts "Warning" and "Standout" which right now are defined the same way, so this code would just pick the first one that matches.

But I think those are edge cases, in 99% what you did here is helpful.

kasemir:
@krisztianloki , since you last worked on the font dialog, can you look at this reported by Mathias @ FRIB:

The "OK" button is only enabled when you actually change the settings.
Open font dialog, assume the "Size" is 20. Change to 21 and "OK" becomes enabled. Change back to 20 and "OK" gets disabled. Makes sense, because there's no change.

However, property changes can apply to more than one widget.
Select one widget with font size 20, then 3 other widgets with various font sizes.
You want all to use the same font size, the one with size 20.
Font dialog appears with the settings of the widget selected first, i.e. 20.
Cannot push "OK", it's disabled.
Change size to anything, "OK" is enabled.
Change size back to 20, since that's what I want, but "OK" is again disabled.

Works OK if user selects some other widget with font size other than 20 _first_, then changes size in dialog to 20, "OK" appears, push it, and all widgets now have font size 20. But when "rubberbanding" widgets, it's unclear what's the "first" selected widget, plus this requires planning ahead.

Should the font dialog just enable "OK" as soon as something is changed, and stay with "OK" enabled even when changing values back to what they were when the dialog got opened?

MathiasSteiner:
Kay is right...

>> Works OK if user selects some other widget with font size other than 20 first, then changes size in dialog to 20, "OK" appears, push it, and all widgets now have font size 20. 

..but keep in mind that now all widgets have size 20 **and all the other font properties** of the first selected widget. 

The whole point was selecting A, then any number of B,C,D,... and asking them to become "just like A".

kasemir:
For your use case, simply always having "OK" enabled, even when nothing was changed, would be most convenient.
But it might generally be more useful to enable on the first change, because I think editing props for just one widget is a very common case.
Or enable "OK" if either there's a change, or more than one widget was selected.
Or find a summer student who can turn this into a capstone project.

kasemir:
Implemented it such that we keep the current behavior when only one widget is selected, i.e. 'OK' enables only when he font differs from the original value.

When multiple widgets are selected, however, 'OK' enables and then stays enabled on any change, even when going back to the original value, since we may assume the intention to apply what's no change to the original value to all widgets.

PR #1617, opened 2020-11-04
Contextaction fixes

krisztianloki:
- Disable widget actions in context menu if 'enabled' property of widget is `false`
- Honor confirmation request of ActionButton when using context menu actions
- Honor the target of Open Display action when using context menu


kasemir (approved):
Was worried that this would require more changes.
Great if this does it.

PR #1621, opened 2020-11-06
Ant: Support building app-channel-*

kasemir:
Adds the channel finder related modules to the ant build.
No change to maven build.


shroffk:
thanks

PR #1622, opened 2020-11-06
Embeddeddisplayrepresentation fixes

### LINKED ISSUE #1590: [Regression] Embedded display scrollbars
Commit 44fc4df48a0a147e38b4f1fe5acc4739951aa5b5 introduced a regression:
an embedded display widget that is set to the exact same size as the display it embeds and has "No resize" as resize behavior will have unnecessary scrollbars. If set to "Size widget to match content" then it will draw outside of the widget.
The old behavior: 
embedded_good
The no-resize behavior:
embedded_bad_no_resize
The size-widget-to-match-content behavior:
embedded_bad_size_widget_to_content
My test case: 
[embedded-regression.zip](https://github.com/ControlSystemStudio/phoebus/files/5372933/embedded-regression.zip)
The main display is main-gas.bob

krisztianloki:
This fixes #1590.
Another test case with all the possible resize values: 
[embedded_display_test.zip](https://github.com/ControlSystemStudio/phoebus/files/5503102/embedded_display_test.zip)

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
@@ -165,6 +165,7 @@ public class EmbeddedDisplayRepresentation extends RegionBaseRepresentation<Pane
         model_widget.propMacros().addUntypedPropertyListener(fileChangedListener);
 
         model_widget.propTransparent().addUntypedPropertyListener(backgroundChangedListener);
+
         fileChanged(null, null, null);
     }
 
@@ -188,8 +189,6 @@ public class EmbeddedDisplayRepresentation extends RegionBaseRepresentation<Pane
 
         final int widget_width = model_widget.propWidth().getValue();
         final int widget_height = model_widget.propHeight().getValue();
-        inner.setMinWidth(widget_width);
-        inner.setMinHeight(widget_height);
 
         final Resize resize = model_widget.propResize().getValue();
         final DisplayModel content_model = active_content_model.get();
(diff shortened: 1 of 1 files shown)


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
```diff
@@ -376,40 +380,42 @@ public void updateChanges()
...
-                    zoom.setX(zoom_factor_x);
-                    zoom.setY(zoom_factor_y);
-                }
-                else // Resize.Crop, SizeToContent
-                {
-                    zoom.setX(1.0);
-                    zoom.setY(1.0);
-                }
+                // During runtime or if the resize property is set to Crop we clip inner
+                // but allow 'overdrawing' in edit mode so the out-of-region widgets are visible to the user
```
kasemir, line 415: Allow 'overdrawing' in edit mode:
On one hand, that sounds useful.
On the other, it means the display looks different in edit mode, showing content outside of the bounds which will then vanish at runtime. Could add another 'outline' border in edit mode to indicate the container size, making it very obvious what's outside and about to be clipped at runtime. But then adding such a border means yet another difference between edit and runtime ..
Let's go with what you have for now; we can tweak the 'overdrawing', additional outline etc. later if it really turns into an issue.

kasemir (approved):

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
krisztianloki: You are right; it is not "nice" to render the content differently during runtime but this way it will be obvious what is causing it when someone opens an issue. Plus it is also a kind of reminder that this needs to be addressed eventually.
The additional border is a good idea but I don't know how to get what is outside `inner`; what is "outside" of the embedded display in the parent display is also "outside" in the child. Maybe we should render a semi transparent rectangle filled with the color "INVALID" over everything that is outside the limits of its display?

Issue #1561, opened 2020-10-08
Alarm datasource expansion


shroffk:
The current alarm channels are primarily suited for textupdate and textinput widgets.
We need to add more subfields to the alarm channels so that they can be more effectively used with other widgets, particularly buttons and leds.

alarm://my/alarm/node.active
Read a VBoolean
Value: true/false
Alarm: Undefined/Major/Minor/OK

Write a VBoolean
ack/unack

alarm://my/alarm/node.enable

Read a VBoolean
Value: true/false
Alarm: Undefined/Major/Minor/OK

Write a VBoolean
enable/disable

alarm://my/alarm/node.state
Return an VEnum

Read only?

~~alarm://my/alarm/node.desc~~
~~Return a VString~~
~~Desc and messages...if it has been defined.~~

shroffk:
@georgweiss @tanviash

georgweiss:
Should alarm://my/alarm/node.desc be VStringArray?

shroffk:
Processing a string array adds complication. 
I am not opposed to it but do you think it is ok if we starte with string and then based on some user feedback update it to VStringArray once we have a few use cases

georgweiss:
But if a string contains multiple parts it would in practice be processed anyway to extract the relevant portions.

kasemir:
What is the description? The alarm guidance?

shroffk:
Yes, the desc was mapping to guidance. 

@georgweiss, I was talking about how the user would have to process the array, I think for the time being I am going to skip the .desc . the other fields are far more clear and important.

kasemir:
Unclear how the dynamic behavior of alarms coming and going, and the potentially large amount of guidance maps to a display where a once-created arrangement of widgets shows alarm PVs.

An alarm config node can have one or more guidance entries. They each have a title and content. In the alarm table, you see nothing when there are no alarms, then those entries which are in alarm. Opening the context menu, you might find guidance with titles like "Overview", "Overpressure Procedure", etc.
Users are presented only with what's in alarm, and for those they can pick the guidance they need to see based on the title.
Putting all guidance titles and contents into a alarm:/path/to/node.description string array, what would you do with that?
Create a display that shows all the guidance for all the alarm nodes, and the user then reads the whole thing? Add rules to the display to show/hide guidance?

shroffk:
> Putting all guidance titles and contents into a alarm:/path/to/node.description string array, what would you do with that?

[...]

PR #1625, opened 2020-11-09
#1561

shroffk:
Alarm datasource with special fields

### DIFF
--- a/app/alarm/datasource/doc/index.rst
+++ b/app/alarm/datasource/doc/index.rst
@@ -68,13 +68,42 @@ matches the alarm state.
    :width: 80%
    :align: center
 
+--------------
+Special Fields
+--------------
+
+The alarm pvs have a few additional fields which can be used to access specific attributes of the alarm node/leaf
+
+These fields can be used by appending the following field tag at the end the alarm pv.
+
+e.g.
+
+| ``alarm://NSLS2_OPR/Linac/Controls.state``
+
+
+| **.state**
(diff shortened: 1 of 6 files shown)


georgweiss:
Took this for a spin and spotted some issues.
1) Use the various fields (state, active, enabled) works if used only once, i.e. an OPI containing widgets for all of them will update only one widget. Seems ordering is a factor: only the widget connecting last will update. Can be verified using separate OPIs for each of the fields.
2) The duration field - when bound to a text update - looks disconnected and always shows NaN.
3) Writing enable/disable seems to work and is reflected in the alarm tree. However, the issue described in 1) will potentially result in a LED showing the enabled field to not update.

shroffk:
Looking into point 1. first

shroffk:
Ok, There is a bug in the pv registration when there is one of these special fields used. It does affect updates to other pvs from the datasources too. I will fix this before we merge this PR

shroffk:
@georgweiss thanks for testing...
Since both `alarm://root/path/node` and `alarm://root/path/node.state` both register to the same node `/root/path/node`.
There was conflict where the new connection was replace the old.

The new implementation maintains a list for both the above cases

georgweiss:
Works fine now. My test of duration was wrong, so what I can see now is the time stamp of the last alarm state transition.
Also tested ack/unack, and it works fine.

georgweiss (approved):
Works as expected.

PR #1629, opened 2020-11-10
Save&Restore: Large PV data storage

geniejhang:
- For large PV data (>64KB, e.g. image), a table for those data is prepared.
- Potential issue fix: When set MariaDB, MariaDB JDBC driver is used instead of mySQL JDBC driver.

@georgweiss This may not be the intended use case of Save&Restore application.
You can reject it if it doesn't fit. I can keep this as FRIB-specific version.

We have about 0.4% of PV data having larger than 64KB. 
Instead of increasing current column format from TEXT to LONGTEXT, which has 3bytes more overhead,
I added another table storing LONGTEXT data only when some PV data are larger than 65KB.

### DIFF
--- a/services/save-and-restore/pom.xml
+++ b/services/save-and-restore/pom.xml
@@ -85,6 +85,11 @@
 			<artifactId>mysql-connector-java</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.mariadb.jdbc</groupId>
+			<artifactId>mariadb-java-client</artifactId>
+		</dependency>
+
 		<!-- https://mvnrepository.com/artifact/com.zaxxer/HikariCP -->
 		<dependency>
 			<groupId>com.zaxxer</groupId>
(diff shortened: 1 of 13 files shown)


georgweiss:
@geniejhang, it makes sense to support PV values > 64kB, but would it not be simpler to change the column type in the snapshot_pv table?

geniejhang:
@georgweiss Right. That is the simplest solution and I also tested that case.
After I read your comment, I recalculated my disk usage estimate.
My calculation was wrong. I'll revert the code as you said.
I already setup for postgres and mysql testing environment. Thanks.

georgweiss:
@geniejhang , for the sake of clarity: has this been verified on all DB engines (MariaDB, Mysql, Postgresql)?

geniejhang:
@georgweiss Yes! All went green. Forgot to mention.

PR #1633, opened 2020-11-12
Preferences: Support annotations

kasemir:
Almost every core and app module uses preferences, often based on this pattern:
```
public class MyPreferences
{
    // Constant for the 'tag' name used in preferences
    public static String SOME_SETTING = "some_setting";

    // Variable that'll receive the current value of the preference
    public static int some_setting;		

    static
    {
        // Parse property file with default settings, apply Java preferences that might override
        final PreferencesReader prefs = new PreferencesReader(MyPreferences.class, "/my_app_preferences.properties");		    	
   
        // .. and then fetch every value
        some_setting = prefs.getInt(SOME_SETTING);
    }
```

That's a lot of boilerplate. The fact that the 'tag' name used in the preferences and the variable name used in the Java code can be different offers some flexibility, but can also add confusion.
While still allowing the above scenario, this adds a `@Preference` annotation and a helper class that maps preferences to the static variables:
```
public class MyPreferences
{
    // Variable that'll receive the current value of the preference with same name
    @Preference public static int some_setting;		

    static
    {
        // Parse property file with default settings, apply Java preferences that might override,
        // set all fields annotated with @Preference
        AnnotatedPreferences.initialize(MyPreferences.class, "/my_app_preferences.properties");		    	
    }
```

In many cases, this significantly simplifies the preference code, handling basic settings of type `int`, `long`, `double`, `String`, `int[]`, `String[]`, `File`, `enum`.
If you need to parse/construct more complex settings from some preference text, the `AnnotatedPreferences.initialize()` method returns the `PreferencesReader` to use it as before.

If for some reason the name of the variable must differ from the tag name of the preference, the `@Preference(name="name_of_tag")` variant can do that.

There's a new unit test to check the new functionality, but I'll use some time to do more testing because so many classes have been updated to use the new API.

### DIFF
--- a/app/3d-viewer/src/main/java/org/phoebus/applications/viewer3d/Preferences.java
+++ b/app/3d-viewer/src/main/java/org/phoebus/applications/viewer3d/Preferences.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 Oak Ridge National Laboratory.
+ * Copyright (c) 2018-2020 Oak Ridge National Laboratory.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -7,7 +7,8 @@
  *******************************************************************************/
 package org.phoebus.applications.viewer3d;
 
-import org.phoebus.framework.preferences.PreferencesReader;
+import org.phoebus.framework.preferences.AnnotatedPreferences;
+import org.phoebus.framework.preferences.Preference;
 
 /** Preferences class for the org.phoebus.app.viewer3d package.
  *  @author Evan Smith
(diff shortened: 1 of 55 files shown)


georgweiss:
I like the simplification.
But... If I understand you correctly the Preference annotation is used in the class referencing the preferences file, right?

kasemir:
Not sure I understand the question. See updated online help:
```
In your application code, you can most conveniently access them like this

package org.phoebus.applications.my_app

import org.phoebus.framework.preferences.AnnotatedPreferences;
import org.phoebus.framework.preferences.Preference;

class MyAppSettings
{
    @Preference public static String my_setting;
    @Preference public static boolean my_other_setting;

    static
    {
        AnnotatedPreferences.initialize(MyAppSettings.class, "/my_app_preferences.properties");
    }
}
```

There's no change to having the end result in `MyAppSettings.my_setting`.
You also remain in control when the fields are initialized, you can debug by setting a breakpoint on that `AnnotatedPreferences.initialize` call.
What's gone is the need for these lines:
```
my_setting = prefs.get("my_setting");
my_other_setting = prefs.getBoolean("my_other_setting");
```

georgweiss:
I was think in "Spring terms":

```
@Preference("my_setting")
private String mySetting;
```

kasemir:
One point is to avoid different names for the preference and the variable used in the code, but you can use
```
@Preference(name="name_used_for_=preference") public static String name_used_for_variable;
```

Don't think it's a good idea, though.

georgweiss:
My point is how the annotation can be used: it seems only in the class loading the preference file. That's fine, but a more general approach would be to reference a preference setting in any class using the annotation.

kasemir:
As I said, another point is that you remain in control when the fields are initialized, you can debug by setting a breakpoint on that `AnnotatedPreferences.initialize` call.
You can't place @Preference just anywhere, unclear when it's handled and using which preference file.
As before, you control when and where you call `AnnotatedPreferences.initialize` or `new PreferencesReader`, and with which preference file.

georgweiss:
OK, point taken.

Issue #1638, opened 2020-11-16
Undo/redo stack cleared on save


georgweiss:
When user saves OPI in editor, the undo/redo stack is cleared => user cannot undo after save. This is different compared to (for instance) RCP-based CSS or Powerpoint where a save operation does not clear the undo stack.

I think we should maintain the undo stack even after save.

georgweiss:
I'd appreciate some feed-back in this matter as I find the current behavior to be wrong. Users might expect undo operations to still be available even after saving a file, as in RCP-based CSS.
Basically the "dirtyness" of an OPI file in edit mode is determined by the state of the undo/redo stack. This I think is wrong as it makes perfect sense to maintain the stack even after saving a resource.

pavel-ch:
From my point this is a nice-to-have feature, fine for me if this is implemented. Need to correctly implement storing the moment of Save action, to see correctly whether is the file dirty or clean when browsing the undo stack back and forth.

georgweiss:
@pavel-ch , thanks for your feed-back. But my point is that the undo stack should not determine the dirty state. Instead, if something is put on the undo stack, or removed from it, then that is a change that should set a separate dirty flag. Only a save operation should clear such a flag. A side effect would be that even if all actions are undone (i.e. undo stack is empty), the file is considered dirty. This seems to be the case in Eclipse-based CSS.

pavel-ch:
My understanding is that the file should be marked as dirty when the saved file is different from the current model state (current state = what is to be saved to a file from the current editor model state). This does not depend on emptiness of the undo stack. Traversing the undo stack, the flag has to be set or cleared according this rule.
This is what I consider the most ergonomic.
But because you implement that, the particular implementation is of your choice.

krisztianloki:
Well, I think that it would be really confusing to me if the display was shown as dirty when I know that I "undid" every change. And sometimes I just want to undo things (to check what changes I made) and I don't want to go further than what is already saved.

georgweiss:
@pavel-ch , I agree that comparing the underlying model is the solution that is conceptually the most correct approach. I could be wrong, but gut feeling this may not be that straight-forward. How should models be compared? Compute and compare hashes based on the widget hierarchy and other properties? Compare XML representations?

pavel-ch:
@georgweiss No idea how to implement that atm.

georgweiss:
I really would prefer the "compare models" approach, but not sure if I can prioritize it right now.

[...]

PR #1646, opened 2020-11-20
CSSTUDIO-1366 Maintain undo/redo stack after save

georgweiss:
#1638


kasemir:
Does this distinguish between a 'redo' and a new change?
I.e. will it do this:

Save (counter=0), undo (-1), redo (0) -> End up clean

Save (counter=0), undo (-1), new change (1) -> End up dirty

ralphlange:
Yes. Given an initial undo stack of e.g. 1000 elements, it will do:

Save (counter=0), undo (-1), redo (0) -> End up clean

Save (counter=0), undo (-1), new change (1001) -> End up dirty no matter

ralphlange:
Sorry - my comment was about my last suggestion for the algorithm.

Commit 7ac89a7 above does what you originally suggested, which will break.

kasemir:
Right. I agree with you that a counter can accomplish the task, but it needs to count differently for 'redo' and 'new action', which this PR doesn't seem to handle.

ralphlange:
See discussion at https://github.com/ControlSystemStudio/phoebus/issues/1638 where I suggested a slightly different algorithm. (That matches my comment above.)

Issue #1654, opened 2020-11-25
organization of app


shroffk:
I need to add 2 additional modules associated with the emails (latter logbook/olog too)

Currently we have a set of applications directly under app and a rather consistent naming scheme too

dir: /app/email
name: app-email

But it is not possible for adding new modules here.
Would it be ok to have the following.

dir: /app/email (parent project/module)
name: app-email

dir: /app/email/util (the main application code... I am open to a better name than util, but want to avoid app-email-email)
name: app-email-util

dir: /app/email/simple-context-menu (optional module for allowing site specific email configuration)
name: app-email-simple-context-menu

dir: /app/email/rich-context-menu  (optional module for allowing site specific email configuration)
name: app-email-rich-context-menu

This is in keeping with the established rules of both dir structure and naming even though some of the names do seem a bit long. With more changes I have planned to reduce the need to inclusion and exclusion preferences clarification and approval of the above format would be good.

georgweiss:
As for the structure I have no objections.
Naming is a different matter... app-email-simple-context-menu is a bit misleading: is the module context menu stuff only? How about:
* app-email-common
* app-email-plain
* app-email-rich-text

shroffk:
Well, for the time being the plain and rich have adapters for databrowser ( I am adding for opi runtime too)
I can remove the context menu part since it has the adapters only

I like common more than util...

@kasemir any thoughts

shroffk:
@georgweiss any idea about organizing the olog peices - the old olog client and the new olog client

/app/logbook/olog/client
/app/logbook/olog/es-client
/app/logbook/olog/ui ( viewer and editor )

georgweiss:
Isn't this the structure we have currently, save for naming?

georgweiss:
Sorry, you are looking from "olog root" perspective, right?
Then why not, separate clients as that could be isolated. But what about ui stuff upon which elog depends?

shroffk:
> Sorry, you are looking from "olog root" perspective, right?
> Then why not, separate clients as that could be isolated. But what about ui stuff upon which elog depends?

the general ui stuff remains where it is, the /olog/ui module will have the new olog specific pieces

kasemir:
As for email, note that there are also headless tools that send email.

core-email is supposed to hold just the basic email support.
app-email is supposed to hold the UI code.

In reality, core-email only has EmailPreferences.java.
app-email has an empty EmailService.java.
[...]

PR #1662, opened 2020-11-30
new organization of olog modules

shroffk:
includes moving the old and new client into a common parent module called olog #1654


georgweiss:
I'm afraid that in the restructured olog* modules package structure in `src` is incorrect, root is `phoebus` instead of `org`.
Also, while we're at it, could we consider renaming the XML* classes in olog-es, and optionally remove XML annotations? Suggest we do:
XmlLog -> LogEntry
XmlLogbook -> Logbook
XmlTag -> Tag
XmlAttachment -> Attachment
XmlProperty -> Property
I'd also propose we put these classes in a "model" sub-package.

shroffk:
Sorry I am working with intellij and was trying to do a refactor with a git move which would preserve history... not sure what I did wrong.
I will fix it and do the rename too.

shroffk:
The rename can't be exactly as you described since the interfaces share the same name. I would like to recommend 
OlogLog
OlogTag
....

shroffk:
Ok, 
I also used this as an opportunity to do a bit of cleanup. There was some reclics of the copy paste from the old olog which needed to be removed.

georgweiss:
Hmm, do we need the @Xml* annotations?

shroffk:
I think the annotations are used by fasterxml... maybe we should expand the serialization and deserialization tests. 
For the time being can we close this PR.

georgweiss (approved):

PR #1664, opened 2020-12-01
CSSTUDIO-1213 Radio and ChoiceButton widget fixes

georgweiss:
For Radio and ChoiceButton widgets the default item size is 2. Increasing once will add an item with the same name as an existing, i.e. "Item 2". This is a fix.

Also changed the default height of the Radio widget to avoid the tracker to be rendered over second item, see screen shot.
Screenshot 2020-12-01 at 08 40 16

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/ChoiceButtonWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/ChoiceButtonWidget.java
@@ -79,12 +79,9 @@ public class ChoiceButtonWidget extends WritablePVWidget
 
     private static ArrayWidgetProperty.Descriptor<WidgetProperty<String>> choiceItemDescriptor =
             new ArrayWidgetProperty.Descriptor<>(WidgetPropertyCategory.BEHAVIOR, "items", "Items",
-                    (widget, index) -> {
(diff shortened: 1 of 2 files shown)


app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/ChoiceButtonWidget.java
```diff
@@ -74,6 +77,15 @@ public ChoiceButtonWidget()
         super(WIDGET_DESCRIPTOR.getType(), 100, 43);
     }
 
+    private static ArrayWidgetProperty.Descriptor<WidgetProperty<String>> choiceItemDescriptor =
+            new ArrayWidgetProperty.Descriptor<>(WidgetPropertyCategory.BEHAVIOR, "items", "Items",
+                    (widget, index) -> {
```
kasemir, line 82: The 'index' is passed here so you can use that for the name instead of having to get the `size`.
So `(size+1)` can be replaced by `index`.

georgweiss: You're right, my bad. But it has to be `index + 1` unless we change scheme to "Item 0", "Item 1" etc.
Will update.

PR #1666, opened 2020-12-01
Change auto update task description while download is in progress

tynanford:
Our users like the auto update feature a lot, but some have become confused because the current text is "Downloaded https://..."

If the user does not make the box bigger, they might think the download has finish when it has not. This commit just updates the text. I will create an issue about some other issues they had that I didn't address in this.


kasemir:
Even better would be if it knew the final size, so it could show progress towards the total...
But yes, just this wording is already better.

PR #1673, opened 2020-12-02
fixed windows error when trying to open embedded examples

kingspride:
Hi,
I have a small contribution to make.
I dont know if this is considered a sloppy hack, but it works....

We noticed that on windows, phoebus generates the following error when trying to open the embedded example displays via File -> Top Resources -> Example Displays and then clicking any of the buttons. (i.e. LEDs)
phoebus would log the following:

```
Exception in thread "DisplayRuntime61" java.nio.file.InvalidPathException: Illegal char <:> at index 8: examples:/01_main.bob
        at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
        at java.base/java.nio.file.Path.of(Path.java:147)
        at java.base/java.nio.file.Paths.get(Paths.java:69)
        at org.csstudio.display.builder.model.util.ModelResourceUtil.normalize(ModelResourceUtil.java:136)
        at org.csstudio.display.builder.model.util.ModelResourceUtil.combineDisplayPaths(ModelResourceUtil.java:199)
        at org.csstudio.display.builder.model.util.ModelResourceUtil.doResolveResource(ModelResourceUtil.java:289)
        at org.csstudio.display.builder.model.util.ModelResourceUtil.resolveResource(ModelResourceUtil.java:253)
        at org.csstudio.display.builder.runtime.ActionUtil.openDisplay(ActionUtil.java:172)
        at org.csstudio.display.builder.runtime.ActionUtil.lambda$handleAction$0(ActionUtil.java:48)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

this seems to be an issue with windows' path names. they apparently cannot contain ":".

- William

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/util/ModelResourceUtil.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/util/ModelResourceUtil.java
@@ -133,7 +133,13 @@ public class ModelResourceUtil
         path = path.replaceAll("\\\\(?!\\\\)", "/");
 
         // Collapse "something/../" into "something/"
-        path = Paths.get(path).normalize().toString();
+        if(path.contains(":")){
+            String[] pathsplit = path.split(":");
+            String pathbefore = pathsplit[0];
+            String pathafter = pathsplit[1];
+            pathafter = Paths.get(pathafter).normalize().toString();
+            path = pathbefore + ":" + pathafter;
+        }
 
         // Pattern: '\(?!\)', i.e. backslash _not_ followed by another one.
         // Each \ is doubled as \\ to get one '\' into the string,
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/util/ModelResourceUtil.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/util/ModelResourceUtil.java
@@ -139,6 +139,8 @@ public class ModelResourceUtil
             String pathafter = pathsplit[1];
             pathafter = Paths.get(pathafter).normalize().toString();
             path = pathbefore + ":" + pathafter;
+        }else{
+            path = Paths.get(path).normalize().toString();
         }
 
         // Pattern: '\(?!\)', i.e. backslash _not_ followed by another one.


georgweiss:
What if path contains multiple ":" chars? Also: it looks like the code is putting back the ":" char where it was before split().

kingspride:
putting the : back where it was is neccessary for phoebus load mechanism to distinguish between local filesystem and inside jar container

regarding the multiple : youre right, thats currently not safe, could be taken care of. although I think on windows it wont happen that there will be a path with any : in it since the system should deny using those (I assume the modified function is also used when opening a file from file open dialog).

pavel-ch:
The solution must be from my perspective safe for all platforms (Win, Lin, Mac) and all sources (not only the open file dialog)

kasemir:
Note the special handling of "examples:/...". It's detected in the ModelResourceUtil and handled by getExampleURL. It's OK to use "examples:/.." for a URL, the "examples" simply becomes the 'schema', but fails when the Windows version of the file/path handler mistakes it for a "C:\" type drive letter. So the fix might be to check for the 'examples:' earlier.

kingspride:
I think fixing it inside of function normalize() is the right way since it seems to be the root of the cause.

maybe treating examples as a protocol instead would be an option?
for this, it would be neccessary to expand it by another "/" . (`examples://`)

just an idea

kasemir:
Meanwhile, your proposal helps on Windows, still works on Linux, passes the unit tests, so I think that's progress, thanks!

Issue #1665, opened 2020-12-01
Email module cleanup


shroffk:
The email modules were one of the first modules to be developed in phoebus. There have been many reorganizations and updates in the framework since their development. 

Thus the email modules could use a reorganization and rewrite.

@kasemir had some suggestions 
 - move the headless EmailService to the core module
 - implement send functionality in EmailService
 - refactor existing usage of "send email" to use the new EmailService

PR #1676, opened 2020-12-02
#1665

shroffk:
First set of cleanup of the email core and app modules.
EmailService.
The email dialog uses the core service to send emails.

Next Steps:
The EmailService and the Controller code can be further cleaned up. Should the controller only work with something like EmailEntry with all the javax.mail code completely compartmentalized in the core-email?


kasemir:
Looks good. When this is merged, I'll try to use the `EmailService` in the alarm server.

PR #1679, opened 2020-12-03
Email deps

kasemir:
Update eclipse and ant files for renamed app-email(-ui)


shroffk:
oh no.. I already have this fixed was just opening a PR

kasemir:
Well, I'm OK with updating the eclipse and ant files. Since I use them in the SNS build, I know if they work or not.
Just like you know best if the pom.xml are correct, where I wouldn't notice issues for a while.

PR #1680, opened 2020-12-03
Alarm Server: Use EmailService

kasemir:
Use the EmailService, seems to work (tested once).
@shroffk , you have a comment "might have to rethink this static initialization of the session."
I don't see a problem with static init, but not sure to which extend the 'session' can be re-used on different threads.


shroffk:
> I don't see a problem with static init
 
This might have to change if you have many mail servers and want to use them simultaneously in your client. In which case we might want to support multiple sessions.

> but not sure to which extend the 'session' can be re-used on different threads.
The method can be synchronized.

kasemir:
> ..if you have many mail servers and want to use them simultaneously in your client

Not impossible, but didn't need that in 20+ years. 
Applications running inside some beam line subnet might need to use a different email server than those on the office network, just like they use different EPICS CA address list etc., but within a given settings.ini there's just one email server.
So I don't think we need to worry about that.

PR #1681, opened 2020-12-03
#1374

shroffk:
#1374
Continued work on how context menu contributions are made.

Instead of hard coded calls, we use core-framework services to keep things flexible.
The changes in the PR ensure
1. sites can choose if to include email support in DB... without the need to preferences... by simply not including the util module in their product.
2. it could be possible to also support your own adaption of displays to emails
3. now the runtime is cleaner... no dependency to app-email-ui

image


kasemir:
This is getting harder to understand, several levels of indirection to basically **not** support email, when in the end every site-specific product is bound to support email...
But OK, I get the idea.
Maybe name the new module 'display-email' instead of display-util? Or 'display-adapters'?

shroffk:
While email is used everywhere it was a good example due to its simplicity.
I had done this with databrowser and now completed it with display builder.

While this example addresses the case of not supporting email. The next step is to allow supporting different logbooks and some of their extra functionalities without forcing everyone to support the same. We will also be able to move away from some hard coded context menu calls.

TODO: Should I add a section in the documentation to clarify this architecture/design pattern.

> display-adapters?

I wanted to keep it a little general in case we need to add something more in there... but I would be happy with display-adapters if you prefer :)

kasemir:
I think 'display-adapters' is good.
As for documentation, that could then also serve as an example:
Just the code needed to adapt something (display) to a context menu service (email, logbook, ..).
'display-util' sounds like it could contain anything, like edm converter etc., but better keep those separate.

shroffk:
I will also use the `<apps>-adapter` as the name template for similar modules in the future.

georgweiss:
In view of the unfortunate 4.6.4 release that contained a bug due to rework on the preferences annotation, may the commits preceding and related to this PR cause any issues?

shroffk:
> In view of the unfortunate 4.6.4 release that contained a bug due to rework on the preferences annotation, may the commits preceding and related to this PR cause any issues?

no worries, I will resolve the conflicts...

shroffk:
Can we merge this, I have will create a new PR with
1. documentation and examples
2. update other uses for hardcoded context menu contributions

image

PR #1683, opened 2020-12-04
Search by snapshot and tag

geniejhang:
@georgweiss I extended tag searching to snapshot&tag searching.
Search button icon is changed as well.

Untitled


georgweiss:
Sorry for delayed response...
@geniejhang , in SearchController#refreshList, can you please log the exceptions rather than calling printStackTrace? If a client is running against an older version of the service - i.e. a service that does not implement the search end-points - a HTTP 404 will propagate and trigger exceptions. That is OK, but logging is better than stack trace print.

geniejhang:
@georgweiss Yup. I agree with you.

PR #1685, opened 2020-12-10
Display 'info' dialog: Show widget counts by type and total

kasemir:
Adds a section to the 'Widget Info' dialog, opened from the top entry of the display runtime context menu when on the display itself, i.e. some background region.

Shows basic counts of widgets by type:

Screen Shot 2020-12-10 at 9 42 40 AM


georgweiss:
Maybe a bit late... is display = embedded display? Are embedded display widgets included?

kasemir:
Yes, embedded displays contribute a 'display' for their content.

kasemir:
In the screenshot, there are 4 'displays's:
One for the top-level display, and 3 for the three 'navtab' widgets which each host a sub-'display'.

Come to think about it, another useful count might be the number of rules and scripts.

Issue #1094, opened 2020-01-31
Display Builder Runtime: log and e-mail context menu entries cannot be configured out


pavel-ch:
Since recent commits (not yet evaluating particular changes), the configuring out of Display Builder Runtime log and e-mail context menu entries does not work.
Entries are still shown, although should not due to the settings.ini statements:
```
# Intentionally empty
org.phoebus.logbook.ui/logbook_factory=
org.phoebus.email/mailhost=DISABLE
```
This worked for me before.

shroffk:
Hello, 
I feel the proper way to address this would be to 

1. Add actions to the context menu using SPI, selection service, adapter service and not hard code them in.
This would also mean that we can clean up the dependencies... Display builder would not depend on logbook and email. The actions for contributing to these would be part of a separate module.

2. Then in your product you choose to include what you need and what you don't. If you don't want logbook and email integration, leave out org.phoebus.displaybuilder.logbook & org.phoebus.displaybuilder.email. 
This is better than including a module, and then adding a preference to basically uninclude the module.

I have added this as a topic for the upcoming developers meeting and I will implement a solution based on the above description if you can wait for another 2 weeks.

georgweiss:
The fact of the matter is that we have been struggling with making the use of classes (fonts, colors...) behave in a predictable manner for some time before we realized that what we observe is a timing/race condition issue. We are determined to deploy a release next week where this works, so no, we cannot wait for another 2 weeks. We can cope with  @pavel-ch 's issue for the time being.
However, I like you approach and suggest you go ahead.

pavel-ch:
@shroffk  yes this looks fine. Waiting for a while is not an issue for me. Priority of this particular issue is low, however, having a more modular product is of a certain priority I think. Thanks.

kasemir:
> This worked for me before.

Sure? I see no change in the code.
In databrowser `Perspective.createContextMenu`, an `SendEmailAction` is always added.
Inside `SendEmailAction`, `EmailPreferences.isEmailSupported` is used to enable/disable.

Setting `org.phoebus.email/mailhost=DISABLE` thus disables the context menu entry, but doesn't remove it.

Same with `SendLogbookAction`.

kasemir:
The display builder checks `isEmailSupported` and `LogbookUiPreferences.is_supported` before adding the context menu items, so there they're indeed removed when not supported.

kasemir:
[...]

PR #1687, opened 2020-12-10
#1094

shroffk:
make email and logbook context menu contributions modular
#1094 resolved.

Add documentation and some example code for explaining Adapters and SelectionService


app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/table/AlarmTableUI.java
```diff
@@ -7,9 +7,12 @@
  *******************************************************************************/
 package org.phoebus.applications.alarm.ui.table;
 
+import java.rmi.activation.Activator;
```
kasemir, line 10: This may be a bogus import, we're not using RMI

kasemir: If we _do_ use RMI Activation, we should reconsider, since it's deprecated, https://openjdk.java.net/jeps/385

shroffk: Fixing this now

shroffk: It was a bogus import and it has been removed

PR #1696, opened 2020-12-15
Eclipse: app-logbook-utility has been removed

kasemir:


shroffk:
Hi Kay, is the .classpath used for any other than the eclipse IDE.
While working on this refactor I set up ant builds and updated the build.xml to try to ensure that both build process continue to work.

but it seems like there are limitations to the IDE refactors... and I think relying on the automatic refactor seems to miss a few things.

kasemir:
It's only used for Eclipse.
Don't worry about them, those who use Eclipse can update them as modules move around.
Pavel already added the new files in #1691, this removes a reference to the now obsolete logbook-util.

shroffk:
Thanks...
I will try and figure how to maintain these too in the future.

Issue #1699, opened 2020-12-16
Eclipse reports Java warnings


MichaelRitzert:
When importing the project into Eclipse, my default settings yield a few Java warnings. I created a branch to remove some of the easy ones.

PR #1700, opened 2020-12-16
Warnings cleanup

MichaelRitzert:
referring to #1699

Besides the closed resource leaks, no change in behavior is intended.

The build has been checked in Eclipse and via maven.


services/scan-server/src/main/java/org/csstudio/scan/server/internal/JythonSupport.java
```diff
@@ -174,15 +174,17 @@ public JythonSupport() throws Exception
...
-            logger.log(Level.WARNING, "Error loading Jython class {0} from {1}",
-                new Object[] { class_name, pack_name });
-            logger.log(Level.WARNING, "Jython sys.path:\n * {0}",
-                       interpreter.getSystemState()
-                                  .path
-                                  .stream()
-                                  .collect(Collectors.joining("\n * ")));
-
-            throw new Exception("Error loading Jython class " + class_name + ":" + getExceptionMessage(ex), ex);
+            try (final var state = interpreter.getSystemState()) {
```
kasemir, line 177: `interpreter.getSystemState()` looks like this:
```
    public PySystemState getSystemState() {
        return systemState;
    }
```

It returns a reference to the shared `systemState` which you must not `close()`.

core/pva/src/main/java/org/epics/pva/client/CreateChannelHandler.java
```diff
@@ -52,7 +52,9 @@ public void handleCommand(final ClientTCPHandler tcp, final ByteBuffer buffer) t
 
             // Reset channel to init state and search again, after delay
             channel.setState(ClientChannelState.INIT);
-            tcp.getClient().search.register(channel, false);
+            try (final var client = tcp.getClient()) {
```
kasemir, line 55: `tcp.getClient()` returns the client that we continue to use.
It must not be closed just because we are adding one more channel to it, quite the contrary, we add the channel because we fully intent to use it.

app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/internal/JythonScriptSupport.java
```diff
@@ -189,29 +189,31 @@ private void addToPythonPath(final String path)
...
-
-            // Already top entry?
-            if (index == 0)
-                return;
-            // Remove if further down in the list
-            if (index > 0)
-                paths.remove(index);
-            // Add to front of list
-            paths.add(0, path);
+        try (final var state = python.getSystemState()) {
```
kasemir, line 192: The global system state must not be closed.

kasemir:
While some of the objects are `AutoCloseable`, that doesn't mean that one must close them all the time.
The PVA client needs to stay open for the duration of its use, the internal Jython system state must remain open for the duration of using jython.

Eclipse can't know if we `get` something that needs to remain available, or if we just `created` something that must now be closed.
Use `@SuppressWarnings("resource")` to avoid warnings in the former cases.

kasemir:
Of these 3 commits, I'm all for 'Remove unused imports' and 'Replace deprecated "new Integer(…)" with "Integer.valueOf(…)"'.

Only the comparably small 'Avoid resource leaks via try-catch.' is a problem since it closes resources that need to remain open.

Meanwhile, the branch has conflicts because the baseline changed.

Do you want to try again with just the imports?

shroffk:
Hi @MichaelRitzert thanks for the code cleanup.
Like Kay suggested could you split this PR with the optimized imports and the use of static constructors.

MichaelRitzert:
Sorry, I took some time to catch up with latest master. Please disregard the old branch and use issue1700.

PR #1703, opened 2020-12-17
Action button to show single action if description is missing

georgweiss:
If an action has one single action with empty description, the properties pane now shows "1 actions" instead of an empty string.
Screenshot 2020-12-17 at 10 49 02


kasemir:
That's a regression.
When there's only one action, it should show the description/label of that action.
Creator of the display can set that description to whatever, including empty.
If the action is configured to have no text, then that's probably by design, for example to get a small 'square' type button next to something, meant to open a detail screen. With this change, such buttons will suddenly show "1 action", likely truncated to "1 a".

georgweiss:
Not sure I understand your objection. At runtime the action button will render whatever text user specifies for the Text property. This PR is about showing in the property pane (the action edit button) that an action button is associated with one action, even if the description of that action is left blank. This is consistent with the case where X actions with blank or non-blank descriptions will render the text "X actions" in the editor pane button for editing actions.

kasemir:
>  At runtime the action button will render whatever text user specifies for the Text property. 

.. which defaults to `$(actions)`. Isn't that the text you're changing? So displays that used to have a blank action button would now show "1 actions"?

georgweiss:
No, if you leave the Text property of the action button empty, it will stay empty. This is not about how the action button renders, but rather the button in the editor pane that launches the actions dialog:
Screenshot 2020-12-18 at 14 56 27

kasemir:
Ah, good.

While you're at it, can you look into configuring a 'choice' based MessageFormat that will correctly handle "1 action" (singular) vs. "5 actions"?
Should be possible via something like 
`{0,choice,0#No actions|1#1 action|1<{0,number,integer} actions}`

georgweiss:
And while I'm at it I can also shape up the grammars for rules and scripts.

kasemir:
Good, thanks!

PR #1713, opened 2020-12-22
Publish artifacts to GitHub Packages

andersharrisson-ess:
Currently the GitHub Workflow build pipeline publishes the jar artefacts to ESS Artifactory instance. Though ESS is happy to accommodate this, it would make sense for the community project to be self sufficient.
This PR merges the two workflows we had into one. For the master branch and tags it will build and publish the artefacts to GitHub Packages. For any other branch, pull requests and forks it will merely do "mvn clean install" to verify that the commit builds successfully.
NOTE: GitHub Packages does not seem to allow publishing of executable jars, like the ones created by Spring Boot with <executable> set to true. In Phoebus there are 3 of these projects: alarm-logger, alarm-config-logger and save-and-restore. Since the documentation states that you run this with `java -jar` I think it should be fine to remove that ability.


shroffk (approved):
This a great... I like the idea of packages being hosted publicly on github

shroffk:
NOTE: GitHub Packages does not seem to allow publishing of executable jars, like the ones created by Spring Boot with set to true. In Phoebus there are 3 of these projects: alarm-logger, alarm-config-logger and save-and-restore. Since the documentation states that you run this with java -jar I think it should be fine to remove that ability.

Ok

I was curious about how the published bundles managed.
Is there a limit to how many bundles we can publish? Will we need to cleanup old published bundles

shroffk:
Merging now.
@andersharrisson-ess can you help me understand the maintenance aspect of this

andersharrisson-ess:
> I was curious about how the published bundles managed.
> Is there a limit to how many bundles we can publish? Will we need to cleanup old published bundles

As far as I can see, GitHub Packages is free for public repositories: https://github.com/features/packages#pricing
For public repos it's not possible to remove the existing packages. This is to ensure that dependent systems won't lose its dependencies.

> Merging now.
> @andersharrisson-ess can you help me understand the maintenance aspect of this

There shouldn't be much maintenance. GitHub Actions Workflow will run for all branches and merge requests. For commits pushed to master and pushed tags, the Workflow will also publish the jars to GitHub Packages. Maybe publishing SNAPSHOT-versions is a bit unnecessary. Maybe we should limit it to tags and rely on tagged commits being non-snapshots?
Either way, the result is a Maven repo available on https://maven.pkg.github.com/ControlSystemStudio/phoebus

Unfortunately you have to authenticate with GitHub to use it, which is not what you would expect with a public repo. Many people have raised their concerns with this: https://github.community/t/download-from-github-package-registry-without-authentication/14407
I am sure this will get solved eventually, but for now we have to authenticate when resolving the jars.

shroffk:
Thanks @andersharrisson-ess 

I think at the next monthly meeting we should discuss the details of how and when to publish however this is what I could consider just minor adjustments to what you have done

Issue #627, opened 2019-04-05
How to reduce Kafka logging for phoebus client, logback, slf4j


tanviash:
@kasemir Hi Kay, Is there a way to reduce logs that show up when Phoebus client is started? Right now, when I start up the Phoebus client using 'java -jar', I see huge number of kafka-related logs. On the service side, I can reduce this logs with .properties file but I can't find a way to reduce logs on the client side.

berryma4:
Just modify the logging level of the originating log entries. Or The whole
namespace
```org.apache.kafka=WARN```

**Comment by kasemir** [comment]
Right. The default log settings are in https://github.com/shroffk/phoebus/blob/master/phoebus-product/src/main/resources/logging.properties

My last commit was actually related to reducing the Kafka log levels:
https://github.com/shroffk/phoebus/commit/76bd041caeb41fcb7660d0fc43fe7b79d7792042

Now that's the default log config for the UI product.
The alarm server has its own file, https://github.com/shroffk/phoebus/blob/master/services/alarm-server/src/main/resources/alarm_server_logging.properties

All Phoebus products, when started from the command line, should understand the option `-logging` option, see 
https://github.com/shroffk/phoebus/blob/master/phoebus-product/src/main/java/org/phoebus/product/Launcher.java#L163 or
https://github.com/shroffk/phoebus/blob/master/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/AlarmLoggingService.java#L34
or
https://github.com/shroffk/phoebus/blob/master/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmServerMain.java#L494
So that way you can use your own log settings when starting tools via the command line.
When starting tools via springboot, I'm sure there's some other way, but I don't know it.

**Comment by tanviash** [comment]
Thanks Kay & Eric.

**Comment by shroffk** [comment]
Hello, 
Since the Kafka libraries use log4j, adding the above changes in the logging.properties does not work.

```
17:16:43.507 [AlarmClientModel] DEBUG org.apache.kafka.clients.consumer.internals.Fetcher - [Consumer clientId=consumer-1, groupId=Alarm-ddb1528b-ecc1-4277-871a-7b081aba010b] READ_UNCOMMITTED fetch request for partition XF23ID_OPR-0 at offset 2732 to node xf23id1-srv3.cs.nsls2.local:9092 (id: 0 rack: null)
```
[...]

PR #1721, opened 2021-01-07
Document logging

kasemir:
.. since I've had several email questions about that this week ..

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -17,6 +17,7 @@ User Documentation:
    running
    docking
    preferences
+   logging
    persistence
    authorization
    applications
--- a/docs/source/logging.rst
+++ b/docs/source/logging.rst
@@ -1,0 +1,18 @@
+Logging
+=======
+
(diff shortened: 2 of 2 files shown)


shroffk (approved):

shroffk:
Do you think we should also include example logging.properties and logback.xml files somewhere?

kasemir:
It's mentioning the ``logging.properties`` file of the ``core-launcher`` module as an example.
I wouldn't want to copy/paste that or anything specific in as static text, since that will soon be obsolete.
Do you want to change that to say a link "https://github.com/ControlSystemStudio/phoebus/blob/master/core/launcher/src/main/resources/logging.properties" to get the actual file?

For logback.xml, I'd consider that a  "temporary" workaround as described in #627.
As we just started to discuss, we might be able to configure slf4j, log4j, ... to all funnel their messages to java.util.logging.

shroffk:
I was specifically referring to the logbakc.xml file however let us hope that we simply get a long term fix.
Thanks

Issue #1690, opened 2020-12-12
Controls.fx version update


pavel-ch:
Can we use latest controls.fx maintained version 11.0.3 instead of 9.0.0 version? As I can see it is used in org.csstudio.display.builder.representation.javafx.widgets.SlideButtonRepresentation. I use it in a site specific code and want to be up to date. Thanks for hints.

kasemir:
I don't see a problem with updating, go ahead and make a PR, especially if you have (site specific) code that can tell the difference between "too old" and "good now".

kasemir:
Done, now have controls.fx 11.0.3.

PR #1722, opened 2021-01-08
Controls.fx version updated from 9.0.0 to active maintained 11.0.3. S…

pavel-ch:
…ome tweaks in app-display-representation-javafx pom.xml have been done due to Controls.fx 11.0.3 dependencies on javafx 11.0.2.

See #1690

### DIFF
--- a/app/display/representation-javafx/pom.xml
+++ b/app/display/representation-javafx/pom.xml
@@ -22,7 +22,26 @@
     <dependency>
       <groupId>org.controlsfx</groupId>
       <artifactId>controlsfx</artifactId>
-      <version>9.0.0</version>
+      <version>11.0.3</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.openjfx</groupId>
+          <artifactId>javafx-base</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.openjfx</groupId>
+          <artifactId>javafx-graphics</artifactId>
+        </exclusion>
+        <exclusion>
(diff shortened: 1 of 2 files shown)


kasemir:
Do we need the exact same exclusions in dependencies/phoebus-target/pom.xml?
Doesn't dependencies/phoebus-target/target/lib otherwise end up with multiple versions of the basic openjfx jar files?

pavel-ch:
Good question, I did this exclusion first in both poms but then realized that the dependencies one does not need exclusions. This was tested on my system, no multiple javafx jar versions in dependencies nor product/target.

kasemir:
>  tested on my system, no multiple javafx jar versions in dependencies nor product/target.

Thanks for checking that!

kasemir:
Well, I get mixed versions:
```
ls -1 dependencies/phoebus-target/target/lib/*fx*
(cd dependencies/phoebus-target/target/lib; ls -1 *fx* )
controlsfx-11.0.3.jar
javafx-base-11.0.2-mac.jar
javafx-base-11.0.2.jar
javafx-controls-15-mac.jar
javafx-controls-15.jar
javafx-fxml-15-mac.jar
javafx-fxml-15.jar
javafx-graphics-15-mac.jar
javafx-graphics-15.jar
...
```

So controlsfx is updated alright, but the openjfx jars are a mix of 11 for base and 15 for the rest.
Result seems to 'run', but I'll look at updating the dependencies/pom to have the same filter, so we get a consistent set of openjfx jars.

PR #1724, opened 2021-01-08
Controlsfx: Update, but keep openjfx15

kasemir:
#1722 updated the version of controlsfx.
In one pom, it had exclusions to keep openjfx15, preventing a downgrade to openjfx11.
In the 'dependencies' pom, there were no such exclusions, resulting in a mix of openjfx jars for version 11 and 15.


pavel-ch:
Interesting... I have really tested this and got rid of duplicate jfx versions with exclusions in just one pom... Maybe Windows vs Linux?

kasemir:
I believe you and leave it to maven affectionarios to try and explain what's happening, all I can say is that
```
```
resulted in `javafx-base-11.0.2.jar` whereas the rest was `javafx-*15*.jar`.

By placing the same exclusion lines in both poms, all is `javafx-*15*.jar`, so no problem there.

Issue #1725, opened 2021-01-11
Alarm formula functions


shroffk:
Add 2 new formula functions which predictably generate an minor and a major alarm when a particular condition is true

=minorAlarm(alarm_condition==true, "Some alarm message")

returns a VString with value = "Some alarm message" and alarm = Minor

While at NSLS2 it is strongly recommended that production alarms are implemented as EPICS pv's. the process of getting to a good trigger alarm pv can be made user friendly by providing them with these formula function to help identify the alarm condition.
Using these formula functions the alarm condition can be tweaked and tested by users ( who usually do not have a lot of EPICS knowledge) and then moved to dedicated alarm IOCs once properly identified.

shroffk:
Also, during operations there are situations where editing existing IOCs or building a new IOC on an epics server is prohibited for safety and reliability reasons... in such situations the above formulas can allow the alarm server to incorporate a potential alarm situation immediately without having to wait till the next downtime.

PR #1726, opened 2021-01-11
#1725 basic formula for conditionally generating minor and major alarms

shroffk:


kasemir:
This is surprisingly little code, mostly stuff that's needed to properly self-describe the new functions, and most of the logic is actually deferred to `VTypeHelper.toBoolean`.

Not sure how you plan to use it, but a useful related addition would be in the PV Tree.
So far it only understands INP links of certain records.
If it could also parse the inputs of an `eq:...` or `=...` equation, you could right click on display items etc. which hold a formula and thus see the result as well as the input values.

shroffk:
> If it could also parse the inputs of an eq:... or =... equation, you could right click on display items etc. which hold a formula and thus see the result as well as the input values.

I did not even think of this but a strong +1
Can we merge this and I will make a new Issue for it.

kasemir:
Sure, go ahead with this one.

The PV tree update shouldn't be too hard: Check for `expression.startsWith("eq:")` or `"="`, then use 
```
     // Parse formula just to determine its variables
     formula = new Formula(expression_text_after_eq, true);
     // Get the 'input links' from the names in  formula.getVariables();
```

PR #1732, opened 2021-01-13
Issue_1730 Error in alarm config logger cmd line parser

georgweiss:
Replaced separator char for list of topics from colon to comma, as echoed in help. This is also consistent with -topics cmd line option in alarm logger service.


shroffk (changes requested):
Can we support both comma and colon for backward compatibility
```
split("[,:]")
```

georgweiss:
Sure. Would you like the command line help content to reflect that both comma and colon are supported?

shroffk:
> Sure. Would you like the command line help content to reflect that both comma and colon are supported?

Yes, I guess it might also be important to note this since these should be considered as special chars that should not be used in alarm topic names

shroffk (approved):

Issue #1714, opened 2020-12-22
Logbook UI's


shroffk:
The logbook ui's should support momento's which remember the user query.

There should also be a preference for the initial default query to the used. Currently, this is hard coded but a preference makes more sense since the initial query will be different for different implementations of logbook.

shroffk:
I have created a new module for contributing olog specific ui components.
The idea is that only sites which use olog will include this module into their products.

This modules will have a lot of overlap with the general logbook ui module, in the future I will restructure both modules to ensure that we do not need to duplicate the general ui pieces. However, for the time being phoebus products will have to choose between the general logbook-ui or the logbook-olog-ui

shroffk:
@georgweiss I think we have done enough to close this broad logbook ui Issue.
Future feature requests and bug fix could be addressed via narrowly defined Issue tickets

PR #1733, opened 2021-01-13
#1714

shroffk:


georgweiss (commented):
Is this correct:
`default_logbook_query=*&start=12 hours&end=now`
I would assume 
`default_logbook_query=desc=*&start=12 hours&end=now`
But maybe I am missing something?

shroffk:
> Is this correct:
> `default_logbook_query=*&start=12 hours&end=now`
> I would assume
> `default_logbook_query=desc=*&start=12 hours&end=now`
> But maybe I am missing something?

done

georgweiss:
Since we decided to support Commonmark for OlogES I have come to realize that the delta to legacy Olog is quite substantial. Should not the new module be targeting Olog ES only?

shroffk:
> Since we decided to support Commonmark for OlogES I have come to realize that the delta to legacy Olog is quite substantial. Should not the new module be targeting Olog ES only?

Yes that will be the case.
However I do not want to make too many modules and confusion. 
The final product can choose 
1. old olog client and the general ui 
or
2. olog-es client and new olog ui

georgweiss:
Fine, but should not the added module be named accordingly, e.g. ui-es/app-logbook-olog-es-ui?

shroffk:
> Fine, but should not the added module be named accordingly, e.g. ui-es/app-logbook-olog-es-ui?

Well, the -es is a placeholder till we completely deprecate and remove the old olog.
I hope that in 1-2 years -es will not be needed

georgweiss:
That's OK as long as we agree that the added module targets Olog ES and nothing else.
However, there is still an "asymmetry" since we have client/client-es

Issue #1734, opened 2021-01-13
Enum formula function 


shroffk:
Request 2 enum formula functions
1. a formula to create a enum
2. a formula to retrive the current index of the enum as numerical value, this is equivalent to the `caget -n pv_name`

The second point is also important for backward compatibility, it is heavily used at NSLS2 in lieu of rules and scripts.

shroffk:
TODO: 

In assume in most cases you simply want to convert a number into an enum, so a format toEnum(value, labels) or actually toEnum(value, label0, label1, ..., labelN) seems much more straight forward to use:

PR #1735, opened 2021-01-13
#1734 adding enum formulas

shroffk:
While the second formula `enumOf(..)` should be deprecated since the functionality is broady available in `loc://pv<VEnum>` I would like to include it for a short period as I transition the old boy screens to bob.

Since we don't have a clear deprecation policy in Phoebus as yet, I would say that the `enumOf` could be removed in 90 days

### DIFF
--- a/core/formula/src/main/java/org/csstudio/apputil/formula/enums/EnumOfFunction.java
+++ b/core/formula/src/main/java/org/csstudio/apputil/formula/enums/EnumOfFunction.java
@@ -1,0 +1,60 @@
+package org.csstudio.apputil.formula.enums;
+
+import org.csstudio.apputil.formula.spi.FormulaFunction;
+import org.epics.vtype.Alarm;
+import org.epics.vtype.EnumDisplay;
+import org.epics.vtype.Time;
+import org.epics.vtype.VEnum;
+import org.epics.vtype.VNumber;
+import org.epics.vtype.VNumberArray;
+import org.epics.vtype.VStringArray;
+import org.epics.vtype.VType;
+
+import java.util.List;
+
+/**
+ * A Formula function to create an enum type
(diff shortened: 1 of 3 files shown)


kasemir (approved):

kasemir:
> ...  should be deprecated ..

I suggest to keep `enumOf(..)`. `loc://pv<VEnum>(1, "Bad", "Good")` can be used to create an enumerated value similar to `enumOf(1, ...`. Both will start out with the same value, but but it's still quite different.

The local PV is, well, a PV that can be used in several sections of the same display, where some write, others read, and all then get the same value. The formula can update when a numeric PV changes,
```=enumOf(`SomeNumericPV`, ...```

I understand that you model the `enumOf(value, intervals, labels)` after an existing implementation for compatibility, so can't change its definition. Still, I assume in most cases you simply want to convert a number into an enum, so a format `toEnum(value, labels)` or actually `toEnum(value, label0, label1, ..., labelN)` seems much more straight forward to use:
```toEnum(`SomeNumericPV`, "Bad", "Good")```

The `intervals` handling seems like a whole other functionality, could be a separate function `bracket(value, intervals)`.

**Comment by shroffk** [comment]
The formula can update when a numeric PV changes,
```=enumOf(`SomeNumericPV`, ...```

+1 I think this is good justification for keeping this.

I agree that the method signature you describe is more intuitive. How about instead of deprecation once I have completed the transition we updated the `enumOf` signature to be similar to the loc enum signature.

**Comment by kasemir** [comment]
> how about .. update the enumOf signature 

Sure, that could be done as long as it remains compatible:
```
if (args[1] instanceof  VNumberArray)
     // (value, intervals, labels)
else if (args[1] instanceof VString)
     // (value, label0, label1, ...) 
```

PR #1736, opened 2021-01-13
Add slf4j binding for JUL and exclude log4j slf4j binding

jacquelinegarrahan:
This PR adds the slf4j binding for JUL and excludes the log4j slf4j binding, resulting in kafka logs surfaced with JUL. See: #627 

Additionally, I have a development question:

I am seeing modifications to my `dependencies/phoebus-target/.classpath` after `mvn clean install`. Why is this a tracked file and how should I be managing it during development?

### DIFF
--- a/app/save-and-restore/ui/pom.xml
+++ b/app/save-and-restore/ui/pom.xml
@@ -15,6 +15,12 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
 			<version>2.1.5.RELEASE</version>
+			<exclusions>
+				<exclusion>
+					<groupId>org.springframework.boot</groupId>
+					<artifactId>spring-boot-starter-logging</artifactId>
+				</exclusion>
+			</exclusions>
 		</dependency>
 
 		<dependency>
--- a/dependencies/phoebus-target/pom.xml
+++ b/dependencies/phoebus-target/pom.xml
@@ -491,6 +491,15 @@
       <artifactId>validation-api</artifactId>
       <version>2.0.1.Final</version>
     </dependency>
+
+    <!--JUL bindings for sfl4j-->
+    <dependency> 
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>1.7.28</version>
+    </dependency>
+
+
   </dependencies>
 
   <build>
--- a/services/alarm-config-logger/pom.xml
+++ b/services/alarm-config-logger/pom.xml
@@ -29,6 +29,12 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework.boot</groupId>
+          <artifactId>spring-boot-starter-logging</artifactId>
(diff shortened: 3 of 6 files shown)


georgweiss:
Good finding, +1.
However, I tried adding the exclusion to app/save-and-restore/ui only and not adding the JUL binding dependency. This also removed the excessive Kafka client logging.
We should maybe not add the exclusions to the service/* modules...

kasemir:
> .. .classpath 

The .classpath files are there for those who use Eclipse, and they're updated by those who use Eclipse.
If you build with maven or use netbeans, intellij, vscose, ... then simply ignore them.

jacquelinegarrahan:
@georgweiss  Is the goal of #627 to completely silence kafka logging or to reduce via JUL logging configuration? 

Actually, if migrating to JUL, this PR as-is incomplete and the binding must also be added to the alarm services. Otherwise the following message is logged:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

The save and restore service exclusion can be dropped.

georgweiss:
Yes, #627 is about silencing the Kafka client in Phoebus.
As for the service modules I'd like save-and-restore pom to be unchanged.

jacquelinegarrahan:
@georgweiss I would guess there was value in keeping Kafka logging at least available for users. With the JUL bindings, the logs can be silenced in the configuration:
`org.apache.kafka.level = OFF`
Is that acceptable?

georgweiss:
@jacquelinegarrahan , not sure I understand.
With this PR the excessive Kafka client logging in the Phoebus app is silenced. However, the changes to the save/restore service pom will remove some logging I'd like to keep.
So, in short, I do not mind any changes to the pom files, except for the save/restore service pom, which I'd like to be unchanged. 
If I'm missing some aspect of your PR, please let me know.

jacquelinegarrahan:
The save/restore service pom changes have been dropped for this PR.

There are two ways to silence the excessive Kafka logging (effecting both the client and services). 

1. Remove all kafka logging by excluding the log4j slf4j binding. This is accomplished by the springboot exclusions.
2. Add the ability to silence the kafka logging by JUL loggers. This is accomplished by the springboot exclusions and the addition of the JUL slf4j binding. Then, kafka logs can be manually silenced with:
`org.apache.kafka.level = OFF`

With my last comment, I am trying to a case for solution number 2. If 2 is preferable, this PR is incomplete and the bindings must also be added to the alarm service poms. I am wondering why the additional dependency may not be wanted in this case.

kasemir:
2 is preferable.

Based on #627, https://github.com/ControlSystemStudio/phoebus/blob/master/docs/source/logging.rst and recent emails we concentrate on JUL, since we can configure it via command-line `-logging` options and for the GUI product there's also an interactive method.

While the immediate desire tends to be to just disable the flurry of INFO type kafka library messages, the better solution is to route them to JUL and then configure them such that you would still see WARNING or SEVERE type errors, should they occur.

jacquelinegarrahan:
Great- I've updates the service poms to include the JUL bindings.

shroffk:
@jacquelinegarrahan thanks

PR #1748, opened 2021-01-26
Changes to support Commonmark in Olog ES UI

georgweiss:
In short, user may use markup as defined by the Commonmark in the log entry text field. In the log table view the text is processed to HTML and rendered in a WebView.

For this to work, the Olog ES service must be of commit 9c1f2da7ff01f1dd41e07651eb25d3c7208e17a7 or later.

### DIFF
--- a/core/logbook/src/main/java/org/phoebus/logbook/LogEntry.java
+++ b/core/logbook/src/main/java/org/phoebus/logbook/LogEntry.java
@@ -13,6 +13,10 @@ public interface LogEntry {
 
     public String getDescription();
 
+    public default String getSource(){
+        return null;
+    }
+
     public String getLevel();
 
     public Instant getCreatedDate();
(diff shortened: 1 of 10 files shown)


shroffk (approved):

core/logbook/src/main/java/org/phoebus/logbook/LogEntry.java
```diff
@@ -13,6 +13,10 @@
 
     public String getDescription();
 
+    public default String getSource(){
```
kasemir, line 16: Not specific to this commit/change, but `LogEntry` is API for code that uses the logbook, or code that implements the binding for a site-specific logbook.
As such, it should have javadoc to describe its methods.
This commit doesn't have to retro-actively add all the missing javadoc, but maybe describe at least the newly added method, since it's not immediately obvious:
Source of what? The body of the entry? In a certain format, or allowing plain text, html, markup, markdown, ..?

PR #1754, opened 2021-01-27
2 Issues related to channel tree

shroffk:


kasemir:
That short info in the online help will be useful!

Issue #1753, opened 2021-01-27
Ordered Selection Dialog


shroffk:
A selection dialog where the selection order is preserved. There is already one such dialog in the channel/ui module
  - [x] fix bug in ordered selection dialog
  - [x] move the dialog to a more commonly used module

PR #1756, opened 2021-01-28
#1753 moving dialog to core-ui, adding javadocs, ui demo

shroffk:
I think this dialog might be useful in other places too
image


kasemir:
Maybe use the shared up/down images, `JFXUtil.getIcon("up.png")`, `JFXUtil.getIcon("down.png")`?

Screen Shot 2021-01-28 at 12 59 13 PM


Those are using the up.png and down.png from the Eclipse icons.
Could look for left/right icons, or in a pinch create them by rotating the up/down images in an image editor.

shroffk:
image

kasemir:
Does double-clicking on an available option add it, and double-clicking on already added option remove it?

kasemir:
.. Looks like the code contains double-click handlers that move left/right, good!

PR #1761, opened 2021-02-03
Save&Restore searching UI issue

geniejhang:
Hi @georgweiss 
This is for fixing sorting issue by created date in searching window


georgweiss:
To be honest, I have not used the search function previously. Now that I do I do not quite understand how it works. There is no submit button, and hitting the Enter key does not seem to do anything.

So I cannot test your PR...

geniejhang:
No submit button because it actively refines the result as you type in.
If you have snapshots and tags, it shows all of them in one window.

Sorting in this PR is clicking table column header in the window.

georgweiss:
Ah, you're right. I was using outdated version of service.

PR #1765, opened 2021-02-08
Check ack auth rule while double-click action

tanviash:
Currently, authorization rules for alarm acks is not checked when double-click is used to ack an alarm in alarm table. This PR fixes it.


kasemir (approved):
Good, thanks!

kasemir:
Do you want to do the same for "un-acknowledge"?
There's no separate authorization item, but I'd say the right to acknowledge includes the right to un-ack.

tanviash:
> Do you want to do the same for "un-acknowledge"?
> There's no separate authorization item, but I'd say the right to acknowledge includes the right to un-ack.

Right, they should stay the same.

PR #1768, opened 2021-02-15
Adding export to file in Widget Info dialog

georgweiss:
As per user request: adding button in Widget Info dialog to export contents of the tab tables.


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
```diff
@@ -134,17 +154,104 @@ public WidgetInfoDialog(final Widget widget, final Collection<NameStateValue> pv
...
+            fileOutputStream.write(buffer.toString().getBytes());
+            fileOutputStream.flush();
+            Alert alert = new Alert(AlertType.INFORMATION);
+            alert.setTitle("Information Dialog");
+            alert.setHeaderText(null);
+            alert.setContentText(MessageFormat.format(Messages.ExportDone, file.getAbsolutePath()));
+            alert.showAndWait();
+        } catch (IOException e) {
+            Logger.getLogger(getClass().getName()).log(Level.WARNING, "Failed to export widget info", e);
+            Alert alert = new Alert(AlertType.ERROR);
```
kasemir, line 244: To display an error with exception, consider using `ExceptionDetailsErrorDialog` which shows the error message and allows users to expand the full exception, with option to copy the text out.

kasemir (approved):

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
georgweiss: Sure, that would make it easier to debug. I keep forgetting about `ExceptionDetailsErrorDialog`...

kasemir: Another good practice for any dialog is using `DialogHelper.positionDialog(...` to locate it relative to for example the button that triggers a dialog, because otherwise users with more than one monitor often get it on the 'other' monitor.

PR #1775, opened 2021-02-22
Delete 'superflous messages' in org.csstudio.trends.databrowser3.Mess…

pavel-ch:
…ages


kasemir:
06aed1ff2ea4cb7d48778da7dfbac6f652ddc521 removed most of these Messages.java entries, so they're indeed no longer used.

PR #1776, opened 2021-02-23
Updated log viewers

shroffk:
image


shroffk:
@georgweiss I think you might have to review the last commit.
I removed the webview from the table and updated the webview in the display with the new Attribute provider

georgweiss:
The attribute provider is not correctly installed. 
LogEntryDisplayController line 138 should read:
`.attributeProviderFactory(context -> new OlogAttributeProvider())`
As a consequence the OlogAttributeProvider inner class in LogEntryCellController could be removed.

georgweiss:
Otherwise looks good, I find the table viewer more useful now. Maybe the height of properties and attachment pane could be reduced in favor of log entry content, but that can be tweaked as we collect feed-back.

georgweiss:
Hmmm, seems ordering in the table view is by ascending date stamp... Is that intentional?

shroffk:
> Hmmm, seems ordering in the table view is by ascending date stamp... Is that intentional?

No, I will fix it so the the newest is on the top

PR #1779, opened 2021-02-24
Support for import of untrusted certificate

georgweiss:
When using WebView from JavaFX 14+ under Java 12+, the usual tricks to configure the SSL layer to trust all hosts and certificates does not work. For instance, an image tag in the WebView pointing to a resource from an Olog service instance over HTTPS with self-signed certificate will simply not work. See https://bugs.openjdk.java.net/browse/JDK-8242077.

This PR implements a strategy where the Olog server certificate - if found to be untrusted - is downloaded and imported into the JVM trust store. Note that the augmented trust store is in-memory only, no file is written to disk. If on he other hand the server certificate is trusted, nothing is changed with respect to the default JVM trust store. 

While the import of the untrusted solves the WebView issue, it also removes the need to configure the Jersey client to skip host validation, and there is no need for a custom trust manager.

The ideas in this PR could possibly be further explored to support a more generic solution where Phoebus as a whole could be configured to import self-signed certificates from all services that for some reason cannot use globally trusted certificates.


kasemir:
core-framework has `ResourceParser` with `public static trustAnybody()`, which might be usable as a common way to trust self-signed certificates. `ChannelFinderClientImpl` contains similar code, might also be able to use the common method.

georgweiss:
The workaround in `ResourceParser#trustAnyone()` works fine except in the case mentioned: WebView from JavaFX 14+ and running under Java 12+.

Also: importing certificates from known service endpoint using untrusted certificates may remove the need for `ResourceParser#trustAnyone()` and classes like DummyX509TrustManager (which by the way exists in multiple packages). This PR shows that the import strategy solves the issue for the Olog ES client.

kasemir:
Do we need a `-import_certificate /path/to/cert` command line option in the `Launcher`?
Or is the best solution to package the product for each site with a bundled JRE that has the local certificates installed?

georgweiss:
In my view customized JRE means additional maintenance, e.g. when certificates expire or should be added or removed.
A command line option could be an alternative, but that would also mean maintenance when the certificate must be exchanged.

georgweiss:
@shroffk, @kasemir, how do we proceed. I need a decision to be able to proceed with Olog roll-out.

kasemir:
With this olog change, I'm OK to go ahead.

Was just wondering if there's something in here that should be generalized.

Use `ResourceParser.trustAnybody()`? No, this is different.

Bundle a JRE? Could do that, but it's a hassle. Well, for the SNS Windows and Mac product I do bundle a JRE because to end users that's much easier than having to install it separately. But each site can already do that as they see fit.

Add a command line option? Maybe, but that would still mean somehow getting the certs to import to the end user. Might as well bundle a JRE.

So maybe there is nothing to generalize, just do this for the olog and that's all.

shroffk:
> Was just wondering if there's something in here that should be generalized.

I think there can be a core module which provides the hostname validators and other ssl resources which are reused by olog, olog-es, channelfinder, etc...

But for the time being this PR can be merged.

shroffk (commented):
approved

georgweiss:
Thanks for your comments.

georgweiss:
Regrettably I need to say that my verification procedure failed when running on Java 12+.
While importing the untrusted certificate the way `OlogTrustManager` does it will enable the client to retrieve log entries from the service, but the WebView will still reject the server certificate. Not sure why I had this working some days ago, but most probably I did not have the correct test setup. 
Note that calling the `OlogTrustManager#setupSSLTrust` in the `Launcher` class will solve the certificate issue also for the WebView. But since Olog is optional, this is not feasible. I would need more time to investigate further.
In the meanwhile I will use the workaround mentioned in https://bugs.openjdk.java.net/browse/JDK-8242077.

kasemir:
I assume that CS-Studio is only one of the olog clients at your site.
What about plain web browsers? Do they have the same problem when accessing an olog that uses self-signed certificates? Does every olog server in the end need to buy normal certificates, and then there's no issue with any client any more?

georgweiss:
We're rolling out Olog, but I guess more may follow.
As for web browsers we bundle certificates with Firefox provided by the IT department. But that does not help mobile clients, which I expect would be eligible clients for the Olog web UI. Ultimately we will need to convince infra staff to install "proper" certificates.

PR #1785, opened 2021-03-02
Fix annunciator active alarm count

tanviash:
Currently, the active alarm count periodically annunciated by annunciator is not correct because the entire alarm tree is not traversed to count all active PVs. This change fixes the issue and reports correct active alarms in annunciations.


kasemir (approved):
Good, thanks!

PR #1793, opened 2021-03-08
Replace Travis and AppVeyor badges with GitHub Actions badge.

willrogers:
Replaces the Travis and AppVeyor badges with a GitHub Actions one. 

I also think we should remove the Travis configuration as it will stop working soon anyway.


kasemir (approved):
Fine with me

shroffk (approved):

PR #1809, opened 2021-03-08
Fix NPE that results from fetching file outside of class's resources

kasemir:
#1784 ends with NPE at runtime
```
   [java] Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "java.net.URL.toExternalForm()" because the return value of "java.lang.Class.getResource(String)" is null
     [java] 	at org.csstudio.display.builder.representation.javafx.actionsdialog.ActionsDialog.<init>(ActionsDialog.java:82)
     [java] 	at org.csstudio.display.builder.editor.properties.ActionsPropertyBinding.lambda$new$1(ActionsPropertyBinding.java:41)
```

because it tries to fetch a resource "../opibuilder.css" one level 'up' from a class's resources.

Using a class within that package name and then just "opibuilder.css" fixes it.


georgweiss:
Thanks @kasemir, did not notice this NPE.

Issue #1814, opened 2021-03-09
common trustmanagers in core-security


shroffk:
There are some trust managers which are reused in multiple places which should be moved to core-ui.

shroffk:
@georgweiss should this Issue be kept open?

PR #1817, opened 2021-03-09
#1814

shroffk:


georgweiss (approved):
Refactoring to eliminate duplicates is fine. However, more work is needed to investigate how to best install the trust manager. But that is a separate ticket.

PR #1823, opened 2021-03-11
Advanced Search Dialog for alarm log table

tanviash:
This PR is to add support for advanced search dialog for alarm log table. #1802


shroffk:
+1
I think we should also add documentation and a screenshot.

But... I am fine with merging this and doing the documentation on the documentation branch

shroffk (approved):

PR #1825, opened 2021-03-12
Documentation

shroffk:
WIP : waiting for all the other documentations


kasemir:
When explaining data sources, formulas, ..: They're more general than just the display builder, so extending core/pv/doc/index.rst would be a good place for details.

shroffk:
I think that is true... formulas are inherently part of the core pv modules so we can expand the documenations there.

However, I think there is a lot of value in adding a section in Display Builder about how formula functions can be used to get a lot of useful dynamic behaviours without the need for rules or scripts. So the section under DB could be more along the lines of the application/usage of formula functions.

georgweiss:
@shroffk, @kasemir , any reason this should not be merged?

kasemir:
Kunal's original WIP comment still holds as long as there's mention of OPI files, opibuilder imports, javascript instead of jython. On the other hand documentation will never be complete, so I'm OK with merging this as a first step. On the 3rd hand, I worry that we'll have to wait for another codeathon in a year to fix the OPI references.

georgweiss:
I have users waiting for documentation, e.g. the sections on scripting. So I'd say incomplete documentation is better than none.

kasemir:
I don't have a good solution. Discussions about scripting are often like this:

 "I want to do open heart surgery!"
 "OK, here's the knife."
 "Well, where do I make the first cut? Is this thing sharp?"

Having offered scripting in the first place was a mistake.

georgweiss:
Should we take a vote on script support I would not hesitate a bit to remove it.

kasemir:
Yes. But too late. So let's merge this and hand out another rusty knife that will encourage more people to create scripts ;-)

georgweiss:
Well, scripting is not the only rusty knife. Crappy IOCs are maybe even worse...

I added this on top of the script doc:

DISCLAIMER

Scripting in CS Studio should be used only if a particular task cannot be achieved with the built-in features of the various widgets, or with formula functions or rules. In particular, CS Studio should not be used as a script execution platform for data processing or complex business logic. It should also be noted that each script referenced by the widgets of an OPI is compiled when the OPI is loaded, which adds to the overall load time, and which may degrade the responsiveness of the UI.

Supported script language versions are Python 2.7 and JavaScript 1.7 (ECMAScript support is very limited).

shroffk:
The reason I have not merged... despite documentation being an ongoing effort... is cause there are empty sections.
Since this documentation is published, I think empty sections look bad.

I think @tanviash and @shroffk have to fill a few of the still empty sections :)

georgweiss:
A comment/question on alarm configuration documentation: the Automated Actions is - correct me if I'm wrong - about actions automatically executed under certain conditions. However, these actions are executed on the alarm server instance, not - as opposed to commands - on the client. 
So I propose I add something to clarify this in the text.

shroffk:
I have tried to fill out most of the empty pages.
We should merge this before this branch gets stale...the documentation project can continue.

georgweiss:
What about clarifying the Automated Actions for alarms, see previous comment.

Issue #1827, opened 2021-03-12
The ListImages dialog editable/viewable 


shroffk:
The ListImages dialog has a bug - it does not correctly enable the "remove" button even in editable mode

PR #1828, opened 2021-03-12
#1827 the remove button is correctly enabled

shroffk:
@georgweiss  this should fix the missing "remove" button for images


shroffk:
Also adding to 2 Demo's to verify the editable and non editable behaviour of the imagelist

georgweiss:
@shroffk , any reason to not merge?

PR #1829, opened 2021-03-12
Passing tag and property key names ending with "!" for NOT searching

geniejhang:
ChannelFinder API needs to be updated for the support.

Discussion #1816


shroffk (approved):
I think this looks a simple enough way to get the "not" search parameters out.

PR #1830, opened 2021-03-15
Add scrollbars to WebView in log entry view

georgweiss:
The changes basically give the WebView a fixed (large) size and add scrollbars. After all, it's the body text of the log entry that should be of main concern in the UI.

In addition the proportions of the different parts of the layout do not change, which makes it (IMO) a bit more user friendly.


shroffk (approved):
Thanks for cleaning up the exception handling too

PR #1834, opened 2021-03-23
NPE fix in ActionsDialog

georgweiss:
To be noted: the NPE is thrown only when running CS Studio installed through a package built with the jpackage tool. Obviously a class loading issue, but I have no details.


kasemir (approved):
Can't debug this and would prefer not using fxml in the first place, but if this is what you want, go ahead

Issue #1855, opened 2021-04-08
pva preferences


shroffk:
Just like the pv_ca_preferences it would be nice to have equivalent preference file for the pva datasource.

PR #1856, opened 2021-04-08
#1855 Adding pva preference handler

shroffk:
@jacquelinegarrahan noticed this and requested a pv_pva_preferences

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/pva/PVA_Context.java
+++ b/core/pv/src/main/java/org/phoebus/pv/pva/PVA_Context.java
@@ -20,6 +20,7 @@ public class PVA_Context
 
     private PVA_Context() throws Exception
     {
+        PVA_Preferences.getInstance().installPreferences();
         client = new PVAClient();
     }
 
--- a/core/pv/src/main/java/org/phoebus/pv/pva/PVA_Preferences.java
+++ b/core/pv/src/main/java/org/phoebus/pv/pva/PVA_Preferences.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Oak Ridge National Laboratory.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ ******************************************************************************/
+package org.phoebus.pv.pva;
+
+import gov.aps.jca.Monitor;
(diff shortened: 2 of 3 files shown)


jacquelinegarrahan:
Thanks for setting this up, @shroffk !

kasemir:
Do we really want this? It's been left out on purpose.

For Channel Access, all EPICS tools used EPICS_CA_.. environment variables.
CS-Studio then introduced its own preferences, and I started to get emails and questions why that is, why we can't simply use the environment settings like every other EPICS tools. Then we kept changing the preference values, and the emails and questions got worse. So for PVA, I kinda liked the idea of simply using the existing environment variables, with the only concession to Java that you may also use Java properties of the same name.

This PR leaves the preference settings empty and then ignores them, so it doesn't change the default behavior.
Maybe this should be emphasized in `pv_pva_preferences.properties` by adding something like this:

```
# By default, these preference settings are empty,
# and the PVA library will then honor the commonly used
# environment variables like EPICS_PVA_ADDR_LIST,
# EPICS_PVA_AUTO_ADDR_LIST etc.
# Defining preference values will override the environment
# variables which allows consolidating PVA settings
# with all the CS-Studio preference settings
```

shroffk:
I added the more detailed explanation that Kay pointed out.

I agree with Kay on the issues associated with a hierarchy of preferences defined in different places. 

Some benefits to the single preference file are that all the settings are consolidated in a single place... also they are a *little* more difficult to mess up.
With shared user accounts, home directories nfs mounted on different subnets, multiple apps also messing around with env settings.. having the separated preference files makes deployment a bit simpler. I understand that this is more of deployment/management issue but having the above feature makes it a bit easier.

 I am also ok with closing this PR if there are strong concerns :)

kasemir:
Note that I personally like the preference handling because it allows me to conveniently bundle all the settings (archive access, EDM converter, CA settings, ...) for SNS users in one file, put that into the SNS-specific product, and it just works, no need to have environment variables defined outside of phoebus.

But then the rest of the world sends emails about not using environment variables.

So I think it's good to have both:

Add a comment like
```
# By default, these preference settings are empty,
# and the PVA library will then honor the commonly used
# environment variables like EPICS_PVA_ADDR_LIST,
# EPICS_PVA_AUTO_ADDR_LIST etc.
# Defining preference values will override the environment
# variables which allows consolidating PVA settings
# with all the CS-Studio preference settings
```
to the PR, so by default it still uses the environment variables, those who like to complain about us not using them can remain quiet, yet those of us who bundle site-specific products can combine all our settings into the preferences file.

kasemir (approved):

PR #1872, opened 2021-04-16
Minor facelift to the log entry table cell

georgweiss:
I have done some changes to the list cells in the log entry table view (Olog), this is based on user feedback and also better aligns to the design chosen for the web client. In my view the changes improves readability of the table.

Most noticible change is the logbooks and tags lines moving to the left. Since the string generated for the list of logbooks and tags may vary, it makes more sense to "left align". Also removed the attachment count.

This PR also fixes a bug in the cell rendering: tags string was "inherited" from preceding cell when tag list is empty.

Screenshot 2021-04-16 at 15 52 17


shroffk:
Ok,
I think the right alignment was an effort to optimize the space usage... but I am fine with this.

shroffk (approved):

PR #1875, opened 2021-04-20
Added deprecation information pv.setValue()

georgweiss:


app/display/editor/doc/access_pv_in_script.rst
```diff
@@ -27,24 +27,32 @@ than one trigger PV for a script and you need to know this execution is triggere
 
 **Examples**:
 
-*Get double value from PV:*
+**Get double value from PV:**
 
 .. code-block:: javascript
```
kasemir, line 32: Regarding JavaScript vs. Jython for scripting, see 
https://github.com/ControlSystemStudio/phoebus/wiki/Display-Builder-Script-Compatibility#jython-vs-javascript

georgweiss: Copy-paste error. I'll update.

georgweiss: On second thoughts.... the syntax `importPackage(Packages.org.csstudio.opibuilder.scriptUtil);` works just fine.

kasemir: While the long-term future of either scripting support in Java is not known, Jython is for now preferred because it has been longer-lived than the changing JS script support. On the wiki I thus only show examples for Jython.

georgweiss: I see. Are you asking me to change to Jython?

kasemir: On the wiki I use Jython because it has been the same since the beginning of BOY, while JavaScript support used different libs, required different import statements over time. At the same time, I cannot guarantee that Jython will remain available, and Javascript starts up faster.

georgweiss (commented):
Can we please decide to merge or take a decision to update all code examples to use either Jython or JS?

georgweiss:
I've updated the doc to user jython.
Also linked to Javadoc instead of repeating the method documentation.
@shroffk, @kasemir, can be merge?

app/display/editor/doc/access_widget.rst
```diff
@@ -6,546 +6,25 @@ Access Widget in Script
...
+A script may obtain a list of all widgets in a display like so:
 
-.. code-block:: javascript
+.. code-block:: python
 
-    display.getWidget("myLabel").setPropertyValue("x", 20); //set x position of the widget myLabel to 20.
+    widget.getDisplayModel().getChildren()
 
-**Common methods to all widgets**
+To find a specific widget the script may then loop over the list and use `getName()` to match on the widget
```
kasemir, line 20: Looping over the child widgets will only consider the direct children, not widgets within a group or tab.
The suggested way to find widgets by name is `ScriptUtil.findWidgetByName`

georgweiss: OK, I will update.

kasemir: Great, then go ahead and merge,

PR #1877, opened 2021-04-22
make logbook implementations optional 

shroffk:
With this change... site specific products will have to explicitly list the logbook implementations they want to use.
so nsls2, ess, and frib can include olog and olog-es while fhi can include the elog one.

the common product will still build and include all which can be good for people using the common product.

Finally, it will be logbook_factory preference which will still be required to select from the included logbook implementations.


shroffk:
This PR must wait till #1875 is resolved (sorry)

tanviash (approved):

kingspride:
I emailed Kay and Kunal about this logbook factory issue, because I was not able to set the elog via my settings.ini.

I hope to narrow down why its conflicting with the olog app, but I am not a java "native speaker"

kasemir (approved):

core/logbook/src/main/resources/logbook_preferences.properties
```diff
@@ -0,0 +1,7 @@
+# ------------------------------
+# Package org.phoebus.logbook.ui
```
kasemir, line 2: I think this is typo, should say "org.phoebus.logbook" without .ui

shroffk: yes this was a typo...thanks for catching it

shroffk:
@kasemir @tanviash @georgweiss 
I think the new preference and its location is more appropriate the how we had it in the past.
The logbook client selection preference is now part of core...

Unfortunately, this might require some changes in your site specific preferences to match the new package name

kasemir:
OK on changing the preference name from my side.

georgweiss (approved):

core/logbook/src/main/java/org/phoebus/logbook/LogbookPreferences.java
```diff
@@ -0,0 +1,33 @@
+package org.phoebus.logbook;
+
+import org.phoebus.framework.preferences.AnnotatedPreferences;
+import org.phoebus.framework.preferences.Preference;
+import org.phoebus.framework.preferences.PreferencesReader;
+
+import java.util.logging.Level;
+
+import static org.phoebus.ui.application.PhoebusApplication.logger;
```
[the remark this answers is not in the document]
kasemir: core-logbook's LogService already has its own logger, so it really makes sense that core-logbook doesn't use the PhoebusApplication logger.

PR #1879, opened 2021-04-22
Use $(pv_name) instead of 'new-PV' as inital value

krisztianloki:
'new-PV' needs to be changed anyway and most of the time it is changed to $(pv_name)


kasemir:
I don't think that's generally true.
The basic PV name of a widget is already used by the widget itself.
For example, a text update widget will show the current value of the PV, with units, and also reflect the alarm state.
In an ideal world, there is no need to use a rule at all, unless you want to create displays that are unnecessarily slow and impossible to port to the next tool.
If a rule is needed, then it's about doing something entirely different to a widget's original purpose, triggered by some other PV. For example, show/hide a widget that's connected to PV X based on some other PV Y.

kasemir:
But having said that, we can leave it to people who want to use rules to figure out how they maintain them in the long run, so let's go ahead with this one.

PR #1883, opened 2021-04-23
Added opacity property to Picture widget

georgweiss:
As requested by users, added an opacity property to the Picture widget so that a rule or script can create a simple animation.

### DIFF
--- a/app/display/model/src/main/resources/examples/graphics_picture.bob
+++ b/app/display/model/src/main/resources/examples/graphics_picture.bob
@@ -1,33 +1,54 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<display version="1.0.0">
+<display version="2.0.0">
   <name>Picture</name>
-  <x>-1</x>
-  <y>-1</y>
   <widget type="label" version="2.0.0">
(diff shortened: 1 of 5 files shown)


georgweiss:
@kasemir , can you please review?

app/display/model/src/main/resources/examples/graphics_picture.bob
```diff
@@ -1,33 +1,54 @@
...
-  <y>-1</y>
   <widget type="label" version="2.0.0">
     <name>Label</name>
+    <text>Picture Widget</text>
     <width>181</width>
     <height>31</height>
-    <text>Picture Widget</text>
     <font>
-      <font name="Header 1" family="Liberation Sans" style="BOLD" size="22.0">
+      <font name="Header 1" family="Source Sans Pro" style="BOLD_ITALIC" size="36.0">
```
kasemir, line 10: Example displays should stick to the "Liberation" fonts that are bundled in the product.

georgweiss: Reverted back.

Issue #1882, opened 2021-04-22
JavaFX classpath vs. module path, "Unsupported JavaFX configuration"


kasemir:
Since JDK 9, JavaFX is meant to be loaded from the module path. It "works" when adding it to the classpath, but according to https://bugs.openjdk.java.net/browse/JDK-8256362 that is not supported, and with JavaFX 16 there is now a warning on startup:
```
WARNING [javafx] Unsupported JavaFX configuration: classes were loaded from 'unnamed module @4809244d'
```

The fix is to move all the javafx-*.jar files from the classpath onto the module path.

During development with Eclipse that can be done in the 'Dependencies' tab of the Run Configuration. Remove all javafx-*.jar from the "Classpath entries' and instead list them under the "Modulepath Entries". In addition, "Add modules:" at the bottom needs to be changed from the default "--" to "ALL-MODULE-PATH".

![Screen Shot 2021-04-22 at 3 09 58 PM](https://user-images.githubusercontent.com/1932421/115772189-d208db00-a37c-11eb-9e6d-fc7cd3da9aa3.png)

Unclear what the easiest way is to remove the warning from the phoebus product.
Add `--module-path` options for the modular javafx jars to the launch script?
Turn everything into modules?

kasemir:
About a year later with OpenJDK 18 there is the same warning.
Command line example:
```
mvn -DskipTests clean install
cd phoebus-product/target
java -jar product-*-SNAPSHOT.jar
```

Runs, but prints
```
WARNING [javafx] Unsupported JavaFX configuration: classes were loaded from 'unnamed module @76c835c5'
```

The 'fix' is to place the javafx jars on the module path. They are located in the `lib` folder of the product together with all the other jar files, but the following fails, presumably because it treats everything in `lib` as modules, while most are still plain jars:
```
java --module-path lib --add-modules javafx.controls -jar product-*-SNAPSHOT.jar
Error occurred during initialization of boot layer
java.lang.module.FindException: Unable to derive module descriptor for lib/elasticsearch-6.8.4.jar
```

One solution is to place all javafx jars in a separate folder:
```
mkdir modules
cp lib/javafx-*jar modules
java --module-path modules --add-modules javafx.web -jar product-*-SNAPSHOT.jar
```

Just listing `javafx.web` adds dependencies like `javafx.controls`, `javafx.core`. So support all phoebus apps, more modules may need to be listed.

PR #1884, opened 2021-04-24
BLD: Add compiler arguments to build with OpenJDK16.

hhslepicka:
Java version:
```
$java --version
openjdk 16.0.1 2021-04-20
OpenJDK Runtime Environment (build 16.0.1+9-24)
OpenJDK 64-Bit Server VM (build 16.0.1+9-24, mixed mode, sharing)
```

Maven version:
```
$mvn --version
Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
Maven home: /Users/slepicka/toolbox/maven/current
Java version: 16.0.1, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-16.0.1.jdk/Contents/Home
Default locale: en_BR, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.6", arch: "x86_64", family: "mac"
```

Patch based on https://github.com/projectlombok/lombok/issues/2681#issuecomment-808895956

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,19 @@
         <configuration>
           <source>11</source>
           <target>11</target>
+          <fork>true</fork>
+          <compilerArgs>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED</arg>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED</arg>
+            <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED</arg>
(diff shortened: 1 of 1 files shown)


shroffk:
@hhslepicka  Thanks for starting to test with jdk16 :) unlike eclipse I like how we can easily be on top of the latest releases of java

Does this issue affect libraries other than lombook? Also, do we need to create a profile for openJDK16 or adding these arguments does not affect builds with older jdks

kasemir:
Will this force a move to JDK 16?
The source code fundamentally works with any JDK between 11 and 16.
.. but JDK 16 is more picky, see runtime warning about JFX module access #1882.

hhslepicka:
@shroffk I only noticed the errors with lombok. According to TravisCI it looks like JDK11 is passing the tests and building properly so it doesn't seem to affect. So I don't think a new profile is required.

@kasemir it will not force a move to JDK 16. TravisCI executed the tests with 11 and all seems fine.
I did not change the minimum compiler version so it should work fine.

georgweiss:
I introduced Lombok as it provides easy-to-use annotations that eliminate the need of a lot of boiler plate code. If it causes more trouble than we can endure then I can remove it.

shroffk:
@hhslepicka thanks... I think this should be fine then.
I am fine with merging this.

@georgweiss I don't think we need to remove this,
but at some point either lombook will have to fix itself or we might want to consider removing it. From the issue that hugo linked I see that in future releases of java even these work arounds will stop working.

PR #1890, opened 2021-04-27
Bug fix attach screen shot when creating log entry from OPI

georgweiss:
Also fixes a bug when using cached credentials: submit button not enabled even though all mandatory fields are valid.


shroffk (approved):

shroffk:
What is the default behaviour of the credentials? I am guessing that by default they do not expire

georgweiss:
Correct, they are maintained in local keystore so even after restart the credential fields are populated. But the flag used to check validity of the new log entry was not updated when credentials are populated from keystore (=cache).

PR #1892, opened 2021-04-27
Some Save&Restore update

geniejhang:
Hi @georgweiss . Please review the following changes.

- When there is '%' character in comment and/or name, it's recognized as utf-8 prefix in MaridDB.
  Replacing it with utf-8 code.

- Some mandatory field's placeholders don't stand out with gray font.
  Changed to red.

- Unacceptable case, but when there're two PVs with the same name in a saveset while one is readonly and the other is restrorable,
  app handles the case correctly.

- Addition of context menu for toggling readonly entry to restorable and vice versa.
  This enables users to temporarily toggle some entry for accidental restoring some critical PVs.

- Replacing GPClient with Phoebus PV class.
  By this, readonly PVs have writable connection while blocked restoring programatically.

### DIFF
--- a/app/save-and-restore/model/pom.xml
+++ b/app/save-and-restore/model/pom.xml
@@ -70,7 +70,7 @@
 			<version>${junit.version}</version>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+    </dependencies>
 
 
 	<build>
--- a/app/save-and-restore/pom.xml
+++ b/app/save-and-restore/pom.xml
@@ -7,9 +7,6 @@
     <artifactId>app</artifactId>
     <version>4.6.6-SNAPSHOT</version>
   </parent>
-  <properties>
-    <gpclient.version>1.0.5</gpclient.version>
-  </properties>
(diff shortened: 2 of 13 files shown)


georgweiss:
Thanks for replacing gpclient, it's been on my backlog for some time already :-)
Please remove commented-out code related to gpclient, and please remove gpclient dependencies in all affected pom files.

As for the text formatting... In the logbook client editor we use red color for the label for mandatory input, see screen shot. When user types in the input field it switches to black. This in my view a more gentle way to request user to specify mandatory data, and I'd prefer if you can change in save&restore so the UI aligns to what we use in the logbook client.

Screenshot 2021-04-28 at 07 15 11

geniejhang:
@georgweiss Thanks for the comment.
All you asked have been done!

georgweiss:
@geniejhang, I'm afraid the migration away from gpclient has broken support for (numeric) arrays in save&restore. I am currently fixing this, but I'd like you to know to save us from double work.

PR #1894, opened 2021-04-29
Fix for array support in save&restore

georgweiss:
Removal of gpclient has broken support for arrays in save&restore, this is a fix.

Also added support for string arrays.


georgweiss:
@shroffk, can you please review this?

georgweiss:
@shroffk, can you please review?

PR #1902, opened 2021-05-05
modernizing the elog module

kingspride:
I've made some modernizations to the elog module.
(replaced deprecated stuff and removed quirks like adding 1900 to the current date...)

hope this helps maintaining cleaner code...

I also replaced a line where the attachments would be downloaded to /tmp,
and replaced it with the correct way of obtaining the system specific temp path.

I also noticed that this occurs in other files aswell, you might wanna change it.


app/logbook/elog/src/main/java/org/phoebus/elog/api/ElogApi.java
```diff
@@ -385,9 +382,7 @@ public File getAttachment( String filename ) throws LogbookException {
           types.add( t );
         }
       }
-    } catch (XPathExpressionException e) {
-      throw new LogbookException( e.getMessage() );
-    } catch (ParserConfigurationException e) {
+    } catch (XPathExpressionException | ParserConfigurationException e) {
       throw new LogbookException( e.getMessage() );
```
kasemir, line 386: Instead of `throw new OtherException(original.getMessage())`, I would suggest `throw new OtherException("Stuff happened", original)`.
For one, `original.getMessage()` may be `null`, so you get nothing useful.
Additionally, passing the 'cause' up in the wrapped message allows for "Caused by: ..." and the full stack trace to appear in the, well, stack trace.

kingspride: You are right. This needs to be adapted.

kasemir: `hrow new LogbookException( "could not parse the elog response", e )` is good, thanks!

shroffk:
@kingspride is this ready to merge now?

kingspride:
I think so, yes.

Issue #1909, opened 2021-05-17
Olog search query adds desc=* after restart


georgweiss:
Searching for an Olog entry where body text is empty does not result in any hits unless desc=* is removed (bug? expected behavior?). Removal desc=* from the query string is persisted in the memento file, but is added in code, resulting in the unexpected behavior where the user modified query string is not the same as when Phoebus was shot down.

shroffk:
So there are 2 things in play here:
1. I think the elastic query mapped to desc=* expects something to be present. 
match anything != match anything including nothing
If we want to update this behaviour we will have to update the service...I fell this should be resolved on the client side which sends the query.

2. The logbook applications treats empty or null search string as a special case where the default search defined in preferences is used. We could change the default search query preference from desc=* to * or something similar

georgweiss:
Is a change in the preference enough, see LogEntryTableViewController:
```
searchParameters = FXCollections.observableHashMap();
        searchParameters.put(Keys.SEARCH, "*");
        searchParameters.put(Keys.STARTTIME, TimeParser.format(java.time.Duration.ofHours(8)));
        searchParameters.put(Keys.ENDTIME, TimeParser.format(java.time.Duration.ZERO));

```

georgweiss:
The AdvancedSearchViewController is also a player, I think.

shroffk:
The above commit ensures that the table and calendar are initialized via the preferences.

shroffk:
@georgweiss can we close this?

PR #1912, opened 2021-05-18
#1909 initialize logbook table and calendar using preferences

shroffk:
#1909

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCalenderViewController.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCalenderViewController.java
@@ -168,11 +168,13 @@ public class LogEntryCalenderViewController extends LogbookSearchController {
         AnchorPane.setRightAnchor(agenda, 6.0);
         agendaPane.getChildren().add(agenda);
 
-        searchParameters = FXCollections.<Keys, String>observableHashMap();
-        searchParameters.put(Keys.SEARCH, "*");
-        searchParameters.put(Keys.STARTTIME, TimeParser.format(java.time.Duration.ofHours(8)));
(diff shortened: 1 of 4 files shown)


georgweiss (commented):
This does not actually solve my issue: if user removes desc=* from search string, it is put back after restart. It's difficult to explain why other modifications to the search string are maintained between restarts, but not removal of desc=*.

shroffk:
> This does not actually solve my issue: if user removes desc=* from search string, it is put back after restart. It's difficult to explain why other modifications to the search string are maintained between restarts, but not removal of desc=*.

I can't recreate this... 
removing desc=* and only keeping the start and end time in the query.. upon restart results in only the start and end being present.

georgweiss:
> > This does not actually solve my issue: if user removes desc=* from search string, it is put back after restart. It's difficult to explain why other modifications to the search string are maintained between restarts, but not removal of desc=*.
> 
> I can't recreate this...
> removing desc=* and only keeping the start and end time in the query.. upon restart results in only the start and end being present.

Using a default search string that includes desc=* will add it. But maybe your point is to remove desc=* from the default in the preferences?

shroffk:
removing from the default preference is one part.

but if you open the log table... changed the query to a non empty query... then exit/restart phoebus... the modified query should be restored from momentos

georgweiss:
> removing from the default preference is one part.
> 
> but if you open the log table... changed the query to a non empty query... then exit/restart phoebus... the modified query should be restored from momentos

Sorry, that is not the case in my test. The default query is not overridden, so desc=*&start=12 hours&end=now. If I change 12 hours -> 120 hours, after restart I do indeed see 120 hours, but desc=* is still there even if I remove that portion before restart.

shroffk:
Looking into this... keep this PR open

shroffk:
Ok, 
I added a caching feature between relaunches of the log table. This is different from the momento which stores the last shutdown state.

georgweiss:
Thanks, works as I expect now.

PR #1914, opened 2021-05-24
Display Editor: small enhancements from users' input

pavel-ch:
Display Editor: Gap Distribution makes sense from 2 widgets selected
Display Editor: Widget move/resize in Editor pane: allow a 10x bigger step using arrow Keys with Shortcut key
In the spirit of #679


kasemir:
Thanks!

Issue #1921, opened 2021-06-07
Alarm: State topic remains when an alarm is deleted


sasaki77:
When I delete an alarm from Alarm tree GUI or `-import` command line option, Config Topic is deleted with "tombstone".
However, State Topic remains as it is.
Following is an example of my test.

```json
config:/Accelerator/TEST/Alarm00010: {"user":"sasaki","host":"sasaki-VirtualBox","description":"Alarm00010"}
state:/Accelerator/TEST/Alarm00010: {"severity":"OK","message":"OK","value":"0.0","time":{"seconds":1623029227,"nano":154814561},"current_severity":"OK","current_message":"NO_ALARM"}
config:/Accelerator/TEST/Alarm00010: {"user":"sasaki","host":"sasaki-VirtualBox","delete":"Deleting"}
config:/Accelerator/TEST/Alarm00010: null
```

The alarm server seems to ignore State topics. On the other hand, alarm client seems to create a new alarm tree item from State topic if the item is not created yet.
I know alarm client deletes the alarm item from alarm tree when it found "tombstone" record and alarm tree should be consistent eventually while "tombstone" record remain.
But alarm client might have alarm item for deleted alarm until it find "tombstone" record.

I think we can simply ignore State topic for deleted alarm if we also mark State topic as "tombstone".

Are there any reason to keep State topic when we delete an alarm?

kasemir:
The `config:` and `state:` messages are sent via the same Kafka topic, so their order is guaranteed.
When an item is deleted, the last message for an item should be the `null` "tombstone" record, similar to your example:
```
# One or more state updates
state:/Accelerator/Top/sim:\/\/sine: {"severity":"MAJOR",...}
# Deletion marker and "tombstone"
config:/Accelerator/Top/sim:\/\/sine: {"user":"...","host":"...","delete":"Deleting"}
config:/Accelerator/Top/sim:\/\/sine: null
```

The end result should be that the alarm server no longer tracks that PV, i.e., it is no longer listed `ls -all` on the alarm server console, and the item is also no longer shown in the alarm tree GUI.
I hope you can confirm that everything "works" from an end-user perspective.

But I agree that you are correct in that the "state:..." message remains in the Kafka topic, so when a client starts up, it will see the last `state:..` update, maybe followed by the "delete" message if that has not, yet, been compacted, and finally the  "null" tombstone. As a result, a client might briefly create and then right away destroy the item.
This could be optimized by deleting like this:

```
[...]

PR #1922, opened 2021-06-07
Alarms: Create "tombstone" Kafka "state:" message for deleted PV

kasemir:
While the "config:" topic contains a "delete" info message followed by a
tombstone which deletes the PV, the last "state:" message remains in the
Kafka stream. The combination of non-null "state:" followed by null
"config:" will briefly create and then delete the PV in GUI clients.
Adding the null "state:" message optimizes this.

#1921

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/client/AlarmClient.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/client/AlarmClient.java
@@ -292,9 +292,15 @@ public class AlarmClient
             else if (type.equals(AlarmSystem.STATE_PREFIX))
             {   // State update
                 if (json == null)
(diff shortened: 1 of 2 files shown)


kasemir:
More info on why this is necessary..

When an alarm tree item is deleted, the log would look like this:
```
.. state:/path/to/some/PV ...alarm_severity...
.. config:/path/to/some/PV .... user .. deleted
.. config:/path/to/some/PV null
```

After compaction, this turns into
```
.. state:/path/to/some/PV ...alarm_severity...
```

A client might not have any configuration detail of the PV, but it shows the current state.

After this PR, a deletion will look like this:

```
.. state:/path/to/some/PV ...alarm_severity...
.. config:/path/to/some/PV .... user .. deleted
.. config:/path/to/some/PV null
.. state:/path/to/some/PV null
```

which compacts into nothing.

Issue #1934, opened 2021-06-17
Alarm logger: Pushing messages into Elasticsearch is a little bit slow


sasaki77:
I'm testing alarm server and alarm logger with 50,000 PVs. All PVs are running on one IOC.
Once the IOC is stopped, 50,000 state change messages are issued.
Alarm logger receives the messages and pushes them into Elasticsearch, but it needs a lot of time to complete pushing all messages.

I configured Elasticsearch with Ingest pipelines to measure a time to complete it as following. We can see the time when the document is stored with this.

```bash
curl -X PUT "http://localhost:9200/_ingest/pipeline/indexed_at?pretty" -H 'Content-Type: application/json' -d '
{
  "description": "Adds indexed_at timestamp to documents",
  "processors": [
    {
      "set": {
        "field": "_source.indexed_at",
        "value": "{{_ingest.timestamp}}"
      }
    }
  ]
}
'
```

Following is the last document stored into Elasticsearch when the 50,000 state change messages are issued. `indexed_at` field indicates that it takes about 7 minutes to store the document.

```bash
$ curl -X GET 'http://localhost:9200/accelerator_alarms_state*/_search?q=*&format=json&pretty&sort=indexed_at:desc&from=0&size=1'
...
{
  "_index" : "accelerator_alarms_state_2021-05-31",
  "_type" : "alarm",
  "_id" : "siVKCXoBi0cjCWSAtncP",
  "_score" : null,
  "_source" : {
    "severity" : "UNDEFINED",
    "pv" : "Alarm48096",
    "message" : "Disconnected",
    "notify" : "true",
    "current_message" : "Disconnected",
    "mode" : null,
    "message_time" : "2021-06-14 06:44:50.660",
    "current_severity" : "UNDEFINED",
    "indexed_at" : "2021-06-14T06:52:09.103327Z",
    "latch" : "false",
    "time" : "2021-06-14 06:44:50.659",
    "config" : "state:/Accelerator/TEST/Alarm48096",
    "value" : ""
  },
  "sort" : [
    1623653529103
  ]
}
...
```

I don't think this is caused by settings of Elasticsearch because I confirmed Logstash can smoothly push all 50,000 messages into Elasticsearch.

Current implementation creates IndexRequest for Elasticsearch per message, right?
I guess we can improve alarm logger with using Elasticsearch Bulk API, but I have no idea to use Bulk API with Kafka stream API.

[...]

PR #1938, opened 2021-06-21
Issue #1934

shroffk:
@kasemir I have also included the change to spli pv names by space and semicolon only.
I tested a few examples and both server and client worked as expected.

Would it be ok to merge this?


kasemir:
OK to split only on space and semicolon, not comma.
Is there a tool tip that needs to be updated because it still suggests comma?

shroffk:
nope, 
the tooltip only mentions spaces
```
                    message.setText("Enter one or more PV names, separated by spaces");

kasemir:
Perfect!

PR #1940, opened 2021-06-28
Fix for file save deadlock

georgweiss:
When user attempts to close application with unsaved edits (display, data browser plot or scan file), the application will prompt user to save, cancel or ignore (yes/cancel/no). If user chooses to save (=yes), an existing file will be saved with the edited content. 

There is however a special case: if data browser plot was triggered from a context menu there is no file to save, so the user must be prompted using a file chooser dialog. Same goes for new scan: user may edit in UI, but no file is associated with the new content.

This special case will as far as I can determine cause a deadlock: the file chooser must be triggered from the JavaFX application thread (see SaveAsDialog#promptForFile), but at this point the thread is in a wait state, see DockItem#prepareToClose.

This PR attempts to fix the deadlock. Verified on display, data browser and scan.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java
+++ b/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java
@@ -521,8 +521,8 @@ public class DockItem extends Tab
     /** Register check for closing the tab
      *
      *  @param ok_to_close Will be called when tab prepares to close.
-     *                     Will be invoked on the UI thread, so it may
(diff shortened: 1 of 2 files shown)


kasemir (approved):
OK, but maybe update the java doc for `DockItem#addCloseCheck`.
It states
```Will be invoked on the UI thread, so it may prompt for "Do you want to save?"```
Should now say ```Will be invoked in background thread, so must not interact with UI without changing to platform thread```

PR #1944, opened 2021-07-01
Update index.rst

thomasms:
Correcting grammatical mistake in alarm-config-logger documentation.


kasemir:
Thanks!
In principle, we let the author of the respective code check if the change makes sense, and await the automated build to verify if stuff still compiles and unit tests still pass, but I think in this case we can assume adding an 's' won't cause too much havoc.

PR #1948, opened 2021-07-04
Adding (missing) 'properties' as logbook search query parameter name

georgweiss:
Currently query parsing and subsequent update of UI throws NPE if "properties" is used in search and used from persisted query upon app restart.

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookQueryUtil.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookQueryUtil.java
@@ -26,7 +26,8 @@ public class LogbookQueryUtil {
         ENDTIME("end"),
         AUTHOR("owner"),
         TITLE("title"),
-        LEVEL("level");
+        LEVEL("level"),
+        PROPERTIES("properties");
 
         // The human readable name of the query key
         private final String name;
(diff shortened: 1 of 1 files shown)


shroffk:
I think the idea was that like channelfinder, if the search parameter does not match the predefined ones then it should be interpreted as property

`desc=*&shift.id=1234`

where shift is the name of the property, id is the name of the attribute and then after "=" there is a search pattern

georgweiss:
Such a search query results in a NPE in the same manner: when app is restarted and cached query is considered.

Also, performing a search using this strategy results in unexpected log entries. Maybe it's because we use property and attribute names containing blank chars?

shroffk:
https://github.com/Olog/phoebus-olog/blob/97834eafbbd6877fafb1f1e5791e301ee9b5f9c9/src/main/java/gov/bnl/olog/LogSearchUtil.java#L162

I guess I was wrong.
Should we add an example or tooltip on how the properties search sting should look?

`properties=shift.id.1234`

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookQueryUtil.java
```diff
@@ -41,6 +42,7 @@
             lookupTable.put("owner", Keys.AUTHOR);
             lookupTable.put("title", Keys.TITLE);
             lookupTable.put("level", Keys.LEVEL);
+            lookupTable.put("properties", PROPERTIES);
```
shroffk, line 45: well Keys.Properties would look consistent but this should be fine :)

shroffk (approved):

georgweiss:
It's already in the docs. Tooltips should be short, right?

shroffk:
ideally yes... but no one reads the docs so we have a catch 22
the tool tip for the search text input could have all key words

georgweiss:
Maybe we should consider an easy-to-use mechanism to launch a particular section of the documentation....

PR #1951, opened 2021-07-07
Add stand-alone image viewer for Olog image attachments

georgweiss:
A simple viewer is launched when user clicks an image attachment from the list of log entry attachments. This is useful if user embeds scaled image, but then needs to view it in full resolution.


kasemir:
A related idea for attachments would be to simply "open" them with whatever tool is registered to handle the file type.
Your preview panel could register as a handler for image files, and in addition  https://control-system-studio.readthedocs.io/en/latest/preference_properties.html#framework-workbench can be used to register various external image viewers.

georgweiss:
Not sure what you mean... The preview panel shows a scaled image attachment, while the log body may embed a scaled image.  The idea was to easily launch a simple dialog that shows the unscaled image, no need to rely on external viewers.

kasemir:
You implemented an image viewer, for only the olog viewer.
That's great.
But if you register that as a tool for handling *.jpg, *.png, ... it could be used to open images in general, including from the file browser, or from the "-resource /path/to/image" command line option.
.. and users who prefer a more sophisticated (external) image viewer, could switch to that.

georgweiss:
OK, now I get it. Will update to make it bit more generic and also move it to a core UI module.

kasemir:
I'd say you can still go ahead and merge this PR. Turning the image viewer into its own 'app' could be a second step.

georgweiss:
Fine, I'm working on an image viewer app, but this will do for the time being.

Issue #1908, opened 2021-05-13
Save Restore service API documentation


shroffk:
It would be useful to have an api documentation for the save restove service like the one we have for olog and channelfinder
https://olog-es.readthedocs.io/en/latest/#rest-api

georgweiss:
I agree.

PR #1952, opened 2021-07-09
#1908 Add basic documentation for save restore REST API

shroffk:


georgweiss (commented):
Line 76:
"A Save set consists of a list ov pvs along with their values at a particular instant in time". My definition of "save set" is basically equivalent with a configuration, i.e.
list of PVs. The data structure holding PV values at a particular instant in time is "snapshot".

Line 151:
"Get a children" should be "Get children" or "Get child nodes"

Line 204:
"The nodeType can be used to specify if we want to create a new folder or a new snapshot configuration". Here "snapshot" can be omitted, or should be replaced with "save set".

georgweiss (approved):

PR #1955, opened 2021-07-09
Annunciator: Only count enabled alarms

tanviash:
Currently, annunciator reports both enabled and disabled PVs in its 'active alarm count' which can be confusing at times since alarm table reports only enabled PVs as active alarms. This PR updates annunciator to count only enabled alarms as active.


kasemir (approved):

kasemir:
Done with the EPICS meeting, it seems...
Thanks!

PR #1963, opened 2021-07-28
Display Editor: better visual recognition of widgets with macros usin…

pavel-ch:
…g an additional icon in the Editor Widget Tree
See #679


kasemir:
Thanks!

PR #1977, opened 2021-08-10
Template/Instance widget

kasemir:
Screen Shot 2021-08-10 at 1 56 28 PM


kasemir:
Screen Shot 2021-08-12 at 8 01 20 AM

PR #1983, opened 2021-08-17
Adding jpackage documentation

georgweiss:
As agreed. Verification on Linux would be valuable. Tried rpm package on CentOS 7, but that failed using both Java 11 and Java 16 for the runtime. Have no other Linux environment for test.

### DIFF
--- a/phoebus-product/README.md
+++ b/phoebus-product/README.md
@@ -55,8 +55,10 @@ The following use cases have been verified:
  
  ##### `jpackage` build step 1
  `jpackage --name <app_name> --input unzipped --type app-image --main-jar product-<version>-<os>.jar
- --icon unzipped/<OS specific icon file> --java-options -Dprism.lcdtext=false --java-options  --java-options
+ --icon unzipped/<OS specific icon file> [--java-options -Dprism.lcdtext=false] --java-options  --java-options
  -Dcom.sun.webkit.useHTTP2Loader=false --runtime-image <jdk_root>`
+ 
+NOTE: the `--java-options -Dprism.lcdtext=false` portion is for MacOS only.
  
 Additional Java options are added using `--java-options <my option>`.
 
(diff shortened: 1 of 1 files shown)


shroffk (approved):

lorenzo-gomez-windhover:
This really cool! Can verify it works on `Ubuntu 18.04.5 LTS`.

georgweiss:
@lorenzo-gomez-windhover, thanks for testing. Did you test rpm or deb? Or both? Can you please update the list of verified use cases?

lorenzo-gomez-windhover:
I should clarify. So I ran the jpackage commands as instructed in the README, and it packages both rpm and deb. It installs it to `/opt/product-name`, which is great. The only problem is when I try to execute `/opt/product-name/bin/exec_name`, it gives me the following:
```
Failed to find JVM in "/opt/commander/lib/runtime" directory.
```

Now I can still run the product by going into `/opt/commander/lib/app` and running `phoebus.sh`, but I was hoping I could just execute the binary under `bin`.

I checked `/opt/commander/lib/runtime/bin` and it has all of the java executables. 

Any idea what the problem might be?

Should I open up an issue about this?

georgweiss:
@lorenzo-gomez-windhover , if the app does not run from installed package then we need to state that we do not know how to make this work on Linux.

I have actually no idea why it fails the way it does in your test. Actually I got the same error message on CentOS7. `jpackage` simply copies the contents of the JDK installation to the runtime directory, so it's all there. Building the installers on MacOS and Windows in the same manner results in a fully functional application.

Tried googling this, but did not find anything useful. 

Please go ahead and open an issue ticket, it's useful for the sake of information sharing and follow up.

georgweiss:
@lorenzo-gomez-windhover, by the way, did your test fail both for rpm and deb? I'm asking as I cannot build deb on my test system, only rpm.

lorenzo-gomez-windhover:
Yes, the behavior is the same for rpm and deb.

PR #2000, opened 2021-08-30
Fix for legacy olog client

tanviash:
This PR is to provide fix for search issues seen in the legacy (non-ES based) olog client.


shroffk (approved):
Thank you.
I am glad that the tests are fixed too

PR #1996, opened 2021-08-27
Split dock unit test

lorenzo-gomez-windhover:
Hi there,

Hope you are all doing well.

These unit tests use TestFX's Robot API to test the behavior of `DockPane` and `SplitDock` classes. Specifically these unit tests test the behavior discussed in  #1986 and  #1991.

I know GUI testing can be tricky because it might very well behave differently on different platforms, but I tried my best to write these tests in a platform-independent way. By all means try it in other platforms as I ran the test only on `Ubuntu 18.04.5 LTS`. It is possible I missed something, but hopefully these tests are reliable enough that we can trust them when it comes to GUI testing.

Thanks
Lorenzo

PR #2002, opened 2021-09-03
Retake on closing empty windows

georgweiss:
This is an attempt to put the logic where it has a better chance to consider all cases with split panes.


kasemir (approved):

lorenzo-gomez-windhover:
Can verify that the tests in #1996 pass with these changes.
Thanks for fixing this!

PR #2003, opened 2021-09-06
Support months and years in log search query

georgweiss:
Using specifiers "months" and "years" in the log search query will trigger exceptions in java.time.Instant.

This is an attempt to provide basic support for these date/time specifiers. Note that the added implementation makes a few assumptions and simplifications in order to calculate the time based on "days":
1) One year is on average 365.25 days when leap years are considered.
2) One month is on average 30.4375 days when leap years are considered.
3) A search string like "2 years 8 months 4 days 12 hours" will consider years, months and days, but ignore hours.


georgweiss:
@shroffk, can you please review?

PR #2025, opened 2021-10-06
Update example for python3

kasemir:
Based on Sinisa's tech-talk message https://epics.anl.gov/tech-talk/2021/msg01776.php


kasemir:
This is a python-based PVA server. The C python that executes it, outside of the GUI, can be python 2 or 3.

PR #2027, opened 2021-10-06
Add button to toggle sort order in log entry table view

georgweiss:
As per user request: make it possible to toggle sort order of log entries in the log table view, e.g. from default descending to ascending (on created date) or vice versa, see screen shots. The tree is re-rendered based on the entries already available, i.e. a new search is NOT performed.

Also some cosmetic updates to the search tool view.

Screenshot 2021-10-06 at 16 39 56
Screenshot 2021-10-06 at 16 40 05


shroffk:
Ok, I will wait for a reply.
We can add `sort` parameter to the service which would allows searching in ascending or descending order... I will create an Issue ticket

PR #2032, opened 2021-10-14
Olog table view sort order backed by new search

georgweiss:
To cover for cases where search result is truncated due to limit of number of hits, a request to sort log entries in reversed sort order is backed by new search in Olog service. Also changed UI a bit to clarify that selecting sort order performs new search.

This functionality depends on Olog server commit bd85619c56121be148d3a45927e391b881f61edc or later.


georgweiss:
Would it be better to use upper case, i.e. sort=DESC?

georgweiss:
Or to spell it out: sort=descending

kasemir:
In SQL, case doesn't matter and the keyword is 'desc'. So sort=desc is somewhat intuitive for software people. To end users, sort=up or sort=down might be as natural.

PR #2039, opened 2021-10-20
Lazy loading of log attachments

georgweiss:
A search for log entries should not retrieve attachments from remote service as this time consuming and puts load on network and server. Moreover, user might not be interested in all attachments anyway.

This PR defers retrieval of attachment data: only when a log entry is selected the attachments are downloaded (asynchronously) and added to the UI.


georgweiss:
@shroffk , please review. This is pretty important as we're now using Olog from the control room and the number of entries is increasing steadily. We should avoid a situation where the search function is perceived as "slow" due to the eager loading of attachments. As the web client loads attachments lazily, we also have an asymmetry that I would prefer to not explain...

PR #2047, opened 2021-10-29
Add tags to alarm configuration

jbellister-slac:
## Description
This PR will allow the inclusion of tags into alarm configurations. Tags are arbitrary tag/value pairs of information that can help operators better handle alarms, but that don't currently fit well elsewhere in the configuration. They can also be helpful when searching for a group of similarly tagged alarms, that would be difficult to group together otherwise. Tags are an optional feature, alarms will continue to work just fine without any set if not needed.

## Screenshots
alarmtagsfull

### DIFF
--- a/app/alarm/Readme.md
+++ b/app/alarm/Readme.md
@@ -239,7 +239,8 @@ The full config topic JSON format for a alarm tree leaf:
         "guidance": [{"title": String, "details": String}],
         "displays": [{"title": String, "details": String}],
         "commands": [{"title": String, "details": String}],
-        "actions":  [{"title": String, "details": String}]
+        "actions":  [{"title": String, "details": String}],
+        "tags":     [{"title": String, "details": String}]"
     }
 
 The full config topic JSON format for a alarm tree node:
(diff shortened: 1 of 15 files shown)


shroffk:
Sorry for the late response.

I think with the alarm system I like being conservative and keeping it as thin as possible.

Form what I understand, the main goal of tags is 
1. to present users with some additional text information 
2. have additional search keywords in the alarm logs

Personally, I link the idea of using something like the directory service for managing tags associated with PVs...like Kay mentioned you can access this info in CS-Studio with channelinfo, but the services is a pure stand alone service and you should be able to include it with any clients.

Regarding logging, we could update the alarm-logger so that we can add additional modules which combine Alarm messages with data from other sources. One of these sources could be tags from ChannelFinder.

kasemir:
> Regarding logging, we could update the alarm-logger so that we can add additional modules which combine Alarm messages with data from other sources. One of these sources could be tags from ChannelFinder.

Just as you say for the alarm system/server, why not keep the alarm logger minimal as well?

A site-specific _viewer_ of the alarm log could merge information from other sources. For instance, add channel finder tags for a channel whenever it hits an alarm log message with PV.

shroffk:
Fair point
I think there is always this balancing act of where to introduce complexity...

If we were going with expanding the Alarm Logger I was thinking to using SPI to allow for the basic version being almost identical to its current state.

Now, arguing against myself... integrating data from various services in the client has the benefit of allowing users to decide what and how much data they want to merge

jbellister-slac:
Thank you both for the feedback. Keeping things minimal and simple where possible is definitely a good thing. I will go ahead and close out this PR for now. Appreciate the suggestions, and will take a look at channel finder.

PR #2057, opened 2021-11-16
Fix for Olog causing application to hang

georgweiss:
Under certain conditions, Phoebus will become unresponsive for a long time. This is caused by the Olog log entry viewer trying to render a preview of a large (video) attachment.

This is a fix such that:

- Only image files are supported for preview.
- Double click on .bob or .plt attachment will launch corresponding Phoebus app.
- If external app is configured for (for instance) text files, that app is launched upon double click.

Also changed: Olog documentation (updated screen shots, clarifications)


shroffk:
Was this closed intentionally or by accident
The fixes and improvements are very useful

georgweiss:
I closed it since I want to add a dialog showing if an attachment cannot launch an external viewer. This is needed to avoid that some attachments can be double-clicked, while others can not.

Will re-open PR soon.

Issue #2074, opened 2021-11-29
The alarm server test docker compose recipe 


shroffk:
The docker compose recipe is a great way to set up kafka and zookeeper for testing and development.
The current recipe seems like it is out dated and fails to start the kafka image ( connections not working )

Issue #2072, opened 2021-11-29
alarm logger 'create_alarm_template.sh' update for elasticsearch 7


kasemir:
Similar to https://github.com/ChannelFinder/ChannelFinderService/issues/20, running https://github.com/ControlSystemStudio/phoebus/blob/master/services/alarm-logger/startup/create_alarm_template.sh will fail for elasticsearch 7 with an error message "Root mapping definition has unsupported parameters".

Unclear about the best long-term solution, but a workaround seems to be adding `?include_type_name=true` to the URLs in there.

Issue #2069, opened 2021-11-24
alarm history table app upgrades


shroffk:
1. add color coding of alarm severity similar to the old alarm message table
2. make the displayed columns configurable
3. add a column showing how much time has passes / time delta
4. update the search based on changes to the elastic templates ( messages, etc... are moved from type text to type keyword)

Issue #2068, opened 2021-11-24
alarm message logging service updates


shroffk:
1. There are fields in the elastic template marked as text with additional mapping to keywords. 
```
"message" : {
            "type" : "text",
            "fields": {
              "keyword": { 
                "type": "keyword"
              }
            }
```
We could update the type directly to "keywords"
@tanviash would this be ok?

2. The statup scripts can be incorporated in the service itself, upon startup if the elastic templates are missing then they are created.

3. The REST api should be expanded so that clients don't have to make direct elastic calls - we will have to update elastic to 7? maybe 8? soon and it might be nice to avoid the situation where we have to update the client every time.

tanviash:
Hi Kunal, yes it is okay to go ahead. Thanks for the update. I'll keep an eye on your changes so I can test it out on our test setup.

PR #2075, opened 2021-11-29
WIP: address a group of alarm logging service and application issues

shroffk:
#2068 
#2072 
#2074 
Resolved

working on matching changes for #2069


shroffk:
Additionally,
I also standardized the elastic client dependency - all services and client now are using 6.8.4
Using a single property, `es.version` ( like a lot of other dependency groups )

kasemir (commented):
While this updates `create_alarm_index.sh` to use `?include_type_name=true` in the URLs for Elastic version 7, there's now also code related to `alarms_*_template.json` which has the logger automagically install the templates.
Should the `create_alarm_index.sh` just be removed, it's no longer necessary? And update the README.md to also not mention it?

shroffk:
@kasemir I am fine with deleting the startup scripts.. .I just wasn't sure if anyone is using them in the deployment process.
I will poll people on Wednesday and then delete it, along with updating the readme.md

I tested the new programmatic way of creating the templates on elastic 6.8.4 and elastic 7.15.2.. it created the templates on both versions.

@tanviash @georgweiss are you guys fine with simply deleting the create scripts?

tanviash:
> @tanviash @georgweiss are you guys fine with simply deleting the create scripts?
Yes, deleting it is fine by me.

georgweiss (commented):
If the scripts are not need to setup the service, then I'm fine with removing them.

shroffk:
- Colour coding the severity columns
 - Adding a new column which displays time deltas
 - Enable selection of column to be displayed.

image

georgweiss (commented):
Columns selection will be appreciated. Propose to add the same in the alarm table view. 
Is the selection persisted in memento and loaded on initialization?

PR #2076, opened 2021-12-01
New app to handle credentials

georgweiss:
As per user request:

I've added a new app - "Credentials Management" - providing means to clear saved credentials. In order to support more than one use case I've also introduced the notion of "scoped authentication". The scope is defined by a string and used to prefix username and password entries in the key store that backs SecureStore.

**It should be noted** that the Java key store API will ignore case and persist entries using **lower case**, i.e. an alias/key named "UPPER" will be persisted as "upper". This peculiar behavior is implementation dependent, but my conclusion from available documentation and various posts out there is to not rely on case sensitivity at all. In other words, the scope identifier should always be lower case.

The Credentials Management app will list all currently saved authentication entries ending in "username", i.e. including "non-scoped" credentials and present a button to delete (or logout) such entries. User may also choose to delete all authentication entries. To be noted: a key store entry is considered to be an authentication entry only if it ends in "username", and if a corresponding "password" entry exists.

The Olog(ES) app will use the scoped authentication "logbook", i.e. saved credential entries will be persisted in the key store under aliases "logbook.username" and "logbook.password". Further, when the log entry editor is launched, credentials saved under that scope are used, if they exist. The side effect is that any previously saved Olog credentials will not be considered in scope for Olog. This presents a minor migration issue, but in practice a user will be prompted only once after Phoebus update. If on the other hand credentials caching is not used, user is of course prompted for all new log entries.

The Credentials Management app is launched from a toolbar icon, there is no menu entry (yet). It presents itself as a modal dialog. The icon will most probably be changed as soon as I find something better.

I also plan to extend the functionality such that the Credentials Management UI would also provide means to log in to a service. 

Screenshot 2021-12-01 at 10 42 16


kasemir (commented):
Functionality is fine, but I really dislike the toolbar button.
It's far too visible for such a specialized functionality.
A menu entry would be much preferred.
While technically a little harder, using the existing user name indicator in the status bar would be neat and somewhat natural. For now, you just see the current user ID in the status bar (https://github.com/ControlSystemStudio/phoebus/blob/a7079426a66ef3c53fb5565c4b6f62eb2c1b357a/core/ui/src/main/java/org/phoebus/ui/statusbar/StatusBar.java#L31-L33) Would be neat to turn that into a button, and when clicked, it opens the credentials manager, so you can see and clear them.

shroffk:
+1 for converting the  current user ID in the status bar into a button... it would match cs-studio and I think aesthetically nice too

georgweiss:
While I do not mind to move the button I do not agree that the current user indicator is the "natural" place for it. In my view there is no obvious relation between the current user identity on one hand, and credentials to (remote) services on the other. While in practice the same account may be used, at ESS we use a group account to login to the workstation, and personal accounts for logbook entries. 

However, comparing with Eclipse CS Studio I suspect you actually propose to use the Credentials Management app to also show the Phoebus "internal" authorizations, right? That's fine with me, but then we need to agree that there is a distinction - conceptually as well as presentation-wise - between Phoebus authorizations and credentials for external services. If we can agree on this then I can update as proposed and implied.

kasemir:
The Eclipse-based version had a button for the status bar user indicator which, when pressed, showed authorization info. But a menu entry is fine as well. Anything but a toolbar button.

kasemir:
> we use a group account to login to the workstation, and personal accounts for logbook entries.

We have the use case as well, the application is running as a generic "beamline user". For that scenario, I suggest to _not_ save any credentials:
```
org.phoebus.logbook.ui//save_credentials=false
```

We only store credentials for office users who run the app on their own computer, and then the convenience of keeping the info is appreciated. But for a shared setting like a control room, it seems wrong.

georgweiss:
Using menu button is less convenient as it implies additional mouse gestures.
Not saving credentials would not be accepted as user would need to provide them for each new log entry. With a dedicated UI to log out we at least offer a way to minimize issues with credentials being used by "wrong" user.

georgweiss:
Updated as discussed.
NOTE: the update also includes an API to support explicit login. Implementations of ServiceAuthenticationProvider can be contributed over SPI. Olog is one such example.
Help content summarizing the features added to ``core/ui/doc/index.rst``.

kasemir (approved):
You started out with a separate credentials management app. That was good.
Now it's been moved into core.ui, I assume mostly because of this line in the status bar code that refers to the `CredentialsManagementApp.name`:
```
ApplicationService.createInstance(CredentialsManagementApp.name)
```

In principle, the credentials management app code could remain in a separate module/jar, as long as the status bar just refers to it by name and checks if it's actually available:

```
// Optionally, link to credentials management app
AppDescriptor cm = ApplicationService.findApplication("credentials_management");
if (cm != null)
{
    Button button = new Button(cm.getDisplayName());
    button.setGraphic(ImageCache.getImageView(cm.getIconURL()));
    button.setOnAction(event -> cm.create());
}
```

At the cost of one 'if' and the hardcoded text "credentials_management", the credentials management app remains uncoupled from core.ui. Those sites which want to include it, can, while others can simply omit the "log out" button.

georgweiss:
I do not quite agree that the "cost" is only an if statement. I'd argue that while we on one hand support modules (apps) that may contribute in a dynamic fashion over SPI, we would need to check the existence of such a module (identified by a hard coded string) to define a behavior in the core module. This is a bit counterintuitive in my view.

Further, there are other "apps" in core, e.g. PVList. How does an app qualify as a core app?

kasemir:
There are other examples for supporting 'optional' apps. The display runtime checks if the display editor is available, and uses that info to enable/hide options for opening a running display in the editor. I agree that there's no hard delimiter between apps that are worthy of being included in core.ui or not, which is why I did push the "approve" button, just added the comment to indicate that it could be done in different ways.

georgweiss:
Thanks for approving. I would argue that it makes sense to add credentials management to the core, in particular as I agreed with @shroffk to later add a view of current authorizations (as in the Eclipse version).

georgweiss:
Decided to revert to separate app after all.
Also added some more documentation.

shroffk (approved):

shroffk:
I think when in comes to the core modules we have followed the following rules
 - is it needed by multiple applications
 - would it be useful or hinder any particular site specific product

In the above case, there might be someone who does not want to allow logout functionality... but since Kay approved the PR I am assuming this was not a show stopper. Kay's suggestion would just allow opting out if needed.

However, now that you have separated the modules we can allow sites to decide if they want to include it... I will be including it in my product and I think it is useful enough that I would encourage including it in the common one too.

Also, thanks for documenting everything... I need to follow your example :)

PR #2079, opened 2021-12-02
Ant and Eclipse updates

kasemir:
Adjust to recent dependency and version changes, only affects ant build and Eclipse


shroffk:
Oh... I am sorry I missed this

kasemir:
No problem at all, I don't mind updating the ant/eclipse build.

PR #2082, opened 2021-12-03
Fixing alarm unit test

georgweiss:
Buys us 10 years.


kasemir:
> Buys us 10 years.

Excellent!

shroffk:
10 years...we will all be using some javascript stuff by then

georgweiss:
> 10 years...we will all be using some javascript stuff by then

Or enjoy retirement.

PR #2085, opened 2021-12-09
Refactoring save & restore UI

georgweiss:
Objective is to reduce code duplication and separate out some code to make controller classes more manageable. This is in preparation for upcoming enhancements and features.


geniejhang (commented):
Yeah. It looks like the main stuff I need to take a look is SaveAndRestoreController*.
I'll try it out today. Thanks.

geniejhang:
@georgweiss Everything works fine. Thanks!

georgweiss:
An operation on the save&restore tree view may fail if an object has been removed by another client. I added some changes to intercept such cases and show suitable message (and update the tree view accordingly).

This PR now also contains some code beautification and added Javadoc.

shroffk (approved):

PR #2088, opened 2021-12-14
Removed remaining dependencies to log4j in save&restore

georgweiss:
Also switched to JUL (instead of slf4j).


mgotz:
I hope it's okay if I comment on this, because I recently tried to navigate the labyrinth of java logging frameworks. And, on some points this touches on what I attempted with my pull request #2083 .

Including the spring-boot-starter-logging dependency pulls in logback-classic and jul-to-slf4j. I believe this means all JUL-logging is redirected via SLF4J to the logback backend. The result would be that you need to configure logging with a logback.xml and not a logging.properties file. Im not sure this is intended here.
In particular, I understand it that jul-to-slf4j can't be present on the classpath at the same time as slf4j-jdk14 (which is in dependencies/phoebus-target/pom.xml), becaue it creates circular redirection.
If the desire is to have all logging go to JUL, I believe you need to fully exclude spring-boot-starter-logging. Spring-boot's Commons Logging (JCL) should then find JUL as the only available logging implementation and use it.

Now, this is where I get really, fuzzy, but I believe as an alternative, one could also exclude the common-logging from the spring framework and instead explicitly depend on jcl-over-slf4j and slf4j-jdk14. This would change it from runtime discovery to compile time binding, which might be faster and more explicit.

georgweiss:
My intention was primarily to get rid of log4j dependencies. On top of that I updated logging statements in the code to use JUL only since that is used in the Phoebus main application.
Now if Spring directs to slf4j is another matter, and I do not really care that much. I get the logging I want without any explicit logback.xml configuration.

georgweiss:
But you have a point with the slf4j-jdk14 reference. I will check and update.

georgweiss:
Actually I am fine with the log setup in save&restore as it stands.

Issue #2026, opened 2021-10-06
Add description field to alarm log table


georgweiss:
As per user request: add the description field from the alarm configuration as a new column in the alarm log table. Also requested: ability to customize the table view, e.g. remove unwanted columns.

shroffk:
Since the description is only associated with the config messages and not the state messages it might not be feasible to have a column...however I can start with adding a context menu option to show the descriptions

PR #2089, opened 2021-12-14
#2026 improving alarm message logging and logging view

shroffk:


georgweiss (commented):
Looks OK to me, except for the failing unit test.

georgweiss:
You will need to merge commit df7b71da903ebbe47d39993380748b44ed6223e4 to fix the unit test.

shroffk:
@kasemir @tanviash 
The elastic service logs the output of the "toString()" as the raw source. I don't think anything uses this field as yet.. but by formatting it as valid json it means it will be easier to parse it.
A prototype action is also added which retrieves the last config.

kasemir:
OK on turning toString into toJSON, but maybe re-use one shared object mapper. Object mappers are comparably expensive, but they're fully thread-safe (https://stackoverflow.com/questions/18611565/how-do-i-correctly-reuse-jackson-objectmapper/18615839), so you can create one once and then use it in all those places where you now have yet another object mapper.

shroffk:
Yes, I like Object mappers cause of them being thread safe.
Since they are expensive.
Currently, the AlarmStateMessage, AlarmConfigMessage, and the generic AlarmMessage use the same object mappers.
`AlarmMessage.objectConfigMapper.writeValueAsString(this);`

The rest have a single static one for each of the message classes...but I can further consolidate them.

shroffk:
@kasemir thanks... when I did a second review it made perfect sense to just combine the mapper into the 3 types I need.
I have made the corresponding changes

Issue #2062, opened 2021-11-18
add possibility to display legend for only selected PVs in a plt


olivierdelahaye:
When displaying a PLT with several PVs, you have the possibility to show or not the graph of a PV with the Properties Table.
When you display the legend, all the PVs are displayed in the legend even not selected in the properties table.
Is it possible to display only the legend for the selected PVs ?
In a same way, on the right of the graph, all the PV names are displayed even not selected in the properties panel. It would be nice to see only selected PVs on the properties panel.
See below.

![phoebusPlt01](https://user-images.githubusercontent.com/94627903/142443303-fbbe8882-a819-47e3-aced-19c540f929cb.png)
t

kasemir:
The fact that 'hidden' traces still show up in the legend or on the right axis is a bug. Do you have time to try and fix it?

olivierdelahaye:
I can have a look on it. I keep you updated.

tynanford:
We had a few users notice this issue recently as well. I went ahead and created a pull request for a possible fix here: https://github.com/ControlSystemStudio/phoebus/pull/2099

kasemir:
Tynan just fixed it, thanks

PR #2099, opened 2021-12-21
#2062 Have databrowser legend respect trace hidden property

tynanford:
Fix for https://github.com/ControlSystemStudio/phoebus/issues/2062

Checks each trace isVisible value before drawing trace in the data browser legend


kasemir:
Thanks!

PR #2112, opened 2022-01-26
Select Tab menu item 

georgweiss:
As per user request:
With a large number of open apps and OPIs it may be difficult to locate a particular tab/OPI. This PR adds a menu item in the Windows menu where sub-menu items is an alphabetically ordered list of open tabs across all application windows. Selecting an item will bring that tab to front.


core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
```diff
@@ -497,6 +502,28 @@ private MenuBar createMenu(final Stage stage) {
         final MenuItem about = createMenuItem(new OpenAbout());
         menuBar.getMenus().add(new Menu(Messages.Help, null, about, content));
 
+        selectTabMenu.getParentMenu().setOnShowing(e -> {
+            JobManager.schedule("List open tabs", (monitor) ->
```
kasemir, line 506: I fundamentally like the idea of generating the list of tabs in a background thread.
The update of the visible menu must of course happen on the UI thread via `Platform.runLater`, as you do.

Generating the list of tabs, however, also interacts with the UI, even though it's only reading. So I wonder if that also should happen on the UI thread, not in a background job. Maybe you're just lucky that it happens to "work" with the background job on the Mac, but will fail on Windows or Linux?

Again I like doing as much as possible off the UI thread, and if it's been tested on all platforms, then go ahead, but maybe add a comment:
```
// Listing tabs in a background job.
// Tested successfully on Linux, Mac, Windows,
// but with future versions of JavaFX we might have to do all this on the UI thread,
// since it is interacting with the UI
```

kasemir (approved):

core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
georgweiss: I'm fine putting all on the UI thread, but since setOnShowing is called by the UI thread, there would be no need to invoke it explicitly through Platform.runLater().

I'll update the PR.

kasemir: Yes, leaving it all on the UI thread might be simpler and even faster.
The menu is invoked on the UI thread. It needs to be updated on the UI thread. Fetching the list of ~10 or ~50 tabs won't take that long, so open menu, list & update on UI thread might be faster than open menu -> Background Job to list -> back to UI to update menu.

Issue #1409, opened 2020-07-01
MySQL Version


kasemir:
Dependabot #1408 suggests updating the MySQL client lib from 5.1.45 to 8.0.16.

That's easy enough to do as in commit 9c8fe2f6f95425712f43aec2860583426a1b39c6, but there are compatibility issues.
The MySQL client 5.1.45 works fine with a very old MySQL 5.1.73 server as well as a slightly newer  MariaDB 5.5.65.
The new  MySQL client 8.0.16 won't connect to them, giving a "CLIENT_PLUGIN_AUTH is required" error.

A brief web search suggests that these server/client versions are simply incompatible.
The operational setups to which I need to interface require the older client, so I need to stay with the currently used MySQL client version.

If you don't use MySQL at all, no worries.

Once the first user comes up with a need to interface a new MySQL server and the old client lib will fail to connect, we can reconsider.

kasemir:
No change:

https://github.com/ControlSystemStudio/phoebus/security/dependabot/2
https://github.com/ControlSystemStudio/phoebus/security/dependabot/20
https://github.com/ControlSystemStudio/phoebus/security/dependabot/21

georgweiss:
I have not verified it, but in order to muffle Dependabot we could maybe treat the MySQL jdbc driver in the same manner as the Oracle jdbc lib?

kasemir:
Our oldest MySQL setup went away, and the current mysql-connector-java 8.0.30 works fine with another one I tested, so I'll just create a `${mysql.version}` property in the parent pom, set that to the current version 8.0.30, and dependabot will be happy.
That should work for most people, and in case a local setup needs to fall back to an earlier version, they can patch the version in the parent pom before building the site-specific binaries.

PR #2116, opened 2022-01-27
Updated dependencies due to vulnerability warnings

georgweiss:
Tried to update protobuf version, but that apparently breaks something as data from the remote archiver appliance was not parsed as expected. Do not know why, requires more investigation. Building against protobuf 3.19.3 is fine though.

The mysql connector dependency is unchanged as I am unable to test the archiver service currently.


kasemir (approved):
As for mysql, see #1409 , it would limit the range of mysql servers with which we can communicate.

PR #2134, opened 2022-02-14
Adding documentation for macros

georgweiss:
As per user request: add in-app documentation for macros. 

I've copied some of the legacy CSS help contents, and added content I think - based on user feed-back - is useful. I am prepared to add/modify/delete as needed.


kasemir (approved):
Looks good to me, thanks

PR #2139, opened 2022-02-17
Patches

shroffk:


georgweiss (commented):
Not sure how to verify this...
The new property log_attribute_desc is not referenced in code. And how are file names specified?

PR #2144, opened 2022-02-18
Add access to display info irrespective of right-click position

georgweiss:
Accessing display info from context menu item can be tricky if the OPI is very busy or if it uses a symbol widget as a background.

This PR adds a menu item for the display info regardless of where user clicks, see screen shot. In case user actually clicks in a position where no other widget is in the foreground, the additional context menu item is not added.

Screenshot 2022-02-18 at 14 37 58


georgweiss:
@kasemir, @shroffk, I'd appreciate if you can add your input.

georgweiss:
Yes, Cmd + Option + L in IntelliJ can be deceptive...
Do you need me to revert formatting in order to review?

kasemir:
Well, yes, I would appreciate that.

georgweiss:
@kasemir, updated as requested.

kasemir:
Thanks!  When I look at the "Files Changed" section on github, the formatting is still all changed, which doesn't line up with the "Reverting formatting" commit. I'm assuming that's a github rendering issue, so go ahead.

kasemir:
.. merge?

georgweiss:
Asking for the record.

kasemir:
Well, looks like it didn't revert the formatting in the end. You win.

Issue #2147, opened 2022-02-22
Handling "null" values in the log properties view


shroffk:
![image](https://user-images.githubusercontent.com/2111304/155164337-b07c1ebb-03a1-4ab2-9a83-1596df21fd46.png)

null values should not be rendered as "null" strings

PR #2148, opened 2022-02-22
fixing #2147

shroffk:
@georgweiss  this should fix both the UI and the serialization of null values.
I allow null in values ( which are mapped as empty strings )

For owner and state, if the value is null should we just ignore/not set those fields?

If the property name or attribute name is null... this represent invalid data and we should throw some error. But the server validation should not allow such a situation to occur in the first place


georgweiss (approved):
If user types "null", then it will be persisted as a string with value "null". I think that is reasonable as it would be confusing to persist it as an empty string...

PR #2150, opened 2022-02-25
improved attachment view

shroffk:
image


georgweiss (commented):
Works fine.
My users would like to remove the auto generated text in the body, which makes sense. Should we add a preference setting to control that?

georgweiss (approved):

Issue #2159, opened 2022-03-07
Supress welcome tab - clean option with resource


mgotz:
I would like to pass a resource to phoebus and open just that. Not the welcome tab, not whatever was in the memento.
If I pass an empty layout (to override the memento) and a resource, I still get the welcome tab.
The `-clean` option, in addtion to ignoring the memento and supressing the welcome tab, also ignores any resources.
Is there a workaround I haven't found yet?

To me the `-clean` option would be much more useful, if it would **_not_** ignore resources, apps etc. Afterall, I am in direct control of what I pass on the command line. So, if don't want it to open a resource I don't pass it.
I am happy to make a PR, but since this behavior is explicitly described, I assume there is some reason behind it.

georgweiss:
My users pointed out that in some cases startup could take a very long time if the application was shut down with a large number of OPIs open. Hence the ``-clean`` option was added to just start the application and nothing else.

One could argue that using ``-clean`` with ``-resource`` (or any other option) is a bit confusing. So I would maybe suggest that ``-nomemento`` would be a more fitting name for an option that only ignores the memento file.

mgotz:
> One could argue that using `-clean` with `-resource` (or any other option) is a bit confusing. So I would maybe suggest that `-nomemento` would be a more fitting name for an option that only ignores the memento file.

Sure, `phoebus.sh -clean -resource my_screen.opi` might be a somewhat contradictory request. However, it would give the caller of the application the freedom to make such a request. My thinking was it might be better to tweak an existing option, than add another one. However, this obviously only works if the change doesn't impact people using the existing option.

> If you want the app to start out in a predefined state, for example just with the top-level start display open, simply save that as a layout and load it with the `-layout` option.

Sorry, I may have created a bit of an x-y problem here. I'll explain what I want in the end:
This is more for collegues that create opi screen, then having one defined way the application comes up in the control room. Some of my users would really appreciate the option to open a file directly from the command line. Ideally with shell globs, like
```bash
phoebus *.opi
```
I can make that work by creating our own launcher. But launching the GUI application invariably brings back the Welcome tab. So saving a specific layout doesn't really work for me, because it will always be different. Unless Phoebus would identify an explicitly loaded empty layout as such and not add the welcome tab.

kasemir:
In the interest of minimizing the need for new options, how about this:

1) Create an empty layout which sets the size/location of the window but is otherwise empty
[...]

PR #2162, opened 2022-03-07
Don't create the welcome if a layout was passed

mgotz:
Address #2159 by not creating a welcome tab, if the `-layout` was passed.


kasemir:
As you say in the last commit, "order matters".
I'm a little worried about the way `-layout /path/to/xxx.memento` is only partially removed, and then there's this check
```
         if (!application_parameters.contains("-clean") &&
             !restoreState(memento) &&
             !application_parameters.remove("-layout"))
```
which will remove the remaining `-layout` from the `application_parameters`, but only if there's no `-clean` flag, because otherwise the shortcut-behavior of `&&` will skip the `application_parameters.remove(`.

Please triple-check that all command line combinations still "work", and/or simplify by introducing some new "suppress_welcome" flag that's set when first handling the `-layout`, then remove the complete `-layout` at that time.

kasemir:
Alternate PR uses an explicit `show_welcome` flag.

kasemir:
Handled by #2184

PR #2166, opened 2022-03-08
Add Dockerfile and use docker in CI

lorenzo-gomez-windhover:
As discussed in #2156, some unit tests do not pass in a docker image since there are multiple issues related to IPV6. This PR addresses those issues.

During this process, I also learned that GitHub Actions has some issues related to Maven network timeouts. See https://github.com/actions/virtual-environments/issues/1499. Specifically in our case GitHub Actions would hang forever(until eventually it times out) when we used the new `docker-tests` profile when running maven. The new flags in `MAVEN_OPTS` seems to fix it. I don't know if there is something in those flags that we shouldn't be doing, so please give me feedback if there is. I pretty much stole the config from here https://github.com/Ericsson/ecchronos/pull/275.

Hope these changes make sense.
Thanks
Lorenzo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,17 +2,24 @@ name: Phoebus build
 
 on: [push, pull_request]
 
+
+env:
+  MAVEN_OPTS: -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3 -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+
+
 jobs:
   build:
     runs-on: ubuntu-latest
+    container:
+      image: lgomezwhl/phoebus-ci:latest
     steps:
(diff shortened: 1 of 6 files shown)


shroffk:
Thanks
Is this image `gomezwhl/phoebus-ci:latest` made using the docker file you have added under /misc/Dockerfile ?

lorenzo-gomez-windhover:
Yes. And I would have liked to build it every time I push it, but it requires a token from my account. It is built every time this [fork](https://github.com/WindhoverLabs/phoebus)  is built.

andersharrisson-ess (commented):
Thank you for the MR, @lorenzo-gomez-windhover .
This looks like a good MR. I added a couple of comments on the Dockerfile.

shroffk:
@lorenzo-gomez-windhover do you want me to wait till you update the PR with anders suggestions

lorenzo-gomez-windhover:
I believe all of the concerns have been addressed. Sorry for delays in responses. Things have been quite busy lately...

Issue #2168, opened 2022-03-09
org.phoebus.framework.autocomplete preferences not working


tynanford:
Our users really like the autocomplete feature in Phoebus. Since we aren't using PVA at this time, I'd like to remove that from the autocomplete suggestions.

These settings are available but setting them has no impact. 

https://control-system-studio.readthedocs.io/en/latest/preference_properties.html#framework-autocomplete

It looks to be a bug here: https://github.com/ControlSystemStudio/phoebus/blob/master/core/framework/src/main/java/org/phoebus/framework/autocomplete/PVProposalService.java#L27

Pull request with a possible fix is incoming but not sure if I went about it the right way

![autocomplete](https://user-images.githubusercontent.com/38359458/157373562-86b7a0a0-140e-40b8-8c91-757b26adae72.PNG)

tynanford:
great thanks!

kasemir:
Thanks for finding this issue in the first place, makes a lot of sense to exclude the PVA suggestions at many sites.

PR #2169, opened 2022-03-09
Fix issue with reading in autocomplete preferences

tynanford:
Possible fix for https://github.com/ControlSystemStudio/phoebus/issues/2168

Copied Preferences class from workbench here: https://github.com/ControlSystemStudio/phoebus/blob/master/core/framework/src/main/java/org/phoebus/framework/workbench/WorkbenchPreferences.java 

And updated it for auto_complete settings found here: https://control-system-studio.readthedocs.io/en/latest/preference_properties.html#framework-autocomplete


kasemir:
Will create new PR with one-word fix

PR #2173, opened 2022-03-14
WIP: Update the olog adapter to add opi resource as a resource property

shroffk:
Do not merge just yet...

This is to Display Builder file info to be added as a property so that the logbook viewer can allow users to launch the screen from the log entry


georgweiss (approved):

shroffk:
This provides a backward compatible way of adding Display Builder info to log entries using properties.

@georgweiss by default the `auto_property` is set to false so you will have to enable in your product.

georgweiss (approved):

Issue #1403, opened 2020-06-26
Gdk-WARNING


libaf:
Hi,
I get the following error when I start up Phoebus and click Applications/Display/New Display:
"(java:20070): Gdk-WARNING **: 16.52:58.167: XSetErrorHandler() called with a GDK error trap pushed. Don't do that."
How can I resolve that?
CentOS 7, java-11-openjdk
Cheers,
Markus

kasemir:
Do you always get this warning?
Is it just the warning, otherwise everything works OK?
Do you still get the warning when you run with `-Djdk.gtk.version=2`?

libaf:
Yes, the warning seems to appear whenever the Display window is open (haven't noticed it otherwise yet, but I've just started trying out Phoebus). From what I can tell, everything works as it should otherwise.

Tried running with ./phoebus.sh -Djdk.gtk.version=2 but it "ignores the launch parameter":
2020-06-27 12:09:02 WARNING [org.phoebus.ui.application.PhoebusApplication] Ignoring launch parameter '-Djdk.gtk.version=2'

kasemir:
`-Djdk.gtk.version=2` is not a launch parameter for the phoebes app but for the JVM.
In phoebus.sh, you need to add it right after 'java':

```
java -Djdk.gtk.version=2 -jar $JAR $OPT "$@" &
```

libaf:
Sorry for the misunderstanding. With that parameter added to the shell script I don't get the warning. Thanks! Should I update some library dependency in order for this to work without that statement?

kasemir:
https://bugs.openjdk.java.net/browse/JDK-8156779

libaf:
Thanks 👍

kasemir:
The warning is still happening with Openjfx18, #2179, but doesn't seem to have any negative effect

PR #2179, opened 2022-03-16
JavaFX 16 -> 18

georgweiss:
JavaFX 18 has just been released and I've run some basic testing on various platforms, see #2177.

As Mac users are moving to Monterrey, JavaFX 18 is a must as Phoebus crashes sooner or later on JavaFX 16.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <properties>
     <epics.version>7.0.7</epics.version>
     <vtype.version>1.0.4</vtype.version>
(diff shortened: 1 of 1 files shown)


kasemir (approved):

kasemir:
Tested with OpenJDK 17.0.2 on RHEL 7.9, GDM via ThinLinc: Looks fine both with and without-Djdk.gtk.version=2. Only symptom is the #1403 message when using GTK3.

PR #2182, opened 2022-03-18
Aligning focus behavior for Spinner widget

georgweiss:
Focus behavior same as for Text Entry widget: when selected explicitly through tab button, the text is selected and overwritten when user enters new value from keyboard or by using up/down button.

Also fixed ESCAPE handling which did not work. Apparently JavaFX Spinner will not receive key event from ESCAPE button using onKeyPressed, need to use event filter instead.

ESCAPE reverts back to unedited value.

Done all testing I can think of, but additional testing would be appreciated.


kasemir (approved):
> Done all testing I can think of, but additional testing would be appreciated.

That never ends with all the possible platforms and future JFX updates, but we can handle issues as they arise.
Looks great for now, just indicating the 'active' state via background color is good progress.

PR #2185, opened 2022-03-21
Plot config dialog optional in data browser

georgweiss:
New preference setting:
``org.csstudio.trends.databrowser3/config_dialog_supported``

Defaults to ``false`` in the context of the data browser.


kasemir (approved):

kasemir:
> Defaults to false in the context of the data browser.

That's a change, supporting it by default would be "smoother", but I agree that we do want to disable it by default.

georgweiss:
> That's a change, supporting it by default would be "smoother", but I agree that we do want to disable it by default.

Ultimately sites will need to override, no matter what we choose as default.

PR #2188, opened 2022-03-23
check clean before restoring

mgotz:
#2184 changed the `-clean` behaviour. Now, the memento is always restored. I added a short-circuit back in to avoid loading the memento when `-clean` is requested, which I believe is the intended behaviour.

My apologies for the inactivity on #2162, I was suddenly ill and many thanks to @kasemir for picking that up with #2184.


kasemir:
Thanks!

PR #2192, opened 2022-03-23
Fix  #2180

### LINKED ISSUE #2180: github actions failing
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project parent: Failed to retrieve remote metadata org.phoebus:parent:4.6.6-SNAPSHOT/maven-metadata.xml: Could not transfer metadata org.phoebus:parent:4.6.6-SNAPSHOT/maven-metadata.xml from/to github (https://maven.pkg.github.com/ControlSystemStudio/phoebus): Not authorized , ReasonPhrase:Unauthorized. -> [Help 1]
```

lorenzo-gomez-windhover:
This should(I think) fix our current issues in Github Actions. What's happening is I think the `env` variable is being overwritten and so the GitHub token is never stored in it. As per [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idenv).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,16 +2,14 @@ name: Phoebus build
 
 on: [push, pull_request]
 
-
-env:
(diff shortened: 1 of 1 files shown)


shroffk:
Great. I will merge this and we should know if it has worked.
Also, it is good to know that the env's can only be set in one place :)

PR #2193, opened 2022-03-24
Removed Lombok dependencies

georgweiss:
Lombok is a nifty tool providing annotations to be used for POJO type classes to save the developer from a lot of boiler-plate  code. However, the project has not been maintained as one would wish, and compatibility with IntelliJ has been a bit problematic with newer versions of the IDE.

This PR removes all dependencies to Lombok.


shroffk (approved):

shroffk:
I agree it was a good option to manage less code....however like you said some of the drawbacks were beginning to out weigh the benefits.
I am hoping we find other better supported solutions.

georgweiss:
Java 16 comes with native support for the "record" type. This provides some of the functionality offered by Lombok.

shroffk:
Perfect... I think given out past experience (eclipse -> pure java ). It seems like a lot of useful features are being pulled into the java stack.

PR #2197, opened 2022-03-28
-Remove container from CI.

lorenzo-gomez-windhover:
See #2180


shroffk:
Oh, I was thinking of just disabling the publish.. but this works too

PR #2200, opened 2022-04-01
Indicate attachments in Olog grouped view

georgweiss:
As per user request: indicate if a log entry has attachments in grouped view:

Screenshot 2022-04-01 at 07 17 24


georgweiss:
@shroffk, can you please take a look on this?

shroffk (approved):

PR #2201, opened 2022-04-04
Minor fixes in ActionButton widget

georgweiss:
If user marks ActionButton as disabled and then edits the associated actions, the enabled state reverts back to "enabled" when user dismisses the actions dialog. 

This PR fixes the issue.

Additionally:

- Cursor changes not applied in edit mode when widget is marked as disabled. Found it a bit confusing and not very useful to see the "not writable" cursor in while  editing the OPI.
- Some dead code removed.


georgweiss:
@kasemir, can you please add your input?

kasemir:
I did, see earlier comment.
The empty attachTooltip() needs to remain.
If you remove it, we fall back to the default which adds a tooltip to the jfx_node, a "Pane", but we already attach a tooltip to the actual button in makeBaseButton(), so you get two tool tips.
Leave `attachTooltip() {}` as it was to only get the one correct tool tip on the Button.

georgweiss:
Precisely, cannot see the review comment.

In any case, will put back `attachTooltip`

Issue #2203, opened 2022-04-04
Legacy rectangle widget with 'None' border style


tanviash:
We have some legacy rectangle widgets that have a 'None' border style but a non-zero border_width. These rectangles when opened with Display Builder have a zero line_width as enforced here https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/OutlineSupport.java#L58.

Can we have the 'None' case to also follow other cases like 'Line' etc. so border_width doesn't get overlooked?

Also, border colors get overridden with NamedWidgetColors.TEXT for 'Line' and other border_styles but what if border color is set to some other color. Why not copy the border_color from BOY rectangle widget to here?

kasemir:
The BOY rectangle has "Line" as well as "Border", which is a bit odd:

![Screen Shot 2022-04-06 at 12 07 46 PM](https://user-images.githubusercontent.com/1932421/162019522-e243628e-faaf-4d80-a6fc-44e72f08b194.png)

The display builder only has one surrounding element for the rectangle, and that's the "Line".
It imports the width of a "Border" as the width of the "Line", and I agree that the color should then be imported as well instead of defaulting to black resp. the "Text" color:

![Screen Shot 2022-04-06 at 12 10 34 PM](https://user-images.githubusercontent.com/1932421/162019549-d123692a-deb0-4a99-a0dc-a1a1b028091c.png)

But not sure what to do if legacy displays use both "Line" and "Border".
Maybe update the legacy displays to just use "Line", since that will work in both incarnations?

kasemir:
As for "border_style" being "None" yet the border_width is set, I'd consider that an error.

Update the legacy displays to set border style to line, or use "line_width" and "line_color", either way it should then work in both.

PR #2205, opened 2022-04-06
Rectangle: Use either legacy "line" or "border"

kasemir:
.. whichever has a larger "width". Used to ignore "line" and always use the "border" settings.
If indeed using the "border", now honors the "border_color" instead of always using black resp. "Text" color.

See  #2203 for screenshots


shroffk:
Thanks for fixing the color part

PR #2210, opened 2022-04-11
Fixing full screen menu item

georgweiss:
Verified on Mac and Windows using menu item. 

Could not verify using F11 on Windows, do not know why. Maybe a keyboard issue, tried two different keyboards, got two different results, none of them full screen transition.


kasemir (approved):

kasemir:
Looks good.
As for F11, I don't think we tie any key combination to the full screen action.
On the contrary, in https://github.com/ControlSystemStudio/phoebus/blob/4a067f85df7dc5f4286dc7c39ab588299a12d420/core/ui/src/main/java/org/phoebus/ui/javafx/FullScreenAction.java#L28 we try to prevent any key combination from exiting full screen mode, because the idea was to use this for "kiosk" displays where the user should not be able to exit via a keyboard. So F11 might be handled by the operating system beyond our control.

PR #2226, opened 2022-04-26
Csstudio 1661

cjenkscybercom:
Refactor: Moved ListSelection fxml and controller into core module
CSSSTUDIO-1661: Updated search such that items are hidden instead of highlighted (replaced selectionModel usage with FilteredList)


georgweiss (approved):
Works as agreed.

PR #2230, opened 2022-04-28
Support for opening save&restore resource

georgweiss:
Changes in this PR:

1. Save & restore made into a "singleton" instance, i.e. opened only once.
2. Added support for opening with resource, e.g. file:/<unique_node_id>?app=saveandrestore. This can be used by Olog properties, i.e. an entry automatically created from a restore context can add the resource identifier thus enabling launch of save & restore from log viewer.
3. Minor bug fix in context menu.


shroffk:
What would happen is ```app=saveandrestore``` is missing when launching the resource

georgweiss:
Then the framework will log that no application was found to handle the "file", and Save & Restore will not be launched unless it is restored from ~/.phoebus.

Actually I'd prefer an API to find application based on scheme, say saveandrestore:/resource-id.

shroffk:
Oh no...
We now have conflicts

georgweiss:
Will fix merge conflict...

georgweiss:
Merge conflicts resolved.

PR #2231, opened 2022-04-28
ContextMenuHelper uses AdapterService to adapt before calling action

shroffk:
The `ContextMenuService` already uses the `AdapterService` to figure out which actions should be included in the context menu... this is great!
I added another step such that when the included action is invoked, the ContextMenuService calls the AdapterService to convert the selection into the required type before executing the action. In the past each action had to retrieve the selection and handle the conversion itself...with this change we should be able to go back and simplify a lot of the context menu actions.


shroffk:
The adaption will only be performed when the user clicks on the action.. so it should not impact the speed at which context menu is created and populated.

kasemir (approved):
There's a certain amount of magic going on in there, but if it works, go for it.

georgweiss (approved):

PR #2232, opened 2022-04-29
AlarmLogTable: sets a TimestampedPV as selection and uses the

shroffk:
contextmenu helper to populate the context menu


shroffk:
@georgweiss can you test this if you can..

georgweiss (commented):
Not sure what I am supposed to see... With the huge indexes we have there might be a timeout issue. Only item I get in the context menu is "Configuration info"...

shroffk:
This actions uses the selection and thus does not need to query... just used the items that are already in the table.

It should look as follows
image

georgweiss:
Thing missing is the service definition file. Once in place all works as expected.

PR #2233, opened 2022-05-02
Enhancements to save & restore snapshot tagging

georgweiss:
As per user request:

- The search tool in save & restore now supports a condition for "golden" snapshots, see screen shot.
- When adding a tag to a snapshot, the dialog will list existing tags to facilitate selection. The AutocompleteMenu API is used  in a manner similar to PV proposals.

Also: minor bug fixes and removal of redundant string resources.

Screenshot 2022-05-02 at 16 55 57

Screenshot 2022-05-02 at 16 12 22


shroffk:
hmm..
I guess the golden snashot makes sense since it is a feature available while tagging

PR #2239, opened 2022-05-06
CSSTUDIO-1660 replaced TimeRangeDialog with popover implementation in DataBrowser

cjenkscybercom:
FXML would have been nice, however the TimeRelativeIntervalPane etc are all dynamic Java nodes. It would have been a moderate-to-sizable chunk of work to convert those to FXML. 

So, this implementation merely tacks on the Apply and Close buttons to the existing TimeRelativeIntervalPane and provides a means to define the behavior of those buttons in callbacks.


georgweiss (commented):
Functionality seems to be in place. 
However, the UI element to which the PopOver is "pointing" differs between the Time Axis tab and the export tab. I prefer the way it works in the export tab, i.e. point to the button launching the PopOver. Currently in the Time Axis tab the PopOver points to the End Time label, which is a bit misleading and confusing.
Also, clicking Apply should dismiss the PopOver to save the user from an extra button click.

PR #2255, opened 2022-05-13
CSSTUDIO-1673 Deduplicate Olog ListSelection Code for LogEntryEditor

cjenkscybercom:
I've pushed some changes onto CSSTUDIO-1673 but there are several questions/observations I have:

- I cannot create a PopOver that uses fxml by extended PopOver; due to circular constructors I need to use Composition. But because PopOver doesn't have an interface I end up creating something that looks like a PopOver (has methods like show() and hide()) but actually isn't
- I created a new Demo class for the PopOver since it needs to be used differently from a modal; e.g. it really ought to be tied to a pane of some kind with a button (and ideally a text field to show the selected values, bound to an observable).

That all said, I'm not sure what to do with:

- the old Demo class for the Popup
- the old olog app

I could swap out usage on the old olog app, however I don't recall how to invoke or test it? Either way I figure if the old usage sticks around then the old Demo class should too (but if removed, then it should also be removed)


shroffk:
@georgweiss could you review this...
+1 for adding the testing/demo

georgweiss:
Sure, will review.

cjenkscybercom:
Updates have been pushed up, addressing the sync issues between the fields and popover. 
Finished rollout of changes to logbooks fields in search and log entry forms.
Found and fixed existing bug, where logbooks always populated with a default value when replying to a logbook entry.

georgweiss (approved):

PR #2265, opened 2022-05-19
Alarm server Kafka error handling

kasemir:
When the alarm server gets disconnected from Kafka, it cannot send alarm state updates.
One possible scenario:
Alarm server gets disconnected from Kafka (and clients as well).
Alarm server still has PV connections, one goes into alarm and then maybe out again. Alarm server latches alarm, but state update to Kafka is not possible (and clients wouldn't see it anyway right now because they are also disconnected).
Once Kafka returns online, it is unaware of the alarms. Newly started clients receive an outdated alarm state. Alarm server still tracks those alarms as active and will even emit the "There are N active alarms" messages every 15 minutes, clients annunciate it but show a different number of active alarms in the UI.

This update offers two additions:

A "resend" command in the alarm server can be invoked at any time to trigger a complete re-send of the state for each item in the alarm tree. It can be used to debug server/client inconsistencies.

More important, the alarm server monitors the Kafka connection. If it's lost, there's nothing it can do about that. But once it's restored, it performs a "resend" so Kafka and clients get updated to the most recent alarm server state, removing inconsistencies because of the dropped messages.
The Kafka client library offers no direct API to check the connection state, so a periodic "listTopics" call is used as suggested in https://stackoverflow.com/questions/38103198/how-to-check-kafka-consumer-state and other places. It can be configured and disabled via a new preference setting.

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
@@ -127,8 +127,11 @@ public class AlarmSystem
     /** Heartbeat PV period in milliseconds */
     public static final long heartbeat_ms;
 
-    /** Nag period in seconds */
+    /** Nag period in milliseconds */
     public static final long nag_period_ms;
+
+    /** Connection validation period in seconds */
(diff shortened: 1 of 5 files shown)


kasemir:
Fixes #2267

Issue #2262, opened 2022-05-17
"Channel Info" resolution of PV names


kasemir:
Assume a channel "X" is registered in the channel finder, for example via recsync.
In most cases, the display will use just "X" as a PV name, but the following are also possible:

 * X.DESC to show description
 * X.HIGH to show or allow runtime adjustment of an alarm limit
 * X.RBV to show a motor's readback value
 * X.VAL$ to fetch a long text as a channel access char array
 * pva://X to use PVA for fetching a long text that plain channel access cannot transfer (alternative to .VAL$)

For all these cases, the "Channel Info" context menu on the PV will fail because just "X" is known to the channel finder, but not the other PV variants for the same underlying record.

@shroffk Should the "Channel Info" tool strip any `xxx://` prefix from the PV name and then stop at the first '.'?

PR #2272, opened 2022-05-24
Channel Info: Look for just the channel 'X' in ca://X.VAL$

kasemir:
#2262


shroffk (approved):

shroffk:
I can see us moving this into a core utility service
Since this might be useful elsewhere too

kasemir:
Maybe. A 'Channel' tends to be a valid 'PV', no conversion needed. 
But for a PV 'ca://X.VAL$' the channel is just 'X', so might have to patch the name. Still, the context menu from PV to the channel info tool seems to be the only place where you need to convert PV to channel, so I put it right there.

PR #2274, opened 2022-05-25
fix: meter widget not using limits defined in editor when pv missing display limits

cjenkscybercom:
To replicate the issue / test first run:

`softIocPVA -d CSSTUDIO-1339.db`

Then, open CSSTUDIO-1339.bob in CSStudio

[Archive.zip](https://github.com/ControlSystemStudio/phoebus/files/8769408/Archive.zip)


shroffk:
Looks like a reasonable check

PR #2281, opened 2022-06-01
Additional options for kafka clients

mgotz:
I wanted to enable ssl encryption and authentication for the kafka communications in the alarm-system. This reqires several additional settings for the kafka consumer and producers.
This PR adds a kafka_properties preference to the phoebus client and a corresponding cmd option to the alarm-server, alarm-logger and alarm-config-logger. This preference is a path to a properties file to load additional kafka-client properties from. The properties file format and content is the same as the one for the kafka-console clients. This should provide a lot of flexibility to configure the kafka clients without adding a lot of extra preferences to phoebus.
In addition, I tried to add this to the existing cmd parse logic in the alarm-server. However, I found it very complicated to add an additional argument in the existing structure, while also ensuring that all combinations and also orders of arguments work properly. So, I rewrote the cmd argument parsing making the processing independent of the order the arguments were given. Hopefully, this makes it a bit more flexible and it removed some of the repetitive code.


shroffk:
Documentation would be great.
I will merge this now since I want to commit changes to the alarm-logger in preparation for elastic 8 and want to avoid conflicts

mgotz:
Wonderful, I will add some documentation and make a new PR

PR #2289, opened 2022-06-08
Increment version of log4j

kasemir:
Updating to 2.17.1 which according to https://logging.apache.org/log4j/2.x/security.html is "fixed".

We're not directly using log4j at all, it's only dragged in via the og4j-to-slf4j connector to route messages from other dependencies, but simply having log4j 2.14 in the file system causes computer security scans to alert.


kasemir:
@shroffk, so you see a problem?

shroffk (approved):

PR #2292, opened 2022-06-10
Save & restore snapshot logging

georgweiss:
Added SaveAndRestoreEventReceiver interface with the purpose of calling implementations about save&restore "events". To start with supported events are new snapshot and restore of snapshot. 

The save-and-restore/logging module is one such implementation that will dispatch a log request to registered log factory implementations. Logging will happen only if valid credentials for the logbook scope are available as provided by SecureStore. Some new settings have been added to customize the log entry in terms of logbook, tags and level. At least one logbook must be specified.

Log entries created contain snapshot meta-data, plus a link in the properties. Clicking the link will open the save-and-restore app and select/show the snapshot.

The save-and-restore/logging module can be excluded from a site product.

This PR also includes some refactoring to simplify the dependencies for core-logbook.


kasemir:
This is similar to https://github.com/ControlSystemStudio/phoebus/tree/master/app/pace which creates a log entry when writing the set of values, you create a log entry when restoring the values.

As for "neither the save-and-restore application nor the logging module will prompt for logbook credentials, ....use .. Credentials Management": The pace app opens a pre-populated log entry dialog, where users then enter their credentials _and_ can tweak the text, for example add comments.

georgweiss:
The intention is to make the logging non-intrusive on both code base and UI. Sites may include save&restore, but exclude logging. 

That said I agree there is a limitation users may dislike. 

@shroffk , can you please comment? I'm fine adding a prompt as described by @kasemir, but then logging becomes more closely integrated with save&restore.

georgweiss:
Additional development needed.

PR #2293, opened 2022-06-10
PVA image compression codecs

kasemir:
Allows displaying area detector images that are compressed with LZ4.
Gives same visual results as ImageJ plugin, tested with byte, short, int, long, float and double data from simdetector, both signed and unsigned variants. Can't compare long (int64, uint64) because ImageJ doesn't handle it, but image looks same compressed vs. uncompressed.

Does _not_ use area detector binaries to decompress but apache-commons-compress, which for now limits it to LZ4.


whs92:
Thanks for implementing this! The LZ4 and JPEG libraries are good enough for most of our applications anyway. I agree with your point about this being easier to maintain than using the areaDetector stuff

kasemir:
Yes. If you're using the area detector, you likely have the compression binaries for Linux and maybe also Windows on the control system network, but getting them to clients running CS-Studio on their office computer still seems inconvenient to me.
In basic tests, LZ4 compressed just fine. If we run into a scenario where BLOSC is orders of magnitude better, we can reconsider supporting it via the native lib.

One thing I haven't figured out is the proper handling of byte order. The compressed data is always a byte array. It contains info about the original data type, but not the original byte order. If the area detector server was on a PPC and the client on Intel, I think that will fail, but don't have a setup to verify this.

PR #2306, opened 2022-06-23
Dialog box choose directory

JeoffreyGi:
I have some doubt about the fonction name, but it seem to work after some test.

openSaveFileDialog give you the possibility to open the dialogBox at the right path.
But it couldn't work because the preference overwrite the file path in : doPromptForFile

So i put earlier the verification of the preference, and except if file = null, it overwrite the preference.
i have add also, 1 more verification because the File was intended to be a file and not necessary a directory


kasemir:
Looks good. Will wait for the build process to run/finish, and then we can merge this.
Thanks!

PR #2319, opened 2022-07-01
CSSTUDIO-1706: Empty widget field (precision) causes warning message

cjenkscybercom:
fix: when blank integer widget properties are parsed (e.g. an empty node in a .bob file such as the precision property of the spinner widget), then an exception is thrown instead of the default value

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/properties/IntegerWidgetProperty.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/properties/IntegerWidgetProperty.java
@@ -63,8 +63,14 @@ public class IntegerWidgetProperty extends MacroizedWidgetProperty<Integer>
     @Override
     protected Integer parseExpandedSpecification(final String text) throws Exception
     {
+
+        if(text != null && text.isBlank())
+        {   // If the value read from the .bob is empty then it should return the default
+            return default_value;
+        }
+
         try
-        {   // Should be integer..
+        {   // Otherwise try to parse to an integer
             return Integer.valueOf(text);
         }
         catch (final NumberFormatException ex)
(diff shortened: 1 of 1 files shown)


kasemir (approved):
Looks good!
How did you get a file with empty values?
From the editor, or are those created by some local script?

kasemir:
Also, if you only care about blank text, then  `if (text != null  &&  text.isBlank())` works.
If you want to guard against `null` as well, this would be more robust:
```
if (text == null  ||  text.isBlank())
   .. use default
```

cjenkscybercom:
Hi @kasemir !

I got blank values in the .bob file by clearing the precision value in the editor and saving the file. It then shows up as an empty node:
`<precision></precision>`

I was hesitant to do a null check because when I checked the value it is parsed as an empty string and not null. I'm not sure what edge case would cause it to parse into a null...I'm not sure what could cause that (I'd think it'd fail hard early on...but I'm also hesitant to find out / find what happens when we allow null Integers ...)

kasemir:
>  .. by clearing the precision value in the editor and saving the file

I see. Another scenario might be setting some number to `$(A_MACRO)` and then passing that in as empty.
You're right that `null` is not expected. I then agree on failing hard early on, and for that you can of course just remove the null check:
```
if (text.isBlank())
   ...
```
With JDK 14+ you can then get a nice "text is null in   text.isBlank()" type of exception, instead of delaying that a few lines down ;-)

cjenkscybercom:
> > .. by clearing the precision value in the editor and saving the file
> 
> I see. Another scenario might be setting some number to `$(A_MACRO)` and then passing that in as empty. You're right that `null` is not expected. I then agree on failing hard early on, and for that you can of course just remove the null check:
> 
> ```
> if (text.isBlank())
>    ...
> ```
> 
> With JDK 14+ you can then get a nice "text is null in text.isBlank()" type of exception, instead of delaying that a few lines down ;-)

Those are great points! I think I'll remove the null check; Integer.valueOf(null) throws an NPE anyway rather than a NumberFormatException. So, if it fails then it'll fail hard somewhere else much earlier.

kasemir:
> With JDK 14+ you can then get a nice "text is null in text.isBlank()" type of exception

Referring to `-XX:+ShowCodeDetailsInExceptionMessages`, https://www.oracle.com/java/technologies/javase/14-relnote-issues.html

PR #2320, opened 2022-07-01
Adding a system time pv, with an offset and multiple format options

shroffk:


kasemir (approved):
Looks good, but I don't know the use case/motivation for `sys://timeOffset(offset, format, update_seconds)`.

I get the `format` to have more flexibility over the fixed format of the existing `sys://time`.
Maybe also the `update_seconds` to control how often it triggers.
But `offset`, is that for displaying from a different time zone?
Seems like it should be the least important, last parameter...

Or did this type of PV exist before, and now we better keep the order of parameters to stay compatible?

PR #2321, opened 2022-07-06
Save&restore tree view improvement

georgweiss:
This is to fix an annoying behavior in the save&restore tree view. Due to the different cell renderers for folder, configuration (save set) and snapshot nodes, the height of a tree cell will vary. Since tree cells are reused by JavaFX, this causes - under certain circumstances - the tree view to automatically scroll when a node is expanded. 

Changed the cell renderers to make the height the same for all node types. To achieve this I removed the secondary line (containing date and user id) from the snapshot node. Also resized the icon for configuration nodes. The result is seen here:

Before:
Screenshot 2022-07-06 at 08 06 19
After:
Screenshot 2022-07-06 at 08 07 02

Long term - if date and user name are considered essential in the view - we should maybe consider a tree table like in the file browser.


kasemir:
>  if date and user name are considered essential in the view 

You could place that info in the tool tip of the cell

PR #2332, opened 2022-07-21
Elastic 8 2 alarm logger

shroffk:
@tanviash @georgweiss @kasemir 
The new alarm logger is ready for testing.
uses elastic 8.2
adds a new REST service so that clients don't have to directly query elastic
updated the logging ui client to use the REST service


app/alarm/model/src/main/resources/alarm_preferences.properties
```diff
@@ -3,7 +3,7 @@
 # --------------------------------------
 
 # Kafka Server host:port
-server=localhost:9092
+server=localhost:19092
```
kasemir, line 6: Is this just from a local test run and should be back to the default of 9092?

shroffk: Yes, I will fix this... thanks

shroffk: reverted all ports to the standard defaults in the next commit.

kasemir:
This updates elastic jars to 7.17.3 in the top-level pom via `<es.version>7.17.3</es.version>` and I'm getting these build errors:
```
   [javac] /opt/jenkins-home/workspace/phoebus/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java:260: error: cannot find symbol
    [javac]                             .create(true)
    [javac]                             ^
    [javac]   symbol:   method create(boolean)
    [javac]   location: class Builder
```

The `Builder` doesn't seem to have a `create` method:

err1

It's the one from elasticsearch-java-7.17.3.jar, it has `name()` and `indexPatterns()` and the other methods but no `create()`:

err2

err3

The problem seems to be that the pom.xml for the alarm logger selects a different version via property `<elasticsearch.version>8.2.0</elasticsearch.version>` (which then isn't used, instead the pom keeps repeating the version in e.g.
```
 <artifactId>elasticsearch-java</artifactId>
      <version>8.2.0</version>
```

That's different from the top-level pom selecting 7.17.3, which is also put into the dependencies/target.

Does the alarm logger need to use a different version of the elastic client?

kasemir:
@shroffk since 8.2.0 is newer, can you set  `<es.version> 8.2.0</es.version>` in the top level pom and then use that everywhere? Or does all but the alarm logger need 7.17.3?

shroffk:
All usage of elastic in Phoebus should now be using 8.2
With the alarm logger now having a REST api, I think the only place which directly uses elastic is the alarm logging service itself

I have opened a new PR with the changes.

PR #2337, opened 2022-07-26
CSSTUDIO-1717: Databrowser displays disconnected / paused PV as if it was valid

cjenkscybercom:
fix: leading point at 'now' is rendered, incorrectly implying it is real data

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PVSamples.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PVSamples.java
@@ -115,15 +115,17 @@ public class PVSamples extends PlotSamples
         if (lock.getReadHoldCount() <= 0  && ! lock.isWriteLockedByCurrentThread())
             logger.log(Level.WARNING, "Missing lock", new Exception("Stack Trace"));
 
+        // If the data point is 'real'/raw then return it
         final int raw_count = getRawSize();
         if (index < raw_count)
             return getRawSample(index);
-        // Last sample is valid, so it should still apply 'now'
+        // Else, create a 'virtual' point by transforming its
+        // timestamp to 'now' to display the currently implied value
         final PlotSample sample = getRawSample(raw_count-1);
         if (Instant.now().compareTo(sample.getPosition()) < 0)
             return sample;
         else
-            return new PlotSample(sample.getSource(), VTypeHelper.transformTimestampToNow(sample.getVType()));
+            return new PlotSample(sample.getSource(), VTypeHelper.transformTimestampToNow(sample.getVType()), true);
     }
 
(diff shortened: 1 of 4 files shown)


kasemir:
Looks good, but should clarify "real".

Consider a setpoint PV. Operators last switched a power supply "On" two weeks ago, and the "VoltageSetpoint" was last adjusted one week ago. Those PVs are still "real", their value is still absolutely valid since I sure hope the power supply is on and regulating to that setpoint. If not, your accelerator is broken and needs to be fixed.
Channel Access and PV Access have been optimized for this scenario. The power supply IOC won't send out a new value for the "On" or "VoltageSetpoint" PVs unless they change. The archive doesn't fill up by repeating the same value all the time. It stores only the values with original time stamp of one or two weeks ago because those exact time stamps, when the value last changed, is important information by itself. If operators show them on a display, they expect to see the current value. When plotted together with for example the "VoltageReadback", they expect to see a flat line for the "On" and "VoltageSetpoint" that extends until "now". If the old values of those PVs were no longer real within ADEL thresholds and sample rates, we should have received new  samples.

You update the special case of showing the data points. By default, the data browser doesn't show data points, only lines between the points. When enabling the display of data points, it's probably correct to suppress the point at the line that's extended to "now" because it's indeed not a point received from the IOC; its time stamp of "now" does not indicate that the value just updated "now".

cjenkscybercom:
> Looks good, but should clarify "real".
> 
> Consider a setpoint PV. Operators last switched a power supply "On" two weeks ago, and the "VoltageSetpoint" was last adjusted one week ago. Those PVs are still "real", their value is still absolutely valid since I sure hope the power supply is on and regulating to that setpoint. If not, your accelerator is broken and needs to be fixed. Channel Access and PV Access have been optimized for this scenario. The power supply IOC won't send out a new value for the "On" or "VoltageSetpoint" PVs unless they change. The archive doesn't fill up by repeating the same value all the time. It stores only the values with original time stamp of one or two weeks ago because those exact time stamps, when the value last changed, is important information by itself. If operators show them on a display, they expect to see the current value. When plotted together with for example the "VoltageReadback", they expect to see a flat line for the "On" and "VoltageSetpoint" that extends until "now". If the old values of those PVs were no longer real within ADEL thresholds and sample rates, we should have received new samples.
> 
> You update the special case of showing the data points. By default, the data browser doesn't show data points, only lines between the points. When enabling the display of data points, it's probably correct to suppress the point at the line that's extended to "now" because it's indeed not a point received from the IOC; its time stamp of "now" does not indicate that the value just updated "now".

Exactly, and well put, thank you! It's a bit challenging to put this one into words, it has to do with making clear that the time stamp of 'now' does not necessarily mean the data point really just updated 'now'.

PR #2339, opened 2022-07-28
feature: text alignment in Spinner widget

cjenkscybercom:


cjenkscybercom:
I've verified this by checking alignment updates in real time in the editor, and that the changes are persisted when the widget is viewed in the Viewer. I'm not sure, however, if I can test this with a real PV or not without write permissions?

kasemir:
Can you try with `sim://ramp` to check if it displays received updates from a read-only PV?
Then try with the PV `loc://x(0)` in both a spinner and a text entry widget.
Should be able to change the PV from a spinner.
Then also change it via the text entry and check if the spinner updates as expected.

cjenkscybercom:
> Can you try with `sim://ramp` to check if it displays received updates from a read-only PV? Then try with the PV `loc://x(0)` in both a spinner and a text entry widget. Should be able to change the PV from a spinner. Then also change it via the text entry and check if the spinner updates as expected.

Thank you! I'll try to remember those. 
Checking it, the value updates in real time (and is read-only), and I can increment/decrement the loc value -- so I think it looks good!

kasemir:
Thanks for checking with those PVs.

PR #2341, opened 2022-08-01
Changes to allow window geometry specification along with target=window

shk78:
Minor changes to allow window geometry, size and position, to be specified when using target=window. Optional geometry parameters can be specified in a similar manner to that used with X windows e.g.

target=window@1038x442%100x100

This enables windows to be popped up with sizes that match the GUI and at standard starting positions.

### DIFF
--- a/core/framework/src/main/java/org/phoebus/framework/macros/Macros.java
+++ b/core/framework/src/main/java/org/phoebus/framework/macros/Macros.java
@@ -42,7 +42,7 @@ public class Macros implements MacroValueProvider
     private final Map<String, String> macros = new LinkedHashMap<>();
 
     /** Regular expression for valid macro names */
-    public final static Pattern MACRO_NAME_PATTERN = Pattern.compile("[A-Za-z][A-Za-z0-9_.\\-\\[\\]]*");
+    public final static Pattern MACRO_NAME_PATTERN = Pattern.compile("[A-Za-z][A-Za-z0-9_.\\@\\+%\\-\\[\\]]*");
 
     /** Check macro name
      *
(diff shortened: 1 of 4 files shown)


core/framework/src/main/java/org/phoebus/framework/macros/Macros.java
```diff
@@ -42,7 +42,7 @@ public class Macros implements MacroValueProvider
     private final Map<String, String> macros = new LinkedHashMap<>();
 
     /** Regular expression for valid macro names */
-    public final static Pattern MACRO_NAME_PATTERN = Pattern.compile("[A-Za-z][A-Za-z0-9_.\\-\\[\\]]*");
+    public final static Pattern MACRO_NAME_PATTERN = Pattern.compile("[A-Za-z][A-Za-z0-9_.\\@\\+%\\-\\[\\]]*");
```
kasemir, line 45: Is this necessary? We don't want to allow '@' or '+' or '%' within a macro name. Within a macro value, yes, but not a macro name.

core/ui/src/main/java/org/phoebus/ui/docking/DockStage.java
```diff
@@ -105,6 +117,39 @@ static String createID(final String what)
         return what + "_" + UUID.randomUUID().toString().replace('-', '_');
     }
 
+    private static Geometry setSizeAndPosition(String geometryString)
+    {
+        Integer x = Integer.valueOf(50),
```
kasemir, line 122: Somewhat unusual to use `Integer`. Try to use plain `int`.

kasemir:
Implemented in #2347 without the MACRO_NAME_PATTERN change.
Also, the format is now like the `-geometry ..` of X11 tools:

`target=window@1038x442+100+100`

PR #2346, opened 2022-08-09
Update Eclipse dependencies .classpath ..

kasemir:
.. to track recent update of elastic #2332 and some other accumulated changes


kasemir:
Only affects Eclipse IDE

Issue #2348, opened 2022-08-09
PV Table -> Right click -> Insert Row doesn't work with empty table


tynanford:
Right click -> Insert Row (above) fails with an error when the table is empty.

![image](https://user-images.githubusercontent.com/38359458/183737558-3c7722d0-dc00-4f4d-a4ee-d980a0e0b900.png)

```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at org.phoebus.applications.pvtable.ui.TableItemProxy.getItem(TableItemProxy.java:85)
	at org.phoebus.applications.pvtable.ui.PVTable.lambda$createContextMenu$17(PVTable.java:615)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:234)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
	at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
	at com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:49)
	at javafx.event.Event.fireEvent(Event.java:198)
	at javafx.scene.control.MenuItem.fire(MenuItem.java:459)
	at com.sun.javafx.scene.control.ContextMenuContent$MenuItemContainer.doSelect(ContextMenuContent.java:1385)
	at com.sun.javafx.scene.control.ContextMenuContent$MenuItemContainer.lambda$createChildren$12(ContextMenuContent.java:1338)
	at com.sun.javafx.event.CompositeEventHandler$NormalEventHandlerRecord.handleBubblingEvent(CompositeEventHandler.java:247)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:80)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:234)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
[...]

PR #2349, opened 2022-08-09
Check for null on PV Table item get method

tynanford:
Possible fix for #2348


kasemir:
On an empty table, the "Enter new PV" hint should make it fairly obvious how to get the first PV in.
Once there is at least one, you might want to insert another one somewhere in the middle instead of continuing to add more at the end.
So "Insert" doesn't really make sense when there's nothing in the table, but fair enough to avoid the NPE. Thanks!

Issue #433, opened 2018-11-15
Display action to "open file" hangs on Linux


kasemir:
Currently calls `java.awt.Desktop.open(..)` to open arbitrary files in OS-defined tool.
May actually work once, then hang when invoked again.

PR #2350, opened 2022-08-10
CSSTUDIO-1607: External viewer of attachments (movie clips, .pdf, office files)

cjenkscybercom:
refactor: delegate opening log attachments to desktop/user default instead of rendering error message

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/AttachmentsPreviewController.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/AttachmentsPreviewController.java
@@ -129,8 +129,15 @@ public class AttachmentsPreviewController {
                     if(parts.length == 1 || !ApplicationService.getExtensionsHandledByExternalApp().contains(parts[parts.length - 1])){
                         // If there is no app configured for the file type, then use the default configured for the OS/User
                         try {
-                            Desktop.getDesktop().open(attachment.getFile());
-                            return;
+                            if(Desktop.isDesktopSupported()) {
+                                Desktop desktop = Desktop.getDesktop();
+                                if(desktop.isSupported(Desktop.Action.APP_OPEN_FILE)) {
+                                    Desktop.getDesktop().open(attachment.getFile());
+                                    return;
+                                }
+                            } else {
+                                ExceptionDetailsErrorDialog.openError(Messages.PreviewOpenErrorTitle, Messages.PreviewOpenErrorBody, null);
+                            }
                         } catch (IOException e) {
                             ExceptionDetailsErrorDialog.openError(Messages.PreviewOpenErrorTitle, Messages.PreviewOpenErrorBody, null);
                         }
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/AttachmentsPreviewController.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/AttachmentsPreviewController.java
@@ -30,6 +30,7 @@ import javafx.embed.swing.SwingFXUtils;
 import javafx.fxml.FXML;
 import javafx.scene.Cursor;
 import javafx.scene.control.*;
+import javafx.scene.control.MenuItem;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
 import javafx.scene.input.MouseEvent;
@@ -46,6 +47,7 @@ import org.phoebus.ui.dialog.ExceptionDetailsErrorDialog;
 import org.phoebus.ui.javafx.ImageCache;
 
 import javax.imageio.ImageIO;
+import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


georgweiss (commented):
The call to the Desktop API should maybe also catch UsupportedOperationException:
Screenshot 2022-08-10 at 13 06 14

georgweiss (commented):
The call to the Desktop API should maybe also catch UsupportedOperationException, see screen shot.
Screenshot 2022-08-10 at 13 06 14

Test on Windows looks good.
Test on Centos7 over NoMachine is not so good. Clicking on MS Office attachments seems to hang Phoebus completely. No stack trace, no clue of what goes on. Will test on workstation (=not over NoMachine) tomorrow.

cjenkscybercom:
Thank you for checking this on Centos! 
I suspect that Desktop isn't supported on that platform, so I've wrapped it in a check; if that fails, then it'll return an error message as before.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/AttachmentsPreviewController.java
```diff
@@ -120,13 +122,22 @@ public void changed(ObservableValue<? extends Attachment> observable, Attachment
...
                     // If not internal apps are found look for external apps
                     String fileName = attachment.getFile().getName();
                     String[] parts = fileName.split("\\.");
                     if(parts.length == 1 || !ApplicationService.getExtensionsHandledByExternalApp().contains(parts[parts.length - 1])){
-                        // If there is no app configured for the file type, show an error message and return.
-                        ExceptionDetailsErrorDialog.openError(Messages.PreviewOpenErrorTitle, Messages.PreviewOpenErrorBody, null);
-                        return;
+                        // If there is no app configured for the file type, then use the default configured for the OS/User
+                        try {
+                            if(Desktop.isDesktopSupported()) {
```
kasemir, line 132: Desktop is from java.awt, while we're running under JavaFX.
That can result in hangups.

The display builder falls back from the ApplicationLauncherService to the JavaFX  `Application.getHostServices()`, 
https://github.com/ControlSystemStudio/phoebus/blob/ab909e323bcfe377511f5cb5df20dda5c746a22b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L1021

georgweiss: Thanks @kasemir, could explain the hand I spotted on CentOS7

@cjenkscybercom, can you please update according to @kasemir 's input? I'll run a new round of tests.

kasemir: We've definitely seen hangups with calls to  AWT `Desktop`, see https://github.com/ControlSystemStudio/phoebus/issues/433

Mind you I don't think we can predict what happens inside HostServices. The documentation mentions opening the file in a web browser,  https://openjfx.io/javadoc/13/javafx.graphics/javafx/application/HostServices.html, and I think that also tends to work for PDF documents. But will it open MS Word, Excel, ...?

georgweiss: The idea is not to provide an exhaustive solution, but to support a "best effort". I'm sure users do not expect MS Office files in general to open on Linux. As for the behavior on Windows/Mac, let's try it out...

cjenkscybercom: Hmmm...I'll give this a shot, thanks @kasemir !

cjenkscybercom: Do we have a standard way of getting access to the Application or HostServices?
The controller doesn't extend Application or have it/HostServices available, and Application ofc doesn't have a static getHostServices method?

kasemir:
> Do we have a standard way of getting access to the Application or HostServices?

> The controller doesn't extend Application or have it/HostServices available, and Application ofc doesn't have a static getHostServices method?



Right, how do you get the JFX Application?


I've put it into PhoebusApplication.INSTANCE:


https://github.com/ControlSystemStudio/phoebus/blob/ab909e323bcfe377511f5cb5df20dda5c746a22b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java#L99-L105

cjenkscybercom:
> Do we have a standard way of getting access to the Application or HostServices? The controller doesn't extend Application or have it/HostServices available, and Application ofc doesn't have a static getHostServices method?
> Right, how do you get the JFX Application? I've put it into PhoebusApplication.INSTANCE: https://github.com/ControlSystemStudio/phoebus/blob/ab909e323bcfe377511f5cb5df20dda5c746a22b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java#L99-L105 ``` /** * JavaFx {@link Application} * * <p>Set on {@link #start(Stage)}, * may be used to for example get HostServices */ public static Application INSTANCE; ```

Was *just* about to ask! 🤣
Sounds like I was on the right track, then; I found it by searching for usages of HostServices

cjenkscybercom:
Copying here too -- wouldn't this be a security issue potentially? I'm guessing maybe we should at least whitelist what files can be opened automatically? (though arguably you can still have issues with e.g. macros in Word files)

georgweiss:
User may already configure through settings file a handler for arbitrary file types, so what we do here is not that different.

kasemir (approved):
Looks good to me

PR #2351, opened 2022-08-10
Restore of snapshot fix for pva

georgweiss:
In save&restore a restore operation will handle write failures when using ca. Since the pva implementation behaves differently, write failures are not handled and user may believe all is well.

This is a fix. Verified both for ca and pva.

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/SnapshotController.java
@@ -95,6 +95,7 @@ import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


kasemir:
Does save/restore keep the order of PVs/values?
The PV Table does write values in row order, and I'm pretty sure older save/restore tools like "BURT" did this, because for some devices it's important that things are written in the correct order, and the "next" PV is only restored after the "previous" PV was completely written.
`pv.asyncWrite(value).get()` is definitely the way to go for that.
But note that when you're writing "4" to the Kelvin setpoint of a cryostat that's at room temperature, the asyncWrite may take two hours to complete since it's actually waiting for the operation to complete.

So the PV table has a "Completion [x]" checkbox for each row to configure if you want to use plain write-and-forget or asyncWrite.

georgweiss:
Users cannot specify the order in the UI, only exclude rows. So I guess the answer is that PVs are written in the order they are listed in the UI.

As for pv.write (fire and forget) vs pv.writeAsync().get() I realize this change will impact the behavior for the case you describe. Currently if a PV takes a long time to complete the write operation users can only tell from the UI that the read-back value has not changed. And that may be the case for a PV at the bottom of a 600 row list...

Question is if the restore operation should support the option - globally or PV by PV - to wait for completion.
@shroffk, what is your take on this?

shroffk (approved):

PR #2352, opened 2022-08-10
#2332 refactoring the elastic search property name and updating to 8.2.0

shroffk:


kasemir (approved):
That should bring all to one and the same elastic jar version, thanks!

PR #2354, opened 2022-08-12
Add value of DESC (if non-null and non-empty) to pv_value in widget tooltip

georgweiss:
If PV defines the DESC field, and if it is non-empty, it's value will be shown in the widget tooltip. Implementation proposes to put the string on a separate line:

Screenshot 2022-08-12 at 15 26 12

NOTE: works only over pva.

The value of DESC is read from the recently added ``description`` field of the ``Display`` class in the VType lib.


kasemir (approved):
Looks good.

You always add the description to the `$(pv_value)`.
In hindsight, we could have supported something like `$(pv_time)`, `$(pv_value)`, `$(pv_alarm)`, `$(pv_description)`, `$(pv_units)`, ... which would allow configuring tool tips to show just the value, or value with units etc.

But that would mean people who want to now see the description need to update all their tooltips to add `$(pv_description)`, which is not practical.
Besides, this is just the tool tip, where having the choice between a default that shows PV name and value vs. some custom text is probably good enough.
No need for detailed configuration of how time, value, alarm, desc. are formatted.

Issue #2301, opened 2022-06-21
OpenFileDialog at a specific target


JeoffreyGi:
Hello,
Today, you can create an Dialog box to choose a file with the Python script. You can launch it in the FileUtil class :

`new OpenFileDialog().promptForFile(window, "Open File", null, null);`

But, you can't choose the first location of the OpenFileDialog. 
To work, tt seem to be :
`new OpenFileDialog().promptForFile(window, "Open File", "My Path", null);`

Whereas, this is not implements.
Maybe we could add this figure.
I can make it.

But if we can do it without change the Java code, i would be glad to know how.
I want to launch a script which settle parameters, and one is the file to use.

kasemir:
You're right, the current set of `FileUtil.openFileDialog`, https://github.com/ControlSystemStudio/phoebus/blob/a3f8c5d2a92d66bfd11f0242e88743d616ed3271/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/FileUtil.java#L154 as well as `FileUtil.openSaveDialog` https://github.com/ControlSystemStudio/phoebus/blob/a3f8c5d2a92d66bfd11f0242e88743d616ed3271/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/FileUtil.java#L168 only take a `boolean inWorkspace` which is there for compatibility with the older Eclipse version that had a 'workspace' and no longer has any meaning.

Feel free to create a pull request which adds for example this:
```
/** Open a file save dialog.
 * @param path_to_file Path to existing file, used to set the initial directory on which the file dialog is opened.  May be null
 * @return Full file path. Or null if dialog is cancelled.
 */
public static String saveFileDialog(String path_to_file)
```

JeoffreyGi:
I've solved the issue but i can't push my branch : 
`git push -u upstream dialogBoxFileUtil`
    `Error : Permission to ControlSystemStudio/phoebus.git denied to JeoffreyGi.`
After looking in the project, i didn't find the solution.
I'm new to GitHub so i don't have know how to get the right.

kasemir:
> I've solved the issue 

Excellent!

> but i can't push my branch :

See github intro https://docs.github.com/en/get-started/quickstart/fork-a-repo

JeoffreyGi:
Thank you, the pull request is ready.

georgweiss:
[...]

PR #2355, opened 2022-08-12
#2301 use parent folder when creating a new (non existent file)

shroffk:


kasemir (approved):
Thanks for digging into this!

georgweiss:
Works as expected: new files created in correct target directory.

PR #2356, opened 2022-08-16
Bug fix save&restore logging

georgweiss:
I have also removed the dialog showing list of PVs that were not written during restore. Main reason is that it interferes with the log entry editor launched. But failures should be communicated differently, e.g. a separate column in the table view indicating row by row about write result. This I will add next.


georgweiss:
The log entry is "pre configured" to list PVs that failed to write. Plus there will be a link to the snapshot so user can open it from the log entry viewer.

Rollback has never been in scope for save&restore. But I do not mind adding such a feature.

georgweiss:
I removed the error dialog as in some cases the log entry editor dialog went hiding behind the main window.

shroffk (approved):

PR #2357, opened 2022-08-16
#715 checking for PVAStructureArray before PVAArray

shroffk:
@kasemir 

I added some changed to the PVAStructureHelper.

I check for PVAStructureArray, before checking for PVAArray, since it is a more specific implementation.

In the future, I think we need to review and move various bits of the PVAStructureHelper to the Decoders.
i.e. the part which handled the special case of the the PVAStructureArray should be a case in the decoders.decodeArray(....)

What are you thoughts?


kasemir (approved):

shroffk:
I have an example PVAServer which servers up an Array of Structures... this was the same service which trigger the need to have support for pva://some_pv_array[i] type pv's

In the end, I think we have to accept that we can reasonably support NT types and incrementally add things when we come across exotic use cases.

shroffk:
https://github.com/shroffk/scope

PR #2359, opened 2022-08-17
fix: send to logbook fails if selection cannot be adapted

cjenkscybercom:


shroffk:
@georgweiss this makes sense to make.

Issue #2275, opened 2022-05-25
Scale, minor ticks and markers with ProgressBar widget


baham4s:
Hello,

I have on CSS (\*.opi file) a ProgressBar with scale, minor ticks and markers.
![CSSProgressBar](https://user-images.githubusercontent.com/61794844/170230967-25cdbbda-77ce-40d7-8496-2a333567b3b8.png)
But i can't find any way to implement this with Phoebus (*.bob file).
![PhoebusProgressBar](https://user-images.githubusercontent.com/61794844/170231888-5f5c3e41-ce91-4c0d-b7e2-8ca6c119e2e6.png)
Will these properties be implemented in Phoebus?

kasemir:
Closing with #2362

PR #2362, opened 2022-08-19
#2275 : add horizontal property to widget Tank

baham4s:


kasemir (approved):
Looks good.

For what it's worth, check the https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java and https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/TextUpdateRepresentation.java for 
their use of `was_ever_transformed`.
By default, those widgets have no transformation. Even `jfx_node.getTransforms().clear()` actually creates a new `NodeTransformation` for the node and then sets it to contain nothing. For labels and text updates which are used many times in pretty much every display we're trying to avoid even adding an empty transformation. For the tank widget that's very likely overkill, so what you have should be fine.

PR #2363, opened 2022-08-22
Removing "split view" in save&restore UI

georgweiss:
To improve code maintainability the "split views" of the save & restore UI have been removed, see screen shots below (which show the split views that are no longer supported).

1) Split view in save&restore data browser where the list view of snapshot objects is no longer supported. Snapshot objects are shown as child nodes of the save set node with which they are associated

Screenshot 2022-08-22 at 14 40 54

2) When user wishes to add a set of PVs (copied from for instance the Channel Finder application) to a save set, or create a new save set from the selection, the save set target selection dialog where list of save sets is shown as a list view is no longer supported. Instead save set nodes are shown as child nodes of the containing folder node.

Screenshot 2022-08-22 at 14 41 11

As a consequence the following preference settings have been removed as they are not needed:

org.phoebus.applications.saveandrestore/splitSnapshot
org.phoebus.applications.saveandrestore/splitSaveset


shroffk:
I think this makes sense in terms of simplifying the code base.

@tanviash do you know who is managing the save restore client UI at FRIB... is anyone really using this feature to justify the cost of maintaining it?

tanviash:
@shroffk @georgweiss I don't see any reason to keep this list view around. Please feel free to merge your changes.

PR #2371, opened 2022-09-07
CSSTUDIO-1725: Properties defined in Widget class files are ignored unless use_class=true

cjenkscybercom:
fix: properties defined in class/template files are ignored unless use_class is set true; now they are used if defined, and use_class can be optionally used to lock properties from editing

As a side effect of this, you can now override the DEFAULT class def by simply defining one with the same name for a given widget, for example:
```
<widget type="label" version="2.0.0">
    <name>DEFAULT</name>
    <text>I've been overridden!</text>
    <width use_class="true">600</width>
</widget>
```

NOTE: Will want to, as followup, cleanup the ess-phoebus-customization files as they set some properties (such as x and y position) that we may be wanting to ignore.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetClassSupport.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetClassSupport.java
@@ -73,10 +73,12 @@ public class WidgetClassSupport
     {
         private final String specification;
         private final Object value;
+        private final boolean use_class;
 
         /** @param property Property with value to use for class */
         public PropertyValue(final WidgetProperty<?> property)
         {
+            use_class = property.isUsingWidgetClass();
             if (property instanceof MacroizedWidgetProperty)
             {
                 specification = ((MacroizedWidgetProperty<?>) property).getSpecification();
@@ -205,9 +207,12 @@ public class WidgetClassSupport
         else if (! DEFAULT.equals(widget_class))
             logger.log(Level.WARNING, "Widget type '" + type + "' class '" + widget_class + "' is defined more than once");
 
-        for (WidgetProperty<?> property : widget.getProperties())
-            if (property.isUsingWidgetClass())
+        for (WidgetProperty<?> property : widget.getProperties()) {
+            if (property.isValueReadFromFile()) {
+//            if (property.isUsingWidgetClass()) {
                 class_properties.put(property.getName(), new PropertyValue(property));
+            }
+        }
     }
 
     /** Get known widget classes
@@ -291,11 +296,16 @@ public class WidgetClassSupport
         }
 
         final PropertyValue class_setting = class_settings.get(property.getName());
-        if (class_setting == null)
+        if (class_setting == null) {
             property.useWidgetClass(false);
+        }
         else
         {
-            property.useWidgetClass(true);
+            if(class_setting.use_class) {
+                property.useWidgetClass(true);
+            } else {
+                property.useWidgetClass(false);
+            }
             class_setting.apply(property);
         }
     }
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetConfigurator.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetConfigurator.java
@@ -134,6 +134,7 @@ public class WidgetConfigurator
             try
             {
                 property.readFromXML(model_reader, prop_xml);
+                property.setIsValueReadFromFile(true); // todo, integrate into readFromXml?
                 property.useWidgetClass(parseBoolean(prop_xml.getAttribute(XMLTags.USE_CLASS)));
             }
             catch (Exception ex)
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetProperty.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/WidgetProperty.java
@@ -48,6 +48,9 @@ public abstract class WidgetProperty<T extends Object> extends PropertyChangeHan
      */
     protected final T default_value;
 
+    /** Was this property's value set from xml, or just from the default value? **/
+    protected volatile boolean is_value_from_file = false;
+
     /** Does property follow the value suggested by class? */
     protected volatile boolean use_class = false;
 
@@ -307,6 +310,14 @@ public abstract class WidgetProperty<T extends Object> extends PropertyChangeHan
      *  @throws Exception on error
      */
     abstract public void readFromXML(final ModelReader model_reader, final Element property_xml) throws Exception;
+
+    public void setIsValueReadFromFile(boolean value) {
+        is_value_from_file = value;
+    }
+
+    public boolean isValueReadFromFile() {
+        return is_value_from_file;
+    }
 
     /** Notify listeners of property change.
      *


kasemir:
Can you give an example of what this is fixing?

This is the expected behavior with the default classes.bcf:

1) Add a label  with some text somewhere.
2) Change class to COMMENT -> now it's using the comment font, FB color black, transparent set, wrap-words enabled and those settings are now locked.
3) Change class to TITLE -> now X and Y position are fixed to 0, 'wrap words' remains set because that was the last value frmo the COMMENT class but it's no longer locked.

Will this now be different?

kasemir:
The current behavior is meant to be this: 
In a class definition display, properties marked with "use_class" will be set by the class and then locked.

Do I understand correctly that you're looking for this:
All properties provided in a class definition are used. It's no longer necessary to mark them with "use_class".
Properties marked with "use_class" are in addition locked, so "use_class" becomes "force_class", while other properties are simply set from the class but can still be edited?

How do you then handle position? Each widget in a class file likely has a non-default X/Y position, otherwise they'd all stack up in the upper right corner. When now applying class settings to a display, will it move the widget to the location of the template, since it's applying all properties from the template?

kasemir:
Also note that class settings are applied to existing displays. Change the font for TITLE, open an existing display that used a widget with Label of class TITLE, and it will now use the changed font.
So "use_class" really means always use the value from the class, whatever that might be in the future.

If you're looking for a weaker copy-these-values-now, but allow me to edit, and then leave them as I set until I edit the file again,  you can use template files that are simply read-only display *.bob files. From those, you can copy/paste widgets into a new editor. You get all the settings from the copied widget, but can now change them.
Updating the template later won't affect displays that you created by copying from an earlier template.

kasemir:
See https://github.com/ControlSystemStudio/phoebus/pull/1340 for template idea

cjenkscybercom:
Hi @kasemir ,

I'll forward this to Dirk for his input. You are correct, the intention for this is to act as templating but as verified it does affect existing displays and I don't think that was the intent.

shroffk:
I think I agree with Kay... The strong binding between the properties defined in the class file is a desired feature but should be used with care... the addition of the "use_class" forces people to give some thought to what they will be forcing onto all existing widgets.

To have widgets initialized with some sensible defaults (which can be edited)... for my users I will have a "top resource" link to an NSLS2 widget convention .bob file from which they can copy paste widgets.
The widgets in this "convention/template" .bob file will use classes for properties which should not be changed and the rest of the properties will be set to some sensible default which they can change. Copying these widgets will give the same functionality you are describing - some properties from classes which are un editable and the rest of the properties initialized to some sensible value.

> As a side effect of this, you can now override the DEFAULT class def by simply defining one with the same name for a given widget, for example:

Is there a use case for this?
Forcing a change on all DEFAULT classes could have a lot of unintended consequences... I prefer the idea that the classes framework only apply to named classes where someone has deliberately decided to apply a class format on a particular widget.

kasemir:
With a read-only *.bob file opened in the editor as a "template", you can not only offer individual widgets to copy/paste as starting points with some suggested settings, but also whole groups or collections of widgets like this one:

xx

Once you paste those into a new display, you can then move/rearrange and it'll stay that way.

If on the other hand you intent to later for example move the LED to the left of the text update and you want all displays to follow that change, you'll need to use an embedded display widget (or a template/instance widget if you want a whole sequence of them) so that later updates to the embedded content are shown in all screens that embed it.
Each embedded display has a little overhead over widgets "right there" in a display, but that was much worse in the original Eclipse CS-Studio BOY implementation, we're now loading all the embedded content in parallel and especially the template/instance widget can be pretty quick to populate.


So we have all these pieces:
Class files that enforce settings, and changes to the class will later update displays that use them.

A copy/paste approach for starting out with suggested settings that you can then change, and they stay as you last set them. You can always do that between editors, but we have added support in the editor to handle read-only *.bob files as copy sources.

Embedded display and template/instance widgets to re-use collections of widgets, where updating the embedded content will result in updates to all users.

There might be a way to somewhat combine the class file and copy/paste approach by updating the class editor. Right now it only has the "use_class" checkbox that sets & locks properties to a class setting. Simply making all other properties suggested defaults could have unintended side effects, but the "use_class" checkbox could change into a tri-state setting:

off) No effect. Typically used for X, Y position
initial_default) New option: In the editor, applies this setting to a widget when the class is selected, but can then still change in the editor
use_class) As now: Forces that setting from the class template onto widgets that use the class

The initial_default is what you were after here, and again it could be implemented as a new option in the class editor and when applying a class to a widget. But there are likely several pitfalls as to when exactly we want to apply this. Only when selecting a class for a widget in the editor? Or also later when checking if a widget property is at default (not set from XML)?
Bottom line is that it won't give us any really new functionality beyond the copy/paste approach which is easy to understand. We might avoid a few mouse clicks by adding an initial_default option but might pay for that by having to explain it over and over.

cjenkscybercom:
> > As a side effect of this, you can now override the DEFAULT class def by simply defining one with the same name for a given widget, for example:
> 
> Is there a use case for this? Forcing a change on all DEFAULT classes could have a lot of unintended consequences... I prefer the idea that the classes framework only apply to named classes where someone has deliberately decided to apply a class format on a particular widget.

From what I've understood, the use case for this comes from a need for applying default/organizational styling automatically; e.g. if you import a .bob file it can be already styled as common in your org. My understanding of the problem now is that you need to select (at most, groups of) widgets individually and apply a custom style class; this can be very time consuming and tedious.

kasemir:
In the class file, you can already define settings for the DEFAULT class.
So you can for example change the font for a Label, which will then be forced on any Label in new or existing displays that don't select another class.
In the editor, any new Label will get that font and it will be fixed until you select a different class. So to make it generally edittable, you then need to enter for example "None" for the class name (and not actually define a "None" class that would again force the font).

Still, if your users really dislike all the default settings and feel the need to change everything, they should remember that this is just one display tool. The general approach should be: Add label, enter text. Add text update, enter PV. That display will generally work, look the same in the web runtime, and in 10 years, when you will have to port all your displays to the next display tool, it won't be too hard. If you instead now spend time on tweaking all the font sizes etc, be prepared to completely re-do all your displays every 10 years.

cjenkscybercom:
I'm going to table work on this for now; this should have been merely a technical review at this point, however this request appears to still need alignment from stakeholders and those decisions are not mine to make. For now I'd like to keep this open a bit longer so the outcome of that can be documented here.

georgweiss:
I an attempt to close this ticket I have checked the validity of the statement
> All properties provided in a class definition are used. It's no longer necessary to mark them with "use_class".
and as far as I can determine it is incorrect.

Defining a class TEST like:
```
<widget type="label" version="2.0.0">
    <name>TEST</name>
    <text>Test</text>
    <height>50</height>
    <width>270</width>
    <foreground_color use_class="true">
      <color red="255" green="0" blue="0">
      </color>
    </foreground_color>
    <font>
      <font family="Source Sans Pro" style="BOLD_ITALIC" size="20.0">
      </font>
    </font>
  </widget>
```
and selecting it for a Label widget, only property ```foreground_color``` is applied:

Screenshot 2023-10-05 at 08 04 05

All other properties from the class are ignored.

With this PR other properties from the class are also applied, as one would expect.

kasemir:
> Defining a class TEST like:
>    <text>Test</text>
>    <height>50</height>
>    <foreground_color use_class="true">
> only property foreground_color is applied:

That is how it's supposed to work.
The "TEST" class for a label is defined as having that background color.
We don't want all TEST labels to also have a text "Test", with 270 and so on.
Well if you want all TEST labels to have the text "Test", you can mark the text as being part of the class definition.

Again, the current class handling works as desired, we can close this PR

georgweiss:
Then I conclude that desires differ.

kasemir:
The classes work as originally designed, as described in several tutorials like https://controlssoftware.sns.ornl.gov/training/2022_USPAS/Presentations/05%20Display%20Builder.pdf, and as in use with existing displays.

This PR mistakes the class idea with the template idea, which is supported since https://github.com/ControlSystemStudio/phoebus/pull/1340, also in use as mentioned by Kunal's comment.

This PR would break existing class behavior and instead duplicate the template idea, so it needs to be closed.

shroffk:
I think we can use the requirements and use cases gathered here to work on a "templating"/"useful initialization" functionality which is separate from the classes

PR #2374, opened 2022-09-08
Fix wrong time stamp in alarm logger

georgweiss:
In the alarm log table, the time stamps were shown in UTC.

### DIFF
--- a/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogMessage.java
+++ b/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogMessage.java
@@ -158,7 +158,7 @@ public class AlarmLogMessage {
 
     public static class AlarmInstantDeserializer extends JsonDeserializer<Instant> {
 
-        private static DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.systemDefault());
+        private static DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"));
 
         public AlarmInstantDeserializer() {
         }


services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogMessage.java
```diff
@@ -158,7 +158,7 @@ public void setEnabled(boolean enabled) {
 
     public static class AlarmInstantDeserializer extends JsonDeserializer<Instant> {
 
-        private static DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.systemDefault());
+        private static DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS").withZone(ZoneId.of("UTC"));
```
kasemir, line 161: You should be able to use the common `TimestampFormats` from https://github.com/ControlSystemStudio/phoebus/blob/master/core/util/src/main/java/org/phoebus/util/time/TimestampFormats.java. They are created with `ZoneId.systemDefault()`. They take an `Instant` that is basically in UTC and display it in the local time zone.

If the alarm message logger display requires `ZoneId.of("UTC")` to display time in the local time zone, something must be wrong with the time stamp. Does the alarm logger not write the time as UTC? Does the retrieval from Elasticsearch already apply a time zone offset?

georgweiss: Elasticsearch stores in UTC, this is unchanged. Elasticsearch  does not apply any time zone adjustment. 

However, now that the client no longer connects to Elasticsearch to retrieve data, the Alarm Logger service instead pulls it. Elasticsearch data is parsed by Jackson, and the AlarmLogMessage has a custom Jackson deserializer to take care of time stamps.

In this deserializer using ZoneId.systemDefault() caused the time stamps to be adjusted to UTC even though they already are in UTC. The effect is -2 hours in my time zone.

The fix is to configure the DateFormatter to treat the time stamps from Elasticsearch as UTC.

kasemir: Ah, this is the 'server' part of the alarm logger which must format the time stamp as text, but in UTC, so that the alarm table GUI, as a client, can then turn the received UTC into local time?

kasemir (approved):

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogMessage.java
georgweiss: One could also put it this way: UTC time stamps from Elasticsearch are parsed twice, once by server - which must preserve UTC - and once by client that does the local time zone adjustment.

shroffk: Thanks for fixing this.

PR #2379, opened 2022-09-12
Csstudio 1675: Font Size Not Applied for ScaledSlider

cjenkscybercom:
Font size is not correctly applied in the ScaledSlider widget. 

The root cause of this is that the font utilities are breaking the font style attributes into their individual pieces, however this is not supported on `-fx-tick-label-font`; it needs to be the short-hand form.


cjenkscybercom:
Hold on a minute, I've accidentally snuck in work from a separate commit/branch; I'll quick correct this...

cjenkscybercom:
This is now sorted out :)

kasemir (approved):
See if you can configure your editor to always use explicit imports, not switching to wildcard `import javafx.scene.layout.*`.

cjenkscybercom:
@kasemir Sure, can look into that. Related, do we have a Java styleguide somewhere? I've not seen one in this repo, and so far have only seen the License templates on Confluence.

kasemir:
#894, #1031, https://github.com/ControlSystemStudio/phoebus/pull/1031/files

PR #2383, opened 2022-09-14
CSSTUDIO-1745: hitting enter after changing a value in the color picker doesn't commit it to the property

cjenkscybercom:
When changing a value in the color picker popover for e.g. a color widget property such as Background, the popover closes without the value changing. 

This occurs because both the root node and the ok button itself are bound to these keyboard events (explicitly and implicitly), and the sanitization is triggered by leaving the field (such as clicking elsewhere, or tabbing to another field) e.g. closing the field's editor. By consuming the Enter key event explicitly, this sanitisation and editor commit/close is bypassed. 

Hence, the fix being one line; to not consume the event captured by the root node and to let it propagate.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetColorPopOverController.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetColorPopOverController.java
@@ -152,7 +152,6 @@ public class WidgetColorPopOverController implements Initializable {
         {
             if (event.getCode() == KeyCode.ENTER)
                 okPressed(null);
-                event.consume();
         });
 
         picker.valueProperty().addListener(( observable, oldColor, newColor ) -> {


kasemir (approved):
I'm fine with doing this, but at the same time I cannot reproduce the issue.

Add Label, click background color to open color popup, click a color, then either press "Return" or click the "OK" button, either way the color is changed. How do you get it to not commit? Do you only see the issue on a specific OS?

cjenkscybercom:
> I'm fine with doing this, but at the same time I cannot reproduce the issue.
> 
> Add Label, click background color to open color popup, click a color, then either press "Return" or click the "OK" button, either way the color is changed. How do you get it to not commit? Do you only see the issue on a specific OS?

I had the same problem initially, haha; it was something that was clearer to see watching somebody do it. 

What the user was doing was clicking the color property to open the dialogue, clicking on either the red, green, or blue spinner, entering a value using the numpad/number-bar (not using the up/down buttons on the spinner), and then hitting enter. (without, for example, tabbing to another control, clicking elsewhere, or hitting ok) They were a very keyboard-centric user.

kasemir:
> entering a value using the numpad/number-bar (not using the up/down buttons on the spinner), and then hitting enter.

That does it! Looks like you're finding all kinds of issues with the color picker, it's a wonder it worked at all.

PR #2385, opened 2022-09-14
Csstudio 1338: ScaledSlider and Scrollbar Widgets Reflect Incorrect PV Value When Dragged Outside Limits

### LINKED ISSUE #2376: CSSTUDIO-1338: ScaledSlider and Scrollbar Widgets Reflect Incorrect PV Value When Dragged Outside Limits
(This mirrors CSSTUDIO-1338)

When the ScaledSlider or Scrollbar widgets are:

- connected to a PV with DRVL/DRVH limits
- configured s.t. those limits can be bypassed (the "Limits From PV" property is false)

If you try to set the PV to a value outside its DRVL/DRVH limits, then the widget displays the new value outside the limits (rather than the actual value, which is ofc set to the DRVL/DRVH values). For example, if the DRVL is 5 and you use the ScaledSlider widget to set the value to -20, then the ScaledSlider displays the value as -20 (even though it is actually set to 5). 

This is not the case for other widgets such as Spinner and TextEntry; if you e.g. attempt the above in the TextEntry widget then the value stays at 5 as expected.

Please see the attached db and bob files. Run the DB file locally via:
```
softIocPVA -d CSSTUDIO-1338.db
```
[CSSTUDIO-1338.zip](https://github.com/ControlSystemStudio/phoebus/files/9527919/CSSTUDIO-1338.zip)

cjenkscybercom:
Fixes #2376


cjenkscybercom:
Something really strange going on here with the commits...

kasemir:
The commits appear to be the ones about the font fixes that have already been applied..

cjenkscybercom:
Hmm, I'm going to try pulling this all fresh, there's some changes that git diff should pick up on that aren't being picked up. Either way, I still want to apply this work to the scrollbar widget as well anyway.

kasemir:
> I don't believe we need to somehow unregister the setOnMouseRelease call

I think that's correct. The JFX listeners don't need to be removed because the complete JFX scene graph gets disposed when we remove the representation.
For the listeners to our model we do remove the property listeners because otherwise we had memory leaks, or would call the listeners in the representation when the GUI was gone but the model still there.... but I can't right now point out exactly how that happened, because when you for example close a display then the model that was represented gets disposed, which should include the listeners.

kasemir:
This is where we added the listener removal: https://github.com/kasemir/org.csstudio.display.builder/pull/479

kasemir:
> .. scrollbar widget as well ..

Excellent, thanks!

Issue #1926, opened 2021-06-10
Detached display width and height


fredrik-soderberg:
When I detach a display in Phoebus, the width and height is calculated from the parent. The width and height shown in the display property panel is not used. Some of our users would prefer to use the values in the property panel. They report that they have to resize a lot of windows all the time when using the detach functionality. What is the main reason for using the current implementation to the determine the width and height? Would it be possible to use the width and height from the display property instead or will it have other consequences/side effects?

kasemir:
> Would it be possible ..

Not easily. The 'docking' mechanism has no idea about the content. It doesn't care if the pane contains a display builder runtime or an editor or probe or ...
At the same time, the apps inside the pane have no idea if the pane is one of many tabs, or in a vertically divided tab, or a standalone window.

It's of course possible to change it, but we should NOT change it by making the docking pane check for a display builder runtime as a child, since that would introduce unnecessary coupling. What if you wanted to build a phoebus with only data browser? Couldn't, would now always need to include the display runtime...
So one way to solve this would be to add something like a "Dimensions getSizeHint()" to the app, which the docking lib would ignore most of the time _except_ when it's detaching a pane. Personally, not sure if it's worth it because you might be able to find as many users who would vote for "use the origin window size" option that's currently implemented.

cjenkscybercom:
@kasemir Started looking into this one again, I see a year later or so (CSSTUDIO-1121) and can see where you're coming from with this implementation-wise. 

I'm curious if you've had any additional thoughts on this, since I'm working on (potentially) implementing it.

cjenkscybercom:
@kasemir It so far wasn't a lot to change, and the ramifications are opt-in...So I made a PR. Would you mind please taking a closer look at it? Thanks!

https://github.com/ControlSystemStudio/phoebus/pull/2387

PR #2387, opened 2022-09-15
fix: DisplayRuntime window doesn't use width set by display properties

cjenkscybercom:
As discussed in #1926 

- does *not* introduce a compile-time dependency to the DisplayRuntime module
- introduces a dimension hint to the AppInstance interface. Since a lot of applications (most, probably?) just want to use the parent window dimensions, the default is to return an empty optional.
- implements the hint method on the DisplayRuntimeInstance impl, using the width and height properties if defined on the display

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
@@ -9,7 +9,8 @@ package org.csstudio.display.builder.runtime.app;
 
 import static org.csstudio.display.builder.runtime.WidgetRuntime.logger;
 
-import java.awt.*;
+
+import java.awt.geom.Rectangle2D;
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
 import java.util.Objects;
@@ -509,15 +510,21 @@ public class DisplayRuntimeInstance implements AppInstance
     }
 
     @Override
-    public Optional<Dimension> getDimensionHint() {
+    public Optional<Rectangle2D> getPositionAndSizeHint() {
         return Optional.ofNullable(active_model).flatMap(displayModel -> {
             Integer width = displayModel.propWidth().getValue();
             Integer height = displayModel.propHeight().getValue();
             if(width != null && width > 0 && height != null && height > 0) {
-                return Optional.of(new Dimension(width, height));
+                return Optional.of(new Rectangle2D.Double(
+                        displayModel.propX().getValue(),
+                        displayModel.propY().getValue(),
+                        width,
+                        height
+                ));
             } else {
                 return Optional.empty();
             }
         });
     }
+
 }
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
@@ -9,6 +9,7 @@ package org.csstudio.display.builder.runtime.app;
 
 import static org.csstudio.display.builder.runtime.WidgetRuntime.logger;
 
+import java.awt.*;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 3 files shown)


cjenkscybercom:
One thing I'm wondering @kasemir is if the dimensions should be bound to a Property on the runtime instance...But as I thought about it, the window dimensions are set when it is undocked so I don't imagine there's much need for it? As it already works, you need to refresh the display anyway when you make any changes to it in the Editor.

kasemir:
For the most part, that looks great!

You used `getDimensionHint` instead of the `getSizeHint` that I had suggested, but I'm perfectly willing to let that slide ;-)
I would, however, use for example `javafx.geometry.Dimension2D`, not the AWT `java.awt.Dimension`, and use an explicit import instead of `import java.awt.*;`

The detached display will now open with a configurable size, but still in a pretty random location.
Do you also want to check the model's X and Y location, so the hint could provide the complete bounding rectangle. That's what EDM does in a similar case.

kasemir:
> ..bound to a Property on the runtime instance...fetching them once is sufficient here.

I think that's fine as it is. Set when pane is created and then leave it.

cjenkscybercom:
> use an explicit import instead of `import java.awt.*;`

Yep, missed it in that one file. Turns out IntelliJ has _yet another_ exception specifically for java.awt and a few other libraries. I nuked em, but forgot to check the other files. That said, I'll use Dimension2D instead thank you!

cjenkscybercom:
Wait, @kasemir using javafx.geometry.Dimension2D would introduce a dependency on javafx-graphics in the phoebus-core module; do we want this?

kasemir:
Good point. We don't want "AWT" per se, but it kinda comes for free.
Also, what do you think about setting all of X, Y, WIDTH, HEIGHT?
Return an `int[]` with 2 (WIDTH, HEIGHT) or 4 (X, Y, WIDTH, HEIGHT) elements??

kasemir:
Maybe the QWT Dimension is fine. We're not touching the AWT event loop which could conflict with the JFX event loop, we're just using that POD type.
Still, use 'Bounds' or 'Rectangle' to set X, Y, WIDTH, HEIGHT?

cjenkscybercom:
Exactly, it should just be a DTO. 
I could throw in x/y position as well, that's a good point for extension later.

kasemir:
Let's return a rectangle or bounds, that is  X, Y, WIDTH, HEIGHT.
If X and Y are zero, as they tend to be when nothing has been entered, ignore those.
Otherwise use them to set the X/Y position of the pane.
That allows opening standalone windows in fixed locations, something EDM could do, and some imported/converted  EDM displays thus have X/Y set.

cjenkscybercom:
Yep!
And in particular I'll use Rectangle2D (as I don't think we get Bounds for free...); it looks like Rectangle's impl has some static blocks that load/init some AWT toolkit/library stuff.

cjenkscybercom:
Do we want to actually implement position in this work as well @kasemir ?
Have we seen requests to be able to set window position in addition to size? Seems like a feature creep for this, unless this is a no-brainer feature that should be tacked on here? Otherwise I can just leave the capability in there for now but not have it used yet.

cjenkscybercom:
@kasemir I've implemented the position stuff now, using position if not the default values (0 and 0). 

But I do wonder, what if people want to use the upper left corner (0,0) of their display?
Could it work better to add a BooleanProperty to the Display widget, such as "Use pointer position" and possibly another "Use window dimensions"?

kasemir:
I think this will do. Yes, can't get exact 0, 0 position, need to use 0, 1 or 1, 0 but this is close enough.
It covers the use-size-from-display request, and now also supports converted EDM screens.

I don't think we need much tighter control or more placement hints because the general trend is towards viewing the displays via a web runtime, and then doing that on a laptop, tablet, phone. So in general trying to place something "just so" for a specific monitor will become a futile effort when every user has a different device.

Showing one screen at a time, "full screen", will become the most portable viewing mode. Opening "tabs" may sometimes be an option, but opening additional panels in specific locations really only works when you're in the control room on the exact same display setup.

Issue #2353, opened 2022-08-10
discussion - have complete settings.ini in target


tynanford:
I personally like to have a settings.ini file with the complete list of settings (everything on this webpage: https://control-system-studio.readthedocs.io/en/latest/preference_properties.html) in my CSS installation package. 

Then the site-specific things I want changed are uncommented and updated to whatever they may be. It also shows everything which is available to change and what the default is set to. Similar to how many config files in Linux under /etc look or for instance how Gitlab does their config: https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template

It's totally fine if no one else likes this. I'm just wondering what others think about it and if the phoebus build should create this full settings file

kasemir:
Way back in #4 I suggested that we use this type of preference access pattern:
```
Preferences prefs = Preferences.userNodeForPackage(...);
value = prefs.get(THE_NAME, default_value);
prefs.put(THE_NAME, value);
```

The idea was that you don't just read the current setting, but also write the setting back.
That way, as soon as your code once runs, the preferences are populated, initially with the default values.
So we would have a list of preferences.

After some discussions, we decided to NOT write the preferences.
You don't really know what's used and what their default values might be.
.. and you would not have any documentation about the use and meaning of preferences anyway.

Instead we have the agreement from https://control-system-studio.readthedocs.io/en/latest/preferences.html to put preferences into files named `*preferences.properties` with documenting comments, and then we locate all those files and combine them into the listing that's in the online help.

Yes, you could locate and parse those files into one long settings file.

What would you do with for example this?
```
# ----------------------------
# Package org.csstudio.archive
# ----------------------------

# RDB URL for archived data
#
# Oracle example
# url=jdbc:oracle:thin:user/password@//172.31.73.122:1521/prod
#
# PostgreSQL example
# url=jdbc:postgresql://localhost/archive
#
# MySQL example
url=jdbc:mysql://localhost/archive?rewriteBatchedStatements=true

# RDB user and password
# Some applications also provide command-line option to override.
user=archive
password=$archive
```

Extract just the settings into a format that can be used in the product, ready to set & un-comment selected lines?
```
#org.csstudio.archive/url=jdbc:mysql://localhost/archive?rewriteBatchedStatements=true
#org.csstudio.archive/user=archive
#org.csstudio.archive/password=$archive
```

Or also include the original descriptions:

```
# ----------------------------
[...]

PR #2390, opened 2022-09-16
Add create_settings_template.py - #2353

tynanford:
https://github.com/ControlSystemStudio/phoebus/issues/2353

Placed the script in phoebus-product but can move it if there is a better place. 

Writes to a settings_template.ini file and is run manually after the build. Unzips the jar files in <product> (defaults to ```./target/lib```) and checks for *preferences.properties files. Allows for including all the comments or not.

```
$ python create_settings_template.py -h
usage: create_settings_template.py [-h] [-c] [product]

Create template of settings.ini with all available settings

positional arguments:
  product         Location of product jars. Defaults to ./target/lib

options:
  -h, --help      show this help message and exit
  -c, --comments  Include settting comments in file
```

Example output files here:

[settings_template_comments.txt](https://github.com/ControlSystemStudio/phoebus/files/9579611/settings_template_comments.txt)
[settings_template_no_comments.txt](https://github.com/ControlSystemStudio/phoebus/files/9579612/settings_template_no_comments.txt)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,8 @@ buildNumber.properties
 # Avoid ignoring Maven wrapper jar file (.jar files are usually ignored)
 !/.mvn/wrapper/maven-wrapper.jar
 
+# Exclude complete settings template file if it is created
+phoebus-product/settings_template.ini
 
 # End of https://www.gitignore.io/api/maven,eclipse
 /.gitignore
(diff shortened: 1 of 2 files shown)


kasemir:
Works fine, thanks!

In my case it takes about 30 seconds and creates a template file for 37 packages with 380 lines respectively 1200 lines when including comments.
Shows how configurable we are, and how extensively the options are commented. 
Sure beats looking for those settings all over the source code.

PR #2391, opened 2022-09-16
Add -v option to log inspected jars, detected properties files

kasemir:
... and packages.
Since tool from #2390  may run for 30 seconds, this option provides feedback on the progress of parsing the files.

### DIFF
--- a/phoebus-product/create_settings_template.py
+++ b/phoebus-product/create_settings_template.py
@@ -3,6 +3,7 @@ import glob
 import zipfile
 import os
 import shutil
+import sys
 
 
 def create_settings_template(product_location: str, include_comments: bool, verbose: bool) -> None:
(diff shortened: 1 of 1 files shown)


tynanford:
I like this update. While the PR is open, can you add a check for ```len(jar_file_list) == 0``` and print a message that no jar files are found?  I thought that could be a hint that a build is needed first

kasemir:
Since it's Friday..
I noticed that you're using type hints `(product_location: str, ...) -> None`.
Do they actually do anything, or is python support for type hints still mostly for documentation?
Are you using an editor where you get warnings when you pass parameters of the wrong type?

tynanford:
Thanks for adding that.

Yeah they still don't do anything. I started to use them about a year ago mostly for documentation and a little hint to others... and because I personally think it would be nice if they were more standard. They can be nice if you use Pycharm too.

But it looks like the python devs specifically have said they won't ever mandate type hints: https://peps.python.org/pep-0484/#non-goals

PR #2396, opened 2022-09-21
Fix import of BOY files that carry no widget version=1.0.0

kasemir:
Old BOY files that just contain `<widget typeId=...` without a `version="1.0.0"` attribute are assumed to have the current version, and patches for older files were then not applied, resulting in wrong displays.

One fix is to simply open the BOY files in the old editor and save them again to write in a BOY format that adds version info.
This patch checks for no "version" but "typeId=org.csstudio..." to assume they are old files.


kasemir:
@tanviash thanks for testing and "this works great" email

PR #2405, opened 2022-09-30
Adding support for searching based on alarm tree root

shroffk:


georgweiss (approved):
Seems to work as expected

PR #2409, opened 2022-10-06
Alarm logging service

shroffk:
cleaning up the alarm server elastic queries

TODO:
how the search query string and parameters are handled can be simplified... for the time being I am just adding the root key word and making the minimum changes needed to get the root search support working


georgweiss:
Sorry, my bad.

PR #2413, opened 2022-10-12
Allow for building the core-pva module with jdk8

shroffk:
DO NOT MERGE

testing support for jdk8 core-pva builds using maven

### DIFF
--- a/core/pva/pom.xml
+++ b/core/pva/pom.xml
@@ -38,6 +38,15 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.6.2</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+          <fork>false</fork>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 </project>


kasemir:
With JCA, wasn't there something network related in the JRE libs that's different between 8 and higher, so binaries built with 8 won't run in a higher JRE?
Maybe add a profile for use of JDK8 when building just core-pva, but otherwise use 11+ from the parent pom?

shroffk:
I was trying to avoid a profile...with the above changes only the core-pva should be built with jdk11.

On my locally setup I was able to build the src but had issues with building the tst for core-pva using jdk8.

shroffk:
@kasemir should we merge this?

kasemir:
You wrote about issues with tests. Does it otherwise build and run?

shroffk:
I does so long as you build without tests ( this means you don't just skip the tests but instruct maven to not build the test classes either )

`mvn clean verify -Dmaven.test.skip`

shroffk:
I have not seen any issues with the rest of the Phoebus build or runtime ( limited testing with local softIocPva )

kasemir:
Go ahead then!

georgweiss:
This change causes issues in IntelliJ: test classes NetworkTest and MulticastDemo fail to compile as they call methods not available in Java 1.8.  Apparently the source and target Java versions are considered when IntelliJ builds these test classes. 

A maven build under Java 1.8 will also fail.

shroffk:
I did not check the IDE setups. Even with intellij I use the maven actions

For the maven build the key is to `-Dmaven.test.skip` with jdk8. I will double check

PR #2416, opened 2022-10-13
Spring rest logging

shroffk:


georgweiss (approved):
Looks fine.

shroffk:
@georgweiss 
A bunch of tests are failing because of the following

Expected :application/json;charset=UTF-8
Actual   :application/json

It seems like setting the content type with encoding info has been deprecated... can I update the tests to match the new content type

georgweiss:
Sadly this seems to be a tricky one...

First: we need to use latest version of flyway-spring-test (7.0.0) to be compatible with latest Spring Boot 2.7. But that is not sufficient. Seems the H2 database syntax we have in the flyway scripts now are invalid, which comes as a surprise. Fixing this would have quite some impact, and also require further migration scripts since database columns would have to be renamed.

Quick fix right now is to annotate DAOTest as Ignore.

Better fix is maybe to put this on ice and await my work with the Elastic migration, which removes all RDB stuff.

shroffk:
@georgweiss 
I think it makes sense to just ignore these tests for now.

Especially since you are working on the non RDB version of save restore which would makes these tests redundant anyway.

shroffk:
@georgweiss should I merge this?

PR #2417, opened 2022-10-14
Csstudio 1414: Enable Grid Color for Plot widgets and Data Editor

cjenkscybercom:
Enable grid+axis color for the Data editor (already has axis color, just make grid color match it)
Enable grid+axis colors on the X/Y Plot and Stripchart widgets

### DIFF
--- a/app/display/convert-edm/src/main/java/org/csstudio/display/converter/edm/widgets/Convert_xyGraphClass.java
+++ b/app/display/convert-edm/src/main/java/org/csstudio/display/converter/edm/widgets/Convert_xyGraphClass.java
@@ -40,7 +40,9 @@ public class Convert_xyGraphClass extends ConverterBase<XYPlotWidget>
 
         convertColor(r.getBgColor(), widget.propBackground());
         convertColor(r.getFgColor(), widget.propForeground());
-        convertColor(r.getFgColor(), widget.propGridColor());
+        widget.propYAxes().getValue().forEach(propYAxis -> {
+            convertColor(r.getFgColor(), propYAxis.color());
+        });
 
         // X Axis
         if (r.getXLabel() == null)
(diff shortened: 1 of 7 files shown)


cjenkscybercom:
One issue I'm still working out is that that the StripChart widget has a "Color" property that *was* being used for the grid (whereas now each y-axis defines its own grid color). 

For legacy widgets, I can see there was some kind of use case for handling that property, but keeping it now may be confusing as it doesn't ofc do anything anymore in this PR. 

What I'm wondering, is if there is a way to retain the property but suppress its display in the DisplayEditor properties window?

kasemir:
> What I'm wondering, is if there is a way to retain the property but suppress its display in the DisplayEditor properties window?

The original BOY widgets used to have version 1.0.0.
Our widgets now tend to be version 2.0.0 (https://github.com/ControlSystemStudio/phoebus/blob/783db2bb481715137973da45d6a843389a4fd491/app/display/model/src/main/java/org/csstudio/display/builder/model/Widget.java#L249), but you can increment the version for selected widgets, like https://github.com/ControlSystemStudio/phoebus/blob/783db2bb481715137973da45d6a843389a4fd491/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/TextEntryWidget.java#L224

In general, however, we want to be very lenient and allow display files where somebody manually created `<display><widget type=label>...` without any version info, so we're "duck typing".

1) Increment the version number of the StripChart if you like
2) Remove the properties that you no longer want
3) Add the new properties
4) Add a custom "WidgetConfigurator" (see for example Action Button) and in there check if the old property happens to be found in the XML. If so, transform it into the new property, like in https://github.com/ControlSystemStudio/phoebus/blob/783db2bb481715137973da45d6a843389a4fd491/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/ActionButtonWidget.java#L142-L143

kasemir:
StripcharWidget already has a custom WidgetConfigurator, so you can simply add to its `configureFromXML`

cjenkscybercom:
Thank you @kasemir ,

The extra property was only on the x/y widget, however your explanation was really helpful! I've pushed up the changes and did a small test manually for e.g. the case where a v2 x/y plot widget is upgraded to v3. Would you mind please taking a look, and let me know if there's anything I've missed?

Thanks,

Tina

kasemir:
I think that looks correct. One thing to check: Can it load a file where the widget has no `version="..."` because it wasn't actually created in our editor but by some screen-generating script that only populated the bare necessities.
While checking the version number is strictly correct, and we certainly have widgets that do that, my leaning now is that just checking if there is a "color" or "grid_color" or .. and making decisions based on that tends to be most robust.

PR #2429, opened 2022-10-27
protobuf dependency update

georgweiss:
Turns out that a version update for protobuf might be sufficient. Basic testing does not show any errors, archiver data is retrieved as expected.


shroffk:
The update of the proto buffer dependency fails.
We are using the archiver appliance snapshot from 17-December-2021

shroffk:
currently, I do not see any error messages or logs on the client side.. it simply fails silently :(

georgweiss:
Sadly I now get the same disappointing result. Obviously my test was not setup properly.
Suggest we close this PR.

shroffk:
Ok,
I have this working too.
Using the prebuilt protoc binaries (for windows) I was able to rebuild the pbraw client and use it to retrieve pbraw data from our production archiver.
TODO: test with archived waveform pv's
@georgweiss we should update the phoebus dependencies to use the new client ( include the new jar in the install dependencies module )

shroffk:
completed testing with waveform pv's too

PR #2430, opened 2022-10-27
Thumbwheel

shroffk:
@kasemir @tanviash new thumbwheel widget


tanviash:
Works well with FRIB's legacy boy thumbwheel usecases.

kasemir:
Good!

In the ThumbwheelWidgetRepresentationService, you don't need the 'unknown_widget'.

The SPI for contributing a widget model/runtime/representation is clearly optimized to contribute multiple widgets, since it asks for a list of types, or a map of type-to-representation. Maybe OK, since it mostly adds `List.of` or `Map.ofEntries` to an API for contributing just a single type.

kasemir (approved):

Issue #2427, opened 2022-10-26
PVA write exceptions


kasemir:
CA will throw exception on write when channel is not connected, PVA does not?
Also check write attempt to read-only channel.

PR #2431, opened 2022-10-27
PVA error handling

kasemir:
The PVA library didn't generate exceptions when reading/writing, #2427 

Now it will, using the same types of exceptions as channel access: `IllegalStateException` when not connected, plain exception with message that mentions an access or permission problem.

@georgweiss note that to get the exception when writing, one must use the `PV#writeAsync` call:
```
pv.asyncWrite(the_value).get();
```
or
```
pv.asyncWrite(the_value).get(5, TimeUnit.SECONDS);
```
For save/restore that should be OK because it needs to confirm that the writes completed anyway.

When using the plain `pv.write(the_value)`, we do now throw an exception if the channel is disconnected right now, but the channel might disconnect in the middle of the call, or there might be errors on the server side. We'll see all such errors in log messages, but the `write` call would appear just fine except for a disconnect at the time of the call.

The issue here is that Channel Access had a `put` and a `put callback` type of operation, and always knows about the current write access permissions. With PVA, there is just one type of `put`, and write access information is only available via an error when writing. There might be some keywords in the "request" that can tell the server to perform a `put callback`, i.e., delay the response until some action has been completed on the server side, but this is not accessible via a designated type of `put` nor documented in the protocol.
When using a `pv.asyncWrite`, we do wait for the reply, we will detect errors in the server reply and report them as exceptions. With a plain `pv.write`, however, we cannot wait for the reply, because otherwise the GUI which calls the plain `write` would block.

To better support exceptions in the plain `write`, a few things need to happen in the PVA protocol, IOC, QSRV:

1) Document a way in which the protocol can select between a plain write-and-forget and a write-complete that waits for some action on the server to complete
2) Add a way for the client to get updates on read/write access, so a GUI can indicate write permissions, and a plain write can right away abort in case of a read-only channel instead of having to wait for the error returned from the server.

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/pva/PVA_PV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/pva/PVA_PV.java
@@ -175,6 +175,16 @@ public class PVA_PV extends PV
         // in case some detail in the 'request' caused the PVA server
         // to perform a put-callback type of operation,
         // and a GUI calling write() expect an immediate return.
+
+        // Perform a disconnect check right now to alert caller
+        // of clearly disconnected channel
+        if (isDisconnected(read()))
+            throw new Exception("Channel '" + getName() + "' is not connected");
+
+        // The channel could still disconnect in the middle of the write,
+        // the channel may be read-only or experience other errors
+        // that we'll only see as log messages since we don't want to
+        // wait in 'get()' here...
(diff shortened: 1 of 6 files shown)


georgweiss:
Thanks @kasemir, works fine.

kasemir:
Will follow up with using plain `write` vs. `asyncWrite` that adds `record._options.block=true` and `...process=true`

PR #2432, opened 2022-10-27
Prepare timestamp userTag for 64 bit Long

kasemir:
EPICS developers meeting suggested update of userTag to 64 bit, and mentioned that phoebus won't handle it, since it expected a 32 bit value.

With this update,  we will handle a received 64 bit userTag without crashing. We still only use 32 bits from the userTag until the IOC and VType get updated, but are prepared for such an update


georgweiss:
Do you need me to test with a custom IOC where utag is 64bit?

kasemir:
Excellent! Again this will only use 32 bits until the VType gets updated, but at least we don't crash. Any number type will now be accepted for the userTag, so we're ready for the move from 32 to 64.

PR #2440, opened 2022-11-01
Migration to JUnit5

georgweiss:
This PR is to migrate all unit tests to JUnit5. The version to use is defined in top-level pom.xml, but future updates of (for instance) Spring Boot may pull in a newer version.

New/updated code depending on some unit test library (e.g. testfx) should make sure to exclude JUnit4, i.e.

```
 <exclusions>
    <exclusion>
        <groupId>junit</groupId>
        <artifactId>junit</artifactId>
    </exclusion>
</exclusions>
```

There are some changes developers should be acquainted with. I list them here for the sake of information. It is not an exhaustive list:

- Some test cases require a newer version of the Surefire Maven plugin (3.0.0-M7). This has been updated in the top level pom.
- ```Assert``` has been replaced by ```Assertions```.
-  ```@Before``` and ```@After``` have been replaced by ```@BeforeAll``` and ```@AfterAll```.
- Usage of ```assertThat``` has been updated to use the Hamcrest API in ```MatcherAssert```.
- Constructs like ```@Test(expected = SomeExceptionClass.class)``` are not supported. One may use ```assertThrows``` instead.
- ```@RunWith(SpringRunner.class)``` is replaced by ```@ExtendWith(SpringExtension.class)```
- A string message in for instance ```assertEquals``` has to be the last parameter, not first.

On top the adaptations to the JUnit5 APIs, this PR also includes a few updates to eliminate warnings.


kasemir (approved):

shroffk (approved):

shroffk:
There are some places where the import statements could be expanded instead of the use of wildcards

PR #2446, opened 2022-11-04
Add resize option to embedded display

georgweiss:
As per user request:

New resize option for embedded display: Size content to fit widget if larger.

The idea is to scale to widget size if the content is larger (width or height), otherwise no resize.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -101,6 +101,7 @@ public class Messages
                          PointType_X,
                          Resize_Container,
                          Resize_Content,
+                         Resize_Content_If_Larger,
                          Resize_Crop,
                          Resize_None,
                          Resize_Stretch,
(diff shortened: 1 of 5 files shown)


kasemir (commented):
This inserts a new option into the enumeration of resize options.
The enumerated options are actually saved as an index.
So this changes the meaning of indices beyond the `Resize_Content` value.

How does this handle older files that used `Resize_Crop` which will now be treated as `Resize_Content_If_Larger`?

Even if you move the new option to the end, how will older versions react if they see the new, previously undefined option?

Finally, will you also implement this in the web runtime?

My leaning is to simply not implement this.
Resizing nothing, or the container to fit the content, or the content to fit the container should suffice. Everything beyond is overcomplicating this.

shroffk:
At the very least I think this option should be the last in the list to maintain the index order. While this would not solve the problem of opening screens in older versions it would ensure that old opis are still properly interpreted 
I am trying to discourage the use of ALL auto resizing.
It is never perfectly what people want... while forcing the use to fixed size ensures that people design the screens and the embedded stuff in a more congruent fashion.

Finally, with many layers of embedded screens... Each with auto resizing... I have seen some terrible load time behaviors

kasemir:
Exactly. The options we have are somewhat necessary to load older *.opi files. But using them in new displays isn't a good idea, let alone adding more options

georgweiss:
Thanks for the feedback, points taken. Based on this feedback I'd question if Phoebus should support resize at all. Personally I find it confusing and have difficulties understanding the need. For legacy/converted OPIs where the resize policy must be updated one could fall back to updating the XML.

PR #2451, opened 2022-11-08
moving the jdk8 compatibility bits to a stand alone profile

shroffk:
This fixes the build issues with IDE's

Also, this profile is automatically activated when the building with jdk8 so no explicit instructions or documentation needed


georgweiss (commented):
Thanks

PR #2452, opened 2022-11-08
Save&restore Elastic migration

georgweiss:


shroffk:
Great...
Will start testing this.

shroffk:
building with openjdk 11.0.2 fails with

```
Compilation failure
An exception has occurred in the compiler (11.0.2). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program and the following diagnostic in your report. Thank you.
java.lang.NullPointerException
	at jdk.compiler/com.sun.tools.javac.comp.Flow$FlowAnalyzer.visitApply(Flow.java:1235)
```

shroffk:
tested with 11.0.17 and the build works...

georgweiss:
Looks like a compiler bug, see for instance https://bugs.openjdk.org/browse/JDK-8203195.

However, I have not been able to locate the code triggering this error. Moreover, building the service-save-and-restore module succeeds if run from that directory. Same with the services module.

Suggest we update the README to state minimum version 11.0.5 (haven't tried 11.0.3 or 11.0.4) be used, plus a reference to AdoptJDK.

shroffk:
@georgweiss I think I have a fix for the compiler bug triggered on 11.0.2

I have added the fix to this branch for you to test

PR #2456, opened 2022-11-10
Clarification of release process

georgweiss:


shroffk:
Hmm... we seem to have pulle din the svg test stuff.

PR #2458, opened 2022-11-10
Add the missing preferences for alarm background

shroffk:
Sorry, 
Missed committing this file


shroffk:
By default all the backgrounds are white (255,255,255) to match the current behaviour

kasemir (approved):

Issue #2461, opened 2022-11-14
Can't hide "Email" context menu


kasemir:
The email configuration allows disabling email support:
```
org.phoebus.email/mailhost=DISABLE
```

The display context menu, however, will still include "Send Email".
When invoked, a call to `EmailPreferences.isEmailSupported()` will notice that it's not supported, and nothing happens, which is confusing. Better would be not showing "Send Email" in the first place, or maybe a dialog "Email is not supported".

shroffk:
makes sense...

In the ContextMenuService, I can add a check for the `isenabled()`

PR #2462, opened 2022-11-14
#2461 check the "isEnabled" before adding context menu actions

shroffk:


kasemir (approved):

kasemir:
Thanks, the "Send Email" entry is indeed gone when I set the host to 'DISABLE'.

Issue #2423, opened 2022-10-17
Boolean button behavior with confirmation


georgweiss:
If the Boolean Button widget is configured as "push" and to show a confirmation dialog (on set), the button does not return to default value after positive confirmation. As this is different compared to running without confirmation I assume this is a bug, but want to confirm before I tackle it.

georgweiss:
Thanks, but not sure this helps me.
What is the intension using the boolean button in "push" mode? From the example OPIs it seems that 1 is written to the specified bit, then immediately afterwards 0 is written.

kasemir:
"Push" writes 1 on mouse-down, then 0 on mouse-release. The intended usage is for example moving a motor as long as the button is pushed down, then stop when released. So typically held down for a little while.

"Toggle" changes between 0 and 1 on each click. For example used to turn a power supply on/off.

georgweiss:
OK thanks.
So I guess using push together with confirmation dialog in practice turns the push mode to a toggle mode.

kasemir:
It can't really work with "push". You push down, and should keep it pressed down for a while. But you can't, because you need to handle the confirmation dialog...

georgweiss:
So maybe we should disable use of confirmation dialog with push?

kasemir:
Ideally, yes, but it's a little tricky. The editor fundamentally just lists widgets and their properties. It doesn't "understand" what they do. You can configure both the background and the foreground color of a text to be black, the editor won't explain to you why you no longer see anything. Likewise, the editor doesn't know about the Bool Button with its toggle mode vs. confirmation behavior. So this would have to be in the bool button model or maybe even representation: If it notices "push" with "confirm", it changes one or both of them to remain compatible, and the editor just shows the current settings.

georgweiss:
Well, if we can agree that "push" shall not support confirmation dialog then I can give it a try...

kasemir:
> "push" shall not support confirmation dialog 

Yes, I'd say there's no practical way to keep pushing down for a while and handle confirmation at the same time.

georgweiss:
So I looked into this and here is a proposal:

If user selects Push/Push inverted, the confirmation dialog setting will automatically set to "no". If user then tries to select confirmation dialog, the value stays unchanged. For this to work I need to change in EnumWidgetPropertyBinding#L42:

```
if (updating ||  jfx_node.isFocused())
``` 
to
```
if (updating)
``` 

[...]

PR #2463, opened 2022-11-15
Update to Boolean Button: confirmation dialog only for toggle mode

georgweiss:
This is for #2423.

An error dialog is launched "close" to the Boolean Button widget if user selects an unsupported combination of mode and confirmation dialog.


kasemir:
At first glance I'm missing some construct including `if (toolkit.isEditMode())`.
Will this always open a dialog, including in run mode?
It seems best to limit the dialog to edit mode.

georgweiss:
> At first glance I'm missing some construct including `if (toolkit.isEditMode())`. Will this always open a dialog, including in run mode? It seems best to limit the dialog to edit mode.

Well, since I assume that a change in mode or confirmation dialog properties happens only in edit mode, a check should not be needed.

kasemir (approved):

kasemir:
Turns out the new listener addition and removal don't match:
```
model_widget.propConfirmDialog().addPropertyListener(confirmDialogWidgetPropertyListener);
```
vs
```
model_widget.propMode().removePropertyListener(confirmDialogWidgetPropertyListener);
```

Need to remove from `propConfirmDialog()` where it was added.

Symptom is runtime error
```
 SEVERE ... Unknown listener
... at org.csstudio.display.builder.representation.javafx.widgets.BoolButtonRepresentation.unregisterListeners(BoolButtonRepresentation.java:226)
```

georgweiss:
My bad, slipped on the keyboard. Thanks @kasemir.

PR #2473, opened 2022-11-22
Fix for alarm background color for state OK

georgweiss:
The addition of alarm background color customization does not consider state OK.

This is a fix.


shroffk (approved):

georgweiss:
Well, we will for sure spend time evaluating any non-white backgrounds...

PR #2449, opened 2022-11-04
Alarm colour

shroffk:
Adding support for alarm backgrounds based on preferences

Currently, adding support in the alarm tree, table, and pvtree... updating more apps soon

alarm-tree-table

pvtree

PR #2475, opened 2022-11-28
Use JavaFX 17 compatible methods

pavel-ch:
As I am for several reasons using JavaFX17 (it is a LTS, 18 and 19 have issues with TableView.scrollTo()), I would kindly ask to use JavaFX17 compatible methods. Here methods of JavaFX18 introduced by #2449 commit e124d2a
Thank you @shroffk for review


kasemir (approved):
I'm for running/building with the more recent JFX (we currently set `<openjfx.version>19</openjfx.version>` in the top level pom), but no problem keeping the source code compatible with JFX 17.

shroffk:
Thank you...
Let us keep the javafx code compatible with javafx17

PR #2484, opened 2022-12-13
Alarm table background

kasemir:
The recent update towards common alarm severity colors #2449 misunderstood thus broke the `alarm_table_color_background` option.
That option was now _required_ to use the common alarm background color in the alarm table:
```
 if (AlarmSystem.alarm_table_color_background)
 {
     setBackground(AlarmUI.getBackground(item));
     setTextFill(AlarmUI.getColor(item));
 }
 else
     setTextFill(AlarmUI.getColor(item));
```

Instead, the alarm table should by default simply use the common alarm severity colors for both text and background.

That option, now renamed as `alarm_table_color_legacy_background`, makes the alarm table deviate from the rest by using the common alarm severity text(!) color for the background of the "Severity" cells. The text color is then simply white or black based on the brightness of that background.
This happened to be the original behavior in the Eclipse-based implementation to which local control room operators have become accustomed. The updated option name, and it being off by default, might discourage new sites from using it.

### DIFF
--- a/app/alarm/Readme.md
+++ b/app/alarm/Readme.md
@@ -25,9 +25,9 @@ kafka in `/opt/kafka`.
     cd examples
 
     # Use wget, 'curl -O', or web browser to fetch a recent version of kafka
-    wget https://dlcdn.apache.org/kafka/3.2.0/kafka_2.13-3.2.0.tgz
-    tar -vzxf kafka_2.13-3.2.0.tgz
-    ln -s kafka_2.13-3.2.0 kafka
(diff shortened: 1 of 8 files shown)


shroffk:
Sorry about that... I guess using the legacy in the property name does make things clearer.

kasemir:
No worries. This is just overall unfortunate.

To get the colored background, I tried using the new preferences, but that results in colored alarm backgrounds everywhere. The PV Tree then looks plain ugly, would need additional preferences to also adjust the color of those icons (similar issue with icons in alarm tree).

PR #2487, opened 2022-12-16
Update elastic mappings

georgweiss:
This update is needed in order to better support search in save-and-restore.

Also included in this PR some documentation on how to do data migration from RDB -> Elastic

### DIFF
--- a/docs/source/preference_properties.rst
+++ b/docs/source/preference_properties.rst
@@ -79,8 +79,14 @@ File ../../app/alarm/model/src/main/resources/alarm_preferences.properties::
    # Allows re-ordering as well as omitting columns
    alarm_table_columns=Icon, PV, Description, Alarm Severity, Alarm Status, Alarm Time, Alarm Value, PV Severity, PV Status
    
-   # Use background color to indicate alarm severity? Default: Text color
-   alarm_table_color_background=false
+   # By default, the alarm table uses the common alarm severity colors
+   # for both the text color and the background of cells in the "Severity" column.
+   #
+   # Older implementations always used the background to indicate alarm severity,
+   # and this options emulates that by using the alarm severity text(!) color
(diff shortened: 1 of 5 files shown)


shroffk (approved):

shroffk:
I am confused why the alarm colours in the ui preferences had to be added again.
Did I miss them in the old PR? I remember Kay mentioning this issue and I thought I had added them as defaults.

In anycase this should fix the missing defaults too then

kasemir:
I think it's not adding the ui preferences but the `docs/source/preference_properties.rst` documentation that lists them.
In previous PRs we updated the prefs but not the auto-generated docs.

shroffk:
Ahh!!!
Ok, I think I have lost the ability pay attention to details...

kasemir:
Since that part of the documentation is generated, we could exclude it from the repo (`.gitignore`)

shroffk:
@georgweiss Just an FYI
now that the PV's and readbackPV's are not keywords and instead text...the default tokenizer will split the PV names which have any of the special char's :{}<> etc... which will impact how searching works especially if you are searching for complete PV names

georgweiss:
Hmmm, as colon is used quite extensively we should maybe reconsider to revert back to keyword? But then users might complain why search on PV name will fail unless the proper case is used. Right?

shroffk:
You can update the elastic search code to be case insensitive 
like this in ChannelFinder
https://github.com/ChannelFinder/ChannelFinderService/blob/70846a2244cc88d799d3ac8927eac0801c4bc20a/src/main/java/org/phoebus/channelfinder/ChannelRepository.java#L412

georgweiss:
OK, so I suggest I revert back to keyword and add the channel finder strategy. Expect a PR soon.

PR #2488, opened 2022-12-16
display enum Labels in Databrowser

thomas1028:
Now Displays Labels for the VEnum data type in the Data Browser.
Also now determines display only for the first Data point and stores that for following data points to use
Screenshot_20221216_143735
original label
Screenshot_20221216_144422
new label


kasemir:
Great idea!
In all the other displays like Display Builder, Probe, PV Tree, .., however, we generally prefer to show the label instead of the number. So maybe the marker should also be "Normal (5.00)" instead of "5.00 (Normal)" to again emphasize that this is a labelled entity where the label is the primary representation, the the number is just used out of necessity to allow plotting it.

kasemir:
Actually, "Normal (5)" would be ideal: Show the label first, then the enum index, and we know that the index is always an integer, so no need to add decimals.

PR #2491, opened 2022-12-21
Support automatic updates from gitlab package registry

MichaelRitzert:
With this code, automatic updates can be retrieved from gitlab CI builds.

The update mechanism are implemented as different plugins detected via the ServiceLoader. All detected plugins are checked to find one that has all configuration in the ini file. This plugin will be used.

The current code that updates from plain URLs has been refactored but is functionally unchanged.

### DIFF
--- a/app/update/pom.xml
+++ b/app/update/pom.xml
@@ -34,5 +34,16 @@
       <artifactId>core-ui</artifactId>
       <version>4.7.2-SNAPSHOT</version>
     </dependency>
+    <dependency>
+      <groupId>javax.jms</groupId>
+      <artifactId>javax.jms-api</artifactId>
+      <version>2.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>javax.json</groupId>
+      <artifactId>javax.json-api</artifactId>
+      <version>1.1.4</version>
+    </dependency>
+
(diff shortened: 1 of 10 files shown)


kasemir:
I had put this into the 'review' section of the github web page, which doesn't always show up for everybody, so here it is again as a comment:

When you add dependencies, then please also add them to the target platform, see https://github.com/ControlSystemStudio/phoebus/blob/master/dependencies/README.md and https://github.com/ControlSystemStudio/phoebus/blob/master/dependencies/phoebus-target/pom.xml. With the current settings, we already have javax.json-api-1.0.jar, while you request 1.1.4. javax.jms-api would be a new one. In the end, please assert that all depedencies that you need are listed in the https://github.com/ControlSystemStudio/phoebus/blob/master/dependencies/phoebus-target/pom.xml and that we end up with one version for javax.json-api


In other words, this PR adds dependencies to javax.jms-api 2.0.1 and javax.json-api 1.1.4, which at the time conflicted with the versions in the target platform.

Meanwhile, #2528 happened. When building the target platform right now, it includes javax.jms-api-2.0.1.jar and javax.json-api-1.1.4.jar, same versions as in this PR, so I think we're all set!

PR #2496, opened 2022-12-23
Load settings from URL

georgweiss:
As per user request: if settings file starts with http(s), load from remote location. The http client side of this implementation is very basic and will only be able to handle an OK response with settings contained in the body. 

Limitations:

- Redirects, basic authentication, and other non-200 responses will not work.
- https must be backed by trusted certificate


kasemir (approved):
Looks good to me

PR #2498, opened 2022-12-27
Import code for Elasticsearch log viewer.

MichaelRitzert:


georgweiss (commented):
@MichaelRitzert, can you please add some background and purpose of the added app?

MichaelRitzert:
Sure.
We are storing all messages sent via the JMS topic used for logging purposes (in our case including logs from our IOCs) in an Elasticsearch database. This is the viewer part for phoebus. It merges archived data from ES with live data received via JMS and offers various filter options.

This corresponds to the code in
https://github.com/ControlSystemStudio/cs-studio/tree/4.5.x/applications/appunorganized/appunorganized-plugins/org.csstudio.logging.es
ported to JFX.

georgweiss:
Thanks for the clarification.

I have but one comment, it concerns the time range specification. In the Olog logbook app as well as in the alarm log app, the "start" and "end" search parameters imply time in the past, or "now". Assuming there is no data with a time stamp later than "now", one need not distinguish between positive and negative numbers. Please disregard this comment if this assumption is wrong.

In the name of consistency I think it would be an advantage to let users adopt the same time specification semantics in all applications supporting time specifiers like "8 hours" or "now".

MichaelRitzert:
> I have but one comment, it concerns the time range specification. In the Olog logbook app as well as in the alarm log app, the "start" and "end" search parameters imply time in the past, or "now". Assuming there is no data with a time stamp later than "now", one need not distinguish between positive and negative numbers. Please disregard this comment if this assumption is wrong.
> 
> In the name of consistency I think it would be an advantage to let users adopt the same time specification semantics in all applications supporting time specifiers like "8 hours" or "now".

You are talking about the "TODO" about the "-" that goes missing, I guess. (Not sure if this link works:
https://github.com/ControlSystemStudio/phoebus/pull/2498/files#diff-606c7d1a304c5e83b6d3e16dca77447cab2b7f0fb77ccec21d5691f66205cc0dR207 ).
Unfortunately, the situation for us is not that simple as "there are no timestamps later than now" (as seen from the user). Given that the log messages do not contain timezone information, there is sometimes confusion with remote access from other timezones. I'm looking at extending the log message format during our upcoming upgrade, but until then, that's where we stand.

You see that for now the button to open the selection dialog for the start/stop times is just foreseen as a comment. This is because I'm not that fluent in writing JFX UI code, but also because I couldn't find something suitable in the phoebus core libs. CSS had org.csstudio.apputil.ui.time.StartEndDialog. How about introducing this dialog and all the time calculation logic in a core module and then using that everywhere?

georgweiss:
I admit I did not consider time zone issues. I'm afraid this is a bigger topic than I can engage with for the moment.
However, for the sake of comparison I checked how Olog works. The service will store a log entry in UTC, and any time range provided by the client will be parsed by the service. So a start time like "8 hours" will be treated relative to the current time on the service, no matter the client's time zone. Client may however specify a "full" date/time string in the future on the format ``yyyy-MM-dd HH:mm:ss``.

shroffk:
> So a start time like "8 hours" will be treated relative to the current time on the service, no matter the client's time zone. Client may however specify a "full" date/time string in the future on the format yyyy-MM-dd HH:mm:ss.

timezones are irritating... and once you add daylight saving it becomes downright annoying.

I think in the past we have chooses to standardize all storage with UTC and then have the retrieval functionality on the service handle relative time descriptions and time zones like Olog does

shroffk:
If we need to support multiple timezones... I think the service should then default to UTC and let each client handle the time zones.
But this would mean the the computation of the relative times on the service side would not behave as expected...so the client would always have to search using absolute times

MichaelRitzert:
In countries with DST, there is one day in the year, where a time between 2am and 3am is ambiguous without the addition information "DST" or "no DST". So even here, time zone information is required. (Luckily, our main operation is in Japan, where there is no DST.)

Just storing everything in UTC by convention solves that problem. But by then, you are already in the business of time zone handling, so you could as well add the "Z" at the hand to make the convention obvious from the data.

Anyway, I think there is nothing bad about time zones, as long as times are properly formatted for the users. 

In this particular case, just changing the timestamp format from `yyyy-MM-dd HH:mm:ss.SSS` to `yyyy-MM-dd HH:mm:ss.SSSX` when querying the server would automatically send the query time range marked as being in the user time zone, and the server would make the correct comparison. So everything would be fine.

Well, mostly anyway. ES will return the date string exactly as sent, so if different formats are used to feed the database, the user might well end up seeing plain millis since the epoch… I'm currently trying a patch to additionally have ES send the time in the fixed "millis" format and then reformat for display to the user.

MichaelRitzert:
I changed the ES access code to use "millis since the epoch" in the query and response. Therefore, this part of the code is now timezone-safe. How things are interpreted on the ES server side depends on the content of the database.

The format of the timestamps displayed to the user is unchanged. Maybe this format could be made a common configuration option.

shroffk:
I think we can merge this
@georgweiss

PR #2505, opened 2023-01-13
CSSTUDIO-1794 Fix bug where the representation of boolean expressions becomes shared in the UI

abrahamwolk:
This PR fixes an issue where the representation of boolean expressions in the rule editor becomes shared between UI elements. To reproduce the issue:

1. Open the editor for control GUIs.
2. Create two buttons, button 1 and button 2.
3. Create a rule for button 1 by opening the rules editor in the "Properties" tab of the editor, and add a boolean expression.
4. Select button 1, and then select button 2 in additon to button 1.
5. Open the rules editor with both buttons selected.
6. Press "OK" in the rules editor window.

Now the program is in a state such that if only button 1 (or only button 2) is selected and the rules editor window is opened, and the boolean expression that was entered in the rule that was created in step 3 above is modified, then the change is reflected in the rule of the other button as well. The expected behavior is that both rules can be edited separately. Editing one the rule of one button should not affect the rule of the other button.

The cause of the problem is that in the function `action_handler()` in the class `RulesPropertyBinding` the result of editing the rules in the rules window is stored in the variable `Optional<List<RuleInfo>> result`, and then each selected widget (i.e., the two buttons in that were selected in step 4 above) is updated and set to `result.get()` using statements of the form `undo.execute(new SetWidgetPropertyAction<>(widget_property, result.get()));` (and similar).

`RuleInfo` contains the field `List<ExpressionInfo<?>> expressions`. At this point in the code, the type of the elements of `expressions` always seems to be `ExprInfoValue` and consequently `ExpressionInfo<?>` always seems to be instantiated to the type `ExpressionInfo<WidgetProperty<?>>` (however, this is not enforced in the types), and each instance carries with it an instance of `WidgetProperty<?>`, representing the entered boolean expression. As a consequence, each selected UI element will be updated so that the expressions entered in the UI are represented by the same instance of `WidgetProperty<?>`, and thus editing the boolean expression of one UI element affects the boolean expression of the other UI elements that share the same instance of `WidgetProperty<?>`. This PR fixes the issue by cloning the widgets before assigning the result to UI elements other than the first UI element that was selected, so that the instances of `WidgetProperty<?>` are not shared. (The PR also handles the case where `ExpressionInfo<?>` is instantiated to `ExpressionInfo<String>` (which is the case when the elements of `expressions` are of type `ExprInfoString`, which is the other possibility that currently exists), by reusing the same instances in that case.)

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/RulesPropertyBinding.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/RulesPropertyBinding.java
@@ -8,12 +8,14 @@
 package org.csstudio.display.builder.editor.properties;
 
 import java.text.MessageFormat;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
 import org.csstudio.display.builder.editor.Messages;
 import org.csstudio.display.builder.editor.undo.SetWidgetPropertyAction;
 import org.csstudio.display.builder.model.Widget;
+import org.csstudio.display.builder.model.WidgetProperty;
 import org.csstudio.display.builder.model.WidgetPropertyListener;
 import org.csstudio.display.builder.model.properties.RulesWidgetProperty;
 import org.csstudio.display.builder.model.rules.RuleInfo;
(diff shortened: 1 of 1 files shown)


kasemir:
Good one.
We definitely want 1-6, or in general: Selecting several widgets means you're then editing their common properties. We also want to then set all their rules to the same ones, but yes, each one needs to receive a copy of the same rules, not a reference to the same rules, because for later edits they could potentially be adjusted separately.
Getting an exact copy of the rule settings could be a little harder, need to check. Maybe support 'clone' on the rule settings?

abrahamwolk:
Thanks! What do you mean with "getting an exact copy of the rule settings"? As far as I could tell, only the "Value" column of boolean expressions becomes shared. (And that's what the PR fixes.) Do you think more fields need to be cloned or duplicated?

kasemir (approved):
This might be more elegant if we implement `RuleInfo.clone()`, but it looks like you're doing the right thing, should certainly fix the issue.
Thanks!

PR #2500, opened 2023-01-04
Add helper class for epics:nt/NTScalar:1.0

jacomago:
Adds helper classes for the [NTScalar type](https://docs.epics-controls.org/en/latest/specs/Normative-Types-Specification.html#ntscalar)

Includes helper classes for:
 Alarm
 Control
 Display
 Enum

With constructors and from PVStructure methods. 
Also refactors the PVATimeStamp class to match.

Tests added:

A few small smoke construction tests.

A Server client test
  Test sets up a server with a NTScalar pv and sends some fake data for all Scalar sub types (except enum)
  Test then checks that a client can recieve events and asserts they match.

An integration test for running with a local version of Epics. Checks that can parse a live ioc with the helper classes.

PR #2507, opened 2023-01-17
Phoebus Olog - Use Group icon instead of color

cjenkscybercom:
style: indicate group / conversation status of search results in olog via icon instead of color.
This is to match the style changes in the web client


cjenkscybercom:
Hmm...Not sure why ServerClientTest#testSinglePV would have failed?

kasemir:
Did you figure out what exactly failed in ServerClientTest?
Once #2500 no longer showed errors, I merged it and then got a notification for an after-merge test failure.
Test runs fine when I run it locally, but something in there must be fragile.

georgweiss (approved):

cjenkscybercom:
> Did you figure out what exactly failed in ServerClientTest? Once #2500 no longer showed errors, I merged it and then got a notification for an after-merge test failure. Test runs fine when I run it locally, but something in there must be fragile.

I could take a look, but it'd probably be better for whoever was working on it last to take a look. I also have other priorities taking precedence at the moment, so I wouldn't get to it for a while.

georgweiss:
I'm seeing the same failure for a different PR. To me it looks like the test cases have some hard coded timeouts that may be too short when tests run in Github.

Will ask developer to reconsider/revisit the failing test case.

PR #2510, opened 2023-01-19
Reduce timeouts in ServerClientTest

jacomago:
Uses a single client and server in the test
for all of the pvAccess data types, rather
than one per data type.


kasemir:
That might help, thanks!

PR #2511, opened 2023-01-20
Use a longer timeout

jacomago:
Hopefully this will reduce chances of test failure due to slower running of tests on the github ci machine.

### DIFF
--- a/core/pva/src/test/java/org/epics/pva/combined/ServerClientTest.java
+++ b/core/pva/src/test/java/org/epics/pva/combined/ServerClientTest.java
@@ -247,7 +247,7 @@ public class ServerClientTest {
 
         PVAChannel channel = client.getChannel(pvName);
         try {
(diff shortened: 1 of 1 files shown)


kasemir:
Just failed with
```
Error:  org.epics.pva.combined.ServerClientTest.testSinglePV(List)[9]  Time elapsed: 0.977 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <9> but was: <10>
```

Is that just about timeouts, or something else?

jacomago:
I now think this might be sometimes the client receives the creation of the pv as an event and sometimes it doesn't. Will investigate more on Monday.

jacomago:
Can reproduce locally now. By adding a timeout between the subscribe and sending the first update, the listener will get the setup of the pv as an event.

kasemir:
"All checks have passed" for once, so this is certainly better

PR #2516, opened 2023-01-24
exclude logback dependency

mgotz:
The added JTango dependency in core/pv pulls in a logback dependency. This causes warnings like
```
SLF4J: Class path contains multiple SLF4J bindings.
```
Phoebus includes the slf4j-jdk14 as a binding to push all logging to java.util.logging. So, I believe logback should be removed here.


kasemir (approved):
Looks good to me

PR #2517, opened 2023-01-24
Fix settings without effect in alarm server

mgotz:
The alarm server includes in its main the AlarmSystem class. This causes all preferences in there to be initialized directly from defaults or previous values.
Passing other settings with `-settings my-settings.ini` has no effect, because they are added to the preferences only after the static variables are already initialized. A re-start of the alarm-server usually resolves this, because then the settings will already be in the user's perferences.
This affects the preferences under org.phoebus.applications.alarm such as the nag_period or the automated_email_sender.

The alarm server main only needs some constants from the AlarmSystem class. So, I introduced an AlarmSystemConstants class to hold the compile time constants. AlarmSystem inhererits these and the alarm server can include only the constants, thus deferring initializiation of the preferences until after the preferences have been updated with any passed settings.

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
@@ -11,7 +11,6 @@ import java.io.File;
 import java.time.LocalDateTime;
 import java.time.temporal.TemporalAmount;
 import java.util.logging.Level;
-import java.util.logging.Logger;
 
 import org.phoebus.applications.alarm.client.IdentificationHelper;
 import org.phoebus.framework.macros.MacroOrSystemProvider;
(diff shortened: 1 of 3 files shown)


kasemir (approved):
> `-settings my-settings.ini` has no effect because ... static variables are already initialized.

Good catch. With the GUI `Launcher` I don't think we have this problem because `-settings..` are handled before creating the `PhoebusApplication` which in turn loads applications and triggers their (static) initialization.

Thezscruning:
Now I'm throwing into a situation where the preference settings fail to load when using the alarm server, I am now running kafka and alarm server on a linux server, then phoebus running on windows, but when the preference ip:9092 and nag_period=00:00:10 of kafka are pointed, I find that kafka's service can be connected, but nag_ period has been the default 15 minutes of the software, I checked the source code and didn't find the problem.

PR #2518, opened 2023-01-24
Missing mapping for plot_area_background_color for BOY XYGraph

tanviash:
When BOY XYGraph is converted to XYPlot or StripChart, BOY property "Plot Area Background Color" is not mapped correctly to the new widget's background color. This PR fixes the issue.


kasemir:
Looks good, thanks!

PR #2519, opened 2023-01-24
Legacy linking container transparent behavior

tanviash:
In BOY, the Linking Container behavior is always transparent, although there is no parameter for it.  During the conversion from OPI to BOB, the “Transparent” setting however is disabled. This PR fixes it.


kasemir:
Never noticed since all our screens use the default background color...
Looks good, thanks!

PR #2523, opened 2023-01-30
Get type pvaccess

jacomago:
Get the type of pvData from a PVAData. This is the top level type, for example structure or double[] rather than enum_t for example.

Also includes an extra integration test to check reading scalars.

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/data/PVAStringArray.java
+++ b/core/pva/src/main/java/org/epics/pva/data/PVAStringArray.java
@@ -124,7 +124,7 @@ public class PVAStringArray extends PVAData implements PVAArray, PVAValue
     @Override
     public String getType()
     {
-        return "string";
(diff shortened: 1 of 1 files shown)


core/pva/src/main/java/org/epics/pva/data/PVAStringArray.java
```diff
@@ -122,10 +122,9 @@ protected int update(final int index, final PVAData new_value, final BitSet chan
     }
 
     @Override
-    protected void formatType(final int level, final StringBuilder buffer)
+    public String getType()
     {
-        indent(level, buffer);
-        buffer.append("string[] ").append(name);
+        return "string";
```
kasemir, line 127: String array should return "string[]", not "string"

kasemir:
Extracting `getType` from `formatType` looks good, just a typo on the string array

Issue #2508, opened 2023-01-17
Databrowser "Statistics" indicate wrong "Mean"


kasemir:
Local user found a situation where the "Mean" appears to be outside the "Min.." .. "Max Value":

![thumbnail_Screen Shot 2023-01-17 at 11 17 50](https://user-images.githubusercontent.com/1932421/212956763-5f6494df-29c5-46eb-b647-97a4b50a0e47.png)

kasemir:
Hard to read, but the red trace shows Min -0.11..., Max 0.12... with Mean 7.18....

georgweiss:
Could it be that there is an exponent for the mean value?

georgweiss:
Of course we could make the columns wider, I'll look into it.

kasemir:
Yes, "7.18E-2" = 0.07 would be an explanation. Unclear if that instance is still running to verify

kasemir:
User was able to change the column width on that instance, and the Mean was indeed "7.18...E-4".

Yes, having the column widths will the width of the pane would have prevented this misunderstanding, though you'll then still be able to make the window size too small to show all values, so there's no perfect solution.

Anyway, nothing severely wrong here.

kasemir:
Yes, thanks!

PR #2524, opened 2023-01-31
CSSTUDIO-1812: Fix width of columns in the "Statistics"-tab

abrahamwolk:
The primary purpose of this merge-request is to increase the width of the columns containing `double`s in the "Statistics"-tab so that the full value is shown. The intent is to prevent the user from mistakenly reading the wrong order of magnitude since the exponent is shown using a postfix-notation at the end of the printable representation of the `double` in question. E.g., if the number to be shown is `1.234E8`, then, if this output is truncated, the user may only be shown `1.23...`, and mistakenly think that the order of magnitude is eight powers of ten less than it actually is.

To prevent this, this merge-request increase the `minWidth`- and `prefWidth`-properties of the columns containing `double`s to high enough values so that any double should be printable.

Other changes that are introduced in this merge-request are:

 - The width-fields of the columns "indicatorColumn", "displayNameColumnn", and "countColumn" are also updated in this merge-request, to improve the user-experience.
 - In "StatisticsTab.fxml", the "GridPane" was changed to "VBox" and a "ScrollPane" was added around the "TableView". This is to enable scrolling in the TableView when the window is too small to show the entire TableView.
 - The types in the view model of the "TableView" in "StatisticsTab.fxml" (i.e., the variables of type `TableColumn<S, T>` in the class `StatisticsTabController`) were changed so that the content of the cells in the table (i.e., the `T` in `TableColumn<S, T>`) is typed according to the data that it represents.
 - The types in the data model of the "TableView" in "StatisticsTab.fxml" (i.e., the class `ModelItemStatistics`) were updated so that the types of the elements reflect the types of the data that they hold. (In other words, the `S` in `TableColumn<S, T>` was also updated.)

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/properties/StatisticsTabController.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/properties/StatisticsTabController.java
@@ -37,6 +37,8 @@ import javafx.application.Platform;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.SimpleDoubleProperty;
+import javafx.beans.property.SimpleLongProperty;
 import javafx.fxml.FXML;
 import javafx.scene.control.Button;
 import javafx.scene.control.Label;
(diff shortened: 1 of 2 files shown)


georgweiss (commented):
Not sure I agree to change String -> Double. Reason is that initial values in the columns become 0.0, which is in a sense incorrect. With String setting values to null would simply not render anything.

@kasemir, what's your take on this?

Screenshot 2023-01-31 at 15 30 41

kasemir:
> Reason is that initial values in the columns become 0.0, which is in a sense incorrect.

Using `Double` for numbers seems better than `String`, but the side effect of starting out with all "0.0" when you really have no data is indeed incorrect. "Count=0" is good for not having any data, but the other stats should then be "NaN" or empty, not "0.0".  So maybe keep the `Double`, but reset some columns to Double.NaN instead of 0.0? Or do go back to String if that was easier to use.

abrahamwolk:
Those are good points. I have updated the code to initialize the doubles to `Double.NaN` instead of `0.0`.

georgweiss (approved):

kasemir:
Thanks, fixes #2508

PR #2534, opened 2023-02-06
CSSTUDIO-1826 Add description of the calculation of the statistical measures to the documentation

abrahamwolk:
This PR adds a description of the calculation of the statistical measures under the Statistics-tab in the Data Browser to the documentation.

The intention is to clarify the meaning of the statistical measures that are shown in the UI. In particular, it is emphasized that the timestamps data-points are *not* taken into account in their calculation.

### DIFF
--- a/app/databrowser/doc/index.rst
+++ b/app/databrowser/doc/index.rst
@@ -443,6 +443,6 @@ indirectly for determining the subset of data-points to calculate the statistica
 individual data-points are part of the calculation.
 
 E.g., the statistical mean will be skewed towards more recent data-points if the sample frequency of recent data-points
-is higher than the sample frequency of less recent data-points. This may, e.g., occur when the Data Browser displays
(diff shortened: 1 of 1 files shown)


georgweiss (commented):
The second paragraph does in my view say the same thing twice. Can we make it a bit more concise? Actually I think the first sentence is quite sufficient.

abrahamwolk:
> The second paragraph does in my view say the same thing twice. Can we make it a bit more concise? Actually I think the first sentence is quite sufficient.

I wanted to emphasize the situation where the data under consideration is both from the archiver (less recent) and from the PV itself (more recent), since it is a very typical situation. The intention is that the user should not have to spend time to "debug" the issue if it occurs, but instead be aware of it after they have read the documentation.

I will try to rephrase the last paragraph, and try to make it more concise.

abrahamwolk:
I have rephrased the second sentence of the last paragraph now; what do you think?

georgweiss (approved):

PR #2538, opened 2023-02-07
Update JavaFX to 19.0.2.1

kasemir:
.. simply because that's the current one, while plain 19 is about 5 months old. Release notes under https://github.com/openjdk/jfx/tree/jfx19/doc-files mostly point to "not public" security fixes.


kasemir:
Superseded by JFX 20, #2638

PR #2556, opened 2023-02-20
Updates in alarm logger

georgweiss:
Some updates to both service and client:

- Search where user specifies "root" (= alarm config) should be limited to matching Elasticsearch indices. 
- Search for config information to search config messages, i.e. messages in <alarm config>_alarms_config* indices, fixed.

Note that search for config messages may return empty result.


kasemir (approved):
@shroffk might want to check, but looks fine to me

shroffk (approved):

shroffk (approved):

PR #2562, opened 2023-02-26
Alarm server XML doc fix

tynanford:
As far as I can tell, the alarm server XML schema doc is incorrect for the "detail" element for Guidance, Display, Automated Action, and Command. It should be `<details></details>`

This updates the XML schema doc and adds a test for Guidance and Display in AlarmModelReaderTest.java


kasemir:
Thanks!

PR #2564, opened 2023-02-27
Add server port to save restore properties

tynanford:
Just to make it clear that this is a setting and can be changed


georgweiss:
Thanks @tynanford, this should of course be present in the properties file.

georgweiss (approved):

PR #2567, opened 2023-03-01
CSSTUDIO-1827: (Data Browser) Fix the check for when exponential notation should be used for the labels of the ordinate

abrahamwolk:
This pull request fixes the check for when exponential notation should be used for the labels of the ordinate. Previously, the absolute difference of the parameters `low` and `high` was considered, and the check failed to account for the case where values are clustered relatively narrowly around large numbers (for example, around 1000000000).

This PR changes the check to check for two conditions:
1. If at least one of the parameters `low` or `high`is of large order of magnitude, then scientific notation is used.
2. If both `low` and `high` are of small orders of magnitude, then scientific notation is used.

If neither of these two conditions hold, then normal notation is used.

The initial value of `exponential_threshold` has been changed from `5` to `4`. In this implementation, `exponential_threshold == 4` means that numbers greater than 10^5-1 (i.e., with more than 4 "trailing decimals") and numbers less than 10^-4 (i.e., with more than 4 leading zeroes) are represented using scientific notation.

Tests have also been implemented for the check for whether scientific notation should be used or not. As part of this implementation, the class `LinearTicksTest` was moved from `org.csstudio.javafx.rtplot` to `org.csstudio.javafx.rtplot.internal`. The fields `buf` and `gc` of the class `TicksTestBase` had to be made `protected` to be accessible from `LinearTicksTest` after the move.

### DIFF
--- a/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/internal/LinearTicks.java
+++ b/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/internal/LinearTicks.java
@@ -50,7 +50,7 @@ public class LinearTicks extends Ticks<Double>
     protected NumberFormat detailed_num_fmt = createDecimalFormat(2);
 
     /** Threshold for order-of-magnitude to use exponential notation */
-    private long exponential_threshold = 5;
+    private long exponential_threshold = 4;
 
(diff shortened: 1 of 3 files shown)


georgweiss (commented):
Looks fine as far as I can determine.

kasemir (approved):
Sure, let's give this a try.

Issue #2506, opened 2023-01-13
Feature Request: Make waveform PVs in 1D table widgets "just work™"


daykin:
Current behavior:

If I plug a waveform PV into the pv_name field of a Table widget, it is interpreted as "Row 1, Column 1 is set to an array of numbers". I can work around this with PVUtil.GetStringArray(), casting to an array of 1-element arrays.

Desired behavior:

I plug in a waveform PV, and each element in the waveform is resolved to its own cell.

daykin:
I will be attending the March codeathon at Diamond. I'll plan on addressing this.

kasemir:
The `VTable` type that the table widget uses internally is column-oriented, https://github.com/epics-base/epicsCoreJava/blob/master/epics-vtype/vtype/src/main/java/org/epics/vtype/VTable.java

This would make it quite easy to turn a waveform PV, for which you already have a `VxxxArray` type, into a `VTable` with that data as the one and only column. Can be done in a zero-copy implementation that just calls `VxxxArray.getData()` from the wrapping `VTable.getColumnData(0)`.

I'm afraid, however, that most users would naively expect to see a single-row table with one column per array element, i.e., a "horizontal" table instead of a "vertical" one. That'd require creating a `VTable` with many columns and copying each array element into a single-element array for that column...

daykin:
@kasemir Is the second commit (a8b76) what you had in mind? I had done this another way then saw your comment.

PR #2571, opened 2023-03-07
Make Table Widget Render 1D arrays as 1-by-n VTable

daykin:
#2506 - 47f3b copies all elements manually, a8b76a wraps the data in a VTable. Let me know which you prefer and I will squash the other.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/TableWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/TableWidget.java
@@ -51,6 +51,7 @@ import org.epics.util.array.ListDouble;
 import org.epics.util.array.ListNumber;
 import org.epics.vtype.VTable;
 import org.epics.vtype.VType;
+import org.epics.vtype.VStringArray;
 import org.phoebus.framework.persistence.XMLUtil;
 import org.w3c.dom.Element;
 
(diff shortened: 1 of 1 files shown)


kasemir:
I slightly prefer the second one which wraps the array into one column of a VTable, but just slightly.
It wraps the array without copying each element. It might actually work with string as well as number arrays.

On the other hand, the data is then copied from a VTable into a `List<List<String>>` anyway, and when you create the `List<List<String>>`  yourself, you can create it such that the array elements turn into columns, which I feel is the expected display for an array, one wide table with columns, and not one high table with just one columns.

daykin:
> On the other hand, the data is then copied from a VTable into a List<List<String>> anyway

I must admit I didn't pay much attention to what is happening after wrapping to a VTable. My logic was "We get a list, wrap it into a VTable, and do whatever is already done with VTables". Giving it a second look, this could be done better.

>you can create it such that the array elements turn into columns, which I feel is the expected display for an array, one wide table with columns, and not one high table with just one columns

Maybe this is my Data Science background talking, but I tend to envision things as the opposite. The convention in, for example, R and Pandas, relational databases, etc. is to have each column be an attribute, and each row a specific instance of that attribute. Plus, if I'm passing in a 2,000 element list, it could quickly get clunky to scroll through an arbitrarily wide table.

daykin:
Looking a bit deeper into this,  the underlying JFX representation is ultimately a `ObservableList<List<ObservableCellValue>>`, so I don't see an easy way to avoid, at some point, copying the data into a list of lists. Maybe it's pointless to wrap it in a VTable first. Stepping back a bit, could you help me understand the rationale for deep-copying all incoming data? Does it have to do with alerting listeners?

kasemir:
The deep copy is there for two reasons:

Avoid changes to widget's value. If you 'get' the value and then change it, while the JFX widget is at the same time using it, who know what'll happen.

Next, if a script fetches the value, modifies it, then sets that as a new value, passing out a deep copy asserts that the 'new' value is indeed recognized as new. When referencing the old value and changing it, the widget would still overall have the same value reference and not notice the change.

kasemir:
There is no need to be hit by a bus, in fact I'd advise against it.
But you may win the lottery, in which case it should be easy to pay somebody else to take over maintenance.
Thanks!

PR #2572, opened 2023-03-08
Provide visual feed-back if alarm actions fail

georgweiss:
A site may configure the Kafka alarm service to allow producer messages only from certain IP addresses. This can be useful if alarm information is of interest to a wide audience, but actions should be allowed only from a limited context.

As this the current setup at ESS we discovered that blocked producer messages failed as expected, but left the user confused as nothing changed in the UI. This PR adds some code to display an error dialog in such cases.

### DIFF
--- a/app/alarm/datasource/src/main/java/org/phoebus/pv/alarm/AlarmContext.java
+++ b/app/alarm/datasource/src/main/java/org/phoebus/pv/alarm/AlarmContext.java
@@ -116,7 +116,11 @@ public class AlarmContext
                     }
                     if (node != null)
                     {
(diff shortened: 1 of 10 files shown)


kasemir:
There's also https://github.com/ornl-epics/alarm-webmon for offering a read-only view of current alarms, but it's really just that.

To open related displays etc. the desktop version is necessary, and displaying errors is certainly good.

It's just look at the `ExceptionDetailsErrorDialog` usage. `ExceptionDetailsErrorDialog.openError` already calls `Platform.runLater` internally, so you can skip that part.

georgweiss:
OK, I'll update.

kasemir (approved):

PR #2577, opened 2023-03-10
phoebus.bat: Don't use server mode

kasemir:
Mention `-server 4918` in comment, but don't include it in the default parameters because it prevents opening multiple instances. In addition, that TCP port may be used by other tools, so leave addition of `-server...` settings to each site.

See also end user questions like https://epics.anl.gov/tech-talk/2023/msg00299.php

### DIFF
--- a/phoebus-product/phoebus.bat
+++ b/phoebus-product/phoebus.bat
@@ -28,7 +28,7 @@ echo off
 FOR /F "tokens=* USEBACKQ" %%F IN (`dir /B product*.jar`) DO (SET JAR=%%F)
 echo on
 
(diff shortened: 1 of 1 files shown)


kasemir:
@shroffk @georgweiss  This is a change for those who simply use the 'phoebus.bat' without any local changes.
I assume, however, that most people would expect a new instance whenever invoking 'phoebus.bat'.
Including `-server 4918` risks issues with _something_ already using that port, including some hung instance of the launcher, so it's likely better to leave adding `-server ...` to whoever creates a launcher for local usage.

shroffk (approved):

PR #2580, opened 2023-03-13
Add a Util class to generate rst widgets document

katysaintin:
see discussion https://github.com/ControlSystemStudio/phoebus/discussions/2269
This documentation could be useful to develop python or jython script in Phoebus for example.
The rst format is used in order to generate html documentation for Phoebus with Sphinx


kasemir:
Thanks!

Issue #2584, opened 2023-03-14
Save and Restore - closing the application can disconnect the service


UrbanBobek:
Hi all,

when using Save and Restore application I noticed that the connection to the S&R service is lost if the application is closed when the tree view is collapsed. 

Steps to reproduce:
1. Open Save and Restore application
2. Collapse the tree view so that only the root folder is visible
3. Close Save and Restore application
4. When reopening Save and Restore the connection to the service is lost

![image](https://user-images.githubusercontent.com/43409317/224996764-f347e10b-4ced-4a83-a2d6-cb13fdc51735.png)

During all the steps nothing is printed in the CLI. 

I could not find the cause of this issue, but to solve it temporary I closed Phoebus, deleted the `.phoebus` folder in my home directory and started Phoebus again.

Best regards,
Urban

UrbanBobek:
Yes, the fix resolved the issue. Thank you!

Issue #2581, opened 2023-03-13
Save and Restore - "Create/add to a Configuration" option not working as expected


UrbanBobek:
Hi all,

I am getting errors when using the `Create/add to a Configuration` option in Phoebus. I am using the latest version of Phoebus and S&R. 

I right click on a widget that is using a certain PV and select the `Create/add to a Configuration` option and a new window appears. If I click on the `Browse` button nothing happens and the following error appears in the CLI:
```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "org.phoebus.applications.saveandrestore.ui.SaveAndRestoreController.matchesFilter(org.phoebus.applications.saveandrestore.model.Node)" because "this.saveAndRestoreController" is null
        at org.phoebus.applications.saveandrestore.ui.BrowserTreeCell.updateItem(BrowserTreeCell.java:155)
        at org.phoebus.applications.saveandrestore.ui.BrowserTreeCell.updateItem(BrowserTreeCell.java:52)
        at javafx.scene.control.TreeCell.updateItem(TreeCell.java:533)
        at javafx.scene.control.TreeCell.indexChanged(TreeCell.java:490)
        at javafx.scene.control.IndexedCell$1.invalidated(IndexedCell.java:85)
        at javafx.beans.property.IntegerPropertyBase.markInvalid(IntegerPropertyBase.java:113)
        at javafx.beans.property.IntegerPropertyBase.set(IntegerPropertyBase.java:148)
        at javafx.scene.control.IndexedCell.updateIndex(IndexedCell.java:130)
        at javafx.scene.control.skin.VirtualFlow.setCellIndex(VirtualFlow.java:1804)
        at javafx.scene.control.skin.VirtualFlow.getCell(VirtualFlow.java:1788)
        at javafx.scene.control.skin.VirtualFlow.getOrCreateCellSize(VirtualFlow.java:3026)
        at javafx.scene.control.skin.VirtualFlow.getOrCreateCellSize(VirtualFlow.java:2998)
        at javafx.scene.control.skin.VirtualFlow.recalculateAndImproveEstimatedSize(VirtualFlow.java:3086)
        at javafx.scene.control.skin.VirtualFlow.recalculateEstimatedSize(VirtualFlow.java:3069)
        at javafx.scene.control.skin.VirtualFlow$5.invalidated(VirtualFlow.java:860)
[...]

PR #2585, opened 2023-03-14
Bug fixes in save&restore

georgweiss:
Addresses #2581  and #2584


shroffk:
I tested the fix for #2584 and it works

georgweiss:
Added fix for #2588

PR #2589, opened 2023-03-14
Display Builder: last save timestamp and user in BOB file comment

pavel-ch:
See #2583

PR #2591, opened 2023-03-14
Fix testWidgetWriting

pavel-ch:
Fix for #2589


kasemir:
I had reverted the commit which added the comment.
Can you add that back to this PR, so we see if all tests run?

pavel-ch:
Sorry no clue how to do that, what I can do is a new commit.

kasemir:
Thanks!

PR #2592, opened 2023-03-15
Add watermark to image in Image Viewer app

georgweiss:
As per user request: to avoid confusing a screen shot of an OPI (e.g. user opens log attachment) with an actual OPI, the Image Viewer app will add a watermark on top of the image, see screen shot.

The WATERMARK font size is computed as a function of the size of the rendered ImageView.

User may hide/show watermark using the Watermark checkbox.

Screenshot 2023-03-15 at 09 39 50


kasemir:
Seems a bit unusual. I'm not aware of any other image or media viewer in general that would default to modifying the content.
Make the default a preference option?

georgweiss:
Agree that it may seem odd. But then again the Image Viewer does not claim to be a general purpose media handler.

Rather than adding a preference setting I'd maybe propose to add an option when launching the app. When launched from logbook viewer - which should be the most common use case where user wants to inspect OPI screen shot - the option would then request watermark. When launched from File Browser the option would be false.

kasemir:
Adding something like "Logbook" as a watermark when launched from logbook viewer sounds fine, while leaving it disabled if you open an image from the file browser

georgweiss:
Not sure I agree. Since the checkbox may show the watermark, it may come as a surprise to see "Logbook". And sites may choose to not use the logbook module(s)...

kasemir:
>  it may come as a surprise to see "Logbook"

A watermark tends to not be the literal text "Watermark". Instead, it conveys some meaning like "Draft", "Preview", "Preliminary" or "Confidential".
For images opened from the logbook viewer, wouldn't it make sense to add "Logbook" as a watermark?

georgweiss:
To facilitate convergence I've made the watermark text customizable.

kasemir (approved):

PR #2593, opened 2023-03-15
Handle tracenames from legacy xygraph widget

tanviash:
The PR is to handle trace names for legacy XYGraph widget when converted to stripchart. This is already fixed for xyplot but was missing for stripchart.


kasemir:
Thanks!

Issue #2596, opened 2023-03-16
services-alarm-server : increase STABILIZATION_SECS when exporting


lcaouen:
In services-alarm-server, when exporting, STABILIZATION_SECS is too small.
With some servers it takes more than 4s to get the data from Kafka.
It would be nice to add a parameter to set the initial value of STABILIZATION_SECS

PR #2597, opened 2023-03-16
 services-alarm-server : increase STABILIZATION_SECS when exporting #2596

lcaouen:
Added an argument export_wait to change the STABILIZATION_SECS default value of 4s.
My server is slow and 4s is too small to get a correct export.
Also added the possibility to set it in the settings.ini (org.phoebus.applications.alarm/export_wait)
Not sure the name I've chosen is the best, let you change it if necessary.

### DIFF
--- a/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmConfigTool.java
+++ b/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmConfigTool.java
@@ -29,13 +29,15 @@ import org.phoebus.util.time.SecondsParser;
 public class AlarmConfigTool
 {
     /** Time the model must be stable for. Unit is seconds. Default is 4 seconds. */
-    private static final long STABILIZATION_SECS = 4;
+    private long STABILIZATION_SECS = 4;
 
     	// Export an alarm system model to an xml file.
-    public void exportModel(String filename, String server, String config, String kafka_properties_file) throws Exception
+    public void exportModel(String filename, String server, String config, String kafka_properties_file, long wait) throws Exception
 	{
         final XmlModelWriter xmlWriter;
 
+        if (wait > STABILIZATION_SECS) STABILIZATION_SECS = wait;
+        
         // Write to stdout or to file.
         if (filename.equals("stdout"))
             xmlWriter = new XmlModelWriter(System.out);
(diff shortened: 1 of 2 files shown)


kasemir:
Those are the idle "state:".. updates which are indeed happening every 10 seconds, or of course whenever there's a state change. The `AlarmConfigMonitor` looks for "config:" messages, and then waits for a 4 (by default) second pause in them.

When you run `monitor_topic.sh Accelerator`, you should see

1) Maybe a 1 second startup delay for the tool to load
2) A very brief burst of all the config (and status) messages
3) Then no more config messages, just a state update every 10 seconds

lcaouen:
Here is what I get with monitor_topic.sh
image
It took around 10s to see the output after running monitor_topic.sh

lcaouen:
If I stop it after 4s I get nothing :
image

If you're ok with my modfications in the PR, it will solve my problem.
I'm already using my fork to get the xml.

kasemir:
I'll create an update that adds an initial wait for the first config message, which is basically a connection timeout, changes the parameter name from "-export_wait" because it's actually also used for "-import", and lists that parameter in the "-help" output.

lcaouen:
Thanks for the merge.
To answer your question about my configuration, I'm using a VM (Hyper-V) on my computer (not a server), that's maybe why it's slower. I have to set a value of 10s for STABILIZATION_SECS to get an answer.
I'll do some other tests on another server (real one this time) next week. I will let you know if it's faster.

kasemir:
OK, there's now a designated `-connect_secs` time that defaults to 10 seconds, and the `-stable_secs` that looks for a pause. The latter can likely be set back to 4 seconds (default), because otherwise the state is never stable with 10 second "idle" state updates.

This applies to import, export as well as alarm server startup

kasemir:
Note the change in names. Instead of `-export_wait`, use `-connect_secs` to address the longer connection time.
Ideally, the default of 10 sec for that new option should suffice.

lcaouen:
I did some tests on another server and I only needed 1s for connect_secs  instead of 10 on my computer, so the problem was more my computer.
Thanks for the news changes , it will help us !

kasemir:
Thanks for the update!  It's certainly good that we now have a configurable delay for the initial connection, because no matter if it's 1 or 10 seconds, getting that initial message always seems to take longer. From then on, we can usually receive 1000 or more per second.

PR #2602, opened 2023-03-20
Save-and-restore : Add snapshot name automatically after a "Take Snapshot"

agaget:
Hi, 
I have modified save-and-restore phoebus app for my own need and add some properties to be able to format an automatic name to the snapshot when clicking "Take snapshot". 
I think the following properties are comprehensible by itself
```
# Configure the default snapshot name when "Take Snapshot"
#default_title_snapshot_configuration=TuneList
#default_title_snapshot_format=<date>_<pv1>_<pv2>
#default_title_snapshot_date_format=YYYYMMHH_hhmmss
#default_title_snapshot_pv1=SL-APP-MCS:BOM-CPU:IonSet
#default_title_snapshot_pv2=SL-APP-MCS:BOM-CPU:BeamDestSet
```

The `default_title_snapshot_configuration` is not used for now. The idea is to configure the format for configurations indicated here. At first implementation I don't find how to make it easily so for now it's a pending possibility.

Maybe the function `getValueVType` can be more efficient, I'm not completely comfortable with the code. But I propose this pull request more as an idea now because I see that @georgweiss made some modification on this part of the code quite recently so it could be deprecated/conflicted quickly if it's not integrated.

image

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/Preferences.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/Preferences.java
@@ -32,6 +32,16 @@ public class Preferences {
     public static int search_result_page_size;
     @Preference
     public static String default_search_query;
+    @Preference
+    public static String default_title_snapshot_configuration;
+    @Preference
+    public static String default_title_snapshot_format;
+    @Preference
+    public static String default_title_snapshot_date_format;
+    @Preference
+    public static String default_title_snapshot_pv1;
+    @Preference
+    public static String default_title_snapshot_pv2;
 
     static
     {
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/Preferences.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/Preferences.java
@@ -33,7 +33,7 @@ public class Preferences {
     @Preference
     public static String default_search_query;
     @Preference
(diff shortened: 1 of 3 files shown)


georgweiss:
In my opinion this looks quite specific to your use case (e.g. use PV names from the snapshot to determine a name). 

If there is a need to automatically set a snapshot name I would possibly recommend for instance the current date/time, and use a property to enable this.

If there is a need for more dynamic behavior, I would prefer to use a Java SPI approach where implementations can be added by various sites. Such an SPI would the define an interface like ```getSnapshotName```, using the snapshot data as input.

georgweiss:
If you opt for the SPI alternative then I'd maybe consider to add an implementation of that in the common code, which would be the "date" implementation. A property could then be used to select the naming SPI implementation based on some identifier exposed by the interface (e.g. name).

georgweiss (commented):
We tend to use the ISO format yyyy-MM-dd HH:mm:ss format for date/time. This should be OK as the snapshot name may contain space chars.

Also, there is no need to instantiate SimpleDateFormat unless a date/time format is set to enable the auto-name of snapshot. Moreover, it should be sufficient to just set the snapshotNameProperty, no need to append the date/time string to an empty string.

agaget:
I'm not sure to understand, you suggest me to add another property field ? 
i.e:
`default_title_snapshot_auto_date=enable`

georgweiss:
No, no additional property. Checking for presence of date format is sufficient.

However, I'd like to propose to rename the property to something like ```default_snapshot_name_date_format```.

PR #2605, opened 2023-03-20
update alarm severity comment and doc

tynanford:
I noticed the online docs had the incorrect index for the INVALID alarm state in pvSev. Probably was referring to the pvLegacySev instead

https://github.com/ControlSystemStudio/phoebus/blob/master/core/pva/src/main/java/org/epics/pva/data/nt/PVAAlarm.java#L54


kasemir:
That older mapping in https://github.com/ControlSystemStudio/cs-studio/blob/2f9c8532c141b47d1d624f8a9f8bd4fe5abdaa6c/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java#L328 was strange...

PR #2607, opened 2023-03-21
Creating a new profile to allow creation of non fat jars for services

shroffk:
Do not merge...
This is primarily for testing the new profile


georgweiss (commented):
Should not the documentation about how to run a release be in the top level README?

To test/verify calling release:perform would also upload artifacts to Maven. How can I avoid this?

georgweiss:
Oh, did not realize dryRun was also available for perform...

georgweiss:
Seems release:perform attempts to push to origin even when using dryRun=true. So I aborted to avoid this and potentially upload to Sonatype.

georgweiss:
Seems I will need to test this with push to Github and Sonatype...

I guess dropping a release still means that a version number is "wasted"...

shroffk:
What happens when you run the release prepare with -DdryRun

georgweiss (commented):
Not sure how to review this without a non-dryrun.

shroffk:
@georgweiss could we at least verify that by default it continues to work as before

georgweiss:
Sure, but to me it seems we need to run this and upload a new version (4.7.2). Right?

georgweiss:
Building on master (no profile) yields:

- alarm-config-logger: fat jar
- alarm-server: thin jar
- alarm-logger: fat jar
- save-and-restore: fat jar

Building with this PR with ```-Dskip-executable-jar=true``` yields:

- alarm-config-logger: fat jar
- alarm-server: thin jar
- alarm-logger: fat jar
- save-and-restore: fat jar

Building with this PR with ```-Dskip-executable-jar=false``` yields:

- alarm-config-logger: thin jar
- alarm-server: thin jar
- alarm-logger: thin jar
- save-and-restore: thin jar

So I guess this is the wanted outcome...

shroffk:
> Building with this PR with -Dskip-executable-jar=true yields:
> 
> alarm-config-logger: fat jar
> alarm-server: thin jar
> alarm-logger: fat jar
> save-and-restore: fat jar

This should happen

The way this profile should work, is that no matter the value specified if the `skip-executable-jar` is used then the fat jars should not be prepared.

`mvn clean verify  -Dskip-executable-jar`
`mvn clean verify  -Dskip-executable-jar=true`
`mvn clean verify  -Dskip-executable-jar=false`

All 3 should result in the same, I have tested the above and they all yield think jars

georgweiss:
Now I am super confused... should not skip-executable-jar=true produce thin jars and vice versa? What other purpose would the profile have?

FYI: I am building with ```mvn clean install```, but I guess that should not be a factor here?

PR #2621, opened 2023-03-30
Update save&restore mapping to allow sorting on name

georgweiss:


shroffk:
Do we need to use the v2? could we just check the existing mapping and if it is different re index it.

georgweiss:
Well, this may boil down to my limited experience in Elastic, but I tried that: add a field to the mapping (which is fine), but then you need to re-index. But re-indexing implies a new index with a different name, right?

PR #2625, opened 2023-04-04
CSSTUDIO-1034 Add tooltips to UI-elements under "Properties" in the Display Builder

abrahamwolk:
This merge-request adds tooltips to the UI elements under "Properties" in the Display Builder that are used to configure widgets.

Event handlers are installed to update the tooltip when the value represented by the underlying UI element is changed.

The following is a screenshot of the tooltip associated with "false" in a widget of type BooleanWidgetProperty. (The mouse disappeared from the screenshot; it is actually hovering over the clickable box on the right of "Highlight Active Region".)

image

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/PropertyPanelSection.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/PropertyPanelSection.java
@@ -250,6 +250,10 @@ public class PropertyPanelSection extends GridPane
             final Button font_field = new Button();
             font_field.setMnemonicParsing(false);
             font_field.setMaxWidth(Double.MAX_VALUE);
+
+            Tooltip.install(font_field, new Tooltip(font_prop.getValue().toString()));
+            font_prop.addPropertyListener((listener, old_value, new_value) -> Tooltip.install(font_field, new Tooltip(new_value.toString())));
+
             final WidgetFontPropertyBinding binding = new WidgetFontPropertyBinding(undo, font_field, font_prop, other);
             bindings.add(binding);
             binding.bind();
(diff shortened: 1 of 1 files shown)


app/display/editor/src/main/java/org/csstudio/display/builder/editor/properties/PropertyPanelSection.java
```diff
@@ -29,20 +31,7 @@
...
-import org.csstudio.display.builder.model.properties.EnumWidgetProperty;
-import org.csstudio.display.builder.model.properties.FilenameWidgetProperty;
-import org.csstudio.display.builder.model.properties.FontWidgetProperty;
-import org.csstudio.display.builder.model.properties.MacrosWidgetProperty;
-import org.csstudio.display.builder.model.properties.PVNameWidgetProperty;
-import org.csstudio.display.builder.model.properties.PointsWidgetProperty;
-import org.csstudio.display.builder.model.properties.RulesWidgetProperty;
-import org.csstudio.display.builder.model.properties.ScriptsWidgetProperty;
-import org.csstudio.display.builder.model.properties.WidgetClassProperty;
+import org.csstudio.display.builder.model.properties.*;
```
kasemir, line 34: Minor issue: Can you revert the wildcard `import ...*` back to the specific imports?
Some IDEs will change over to wildcard imports after a certain number of imports from the same package, but we'd like to keep all imports explicit to avoid confusion down the road (https://github.com/ControlSystemStudio/phoebus/pull/2440#issuecomment-1299034799)

abrahamwolk: Sure. I have changed the wildcard import statement to individual import statements now.

Issue #2629, opened 2023-04-10
Choice button issue when custom options differ from enum options


kasemir:
Assume an enumerated PV with options A, B, C, D, E.
When configured with "Items from PV", the choice button will indicate all enum options and function as expected.
When not using "Items from PV" and instead configuring different items, for example only "A, C, E", the choice button does not operate as expected. Pressing "C" will set the PV to "B" because that is the second option (index 1). When the PV on the other hand changes to "C", the button will show "E" because that is the 3rd option (index 2). 
Reported by @tanviash 

To fix, the choice button needs to match values against the labels of enum PVs as done in the combo box widget.
In addition, the choice button needs to somehow indicate when the PV value does not match any of the configured options.

PR #2630, opened 2023-04-10
Choice button: Use proper enum index even if options differ from VEnum

kasemir:
Allow choice buttons with custom options A, C, E for PVs with enum labels A, B, C, D, E

#2629

Without "Items from PV" selected, the PV's value may not match any of the custom choice button options.
Indicate with same border as "disconnected". This may be an overload of the border meaning, but avoids having to invent another way to indicate the issue. The border is quite obvious in indicating a problem, the tool tip will show the current value and help explain why no button option is selected, and in a way the value of the PV is disconnected from the available choices.


tanviash:
@kasemir Thank you for the fix.

PR #2633, opened 2023-04-11
CSSTUDIO-1440 Remove saveFileDialog().

abrahamwolk:
This merge-request removes the function `FileUtil.saveFileDialog()`.

The motivation for removing the function is that a function call to `FileUtil.saveFileDialog()` cannot work: `FileUtil.saveFileDialog()` calls `ScriptUtil.showSaveAsDialog()` with the parameter `widget` set to `null`. The function `ScriptUtil.showSaveAsDialog()`, in turn, tries to invoke `widget.getDisplayModel()` which will necessarily result in an exception being thrown, the consequence of which is that an error is logged.

The user can use the function `ScriptUtil.showSaveAsDialog()` directly instead of calling `FileUtil.saveFileDialog()`.

There are no usages of `FileUtil.saveFileDialog()` in the Phoebus source code. It is possible that scripts reference it, but they are most likely buggy in that case, since the function call cannot succeed.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/FileUtil.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/FileUtil.java
@@ -175,14 +175,18 @@ public class FileUtil {
         return selected.getPath();
     }
 
-    /**Open a file save dialog.
-     * @param inWorkspace true if it is a workspace file dialog; Otherwise, it is a local
-     * file system file dialog.
-     * @return the full file path. Or null if it is cancelled.
+    /** Show file "Save As" dialog for selecting/entering a new file name
+     *
+     *  <p>Call blocks until the user closes the dialog
+     *  by either either entering/selecting a file name, or pressing "Cancel".
+     *
+     *  @param widget Widget, used to create and position the dialog
+     *  @param initial_value Initial path and file name
+     *  @return Path and file name or <code>null</code>
      */
-    public static String saveFileDialog(boolean inWorkspace)
+    public static String saveFileDialog(Widget widget, String initial_value)
     {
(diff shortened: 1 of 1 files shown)


abrahamwolk:
1. I am not familiar with the intention of the APIs implemented in `FileUtil` and `ScriptUtil`; is the reason to implement the same functionality in both of these classes backwards compatibility with the Eclipse-based Control System Studio?

2. Currently, `FileUtil.saveFileDialog()` has the signature
<p align="center">
<code>String saveFileDialog(boolean inWorkspace)</code>.
</p>
Should this then be changed to:

<p align="center">
<code>String saveFileDialog Widget widget, String initial_value)</code>,
</p>

to match the signature of `ScriptUtil.showSaveAsDialog()`?

tanviash:
@abrahamwolk 1) Yes, you are correct. This was done to support legacy OPI scripts.
2) Yes, we should just correct the api definition for FileUtil api.

abrahamwolk:
I have now modifed `FileUtil.saveFileDialog()` (and its description) to correspond precisely to `ScriptUtil.showSaveAsDialog()`. However, doesn't this break the existing API, as `FileUtil.saveFileDialog()` now takes the two input parameters `Widget widget` and `String initial_value`, whereas before it took only the parameter `boolean inWorkspace`?

If, on the other hand, the parameter list is left unchanged (i.e., `FileUtil.saveFileDialog()` has as its only parameter `boolean inWorkspace`), where can the object of type `Widget` be found that needs to be passed to `ScriptUtil.showSaveAsDialog()`? (I believe that the second argument `String initial_value` of  `ScriptUtil.showSaveAsDialog()` could be set to `null`, and that it is not necessarily required.)

abrahamwolk:
@tanviash You're welcome.

PR #2636, opened 2023-04-12
Legacy longString textUpdate interactive property 

tanviash:
This PR is to have 'interactive' property set to true (to get a scroll) for legacy TextUpdate widgets with PV configured as LongString.


kasemir:
Good idea!

Issue #2465, opened 2022-11-18
Apache Derby target is byte code version 61


georgweiss:
PR #2381 pulled in derby 10.16.1.1. This is built for class version 61, i.e. Java 17+. Will not run on Java 11.

kasemir:
Yes, we updated derby in #2381 because of a dependabot warning.
We also have an open issue #2300 to generally try and reduce dependency vulnerabilities.

When I check the derby jar that I magically get via maven, that indeed seems to be a binary for Java 17 / major version 61:
```
javap -v 'jar:file:derbytools-10.16.1.1.jar!/org/apache/derby/jdbc/EmbeddedDriver.class'

Classfile jar:file:derbytools-10.16.1.1.jar!/org/apache/derby/jdbc/EmbeddedDriver.class
  Last modified Oct 31, 2022; size 1616 bytes
  SHA-256 checksum a270db1da00b3bc03e8505ac11d4e8b21b98b54393b845503b9bb7f018bf4604
  Compiled from "EmbeddedDriver.java"
public class org.apache.derby.jdbc.EmbeddedDriver implements java.sql.Driver
  minor version: 0
  major version: 61
```

If we don't use maven but fetch binaries from https://db.apache.org/derby/derby_downloads.html, the latest version also requires Java 17 or higher, so I don't think there's a way to get a binary for older JREs via maven, but Kunal has been having fun with using Java 8 for some libs and not others recently, maybe he has an idea.

The simple fix should be to run with Java 17 or higher. OpenJDK is currently at 18.

Alternatively, you could try to find a derby version between 10.14.1.0 as we used to have and the latest one that removes the dependency warning and still runs with older JREs. Like 10.15.2.0 looks like its binaries are for JDK 9, and should still fix the dependapot issue with 10.14.1.0. 

With #2381, there was no change to any of our sources, just the derby version changed and unfortunately also the packaging as one jar vs. 3 jars.

georgweiss:
Then I suggest to move all references (e.g. version variable in top level pom) to that module. Also suggest we add a README to state that scan server requires Java 17+ runtime.

kasemir:
JavaFX 20 just came out which also requires JDK 17+

georgweiss:
So the day has finally come... Guess sticking to JavaFX 19 to avoid JDK update is not acceptable.

[...]

PR #2638, opened 2023-04-14
JavaFX 20

kasemir:
Updates JavaFX version to 20, the current one, https://github.com/openjdk/jfx/blob/jfx20/doc-files/release-notes-20.md

On Mac OS, this fixes the menu icons, https://bugs.openjdk.org/browse/JDK-8181084. Icons provided in high-res variants used to render in a wrong size, note data browser, file browser, logbook entry items:

Screenshot 2023-04-13 at 4 18 15 PM
Screenshot 2023-04-13 at 4 18 38 PM

With JavaFX 20, they render at the correct size:

Screenshot 2023-04-13 at 4 25 47 PM
Screenshot 2023-04-13 at 4 26 00 PM

On Mac OS, there is now also dedicated support for "Apple Silicon" aka "Arm", "M1", "M2", "aarch64", although as far as I can tell products built on/for the Intel x64 based version of OpenJDK for Mac OS will run fine on both Intel and Arm macs.

**An update to JFX 20 does require at least JDK 17!**
This requirement is shared with the apache derby lib as used by the scan server, #2465
So right now the github test builds which use JDK11 fail for the JFX 20 update.
JDK11 was released 2018. JDK 17 from Sept. 2021 is the current LTS release.
Might be time to overall require JDK 17 and then update JFX to 20.

### DIFF
--- a/dependencies/phoebus-target/.classpath
+++ b/dependencies/phoebus-target/.classpath
@@ -35,7 +35,6 @@
     <classpathentry exported="true" kind="lib" path="target/lib/cal10n-api-0.8.1.jar"/>
     <classpathentry exported="true" kind="lib" path="target/lib/cglib-full-2.0.jar"/>
     <classpathentry exported="true" kind="lib" path="target/lib/checker-qual-3.12.0.jar"/>
-    <classpathentry exported="true" kind="lib" path="target/lib/commonj.sdo-2.1.1.jar"/>
     <classpathentry exported="true" kind="lib" path="target/lib/commonmark-0.15.2.jar"/>
     <classpathentry exported="true" kind="lib" path="target/lib/commonmark-ext-gfm-tables-0.15.2.jar"/>
     <classpathentry exported="true" kind="lib" path="target/lib/commonmark-ext-image-attributes-0.15.2.jar"/>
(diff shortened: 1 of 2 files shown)


kasemir:
Haven't tried to cross-build.
What I've done:
 * Build with JFX20 on Intel mac
 * Run the result on Intel mac
 * Run the result on M2 mac --> The Intel build result is a good common binary
 * Build on M2 mac
 * Run that result on M2 mac --> Works, but not sure I want to start publishing both Intel and aarch64 binaries now because I can see people on intel macs download the aarch64 binary and then complain

kasemir:
> .. testing of current code base with Java 17 ..

Definitely a good idea, since that's the LTS version. I'm OK with delaying the JFX20 update a little

georgweiss:
What is your build output? A JPackage generated installer?

kasemir:
My build output is a zip file

kasemir:
Javafx-21 is about to be released.

https://github.com/openjdk/jfx/blob/jfx21/doc-files/release-notes-21.md didn't point me to essential bug fixes or new features, but GTK2 is gone, Linux must use GTK3. Would that be a reason for us to move to JFX-20, delaying JFX-21?

shroffk:
hmm... the GTK3 *only* is a bit scary, I remember javafx (efxclipse) constantly crashing when used with GTK3. ESS has reported rendering issues.
Annoyingly the crashes were linked to a combination of UI operations, which makes it difficult to feel confident that we have tested all possible scenarios. I can start testing but would vote to wait before we remove GTK2 support.

kasemir:
Agree on keeping GTK2 supported, I still select GTK2 on several Linux setups because it "works".
Still interested in JFX 20 because it fixes the ugly Mac OS icons.
So for JFX 20, just not JFX 21, yet.

georgweiss:
Ran some tests on Ubuntu 20.04 with Java 17 and JavaFX 21. Could not see any GTK rendering issues.

PR #2641, opened 2023-04-18
Truncate log title but avoid ending in partial word

tanviash:
This PR is for the old olog client to truncate its log's title but avoid ending in a partial word.


kasemir:
Does that work in case the text contains no space at all, so `lastIndexOf(" ")` returns -1?

PR #2642, opened 2023-04-19
CSSTUDIO-1903 Use colors specified in the configuration file for alarm indications, and add separate color options for the application "Alarm Area Panel".

abrahamwolk:
This merge-requests improves the graphical representation of alarm indications by improving the use of the colors as specified in the Phoebus configuration file in the following applications/dialogs:

- The application "Alarm Area Panel"
- The application "Alarm Log Table"
- The information window "Widget Info" (Accessed through right-clicking on a widget in an OPI)
- The application "Probe"
- The application "Alarm Table"

In some cases, the text is also centered. (Previously, it was left-aligned.)

It also adds and implements separate configuration parameters for the application "Alarm Area Panel". The new parameters (whose names are self-descriptive), are (together with their default values):
 
> ```org.phoebus.ui/alarm_area_panel_ok_severity_text_color=255,255,255
> org.phoebus.ui/alarm_area_panel_ok_severity_background_color=0,255,0
> 
> org.phoebus.ui/alarm_area_panel_minor_severity_text_color=255,255,255
> org.phoebus.ui/alarm_area_panel_minor_severity_background_color=255,128,0
> 
> org.phoebus.ui/alarm_area_panel_major_severity_text_color=255,255,255
> org.phoebus.ui/alarm_area_panel_major_severity_background_color=255,0,0
> 
> org.phoebus.ui/alarm_area_panel_invalid_severity_text_color=255,255,255
> org.phoebus.ui/alarm_area_panel_invalid_severity_background_color=255,0,255
> 
> org.phoebus.ui/alarm_area_panel_undefined_severity_text_color=192,192,192
> org.phoebus.ui/alarm_area_panel_undefined_severity_background_color=200,0,200,200´´´

The motivation for having separate configuration options for the application "Alarm Area Panel" is that oftentimes the requirements are different, since this application is often running on a separate screen with the background of the individual items signalling the status of the corresponding sub-systems. In contrast, in other places where alarm statuses are indicated, the background color is not always used to signal the alarm-status. E.g., elsewhere in Phoebus, the status "OK" could be represented by black text on white background, while the "Alarm Area Panel" might represent the same status using a green background.

### DIFF
--- a/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
+++ b/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
@@ -169,8 +169,8 @@ public class AlarmLogTableController {
 
                 if (empty  ||  item == null)
                 {
+                    setStyle("-fx-text-fill: black;  -fx-background-color: transparent");
                     setText("");
-                    setTextFill(Color.BLACK);
                 }
                 else
                 {
@@ -242,8 +242,8 @@ public class AlarmLogTableController {
 
                 if (empty  ||  item == null)
                 {
+                    setStyle("-fx-text-fill: black;  -fx-background-color: transparent");
                     setText("");
-                    setTextFill(Color.BLACK);
                 }
                 else
                 {
--- a/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
+++ b/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
@@ -40,13 +40,13 @@ import org.phoebus.framework.jobs.Job;
 import org.phoebus.framework.selection.SelectionService;
 import org.phoebus.ui.application.ContextMenuHelper;
 import org.phoebus.ui.dialog.ExceptionDetailsErrorDialog;
+import org.phoebus.ui.javafx.JFXUtil;
 import org.phoebus.util.time.TimeParser;
 import org.phoebus.util.time.TimestampFormats;
 
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Instant;
-import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


georgweiss:
Will perform some tests before requesting review.

kasemir:
Designated color settings for the alarm area panel are fine.

At the same time, this would replace the simple re-use of pre-generated `Background` objects with creating a `new Background(new BackgroundFill({color}, CornerRadii.EMPTY, Insets.EMPTY))` for every GUI element on each change, or it updates a style like `"-fx-alignment: center; -fx-border-color: transparent; -fx-border-width: 2 0 2 0; -fx-background-insets: 2 0 2 0; -fx-text-fill: " + {color1} + ";  -fx-background-color: " + {color2}` that the graphics library needs to parse and interpret, so it could use more CPU or delay updates.

One thing to test: Run the https://github.com/ControlSystemStudio/phoebus/blob/master/app/alarm/model/src/test/java/org/phoebus/applications/alarm/AlarmConfigProducerDemo.java to create a setup with 100k PVs where about 1000 of them change and see if that is still displayed alright.

abrahamwolk:
> At the same time, this would replace the simple re-use of pre-generated `Background` objects with creating a `new Background(new BackgroundFill({color}, CornerRadii.EMPTY, Insets.EMPTY))` for every GUI element on each change, or it updates a style like `"-fx-alignment: center; -fx-border-color: transparent; -fx-border-width: 2 0 2 0; -fx-background-insets: 2 0 2 0; -fx-text-fill: " + {color1} + "; -fx-background-color: " + {color2}` that the graphics library needs to parse and interpret, so it could use more CPU or delay updates.

That is a good point. I will change this so that there are not allocated many identical objects unnecessarily.

> One thing to test: Run the https://github.com/ControlSystemStudio/phoebus/blob/master/app/alarm/model/src/test/java/org/phoebus/applications/alarm/AlarmConfigProducerDemo.java to create a setup with 100k PVs where about 1000 of them change and see if that is still displayed alright.

I think that even if it runs fine, there is no need to allocate many identical objects in this case. I will update the code.

kasemir (approved):

kasemir:
Wonder if it's related to this: 

> .. "Alarm Table", set the background of cells to transparent when they are empty

The "active alarms" section is sometimes empty, so cells are set transparent, and maybe that remains so even when they're then populated? That would explain why the top cells lose their colors.

abrahamwolk:
@kasemir: I am looking into this.

abrahamwolk:
I have unfortunately not been able to reproduce this problem.

I think you may be right that this PR is the reason for the bug, and that it may be caused by setting the transparency, as you suggested: when the option `org.phoebus.applications.alarm/alarm_table_color_legacy_background` is set to `true`, then the background and text colors are set using Java methods. However, all other colors are set using CSS in this PR, including the transparency. Perhaps this could be the cause of the problem.

I have pushed the branch `CSSTUDIO-1903-bugfix` on which I have changed the code path for when `alarm_table_color_legacy_background == true` to use CSS instead of Java methods.

Unfortunately, I cannot verify if this fixes the issue or nor, since I have not been successful in reproducing it. Would you mind to test the branch (in case you can reliably reproduce the issue)?

kasemir:
I can reproduce it by simply keeping an alarm table open for a day, so takes a little time.
A conflict between both css and api commanding the colors would perfectly explain what's happening.
We'll know next week if the bugfix works.

kasemir:
Didn't have to wait long.

Older version on the left has text that lacks colors, while the version on the right from CSSTUDIO-1903-bugfix looks good:

Screenshot 2023-09-08 at 11 12 53 AM

abrahamwolk:
OK, that's good news!

PR #2646, opened 2023-04-20
Replace logbook advanced search button

georgweiss:
Based on user request: it is not obvious what the current button showing a plain ">" character actually does. This PR replaces that button with a label that can be clicked to open/hide the advanced search view. That way user will know up-front what to expect, plus we gain a little more horizontal space for the search query.

Screenshot 2023-04-20 at 1 09 27 PM
Screenshot 2023-04-20 at 1 09 36 PM


shroffk:
I just thought it was more in keeping with applications which have a outline/detail view to the left... but I can see how explicit buttons might be better

Issue #2647, opened 2023-04-20
Databrowser refresh issue


georgweiss:
When launched - from OPI or from plt file - with one single PV the Data Browser connects to the PV and the first value updated will determine the ```border_time``` field in ```org.csstudio.trends.databrowser3.model.HistoricSamples```. This time will then determine - as commented in the code - the time "beyond which no samples are returned from the history". This has a side effect:

If the PV in question updates rarely and there are no data in the archiver beyond the ```border_time```, a refresh of the plot will simply not plot anything even though there was a plot rendered before the refresh. Another way to see this side effect is to set a time range (where end != now) and then do a refresh. Result between initial plot and after refresh is shown in screen shots below.

Not sure where and how to fix this. Set a new ```border_time``` when user has selected a time range?

![Screenshot 2023-04-20 at 4 33 34 PM](https://user-images.githubusercontent.com/35602960/233399227-d0e96ed7-13f4-42b0-94c5-114f7c25ab23.png)
![Screenshot 2023-04-20 at 4 33 48 PM](https://user-images.githubusercontent.com/35602960/233399236-e76575c3-ca4d-4960-9808-e8cdaf351877.png)

georgweiss:
To clarify the issue:

User launches databrowser for "myPV". Default time span is 1 hour, but last time myPV was updated (and archived) is > 1 hour. Since databrowser connects to myPV it will pick up the latest update and use that as a live sample, which is also copied to the sample for time "now". Querying the archiver does not return any data as the last archived is outside default time span. Databrowser has thus two "live" samples and plots a line.

Next user selects refresh from context menu. This time only the archived data is considered, and since this is empty there are no samples to plot. The previously collected live sample - created when myPV connects - is discarded it seems.

My second example - demonstrated by the screen shots - is different in that the time span is wider and does not contain "now". In this case when user selects refresh, the "live" sample acquired when myPV connects will set the ```border_time```, which has the side effect that the archived sample (value 1, right-most sample in first screen shot) is excluded. Since the sample preceding that has value 0, the last sample is 0 and is plotted (second screen shot).

All in all, user gets a different plot after refresh.

georgweiss:
Well, if that is an acceptable solution I can look into it.

georgweiss:
#2654

PR #2654, opened 2023-04-26
Clear border_time for historic samples in Data Browser

georgweiss:
Handles a corner case where refreshing a plot may produce a different result compared to the plot rendered before refresh.
See #2647.


kasemir (approved):

kasemir:
Not sure if that will solve all our problems, but clearing the `border_time` definitely looks like something that should have been done all along.

georgweiss:
>Not...solve all our problems

For that we need God mode.

Issue #2659, opened 2023-05-01
NTNDArray Decoding Failure


jeinstei:
I'm not sure where to start with this, as it looks like this could just be user error as much as an issue with NTNDArray decoding of a single-axis waveform. We have a waveform of RF data coming back to us in an NTNDArray, but the decoding seems to fail. 

The error below is created in the terminal when attempting to connect using the PV List in the most recent head on master (d8ed8062) built in my local system and passing all the tests. I'm probably on an "unsupported" java version (openjdk 20.0.1), but I also can't see how that could be causing this issue.

Happy to debug further if there is a debug guide I should be looking at for phoebus to enable some more details or we can modify the data structure being sent back.

```
2023-05-01 15:31:57 WARNING [org.phoebus.pv] Cannot decode 'my_ioc:Root:PvProcessor:AdcI[0]' [CID 2, SID 117768961 CONNECTED] = epics:nt/NTNDArray:1.0 
    union value
[...]

PR #2664, opened 2023-05-04
Updating ImageDecoder to allow null names from #2659

jeinstei:
This might actually be contentious if others want to catch "improper" NTNDArrays, but it already is allowing blank names, so a null name isn't too different.

This is a PR in reference to #2659


kasemir:
I agree, https://docs.epics-controls.org/en/latest/specs/Normative-Types-Specification.html#ntndarray allows an "empty" name and it's not obvious to me if that means `""` or simply no "name" at all.

Issue #2662, opened 2023-05-04
Two issues in macro expansion


abrahamwolk:
I think there are two issues with the macro expanding phase.

**Issue 1**

When issuing a re-load of an OPI, the macro values of the OPI are not re-initialized to just the top-level macro-variable definitions of the OPI. Instead, the macro-variables defined at the top-level of the OPI are merged with the existing macro-variables of the running instance of the OPI. In case of naming conflict, top-level macro-variables overwrite existing macro values.This behavior makes it difficult to reason about the effect on macro-variables when issuing a "re-load" command to an OPI.

The current implementation is located in the function `loadModel()` in the class `DisplayRuntimeInstance`:
https://github.com/ControlSystemStudio/phoebus/blob/54c3eecda87dfe70885879ea73af2a5b3f862db7/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java#L360-L386

**Issue 2**
Consider the following function `expandValue()` in the class `Macros`:
https://github.com/ControlSystemStudio/phoebus/blob/54c3eecda87dfe70885879ea73af2a5b3f862db7/core/framework/src/main/java/org/phoebus/framework/macros/Macros.java#L272-L288

On line 284 it is checked whether the expanded form of the contents of a macro-variable is equal to the non-expanded form. Only if it is not equal to the expanded form is the binding of the macro variable updated (line 285).

Suppose the argument `input` of the function `expandValue` already contains a binding `B` for a macro-variable `$(M)`. Suppose further that the Java-variable `macros` contains a binding of `$(M)` to a literal value `V` with no free variables (i.e., no substitution will occur in the call to `MacroHandler.replace()`). After `expandValue()` has returned, `$(M)` is still bound to `B`, and not `V` which it, I believe, should be bound to.

**An example of how issue 1 and issue 2 cause a problem**
Counter-intuitive/wrong behavior may be observed in the following case (this is written using pseudo-code, not the actual syntax for macro assignments; `+` denotes string concatenation):

 - The OPI itself contains no top-level macro definitions.
 - A sub-OPI that is loaded when the main OPI is loaded contains two declarations of the form (this code is only loaded once when the OPI starts):
    - `$(A) = "SOMETHING`
    - `$(B) = $(A) + ":POSTFIX"`.
    
   (So that at this stage, it `$(B)` is set to `SOMETHING:POSTFIX`)
    
[...]

PR #2665, opened 2023-05-05
CSSTUDIO-1911 Changes to macro expansion

abrahamwolk:
This MR changes macro expansion as described under "Proposed changes" in issue https://github.com/ControlSystemStudio/phoebus/issues/2662.

I am not very familiar with the macro implementation and with how macros are "supposed" to behave in Phoebus, and therefore I cannot predict very well whether the changes can have adverse effects for existing OPIs or other unintended side-effects. I have created this merge-request to have something concrete to discuss, as well as to facilitate testing.


kasemir:
At first glance, removing that "info" handling will cause problems.
The idea was this:
Assume a display that's opened from an "Open Display" action. That action can add macros, which will be in the "info". When we later restore such a display, it's not sufficient to simply use the macros that are in the display itself. We do need to add those from the "info".

PR #2668, opened 2023-05-09
MEDM Converter: Turn "oval" or "rectangle" with alarm-sensitive fill into MultiStateLED

kasemir:
Tweak converter for example provided by @MarkRivers
If the shape is not filled but just an outline, support alarm-colored outline in the script.
For filled "oval" and "rectangle", avoid the script and instead create a MultiStateLED with `=highestSeverity(..)` as a PV

### DIFF
--- a/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/AbstractADL2Model.java
+++ b/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/AbstractADL2Model.java
@@ -177,7 +177,7 @@ public abstract class AbstractADL2Model<WM extends Widget>
         if (dynAttr.getClrMode().equals("alarm")  &&  dynAttr.get_chan() != null)
         {
             // Attach script that sets background color based on alarm severity
-            final String embedded_script =
+            String embedded_script =
                 "from org.csstudio.display.builder.runtime.script import PVUtil\n" +
                 "from org.csstudio.display.builder.model.persist import WidgetColorService\n" +
                 "sevr = PVUtil.getSeverity(pvs[0])\n" +
@@ -193,6 +193,9 @@ public abstract class AbstractADL2Model<WM extends Widget>
                 "    name = 'OK'\n" +
                 "color = WidgetColorService.getColor(name)\n" +
                 "widget.setPropertyValue('background_color', color)\n";
+            // Also set optional line color
+            if (widgetModel.checkProperty("line_color").isPresent())
+                embedded_script += "widget.setPropertyValue('line_color', color)\n";
             final ScriptInfo script = new ScriptInfo(ScriptInfo.EMBEDDED_PYTHON,
                                                      embedded_script,
                                                      true,
--- a/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/Oval2Model.java
+++ b/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/Oval2Model.java
@@ -8,13 +8,17 @@ package org.csstudio.opibuilder.adl2boy.translator;
 
 import org.csstudio.display.builder.model.ChildrenProperty;
 import org.csstudio.display.builder.model.Widget;
+import org.csstudio.display.builder.model.persist.NamedWidgetColors;
+import org.csstudio.display.builder.model.persist.WidgetColorService;
 import org.csstudio.display.builder.model.properties.WidgetColor;
 import org.csstudio.display.builder.model.widgets.EllipseWidget;
+import org.csstudio.display.builder.model.widgets.MultiStateLEDWidget;
 import org.csstudio.utility.adlparser.fileParser.ADLWidget;
+import org.csstudio.utility.adlparser.fileParser.widgetParts.ADLDynamicAttribute;
 import org.csstudio.utility.adlparser.fileParser.widgets.ADLAbstractWidget;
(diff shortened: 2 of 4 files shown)


tynanford:
> avoid the script and instead create a MultiStateLED

Hi @kasemir, could this also apply to the EDM converter? This seems to ring a bell for when I tried the converter with our EDM screens. But maybe that is outdated information

kasemir:
Yes, need to check into that tomorrow.

kasemir:
There's one problem..

EDM dynamic colors are defined as ranges like this:
```
>=-0.5 && <0.5: "color0"
>=0.5  && <1.5: "color1"
>=1.5  && <2.5: "color2"
>=2.5  && <3.5: "color3"
>=3.5  && <4.5: "color4"
>=4.5  && <=5.5: "color5"
```


In most cases, they are then used with enumerated or integer PVs
so values 0, 1, .., 5 map to color0..5.
We assume that case and try to create an LED or MultiStateLED

If the EDM display was meant to be used with double values, this will fail.
For example, a value of 1.9 that was resulting in color2 for EDM will now show color1...

In practice, I have only seen use with enum or int PVs, so I think we'll be fine.

tynanford:
Thanks for looking into that. From what I can tell here this would also be fine. I see EDM widgets using double values for the visible/invisible but not for the color mapping. But I can't say if this is a good assumption for all EDM users so I will defer to your judgement

kasemir:
I think it's the same underlying type of code, but for "visible" you can see the start..end values of each range in the properties for each widget.
The dynamic colors have that text in the colors.list file.

tynanford:
I see. Well at the ALS the colors.list file contains only integers so that would be okay with us. But not critical if you think this would break anyone's conversions.

kasemir:
Can you post one or two example dynamic colors from your colors.list?
As shown above ours here all bracket the 0, 1, 2, ... states as   `>=(state-0.5) && <(state+0.05)`.
Do you have full numbers like this?
```
>=0 && <1: "color0"
>=1  && <2: "color1"
...
```

tynanford:
Sorry I didn't fully explain in the last post. It actually looks like we never used dynamic colors much. There are only 2 in the file and based on their names it looks like they were never used (although based on the rule it looks like it would expect floating point numbers)

```
rule 81  test1 {
 >=20        : purple-46
 >=0 && <10  : "UNMS help"
 >=10 && <20 : invisible
 default     : blue-26
}

rule 82  test1Inv {
 >=40 && <45 : 79
 >=45 && <47 : blue-26
}
```

It appears that instead of dynamic colors we sadly used visibility rules and stacked a whole bunch of LED widgets on top of each other. So usually the EDM -> phoebus conversion is messy here either way.  But I had noticed some of our widgets that display the alarm state as an LED used a script in the phoebus conversion instead of multi state LEDs so this at least could help there.

kasemir:
Bummer. Those color rules clearly don't map to states 0, 1, 2, 3, so they won't turn into LEDs.

But yes, the circles or rectangles that use the alarm severity will now turn into MultiStateLEDs with `=highestSeverity(..)` as a PV, so that's some progress.

In principle, one could try to detect circles with (almost) the same size and position, stacked on top of each other, all with conditional visibility, then check if the visibility ranges map to basic indices 0, 1, 2, ... and replace with LEDs.
But that's not trivial, would take some time to implement and likely be error prone.

Well, from 2009 on, like "Not another EPICS Display Manager" on https://epics.anl.gov/meetings/2009-10/index.html, we pointed out that it's better to have designated widgets. "Text Update" instead of a "Label" that can also change its text, "LED" instead of somehow changing the color. Otherwise you're stuck with one tool and conversions are hard.

tynanford:
I tried your PR and it works great for the alarm severity widgets, thanks! Saves us a lot of time on those.

The stacked EDM widgets with visibility conditionals mean we have hundreds of Phoebus rectangle widgets with rules but we should fix those using summation CALC PVs and a single LED anyways. Because of how the EDM screens were created here we are trying to update the screens manually in these cases since as you said it's going to be better in the long run when "The next EPICS Display Manager" arrives.

kasemir:
Yes, summarizing in PVs and simplifying the EDM displays is a good idea.
Might also be possible to handle some site-specific arrangements with post-processing.
A python tool that reads the *.bob XML, locates such rectangles, deletes them all and instead adds an LED.
Either way, not too much fun.

shroffk:
So I have been staying out of this conversation despite encountering very similar issues...cause I have not tried to help my users like you guys are trying :)

Alarm sensitive shapes were/are used widely as status representation ( along with left click open details view )

I am urging users to fix these manually.
 - Convert these shaped to multi state LED with the use of PV.SEVR or one of the alarm formulas. I have also managed to convince a few to display the alarm on the border and "OPEN"/"CLOSE" type state info using the LED colour... thus being consistent with all other widgets.
 - Add description to the actions which used to happen on left click, this way they make sense when they show up in the context menu.
 - Also, for quite a few of them we found justification to create new PV's instead of computing status in OPI screens. The summarizing PV's is useful not just for the sake of simple UI, but also to make things more clear to the end user.

As annoying as this has been, due to the high use of linking containers to embed screens.. the number of screens that had to be fixed is still manageable ( a few dozen vs hundreds/thousands )

shroffk:
> Access to PV.SEVR or PV.EGU or PV.PREC means that it only works with an IOC and actual records, not with some other channel access or PV access server.

+1 
We mainly use formulas... and try to move things into IOC's if possible.

> Of course we don't have formula support for all the pieces, but we can add them as needed.

+1

PR #2679, opened 2023-05-17
#2667 Add support for java8 Instant to the objectmapper

shroffk:
Fix json serialization issue with alarm command messages


shroffk:
To fix the parsing of the message time which is a Instant

> ``` com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.Instant` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: org.phoebus.applications.alarm.messages.AlarmCommandMessage["message_time"])```

georgweiss (approved):

Issue #2535, opened 2023-02-07
Setting es_sniff to false doesn't disable auto-configured Elasticsearch sniffer


jbellister-slac:
### Context

When doing a deployment of the latest 4.7.1 version of the alarm-logger, this error was logged every 5 minutes despite leaving es_sniff to the default of false:

```
Feb 06, 2023 4:57:10 PM org.elasticsearch.client.sniff.Sniffer$Task run
SEVERE: error while sniffing nodes
java.net.ConnectException: Connection refused
	at org.elasticsearch.client.RestClient.extractAndWrapCause(RestClient.java:918)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:299)
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:287)
	at org.elasticsearch.client.sniff.ElasticsearchNodesSniffer.sniff(ElasticsearchNodesSniffer.java:106)
	at org.elasticsearch.client.sniff.Sniffer.sniff(Sniffer.java:209)
	at org.elasticsearch.client.sniff.Sniffer$Task.run(Sniffer.java:140)
	at org.elasticsearch.client.sniff.Sniffer$1.run(Sniffer.java:81)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:777)
	at org.apache.http.impl.nio.reactor.DefaultConnectingIOReactor.processEvent(DefaultConnectingIOReactor.java:174)
	at org.apache.http.impl.nio.reactor.DefaultConnectingIOReactor.processEvents(DefaultConnectingIOReactor.java:148)
[...]

PR #2684, opened 2023-05-23
#2535 @jbellister-slac's fix to disable duplicate springboot es client

shroffk:
@jbellister-slac
adding your fix to the default alarm logger properties


jbellister-slac:
Looks good to me, thanks!

PR #2685, opened 2023-05-24
CSSTUDIO-1905 Improve caching of ImageView of the widgets "Picture" and "Symbol"

abrahamwolk:
This merge-request fixes an issue with excessive VRAM usage when zooming in on a Picture or Symbol widget using the D3D library for rendering under Windows.

**Picture Widget.** When working on an OPI that contains a large image as a background that is rendered by an instance of the Picture widget, Phoebus would freeze and become unresponsive when zooming in on the OPI, when running Phoebus under Windows using the D3D library for rendering. Under Linux (using the ES2 driver for rendering), the Phoebus application remained responsive under the same circumstance.

A workaround was to launch Phoebus with the flag **-Dprism.maxvram=2048M** or higher, depending on the computer (the default value is 512M).

In contrast, Phoebus remained responsive under Linux even when launching Phoebus with the flag **-Dprism.maxvram=50M**.
(On the other hand, exceptions are thrown also under Linux with this low amount of VRAM in this circumstance, but the application doesn't freeze.)

Setting the cache hint of the `ImageView` that holds the displayed image to `Cache.Hint.SCALE` seems to solve the problem: Phoebus remains responsive under Windows even when launching Phoebus with the flag **-Dprism.maxvram=100M**. It therefore seems likely that the problem is related to the default caching policy of D3D somehow. As an additional benefit, zooming is much smoother under both Linux and Windows under this change.

**Symbol Widget.** Freezes are also observed when zooming in on OPIs that contain instances of the Symbol widget (even though the widget sizes are small). Setting the caching policy in the same way as for the Picture widget seems to solve this issue, as well, and it seems likely that the cause of the problem is the same as for the Picture widget.

**Reduced Image Quality?** It is of course possible that the quality of the rendered image is reduced, but to me it looks acceptable in the instance I am looking at. I don't know whether it would be preferable to disable caching after a zoom operation has completed?

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
@@ -58,7 +58,7 @@ public class PictureRepresentation extends JFXBaseRepresentation<ImageView, Pict
         final ImageView iv = new ImageView();
         iv.setSmooth(true);
         iv.setCache(true);
-        iv.setCacheHint(CacheHint.SCALE);
+        iv.setCacheHint(CacheHint.SCALE);  // Prevents excessive VRAM usage when zooming in using the D3D library for rendering under Windows.
         iv.getTransforms().addAll(translate, rotation);
         return iv;
     }
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
@@ -11,6 +11,7 @@ import static org.csstudio.display.builder.representation.ToolkitRepresentation.
 
 import java.util.logging.Level;
 
+import javafx.scene.CacheHint;
 import org.csstudio.display.builder.model.DirtyFlag;
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.UntypedWidgetPropertyListener;
@@ -56,6 +57,8 @@ public class PictureRepresentation extends JFXBaseRepresentation<ImageView, Pict
     {
         final ImageView iv = new ImageView();
         iv.setSmooth(true);
+        iv.setCache(true);
+        iv.setCacheHint(CacheHint.SCALE);
         iv.getTransforms().addAll(translate, rotation);
         return iv;
     }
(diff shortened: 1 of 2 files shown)


georgweiss (approved):

kasemir (approved):

kasemir:
Interesting. Our operational setup is on Linux, and development on Mac, so can't check this out.
When you say zooming in, do you refer to the overall display zoom from the display's toolbar? Do users frequently zoom in/out that way at your site?
Also, would the same apply to the plots based on `org.csstudio.javafx.rtplot.internal.PlotCanvasBase` which extends `ImageView`?

shroffk:
I have a windows development setup... since we rarely use dynamic zooming ( usually users know a fixed zoom size they always use to fit something on a screen and they keep using that ) I had not come across this issue.

> Picture widget, Phoebus would freeze and become unresponsive when zooming in on the OPI

I tried recreating it but was not successful... is there something specific I need to do to recreate the freezing issue?
I am using DirectX 12

0lkemglp

georgweiss:
In my test case I hade several Picture and Symbol widgets. Removing some of them resolved the issue, but the freeze behavior was quite reproducible on several Windows boxes.

georgweiss:
Tried both.

georgweiss:
SVGs are transcoded to PNGs in code, if I recall correctly.

kasemir:
So it's not actually an SVG issue but indeed an ImageView thing?
Kunal, nice screen recording, but the combination of people climbing stairs and the in/out zoom gives me vertigo..

shroffk:
> Kunal, nice screen recording, but the combination of people climbing stairs and the in/out zoom gives me vertigo..

Brilliant... I will attach this to all my Issues/PR's now :)

> In my test case I hade several Picture and Symbol widgets. Removing some of them resolved the issue, but the freeze behavior was quite reproducible on several Windows boxes.

I was just working on a larger test case... I did one with 15 images... each image is 3MB. Still no luck reproducing this :(
These are all png images ( but I understand that we can rule one the image type here )

brhz2eue

I don't have an issue with setting the cache hint to scale for picture and symbol... so we can merge this PR before I complete the release today

shroffk:
I was just hoping to reproduce the issue... not having reproduceable issue is quite annoying.

kasemir:
Yes, since if it's indeed an ImageView issue, all the plots should show the same excessive VRAM usage on Windows.

abrahamwolk:
It is possible that it relates to the way SVG images are loaded: they use a slightly different code path (invoking `SVGHelper.loadSVG()`) than PNG images, and it is possible that a different caching behavior is triggered somehow when using the D3D library. This is only one possible hypothesis, however; I did not investigate further after I found this solution to the problem.

abrahamwolk:
The issue is not restricted to SVG images (and maybe it doesn't apply to SVG images): I have reproduced it using only PNG images (I use 130 small Picture widgets (120x100) and 1 large (2110x1270) Picture widget as a background, all with PNG images (the small ones use the same image file), and then zoom in and out several times).

I have not reproduced the issue using only SVG images yet, but I will continue to experiment to see if it can be reproduced using only SVG images. Adding one small PNG image to an OPI with many SVG images triggers the bug, however.

I run Phoebus with the flag **Dprism.maxvram=100M** in order to more easily reproduce the bug.

PR #2697, opened 2023-06-01
Macro Update

kasemir:
Comprehensive update of macro handling.

Most macros are straight forward. For example, a display is opened with macro `DEV=Motor`. Inside the display are two groups that set `N=1` respectively `N=2`. A widget in each group then uses the PV `$(DEV)$(N)`.

Macros originally used late evaluation. As an inner widget evaluates  `$(DEV)$(N)`, it queries its parent widgets for the values of `$(DEV)` and `$(N)`.
The immediate parent, the group, returns `N=1` respectively `N=2`, and obtains `DEV=Motor` from its parent, the display, overall resulting in `Motor1` respectively `Motor2` for the PV.

This approach works fine for most displays and is memory-efficient, but breaks when macros are re-defined across the hierarchy.
Display tools like EDM and BOY treat macros like environment variables in nested processes, expanding them early.
`export PATH=“$HOME/bin:$JAVA_HOME/bin` will right away expand `$HOME` and `$JAVA_HOME` and then assign the result to the `PATH`.
For the above macro example, the end result would be equivalent, except that the inherited macro values are copied down the widget hierarchy.
The display receives a macro `DEV=Motor`. Each group then inherits that macro and adds a value for `N`, so one group will hold macros `DEV=Motor,N=1` and the other group `DEV=Motor,N=2`. The widgets within each group finally get macros `DEV=Motor,N=1,PV=Motor1` respectively  `DEV=Motor,N=2,PV=Motor2`

The crucial difference appears when macros are re-defined down the widget hierarchy. For example, start with `X=x` and later `X=$(X)$(X)$(X)$(X)`. With late evaluation this results in a recursive definition within the child widget, starting with `X=$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)$(X)`. With early evaluation, we get `X=x` in the parent widget and  `X=xxxx` in the child widget.

Another example would be `A=a,B=b` in a parent widget and `X=$(A),A=$(B),B=$(X)` in a child widget. With early expansion, the child widget gets macro values `A=b,B=a,X=a`, whereas late evaluation gets an infinite recursion in the form of `B=$(X)=$(A)=$(B)`, i.e., `B=$(B)`.
Since legacy BOY displays and auto-converted EDM displays may contain such macro-redefinition, a partial early expansion of macros has been added to handle specific cases, but as shown in #2662 there will always be more cases.

This update thus turns the `Macros` class into one that separates specifications like `A=a,B=b,X=$(A),A=$(B),B=$(X)` from the resulting expanded values `A=b,B=a,X=a`. After a display has been loaded, all macros are expanded down the widget hierarchy.

As mentioned in #2662, early expansion as in environment variables and macros in EDM, BOY, .. is simply the way it has to be, the late expansion was a bad idea. The example displays as well as some local display tests work fine, but since it's a comprehensive update there could be some impact on existing displays that we'll have to handle as they become known.


georgweiss:
Yes:
Top level OPI defines macro "bobfile". The embedded display of a tab widget then uses this as its filename property, which fails as it is expanded to $(bobfile).

kasemir:
[example.zip](https://github.com/ControlSystemStudio/phoebus/files/11725319/example.zip)

georgweiss:
Exactly!

Issue #2695, opened 2023-06-01
Translations are observed instead of rotations when rotating Picture or Symbol widgets


abrahamwolk:
The merge-request https://github.com/ControlSystemStudio/phoebus/pull/2685 can cause issues when Picture or Symbol widgets are rotated: instead of a _rotation_, a _translation_ of the image can be observed.

This issue has been observed on Mac, Linux, and Windows. Under Windows it has been observed to depend on the computer that runs Phoebus: in at least one instance the problem has _not_ been observed, and in at least one instance (on a different computer) the problem _has_ been observed. Under Windows, at least, it therefore seems to depend on the specific setup.

The cause is the setting in https://github.com/ControlSystemStudio/phoebus/pull/2685 of `imageView.setCacheHint(CacheHint.SCALE)`. Why this setting causes the problem is not understood. A guess is that instead of _optimizing_ for a scaling transformation, the graphics system in these cases continues under the assumption that _only_ scaling transformations will be applied.

Experimentation has shown that instead setting `imageView.setCacheHint(CacheHint.SPEED)` seems to work.

The fact that the original option (`CacheHint.SCALE`) didn't work on some setups, suggests that it is conceivable that `CacheHint.SPEED` may also fail on some setups. Such a failure has not been observed yet, but only very limited testing has been carried out.

The question is how to fix this. Options include:

 1. Change from `CacheHint.SCALE` to `CacheHint.SPEED`
 2. Add a boolean option to `Phoebus.ini` that when set to `false` turns off the caching (i.e, it leaves the caching as it was before https://github.com/ControlSystemStudio/phoebus/pull/2685), and when set to `true` enables `CacheHint.SPEED`. It is an open question what the default value of the option should be.
 3. Add an option to `Phoebus.ini` that allows for a selection between all available `CacheHint` options: `CacheHint.DEFAULT`, `CacheHint.SPEED`, `Cachehint.SCALE_AND_ROTATE`, `CacheHint.SCALE`, `CacheHint.ROTATE`, and `CacheHint.QUALITY`. (The different options are documented at [1].) The rationale for this option would be that since the behavior is outside of our control and we cannot possibly test all kinds of setups, we cannot rule out that any option we choose doesn't fail on some setup.  Under this option, the user can test the available options and choose the one that works on their setup. It is an open question what the default value of the option should be.
 
[1] https://docs.oracle.com/javase/8/javafx/api/javafx/scene/CacheHint.html

[...]

PR #2701, opened 2023-06-07
CSSTUDIO-1948 Add an option to the Phoebus-initialization file that sets the cache hint of images in the Picture- and Symbol widgets

abrahamwolk:
This pull-request implements a new option for the Phoebus-initialization that allows the setting of a cache hint for images in the Picture- and Symbol widgets. By default, the default caching behavior is used.

This merge-request is intended to fix the issue https://github.com/ControlSystemStudio/phoebus/issues/2695.

The code-structure is perhaps not optimal: the `protected static` method `setCacheHintAccordingToPreferences()` is placed in the class `PictureRepresentation`. This method is not only called from `PictureRepresentation`, but also from the class `SymbolRepresentation`. Placing the code in a separate class would be preferred from a structural point-of-view, however, it would introduce an additional class just for the purpose of declaring this method. It is also not difficult to refactor the code if the method needs to be used in other places, as well. If declaring the method in a separate class is preferred, I will change this.

The option that has been added to the Phoebus initialization file is called `cache_hint_for_picture_and_symbol_widgets`, which is very specific. The intention is that the meaning of the option is clearly defined, since it is a workaround around an issue that is not fully understood (and which we don't seem to have control over). A drawback with such a specific name is that if the workaround is to be extended to, e.g., other widgets in the future, then the name is not valid anymore.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PictureRepresentation.java
@@ -12,6 +12,7 @@ import static org.csstudio.display.builder.representation.ToolkitRepresentation.
 import java.util.logging.Level;
 
 import javafx.scene.CacheHint;
+import javafx.scene.Node;
 import org.csstudio.display.builder.model.DirtyFlag;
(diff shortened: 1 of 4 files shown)


abrahamwolk:
@kasemir: I have now changed the parsing to use `CacheHint.valueOf()`.

PR #2707, opened 2023-06-12
Tabs: Expand macros into widgets of each tab

kasemir:
Fixes #2697


georgweiss (commented):
While this fixes the runtime behavior, in edit mode the bobfile macro in the example is not resolved.

kasemir:
Try now

georgweiss:
That's it! Thanks.

PR #2712, opened 2023-06-15
CSSTUDIO-1928 Re-implement the computation of the values on the scale on the ordinate when using a logarithmic scale in the data browser.

abrahamwolk:
This pull-request re-implements the computation of the values on the scale on the ordinate when using a logarithmic scale in the Data Browser. The intent is that the new implementation provides more intuitive behavior.

In addition, it fixes the following two issues that were present before:
1. `zero_threshold` is set to `Double.NEGATIVE_INFINITY` (i.e., the threshold is "disabled"), since on a logarithmic scale, values arbitrarily close to zero are meaningful to consider.
2. A check that the parameter `low` (i.e., the lowest value on the scale on the ordinate) is strictly greater than `0.0` has been added (since the values in the range (-∞, 0] may not be displayed on a logarithmic scale).

The newly implemented scale operates in one of three different modes, depending on how "zoomed-in" the scale is:

1. If the interval the scale covers is so large that there is not sufficient space to have a major tick for every power of 10, then a number of powers of ten are displayed that fit in the available space.
2. Otherwise, if the interval the scale covers is small enough that at least two powers of ten fit in the available space, then a major tick is created for every power of ten that fits on the scale, and minor ticks (with no label) are created for every integer multiple of powers of ten that fits in the available space.
3. Otherwise (1 or 0 powers of ten fit in the available space), a linearly spaced scale is created, based on the lowest and highest value of the scale and the number of major ticks that fit on the screen.


abrahamwolk:
I see now that there are failing tests. I will fix this.

abrahamwolk:
The pull-request is ready for review now.

kasemir:
Generally very happy to see that somebody else it taking this on!
My concern is erratic axes ranges.
For example, assume range was 0 .. 1000, then user "zooms out" so range changes to say -500 .. 1500.
Or vacuum readings which tend to be in the range 1e-2 to 1-e9, but then there's a communication error with the vacuum gauge controller and the PV becomes 0 or -1, so the axis range becomes 0 .. 1e-2 or -1 .. 1e-2.
What will the axis labels be? There was no unit test for this because it's unclear what should be done, but it has to show _something_ and not crash in those scenarios. Can you try those?

kasemir:
> a range for a logarithmic scale must be a sub-interval of (0, ∞); negative values are not allowed

That's the correct answer if you're in a math exam.
In the control system, "Vacuum" PVs can report 0, -1, 0xFFFFFFFF or anything they want when disconnected from the device. You'll find those values in the archive, and we need to somehow show them, not crash, not freeze the GUI in an infinite loop, not show an empty value range.

kasemir:
See comment in "Log10" and reference to how it's handled in JFreeChart.

abrahamwolk:
It's just a fact that a logarithmic scale cannot show negative values, just like an "ordinary" 2-dimensional linear scale cannot show complex numbers. Of course, one can change the logarithmic scale into a non-logarithmic scale and display more values, but it is no longer so clear what one is displaying, since it is no longer really a logarithmic scale.

Currently, `Log10.log10()` maps `(-∞, 0]` to `-1e100`. This seems inadvisable to me, since `Log10.pow10(-1e100)` will be mapped to the same value, and hence cannot be distinguished from "invalid" inputs. It also makes the transformation discontinuous, and continuous input can lead to a discontinuous graph, which I think is counter-intuitive.

I propose that we instead extend `Log10.log10()` to map `(-∞, 0]` to `Double.NEGATIVE_INFINITY`, and add a mechanism to the Data Browser that indicates where values are out of range. E.g., if someone attempts to plot `-1` on a logarithmic scale, then the lower horizontal border of the graph could indicate that there is a value that is below range of the scale at that point. This would apply also in situations like min = 0, max = 10, and y = 11, where the upper horizontal border of the graph could indicate that there is a value above the range of the scale at that point. Then the user can see where there is data that is not represented on the screen, and investigate further if something seems wrong.

We could distinguish between two cases, by having two possible indications:
- One where values are out-of-range, but in principle displayable (i.e., by scrolling, the data-point will eventually appear) (e.g., linear scale with min = 0, max = 10, y= 11)
- One where values are in principle not displayable with the chosen kind of scale, no matter how far one scrolls (e.g., logarithmic scale with min = 0,1, max = 10, y = -1)

kasemir:
Sounds great. So we merge this one as is?

kasemir:
If you leave your email and phone number, you're free to do anything to the log axis.
Yes, fixing the freezeup sounds good.
Also consider leaving keeping like this at the end:
```
if (major_ticks.size() < 2)
{   // If the best-laid plans of mice and men fail
    // and we end up with just one or no tick,
    // add the low and high markers
    major_ticks.add(0, new MajorTick<>(low, format(low)));
    major_ticks.add(new MajorTick<>(high, format(high)));
}
```

abrahamwolk:
The pull-request is ready for review again.

I believe I have now fixed the freezing, through two changes:

- In the abstract class Ticks, the method public abstract boolean isSupportedRange(XTYPE low, XTYPE high) has been changed to public abstract Pair<XTYPE, XTYPE> adjustRange(XTYPE low, XTYPE high), with the change that the method adjusts the range given as its input to a range that is meaningful for the chosen type of axis (instead of just checking whether a range is meaningful). E.g., for a logarithmic scale, a negative min-value is not meaningful, and it is adjusted by LogTicks.adjustRange to Math.ulp(0.0).
- The computation of the linear scale now aborts if certain values involved in the calculation are not representable by the type `double`. (Previosuly, the code would enter an endless loop when the precision of `double` was not sufficient to represent these values in the calculation.)

I have also added back the fallback case, which can be relevant in the cases the linear scale cannot be computed.

kasemir (approved):
Looks like good progress, I'm for merging this

PR #2720, opened 2023-06-21
CSSTUDIO-1964 Add the option "org.csstudio.display.builder.model/enable_saved_on_comments"

abrahamwolk:
This PR adds the boolean option `org.csstudio.display.builder.model/enable_saved_on_comments` for enabling and disabling the comment of the form `<!--Saved on ... by ...-->` in the output when saving an OPI in the Display Editor.

If the option is not present in the Phoebus.ini file, the default value is `true`.


georgweiss (commented):
I have no objections, but I'd like to clarify the background. We maintain bob files in git, and the comment triggers a diff considered irrelevant when comparing versions. Hence our users would like do disable the comment, if possible.

kasemir (approved):
Looks good, specifically how AllWidgetsAllProperties will always enable comments.

georgweiss (approved):

Issue #2727, opened 2023-07-05
Add horizontal & vertical alignment properties of content in Action/Choice/Bool button


Mathis-Hu:
Add horizontal & vertical alignment properties of content in Action/Choice/Bool button

shroffk:
This functionality was added to dbwr too

PR #2728, opened 2023-07-05
Add horizontal & vertical alignment properties #2727

Mathis-Hu:


shroffk:
@kasemir @georgweiss 
Some basic text alignment properties for 3 widgets

georgweiss (approved):
Seems to work.
Only remark is variable names: Java convention is camelCase.

kasemir (approved):
Doesn't seem to be an essential set of properties, but if you have time for this, can you also support them in the web runtime, https://github.com/ornl-epics/dbwr ?

PR #2732, opened 2023-07-11
CSSTUDIO-1133 Change condition when selecting distance between major ticks in `LinearTicks.selectNiceStep()`.

abrahamwolk:
This pull-request fixes an issue with the step-size selection for linear scales.

`LinearTicks.selectNiceStep()` is called (except for test-cases) only in `LinearTicks.compute()` with the argument `min_distance` set to the minimum distance that doesn't result in too many major ticks on the screen (in the sense that they would be too dense). The call is preceded by the comment `// Round up to the precision used to display values`.

However, `selectNiceStep()` can sometimes also round _down_ (not just _up_) the distance between major ticks.  For example:
 - Suppose that `distance` has been calculated to be `0.29`. (I.e., with `Ticks.FILL_PERCENTAGE` set to `60`, a distance between major ticks of `0.29` would fill up around 60% of the available space with labels of major ticks.)
 - Then on line `272`, `order_of_magnitude` is calculated to be `0.1`.
 - Since `3.0 >= 2.9 >= 1.2` holds, `selectNiceStep()` will return the step size `2.0`.
 - The consequence is that now around 90% of the available space is filled with labels of major ticks, since the distance between the major ticks is only `0.2`, not `0.29` or greater.
 
Too dense scales have been observed when testing the Linear Meter (which uses `NumericAxis` (and therefore `LinearTicks`) for its scale). E.g. (the displayed warning is not directly relevant to the described issue):

linearTicks

 
This pull-request changes the implementation of `selectNiceStep()` to instead select the smallest nice step that is larger or equal to the calculated minimum distance.
 
To compensate for the rounding down that sometimes occured previously, `Ticks.FILL_PERCENTAGE` has been increased from `60` to `70`.
 
Test cases had to be updated, since the behavior of `selectNiceStep()` has changed.

### DIFF
--- a/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/internal/LinearTicks.java
+++ b/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/internal/LinearTicks.java
@@ -254,8 +254,7 @@ public class LinearTicks extends Ticks<Double>
      *  A computed distance of 6.1 turns into 10.0, not 5.0.
      *  @see #selectNiceStep(double)
      */
-    final private static double[] NICE_STEPS = { 10.0, 5.0, 2.0, 1.0 },
-                             NICE_THRESHOLDS = {  6.0, 3.0, 1.2, 0.0 };
+    final private static double[] NICE_STEPS = { 1.0, 2.0, 5.0, 10.0 };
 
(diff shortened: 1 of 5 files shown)


kasemir (approved):
Looks good to me

PR #2737, opened 2023-07-18
Olog editing support

shroffk:
associated with accompanying changes to the Olog service in PR https://github.com/Olog/phoebus-olog/pull/171


georgweiss:
Looks good. Users will probably ask for an "Edit" button in the toolbar above the log entry details...

To consider:
When edited log entry has been submitted, the search result list should be refreshed to reflect changes (e.g. title, tags, logbooks). Same if the edited log entry is the one currently shown in the details view, otherwise user may question whether the entry was actually updated.

shroffk:
+1 on the auto refresh

Issue #2748, opened 2023-07-26
Logging Config: Add 'Refresh', sort entries?


kasemir:
The logging config display/configurator could be improved in two ways.

1) Sort the entries, so it's easier to locate an entry you want to adjust. Right now they are in no predictable order.

<img width="551" alt="Screenshot 2023-07-26 at 11 56 05 AM" src="https://github.com/ControlSystemStudio/phoebus/assets/1932421/7def5c50-05be-45bc-b638-7e6a357dbd07">

2) Add a "Refresh" button. Each logger only shows up after it has been created/accessed, which might require running some part of the application. To then actually see the logger, it's currently necessary to close & re-open the logging config. Ideally, a refresh would update the tree such that currently expanded/selected entries remain, but a simple "Refresh" button that replaces the tree and starts over with all items expanded would still be more convenient then close/re-open.

kasemir:
Turns out sorting is possible by clicking on the "Logger Name" column.
Still, being sorted by name from the start might be a good idea, i.e., do a sorted insert when adding the new nodes.

PR #2749, opened 2023-07-27
#2748 sort the logging configuration tree by logger name

shroffk:
fixes the sorting part of the issue #2748


kasemir (approved):
Meets and exceeds all I ever wanted, at least regarding a logging configuration GUI. Thanks!

PR #2752, opened 2023-08-03
CSSTUDIO-1998 Add the keyboard shortcut "Ctrl + Shift + S" for "Save As...".

abrahamwolk:
This merge-request adds the keyboard shortcut "Ctrl + Shift + S" for "Save As...".


kasemir (approved):
even better!

PR #2754, opened 2023-08-03
Change formula autocomplete to match others autocomplete

Mathis-Hu:
Autocomplete for formula would only display proposals if the name was matching entirely a function name. I wasn't very pratical when searching for a formula approximative name because you forget what's the exact name. I modified the autocomplete behavior for formula to match the behavior for the others autocomplete (like sim://), wich is displaying proposal as long as the text entered match some part of the function name. @shroffk


Mathis-Hu:
I'll modify for the test to succeed.

Mathis-Hu:
Autocomplete is complicated for formula, working on alternative to better support nested formulas and multiple formulas combined with maths and operators.

PR #2756, opened 2023-08-07
CSSTUDIO-1995 New Widget Property "Run Actions on Mouse Click" for the Symbol widget

abrahamwolk:
This merge-request adds the option "Run Actions on Mouse Click" to the Widget Properties of the Symbol widget.

Enabling this option results in the running of specified Actions of a Symbol widget by clicking on the widget. The widget can also be selected using the "tab" key, and when a clickable Symbol widget is selected, it can be pressed also by typing the "enter" or the "space" keys.

There are two points I would like to ask for input on:
 1. The default tool-tip of the Symbol widget has been updated. Before this merge-request, the default tool-tip was `$(pv_name)\n$(pv_value)`. This merge-request updates the default tool-tip to be: `$(pv_name)\n$(pv_value)\n$(actions)`, so that any actions are also listed in the tool-tip. On the one hand, this is not optimal, since it adds extra information also in the case where a symbol widget is not clickable. On the other hand, it is already the case that symbol widgets can have associated actions that can be activated in their context menu (which is opened by right-clicking on the symbol widget in question), and perhaps it is therefore a good idea to include the list of actions in the tool-tip even when a Symbol widget is not clickable.
 
 2. The specified actions are invoked by the following code:
 ```model_widget.propActions().getValue().getActions().forEach(actionInfo -> toolkit.fireAction(model_widget, actionInfo));```
 In comparison, for instances of `ActionButtonWidget` the code for invoking the set of specified actions is considerably more complex:
 https://github.com/ControlSystemStudio/phoebus/blob/18ca8ef39ddd37b4786dd6cb527c903b20c7653e/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ActionButtonRepresentation.java#L319-L347
Is there a reason to invoke the same or analogous code when invoking actions from the Symbol widget, or is the current implementation in the merge-request correct (and the complexity of the invocation of actions in instances of `ActionButtonWidget` is due to other reasons)?

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -279,6 +279,7 @@ public class Messages
                          WidgetProperties_RingWidth,
                          WidgetProperties_Rotation,
                          WidgetProperties_Rules,
+                         WidgetProperties_RunActionsOnMouseClick,
                          WidgetProperties_Running,
                          WidgetProperties_ScaleFactor,
                          WidgetProperties_ScaleFormat,
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/SymbolWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/SymbolWidget.java
@@ -304,4 +304,9 @@ public class SymbolWidget extends PVWidget {
 
     }
 
+    @Override
+    protected String getInitialTooltip()
+    {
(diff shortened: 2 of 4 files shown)


kasemir (approved):

smngr:
Hi,

Sorry for duplication - I've commented on a few things today and also the EPICS mailboard.

IMO this Widget property would be useful across more widgets to replicate functionality we had in CS-Studio/.opi. Is there anything that would/should prevent me adding this to other widgets and submitting a PR later?

Issue #2743, opened 2023-07-24
Change color title of group container widget independently


Mathis-Hu:
It would be practical to add the possibility of changing the title color of a group container independently of the background color and the color of the line/title bar, because sometimes for the title bar style, the title label is barely visible and we do not want to change all the background or bar color. I did a first version in the code, but I'm creating an issue first rather than creating a pull request to ask if the design is ok with the rest of phoebus.
Before :
![image](https://github.com/ControlSystemStudio/phoebus/assets/100083521/25616b23-a03b-4a8a-b086-1b3ca4da34f6)
![image](https://github.com/ControlSystemStudio/phoebus/assets/100083521/25991794-44db-4e55-997d-cef3f372b0dd)

After :
![image](https://github.com/ControlSystemStudio/phoebus/assets/100083521/ee3b2048-38d9-492c-a06d-7cd02828bab0)
![image](https://github.com/ControlSystemStudio/phoebus/assets/100083521/81b9d450-e555-4994-a60e-89655238c72f)

Actually, foreground property is for the line and bar color (and label if the style is not TITLE) and background is for the background color (and label if style is TITLE)
In my modification, I looked how it's done with other widget and I'm doing foreground for label (no matter the style), line color property for lines & bar color, and background is still for background color.
To do the conversion with older view version, I had increase the GroupWidget version from 2.0.0 to 3.0.0.

Of course, if this update is implemented to Phoebus, I'll try adding it into DBWR also

Mathis-Hu:
@shroffk

mcnanneyd:
It looks like this commit is what is breaking our BOY to Phoebus conversions. I really like the addition of an independent color title. However, it does seems to break almost all of our screens during conversion. Here is the opi screen before conversion:
![image](https://github.com/ControlSystemStudio/phoebus/assets/91760294/f785eba9-c54a-44f1-af7e-a372652d9e6e)
And its attributes:
![image](https://github.com/ControlSystemStudio/phoebus/assets/91760294/81a78d08-e69e-4ecb-85e2-ce2772f6ed74)

After converting to bob, the screen looks as follows:
![image](https://github.com/ControlSystemStudio/phoebus/assets/91760294/84258317-7ace-4d8c-b921-97ea1ca3ac25)
And its attributes:
![image](https://github.com/ControlSystemStudio/phoebus/assets/91760294/8ba90149-d198-42de-8e2f-8fe291f5d5e8)

[...]

PR #2757, opened 2023-08-07
Change color title of group container widget independently #2743

Mathis-Hu:
Change color title of group container widget independently.
see issue https://github.com/ControlSystemStudio/phoebus/issues/2743.

If this pull request is merged, I'll implement the same behavior in DBWR.


Mathis-Hu:
@kasemir I change the value of the colors directly in the GroupWidget#configureFromXML method as you asked (if I understand it well), thanks for your feedback. Do not hesitate to give more.

Sorry for the unrelated commit, I still have some struggle with GitHub for chosing some commit and not the others.

kasemir (approved):
That looks good: Widget has new 'line color', and representation simply uses line color for line, fore/back colors for text.

The conversion from older versions is handled in `configureFromXML`, which sets the new line color from the old foreground, for "TITLE" mode the background becomes the foreground etc.

It's possible that this conversion needs to be tweaked as we open a bunch of older displays and find that we missed something, but that would then be a follow-up update.

PR #2758, opened 2023-08-08
CSSTUDIO-1987 New "Unsaved Changes" confirmation dialog

abrahamwolk:
This merge-request implements a new dialog when closing multiple tabs, as for instance when exiting "Phoebus" or running the command "Close All Tabs".

Instead of each tab displaying its own confirmation window (which can be difficult to locate), the changes to be saved or discarded are summarized in one window (there are several variations depending on which command is run):

screenshot

If an individual tab is closed, the original confirmation dialog is still shown. The text of the dialog, however, has been slightly updated.

Previous to this merge-request, the confirmation dialogs of instances of `DockItemWithInput` were added as event handlers when a close-request was received. This has been changed: now the code for the confirmation dialogs is explicitly called when commands for closing tabs are called. I believe that this improves the behavior in cases where the user cancels the close-action, since either all `prepareToClose()` methods will be invoked or none (in the case of cancellation).

**Additional screenshots**
When some of the instances with changes have been saved:
screenshot2

When all instances with changes have been selected:
screenshot3

When a save operation is in progress:
screenshot4

When a "Save as..." operation was cancelled by the user:
screenshot5


app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
```diff
@@ -33,6 +34,7 @@
 import org.phoebus.framework.persistence.Memento;
 import org.phoebus.framework.spi.AppDescriptor;
 import org.phoebus.framework.spi.AppInstance;
+import org.phoebus.ui.application.PhoebusApplication;
```
kasemir, line 37: Does DisplayRuntimeInstance need to access PhoebusApplication? It should only be aware that it's a DockItem and not care about the higher level PhoebusApplication. Maybe this is a leftover import that's not actually used?

kasemir (approved):
The end result looks great, and I like how the changes to applications are minimal.

Slightly worried because the shutdown is complicated: Closing of one or all apps is initiated on the UI thread, then we perform the actual writing of files in background threads,  eventually reacting to success or failure back on the user thread.
I find it impossible to tell if all is "correct" just from looking at the changes.

Have you tried pushing all the buttons for like an hour in various ways and found no hangup?

app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
kasemir: Ideal!

abrahamwolk:
> The end result looks great, and I like how the changes to applications are minimal.

Thank you!

> 
> Slightly worried because the shutdown is complicated: Closing of one or all apps is initiated on the UI thread, then we perform the actual writing of files in background threads, eventually reacting to success or failure back on the user thread. I find it impossible to tell if all is "correct" just from looking at the changes.
> 
> Have you tried pushing all the buttons for like an hour in various ways and found no hangup?

No, I have not, and during development I _did_ encounter unexpected bugs. I can spend some time tomorrow to test the implementation, and also to review it myself again. But bugs can of course remain.

One issue that I can think of is: suppose the the confirmation dialog is displayed, but then some background process causes an application instance that previously (when the confirmation dialog was created) did _not_ have unsaved changes to _have_ unsaved changes. (For instance, previously, incoming data points could trigger the autoscale-functionality of the Data Browser to mark the Data Browser as having unsaved changes.)

In the current implementation, these new unsaved changes will simply be ignored if the user proceeds to close the application. Is this something that needs to be taken into account?

kasemir:
> I can spend some time tomorrow to test ..

Yes, that sounds good, gives me a warm fuzzy.

> new unsaved changes will simply be ignored 

That's fine. "Close" checks what needs to be saved at that time, saves, done. If new changes come in, they're lost, which is fine because they're after the point in time where we started the closing.

core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
```diff
@@ -1263,22 +1293,347 @@ private void closeMainStage() {
...
+                                hBox.addEventHandler(MouseEvent.MOUSE_CLICKED, mouseEvent -> checkBox.setSelected(!checkBox.isSelected())); // Enable toggling checkbox by clicking on its label.
+
+                                Supplier<SaveStatus> saveIfCheckboxEnabled = () -> {
+                                    if (checkBox.isSelected()) {
+
+                                        Text saving = new Text("[" + Messages.UnsavedChanges_saving + "]");
+                                        saving.setFill(Color.ORANGE);
+                                        saving.setStyle("-fx-font-weight: bold;");
+                                        hBox.getChildren().set(0, saving);
+                                        boolean saveSuccessful = dockItemWithInput.save(monitor, prompt.getDialogPane().getScene().getWindow());
```
abrahamwolk, line 1517: Is this call to dockItemWithInput.save() blocking? (The return type boolean suggests it is, but I do not know what the parameter monitor does.)

abrahamwolk, line 1517: Is this call to `dockItemWithInput.save()` blocking? (The return type `boolean` suggests it is, but I do not know what the parameter `monitor` does.)

abrahamwolk:
> Suppose there are two different instances of the Data Browser, and they both are saved to the same file. Currently, both instances will be saved to the same file, the second one overwriting the first one. Should the "Unsaved Changes" dialog attempt to detect this problem?

One alternative could perhaps be to modify the save procedure such that it is not possible to save to the same file as another opened Data Browser.

It is problematic when there are two running instances of the Data Browser associated with the same file, also without the "Unsaved Changes" dialog: for instance, if one such instance runs in the Main Window and the other such instance runs in a Secondary Window without any other application instances in the Secondary Window, then when I close Phoebus and open it again, the second window will be restored but without any content. I am also unable to close the window, but must instead remove or edit the file `~/.phoebus/memento`. Therefore, perhaps the program state where two instances share the same input file should be avoided in the first place.

core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
kasemir: Yes, it's blocking. It's supposed to be called from a background job, and it can then use the 'monitor' to update the job's progress.

kasemir: See JobManager.schedule for submitting a JobRunnable, and the JobRunnable being 'run' with a JobMonitor

abrahamwolk:
I discovered a bug https://github.com/ControlSystemStudio/phoebus/pull/2759 where sometimes the wrong tab would be selected. I have added the fix to this pull-request, since it is part of the code that is part of the event handler that I have combined in this PR. The fix consists of always running the tab-selection code on the JavaFX-thread using Platform.runLater().

abrahamwolk:
This PR relies on the assumption that instances of `DockItemWithInput` are the only tabs that have changes that can be saved. Is this assumption correct?

kasemir:
Yes,  DockItemWithInput are the only tabs that may have changes that can be saved. Mind you, any DockItem can store/restore state from a memento, but the clean/dirty state that goes with save/saveAs is limited to DockItemWithInput.

abrahamwolk:
Thanks! I believe that this PR is ready to be merged then.

PR #2759, opened 2023-08-08
CSSTUDIO-1997 Add option to open most recently focused tab when closing a tab

abrahamwolk:
This merge-request adds the option `org.phoebus.ui/open_previous_tab_when_closing_tab` to Phoebus. When `org.phoebus.ui/open_previous_tab_when_closing_tab=true` is set (the default is `false`), then Phoebus will set the focus on the most recently focused tab when closing the active tab. If a tab is closed that is not active, the focus remains unchanged.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/Preferences.java
+++ b/core/ui/src/main/java/org/phoebus/ui/Preferences.java
@@ -53,6 +53,7 @@ public class Preferences
     @Preference public static int[] alarm_area_panel_invalid_severity_background_color;
     @Preference public static int[] alarm_area_panel_undefined_severity_background_color;
     @Preference public static String cache_hint_for_picture_and_symbol_widgets;
+    @Preference public static boolean open_previous_tab_when_closing_tab;
 
     static
     {
--- a/core/ui/src/main/java/org/phoebus/ui/Preferences.java
+++ b/core/ui/src/main/java/org/phoebus/ui/Preferences.java
@@ -53,7 +53,6 @@ public class Preferences
     @Preference public static int[] alarm_area_panel_invalid_severity_background_color;
     @Preference public static int[] alarm_area_panel_undefined_severity_background_color;
     @Preference public static String cache_hint_for_picture_and_symbol_widgets;
-    @Preference public static boolean open_previous_tab_when_closing_tab;
 
     static
     {
--- a/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java
+++ b/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java
@@ -176,6 +176,18 @@ public class DockItem extends Tab
         createContextMenu();
 
         setOnClosed(event -> handleClosed());
+        setOnCloseRequest(event -> {
+            // Select the previously selected tab:
+            var dockPane = getDockPane();
+            var recentlyOpenedTabs = dockPane.tabsInOrderOfFocus;
+            recentlyOpenedTabs.remove(this);
+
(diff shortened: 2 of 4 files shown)


shroffk:
This is quite intuitive, I would even be in favour of making the default true.

kasemir:
>  I would even be in favour of making the default true.

I'd go even further: Make that THE behavior. No added preference, that's just how it'll be.

For the implementation, I'm worried that `List<DockItem> tabsInOrderOfFocus` grows with reference to tabs that no longer exist, creating a memory leak. Maybe change it to a list of weak references?

abrahamwolk:
> For the implementation, I'm worried that `List<DockItem> tabsInOrderOfFocus` grows with reference to tabs that no longer exist, creating a memory leak. Maybe change it to a list of weak references?

Can tabs be closed without invoking the event handler set by `setOnCloseRequest()` (which will remove the `DockItem` from `tabsInOrderOfFocus`)?

abrahamwolk:
I have now made the behavior the only behavior.

abrahamwolk:
> > setOnCloseRequest() ..will remove the DockItem from tabsInOrderOfFocus
> 
> I'm not 100% sure: Is that only called when clicking the 'x' on the tab? Or also when closing the window?

You are right that the OnCloseRequest event-handler is not called when closing the window. I have added the following code to `handleClosed()` which _is_ called when the window is closed:

```
var dockPane = getDockPane();
if (dockPane != null) {
    dockPane.tabsInOrderOfFocus.remove(this);
}
```

However, isn't the whole `DockPane` instance de-allocated when a window is closed? It is good to have redundancy, however. And perhaps a tab can be closed in other ways, also, without invoking the OnCloseRequest event-handler.

> Regarding setOnCloseRequest, you add that call in DockItem line ~179, but there's also a call around line 597 for addCloseCheck which will now be disabled because `getOnCloseRequest() != null`:
> 
> ```
>     public void addCloseCheck(final Supplier<Future<Boolean>> ok_to_close)
>     {
>         if (getOnCloseRequest() == null)
>             setOnCloseRequest(event ->
>             {
> ```
That is a good point! In the PR https://github.com/ControlSystemStudio/phoebus/pull/2758 (_CSSTUDIO-1987 New "Unsaved Changes" confirmation dialog_) I have changed this behavior: instead the close-request handler is installed unconditionally in the constructor of `DockItem`. The two handlers in the two PRs will have to be combined. When I have merged one of these PRs, I will update the other so that the two handlers are combined into one.

For the time being, I have now modified `addCloseCheck()` in this PR to call any event handler that was already installed. `setOnCloseRequest()` is now invoked by `addCloseCheck()` as follows:
```
var alreadyExistingEventHandler = getOnCloseRequest();

setOnCloseRequest(event -> {
    // For now, prevent closing
    event.consume();
    // Invoke all the ok-to-close checks in background threads
    // since those that save files might take time.
    JobManager.schedule("Close " + getLabel(), monitor ->
    {
        if (prepareToClose()) {
            if (alreadyExistingEventHandler != null) {
                alreadyExistingEventHandler.handle(event);
            }
            Platform.runLater(() -> close());
        }
    });
});
```
I propose that this temporary fix be merged in this PR, and that I then combine the two event-handlers into one event-handler that is installed unconditionally in the constructor of `DockItem` in https://github.com/ControlSystemStudio/phoebus/pull/2758.

abrahamwolk:
With the new fixes in place, are there still concerns about memory leakages? And are there any other concerns, or can I go ahead and merge this version?

kasemir:
Go ahead

Issue #2760, opened 2023-08-08
Bitwise operations support


shroffk:
EDM CALC supports bitwise operations. We can inlcude bitwise operations support using Phoebus formula functions.

The main operations to be supported are
 - bitwise OR
 - bitwise AND
 - bitwise XOR
 - bitwise NOT
 - bitwise left and right shift

ideally we should use the java operands (|, &, ^, ~, >>, <<) however, these are already in use in logical and power operations... so we can just use
 =bitwiseOR
 =bitwiseAND
 ....
 
 Not ideal, but this would not raise any backward compatibility issues

PR #2761, opened 2023-08-08
Add bitwise operations in formula and add round() in spi

Mathis-Hu:
see issue https://github.com/ControlSystemStudio/phoebus/issues/2760
@shroffk


kasemir:
The check of `(long) value == (double) value` seems like a good idea, since the bitwise operations aren't obvious for floats.
The EPICS calc support used in the calc record or EDM doesn't give any warnings, it simply truncates to an integer.

Mathis-Hu:
Do you prefer that I truncate the value to integers or is it good that way ?

kasemir:
I think what you have now is good.
I'm a little concerned that it's not 100% compatible with the calc record or EDM, so there's a slight chance that for example displays converted from EDM will give different results; error instead of silent truncation.
Still, what you have now seems "correct", so better go with that until we see an actual issue with it.

PR #2767, opened 2023-08-15
Olog client to use new REST endpoint when creating entries

georgweiss:
Olog client to use the new REST endpoint whereby all attachments are bundled with the log entry request. Client also shows upload size limits (configured on server).


shroffk (approved):

georgweiss:
@shroffk, we need to merge https://github.com/Olog/phoebus-olog/pull/171, right?

shroffk:
I would like to merge https://github.com/Olog/phoebus-olog/pull/171

PR #2768, opened 2023-08-15
CSSTUDIO-1995 Bugfix: Clicking on a clickable Symbol Widget with mouse buttons other than the primary mouse button should not run the specified actions

abrahamwolk:
This pull-request implements a bugfix and improvements for https://github.com/ControlSystemStudio/phoebus/pull/2756. 

 - The bugfix is that clicking on a clickable Symbol Widget with mouse buttons other than the primary mouse button should not run the specified actions.
 - The improvements are to the intuitiveness of mouse- and keyboard-presses on clickable Symbol widgets.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SymbolRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SymbolRepresentation.java
@@ -27,6 +27,7 @@ import javafx.scene.effect.ColorAdjust;
 import javafx.scene.effect.DropShadow;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyEvent;
+import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
 import org.csstudio.display.builder.model.ArrayWidgetProperty;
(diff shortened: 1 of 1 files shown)


kasemir (approved):
The title is way too long, and instead of the array for buttonIsClicked you could use an AtomicBoolean (not for atomicity but for the closures), but otherwise perfect.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SymbolRepresentation.java
```diff
@@ -495,34 +496,47 @@ private void enableRunActionsOnMouseClick() {
         ColorAdjust increaseBrightness = new ColorAdjust(0, 0, 0.3, 0);
         ColorAdjust decreaseBrightness = new ColorAdjust(0, 0, -0.3, 0);
 
+        boolean[] buttonIsClicked = { false }; // Value is wrapped in an array as a workaround of the fact that Java doesn't allow non-final variables to be captured by closures.
```
georgweiss, line 499: You could consider using AtomicBoolean instead of boolean array, and then use AtomicBoolean.get() and AtomicBoolean.set().

georgweiss (commented):
I have commented on line 499

PR #2779, opened 2023-08-21
CSSTUDIO-2014 Sort PVList table after "refresh" & add default sorting according to name.

abrahamwolk:
This pull-request fixes a bug where the PVList application didn't sort the table with PVs after the "refresh" action, and this PR also adds a default sorting according to the name-column of the table when the application is first opened.


shroffk (approved):

shroffk:
small change.. but makes such a big difference +1

PR #2785, opened 2023-08-24
CSSTUDIO-2015 Remember open tabs in the "Navigation Tabs" widget.

abrahamwolk:
This PR adds support to Phoebus to remember the open tabs in nested occurrences of the Navigator Tabs widget.

When switching between tabs, the Navigator Tabs widget closes/disposes the OPIs that are not the currently opened one. If an OPI that is inside of a tab of a Navigator Tabs widget ("A") in turn contains a Navigator Tabs ("B"), then the information about which tab is currently opened in "B" is lost when another tab of "A" is opened.

This PR adds the functionality to the Navigator Tabs widget to remember which tabs are opened in nested Navigator Tabs widgets, and re-activates the corresponding tabs when a tab is selected and its OPI has been loaded.

**Note:** There is another widget that is similar to the Navigator Tabs widget: the Tabs widget. The functionality has not been implemented for the Tabs widget in this PR.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/NavigationTabsRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/NavigationTabsRepresentation.java
@@ -57,13 +57,13 @@ public class NavigationTabsRepresentation extends RegionBaseRepresentation<Navig
     private final DirtyFlag dirty_tabs = new DirtyFlag();
     private final DirtyFlag dirty_tab_look = new DirtyFlag();
     private final DirtyFlag dirty_active_tab = new DirtyFlag();
-    private class SelectedNavigationTabsWidget extends MutablePair<Integer, HashMap<String, SelectedNavigationTabsWidget>> {
-        public SelectedNavigationTabsWidget() {
+    private class SelectedNavigationTabs extends MutablePair<Integer, HashMap<String, SelectedNavigationTabs>> {
+        public SelectedNavigationTabs() {
             left = 0;
             right = new HashMap<>();
         }
     };
-    protected SelectedNavigationTabsWidget selectedNavigationTabsWidget = new SelectedNavigationTabsWidget();
+    protected SelectedNavigationTabs selectedNavigationTabs = new SelectedNavigationTabs();
     private final UntypedWidgetPropertyListener sizesChangedListener = this::sizesChanged;
     private final UntypedWidgetPropertyListener tabLookChangedListener = this::tabLookChanged;
     private final WidgetPropertyListener<Integer> activeTabChangedListener = this::activeTabChanged;
@@ -186,7 +186,7 @@ public class NavigationTabsRepresentation extends RegionBaseRepresentation<Navig
         dirty_active_tab.mark();
         toolkit.scheduleUpdate(this);
         tab_display_listener.propertyChanged(null, null, null);
-        selectedNavigationTabsWidget.left = tab_index;
+        selectedNavigationTabs.left = tab_index;
     }
 
     /** Update to the next pending display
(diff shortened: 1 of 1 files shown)


kasemir:
The NavigationTabs widget has an `active_tab` property. By default it's zero, so by default we first see the first tab, but it can be set to 1, 2, 3 or even a macro to control what tab is initially opened.

Will that basic functionality remain?

Open display and then nested tabs for the first time, and they'll open the `active_tab`?

When user then clicks through the various tabs, it may remember what was last open, not always starting over on the `active_tab`?

abrahamwolk:
@kasemir: Thanks, that is a very good point.

The Navigation Tabs widget should definitely start on `active_tab`.

Do you think that there are use-cases where `active_tab` should always be loaded? In that case it makes sense to add a boolean to the Widget Properties that controls whether the last opened tab is remembered or not.

I have just pushed an update to the branch that initializes the selected tab to `active_tab`. This PR cannot be merged yet, however, since I just found a bug that I need to fix.

kasemir:
> Do you think that there are use-cases where active_tab should always be loaded? In that case it makes sense to add a boolean to the Widget Properties that controls whether the last opened tab is remembered or not.

When a display is opened, the nav tabs should start on the `active_tab`, which I think you just added. When then clicking around inside the open display, it's OK to keep the last-open-tab, which I think was your main idea.
So no, it shouldn't always fall back to the `active_tab`, that's only used when opening the display. No need for another option to select "active or last".

Mind you the last-tab-open memory should be limited to a display.

 1) Open x.bob, and nav tabs start on their `active_tab`
 2) Click around, and nav tabs remember the last tab, not going back to 0 or  `active_tab`
 3) Close x.bob display, and the last-tab info goes away
 4) Open x.bob again, and you start over on `active_tab` as in step 1

abrahamwolk:
> When a display is opened, the nav tabs should start on the `active_tab`, which I think you just added. When then clicking around inside the open display, it's OK to keep the last-open-tab, which I think was your main idea. So no, it shouldn't always fall back to the `active_tab`, that's only used when opening the display. No need for another option to select "active or last".

It was not my idea, I have only implemented it. However, yes, this is what I have implemented in this PR currently.

> Mind you the last-tab-open memory should be limited to a display.

I agree, and this is also how it currently works.

**Issue.** I have just improved on the following issue: Before, only the widget name of nested Navigation Tabs widgets was used to retrieve the information about which tab was opened last in a nested Navigation Tab widget. This lead to a problem when two tabs of a Navigation Tabs widget in turn each contained a Navigation Tabs widget whose names were the same.

Now, the tab index as well as the tab name has been added to the domain of the mapping that returns the information about which tab was opened last. This is not perfect, since the set of tabs can change programmatically: it is possible (but I believe it is unlikely) that a tab T1 is programmatically replaced by another tab T2 with the same name and with the same tab index, such that T2 contains a nested Navigation Tabs widget with the same name as one contained in T1. In this case, the mapping remembering which tab was last opened will conflate the two nested Navigation widgets.

kasemir:
Changing the selected navigation tab is basically replacing the content of an embedded display. It's opening/loading an embedded display. The content of that display file might actually change:

Select tab 1 to load "tab1.bob". Select tab 2 to load "tab2.bob". Somebody updates "tab1.bob".  Select tab 1 to load "tab1.bob" -> It's now different from what used to be in tab 1 when we first looked at it. It might no longer contain a navtab that you want to restore, or contain a new navtab for which you haven't saved the selected tab.
As long as you don't crash and simply use the `active_tab`, that should all be fine.

kasemir:
> this behavior is a consequence of the intended functionality of the Navigation Tabs widget

Right. The content of embedded displays and the nav tabs is by design a black box.
While BOY would block the whole UI until all content had been loaded, combining it into one large model, we load & represent the "main" display, then load the embedded content in parallel threads, so the main display isn't supposed to know about or rely on anything in the embedded content.

Using a "path" constructed of tab indices, tab names, widget names as you describe sounds reasonable. Worst case, this might require changing some displays where all nav tabs are currently called "Navigation Tabs" (default name) into something more specific, but that's only needed for displays with many nested nav tabs that otherwise show up with the same "path".

abrahamwolk:
It is not perfect, but I believe that for practical purposes it will work well enough. If there is a better way to identify nested Navigation Tabs, however, I would prefer to change to that.

kasemir (approved):
OK, then let's go with this

PR #2798, opened 2023-08-31
Save restore

shroffk:


georgweiss (commented):
Maybe add "SaveRestore" to the default URL in the client app preferences?

georgweiss (approved):

Issue #2795, opened 2023-08-30
[Codeathon]Fix errors on javadoc generation


katysaintin:
Fix javadoc generation 
`mvn compile javadoc:javadoc`
generates several errors

- [x] core-formula
- [x] core-framework
- [x] core-util 
- [x] core-pva 
- [x] core-security
- [x] core-ui 
- [x] core-logbook 
- [x] app-pvtree
- [x] app-update
- [x] app-filebrowser
- [x] app-logbook-ui
- [x] app-logbook-olog-client
- [x] app-logbook-olog-client-es
- [x] app-logbook-olog-ui
- [x] app-logbook-elog
- [x] app-rtplot
- [x] app-display-representation-javafx
- [x] app-display-convert-medm
- [x] app-display-convert-edm 
- [x] app-display-linearmeter 
- [x] app-alarm-model 
- [x] app-alarm-logging-ui
- [x] app-scan-model
- [x] app-scan-ui 
- [x] app-channel-channelfinder 
- [x] app-channel-views 
- [x] save-and-restore
- [x] app-eslog 
- [x] service-alarm-server
- [x] service-archive-engine
- [x] service-scan-server

katysaintin:
Finish, 

```
> mvn compile  javadoc:javadoc
[INFO] Reactor Summary:
[INFO]
[INFO] phoebus (parent) ................................... SUCCESS [  0.715 s]
[INFO] dependencies ....................................... SUCCESS [  0.015 s]
[INFO] install-jars ....................................... SUCCESS [  0.561 s]
[INFO] phoebus-target ..................................... SUCCESS [  1.082 s]
[INFO] core ............................................... SUCCESS [  0.007 s]
[INFO] core-vtype ......................................... SUCCESS [  1.465 s]
[INFO] core-formula ....................................... SUCCESS [  3.747 s]
[INFO] core-framework ..................................... SUCCESS [  8.099 s]
[INFO] core-types ......................................... SUCCESS [  1.729 s]
[INFO] core-util .......................................... SUCCESS [  3.223 s]
[INFO] core-pva ........................................... SUCCESS [ 11.986 s]
[INFO] core-pv ............................................ SUCCESS [  4.652 s]
[INFO] core-security ...................................... SUCCESS [  2.267 s]
[INFO] core-ui ............................................ SUCCESS [ 11.395 s]
[INFO] core-logbook ....................................... SUCCESS [  2.710 s]
[...]

PR #2796, opened 2023-08-30
[codeathon]Fix errors on javadoc generation for core-formula module

katysaintin:
[codeathon]Fix errors on javadoc generation for core-formula module see issue #2795

PR #2799, opened 2023-08-31
[codeathon]Fix errors on javadoc generation on core-framework #2795

katysaintin:


core/formula/src/main/java/org/csstudio/apputil/formula/math/ACos.java
```diff
@@ -1,9 +1,14 @@
...
+/**
+ * Acos class see @Math acos function
+ */
 public class ACos extends OneArgMathFunction
 {
-    public ACos()
+	 /**
+     * Constructor
+     */
+	public ACos()
```
tynanford, line 11: @abrahamwolk comment from previous PR still applies here and a few other spots

https://github.com/ControlSystemStudio/phoebus/pull/2796#discussion_r1311244806

georgweiss (commented):
Not sure how to review. Enabling javadoc lint (disabled by default) still generates errors.

katysaintin:
> Not sure how to review. Enabling javadoc lint (disabled by default) still generates errors.
> 
> @tynanford has a point in this is work in progress, so I do not mind a merge.

Hie, actually, I commit module by module, there is still 5 modules in core that generates javadoc error. And the module in app also in errors. Today, I'm trying to finish all, I will make my pull request at the end that enscope severals commit. 

Katy

PR #2802, opened 2023-09-01
documentation update : services architecture update + added troubeshooting

lcaouen:
[documentation update : services architecture update + added troubesho](https://github.com/ControlSystemStudio/phoebus/commit/4b7035c456212cf86c7d5703eae7406a9db00cb7)


georgweiss (commented):
Actually the sections on JavaFX issues to me look a bit site specific. If Phoebus is built on same OS as target, the correct JavaFX libraries are included in the product tar.gz (or zip) file. 

It is also possible to do a cross-build by specifying -Djavafx.platform=\<os\> on the Maven command line, where os is linux, win or mac.

lcaouen:
We had the problem beacause we build Phoebus under windows and we deloy it on linux and windows.
However, if you think it's too specific, I can remove the section.

georgweiss:
Yes I think it is a bit too specific.

Again, if you need to build for both linux and win you can do that on Windows using -Djavafx.platform, but those would be two separate builds and two separate packages.

georgweiss:
Maybe you can add the option to do cross-builds to the troubleshooting?

georgweiss (approved):

PR #2803, opened 2023-09-01
Datasources and process variables examples

Mathis-Hu:
Without Alarm Datasource.

Need further discussion before implementing Alarm datasource.

@shroffk


shroffk:
I think it is great to include examples on how to use the formulas... given our general recommendation to avoid scripts it would be nice to have alternatives highlighted in the default examples

image

image

kasemir (approved):

PR #2805, opened 2023-09-04
[Codeathon]Fix errors on javadoc generation on all the project see issue #2795 

katysaintin:
That's it I finally finish this task. All the javadoc is generated for core, app and service


georgweiss:
Good job.
Looks good, only warnings remaining.
We should remove the doclint exclusion (line 250 in top level pom.xml) to be able to find offending javadoc going forward.

katysaintin:
> Good job. Looks good, only warnings remaining. We should remove the doclint exclusion (line 250 in top level pom.xml) to be able to find offending javadoc going forward.

There is a lot of warnings I start to fix them at the beginning, then I volontary focus on the errors. When I will have time, I will remove warnings also but the difficulty for warning is to put a description for a parameter or a method.

georgweiss:
Well, getting rid of warning would be a huge task, especially for classes like Messages, which are only meant as static references to localized text defined elsewhere. But it would certainly look better in our javadocs.

PR #2810, opened 2023-09-07
Improvements to the scoped authentication APIs

georgweiss:
Supported scopes now managed as enum values (instead of string values scattered around various locations...)

Also added a service interface such that clients/application may be notified when user signs in/out. Useful if UI elements need to be enabled/disabled depending on authentication status.

### DIFF
--- a/app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
+++ b/app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
@@ -35,6 +35,7 @@ import javafx.util.Callback;
 import org.phoebus.framework.jobs.JobManager;
 import org.phoebus.security.authorization.ServiceAuthenticationProvider;
 import org.phoebus.security.store.SecureStore;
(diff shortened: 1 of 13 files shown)


shroffk:
Looks good to me but I am not sure I understand all the changes here.

shroffk (approved):

georgweiss:
It's about:

1. Using enum values instead of strings to identify an authentication scope.
2. Provide a mechanism to notify when user has signed in/out. For instance, in save&restore the "restore" button can be enabled/disabled if user logs in/out. That way we can save us an error dialog in case restore fails for unauthenticated user.

kasemir:
"user logs in/out" gives me a headache. For Eclipse/CSS we tried to support logging in/out. Administration was hard; in our case LDAP servers morphed, certificates expired. While the Linux systems have a group of admins to handle such changes, I was left with "I cannot log into CSS". Next, end users really expected that "logging in" changes who they are for channel access, but CA (and PVA) use the current user as provided by the OS.

==> For phoebus I was hoping we stick with just authorization, and leave the authentication to the OS, as stated in https://control-system-studio.readthedocs.io/en/latest/authorization.html

PR #2813, opened 2023-09-08
CSSTUDIO-1903 Bugfix: Set colors using CSS when alarm_table_color_legacy_background enabled

kasemir:
See screenshots in comments for color handling update #2642

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmUI.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/AlarmUI.java
@@ -124,17 +124,17 @@ public class AlarmUI
         createColor(Preferences.alarm_area_panel_undefined_severity_background_color),                                                                     // UNDEFINED
     };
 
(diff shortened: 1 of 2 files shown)


abrahamwolk:
When `org.phoebus.applications.alarm/alarm_table_color_legacy_background == false`, there is a (transparent?) margin of 2 pixels above and below the background color: the result is that it is easy, in the vertical direction, to see the borders between the individual cells displaying severity levels.

In contrast, when `org.phoebus.applications.alarm/alarm_table_color_legacy_background == true` (the case that this PR affects), there is no margin around the background of cells in the severity columns. The consequence is that when the severity level of two vertically adjacent cells are the same, one cannot see where one cell ends and another cell begins in the vertical direction without comparing with adjacent columns.

Should this be changed?

kasemir:
Can you easily add the 2 px margin? That would of course look a little better.

kasemir:
Otherwise, it "works" well. I've had the old and patched version run over the weekend, and the patch indeed fixes the problem, no more missing colors.

kasemir:
Perfect, thanks!

PR #2814, opened 2023-09-11
Fix formatting of alpha value

georgweiss:
In ```JFXUtil#webRgbOrHex``` the conversion to a decimal value for the alpha channel is actually sensitive to the region setting on OS level. For regions using comma as decimal character  an alpha of 254 will be converted to ```1,00```, which results in CSS string like ```rgba(r, g, b, 1,00)```, i.e. alpha is 1. But for alpha < 254 the converted value comes out as  ```0,xx```, which feeds ```rgba(r, g, b, 0,xx)``` to CSS, i.e. alpha is 0.

This PR is a fix where the string format uses explicit locale setting (US).


kasemir (approved):
Looks good. I wonder if there is a more technical locale that will always give us the desired "dot", even if the US decides to adopt the metric system, change decimal separators to ':' etc. But then that's very unlikely to happen...

georgweiss:
> US decides to adopt the metric system

That's a good one...

PR #2816, opened 2023-09-11
Refactoring of save&restore modules for the sake of simplification

georgweiss:


app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/service/SaveAndRestoreClient.java
shroffk, line 1: Can we name the package `o.p.a.saveandrestore.client`

georgweiss: Sure

georgweiss: Done

georgweiss: Updated, done now.

PR #2818, opened 2023-09-12
CSSTUDIO-2025 In the Alarm Log application, always use the Kafka server specified in `phoebus.ini`

abrahamwolk:
This PR implements the following two changes to how the Alarm Log application is restored:

 1. Currently, the server stored in the file `.phoebus/memento` is used for restoring instances of the Alarm Log application. This PR changes the initialization to instead always use the Kafka server specified in `phoebus.ini`. (I.e., to always use the server specified under `org.phoebus.applications.alarm/server`.)
 2. It adds a check for whether the config name stored in `.phoebus/memento` is available in the list of config names specified in `phoebus.ini` (i.e., the option `org.phoebus.applications.alarm/config_names`): if it is, then the stored config name is used, otherwise the default config name specified by the option `org.phoebus.applications.alarm/config_name` is used.


georgweiss (approved):
Works as expected.

shroffk (approved):

PR #2837, opened 2023-10-06
All fields of PVAStructure must be added to BitSet in a put request

georgweiss:
Attaching an Epics db example and bob file.

In the bob file an Action Button is writing hard coded values to the PVATable record implemented in the db example.

[Archive.zip](https://github.com/ControlSystemStudio/phoebus/files/12831513/Archive.zip)

Initial values for the table can be set like so:

```pvput TST:ResultsTable value.A=[1,2] value=[7,8]```

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/PutRequest.java
+++ b/core/pva/src/main/java/org/epics/pva/client/PutRequest.java
@@ -11,6 +11,7 @@ import static org.epics.pva.PVASettings.logger;
 
 import java.nio.ByteBuffer;
 import java.util.BitSet;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by kasemir** [comment]
There is already
```
    /** @return Structure elements */
    public List<PVAData> get()
    {
        return Collections.unmodifiableList(elements);
    }
```

Can you use that instead of adding a `getElements` that would conceivably allow changing the structure?

georgweiss:
Done

kasemir:
Was wondering if maybe `getIndex` doesn't work on the `unmodifiableList` wrapper of the elements, but in the end it's about the same elements, so should be OK.

kasemir (approved):

georgweiss:
Quick test shows it is OK.

PR #2845, opened 2023-10-19
Execute command when parent_dir not on local filesystem

tanviash:
This PR handles the situation when execute command operation is needed to be performed for a BOB file that exists on external webserver. In such case, commands will be executed in current directory.


kasemir (approved):

kasemir:
Some clarification: Display runtime will try to execute commands within the folder that contains the display file.
So a display "/some/path/display.bob" can run a "script.sh" that's in the same folder.

But when the display is fetched as "https://my_host.site/some/path/display.bob", there were two issues:
1) command executor can't make "https://my_host.site/some/path" the current directory
2) command executor can't execute a script "https://my_host.site/some/path/script.sh" without first downloading it, which brings up the follow-up questions of where to put it, when to remove it, and if downloading and running stuff might not be a security concern.

So to execute scripts, they must be present in the local file system, be it actual disk or network file mount. To then allow a display "https://my_host.site/some/path/display.bob" to execute a script "/some/local/or/mounted/path/script.sh", this PR falls back from trying to run the script in the display's parent folder "https://...", which is still impossible, to just whatever the current directory might be.

Issue #2853, opened 2023-10-26
Shortcut for re-load displays, run displays and middle click for open in new tab


Mathis-Hu:
There were several shortcuts in CSS, like :
- Run a Display with F5
- Re-load a Display in editor and in runtime with Ctrl+G

We can also add the possibility of opening a display in a new tab with the middle click onto an action button with open display action (in addition to the shortcut Ctrl + Left Click that already exist for that purpose), it would give another way to do that, just like we can do it in a web browser.

kasemir:
> middle click

Middle-click is somewhat a Linux/X-only idea.
Plus it's already used to copy the PV name, since that's what EDM did.

Mathis-Hu:
> Plus it's already used to copy the PV name, since that's what EDM did.

The middle click is just when clicking on an action button, so it should not be conflictual with EDM, right ?

The goal of that shortcut is when you want to open a large number of displays, it's more fluid with middle click and there's a lot of shortcuts with Ctrl, so the user can use unwanted shortcuts.

PR #2854, opened 2023-10-26
Adding shortcuts : re-load, run displays, open tab

Mathis-Hu:
see issue : https://github.com/ControlSystemStudio/phoebus/issues/2853


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ActionButtonRepresentation.java
```diff
@@ -139,7 +139,7 @@ private void checkModifiers(final MouseEvent event)
         }
 
         // 'control' ('command' on Mac OS X)
-        if (event.isShortcutDown())
+        if (event.isShortcutDown() || event.isMiddleButtonDown())
```
kasemir, line 142: Middle click already copies a PV name, same as it did in EDM. For actions buttons that write a PV, middle click gets that PV name.
Now it would open the display?

app/display/editor/src/main/java/org/csstudio/display/builder/editor/DisplayEditor.java
```diff
@@ -19,7 +20,9 @@
 import java.util.logging.Level;
 import java.util.prefs.Preferences;
 
+import javafx.scene.control.*;
```
kasemir, line 23: We generally use individual imports, don't let the IDE collapse them into wildcard imports

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ActionButtonRepresentation.java
Mathis-Hu: If it's conflictual, I remove it.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/DisplayEditor.java
Mathis-Hu: Yes, I did not see my IDE collapse that, I did the correction, I'm just waiting for the middle click response and I'll re-push.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ActionButtonRepresentation.java
kasemir: Ideally, you could get both:
Middle-click generally copies the PV name. Middle-click on text update, bool button, action button that writes PV, and you get the PV name. That pleases those who are used to fetching PV names that way from previous EDM usage.

Middle-click on action button that opens display, on the other hand, opens that in new tab, same as ctrl-click.
So you'd have to peek into the action: It is only one, is it the type that opens display -> do same as ctrl-click. Otherwise ignore the middle click, because it's already fetching a PV name.

Maybe you get that for free, because action with PV doesn't open anything, while action that opens display doesn't have a PV?

Mathis-Hu: In the code, middle click only modify the target of the OpenDisplayActionInfo, so it only affect a action button with the open display action. But when testing the version with the middle click, you can middle click the button with a write PV action and it does the same as a left click (it execute the action), so it could be conflictual (in the version without, middle click doesn't execute the action).

Do you want me to put a condition like "if (action instanceof OpenDisplayActionInfo)" before the middle click just to be sure ?

kasemir: Thanks for checking.
So middle-click to open display works as designed.
Copy of write-PV also still works.
Yes, if you can make middle-click on write-PV to _only_ get the PV name and not actually write, that'd be great.

Mathis-Hu: I added a verification before the middle click, so now, it behave like before if the button is not a open display only.

While testing, I noticed (version without the changes) that when you have multiple actions on the same button, it behave differently, regardless of the shortcuts. You can left / right click or middle click, it will execute the action selected (without considering shortcuts). I let that behavior in the version with my changes. So you can middle click to open displays in new tab on button with open display action only.

kasemir: sounds good, thanks

Issue #2859, opened 2023-11-06
Sort or not alarm items


lcaouen:
We would like to have the possibility to keep the order we decided when creating the alarm items and not always sorting them by alphabetical order.
Would be nice to add a parameter "sort_alarm_items" to decide if we want to sort or not the items.

kasemir:
Associated PR has some details as to why it was abandoned. Close this?

lcaouen:
Yes it's abandoned. You can close it.

PR #2860, opened 2023-11-06
Parameter to sort or not alarm items

lcaouen:
Parameter to sort or not alarm items

https://github.com/ControlSystemStudio/phoebus/issues/2859

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/AlarmSystem.java
@@ -121,6 +121,9 @@ public class AlarmSystem extends AlarmSystemConstants
     /** Macros used in UI display/command/web links */
     public static MacroValueProvider macros;
 
+    /** Disable notify feature */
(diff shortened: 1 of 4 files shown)


kasemir:
> keep xml import / creation order for alarms

Well, the result will be somewhat unpredictable. When you edit the configuration or export and then re-import, the order will change. For operators, it's bad to present them with randomly moving alarm items. But they you may also nail a salami to your knee, doesn't make sense, but you're free to do it.

Please check the file endings, for some reason you added "Ctrl-M" (windows file endings) to some of the files.

lcaouen:
In my tests the "kafka" order is kept but I will do some more to be sure it works.
I've marked the PR as draft and corrected the file ending character.
I will mark it as "ready for review" after my tests.

kasemir:
I understand that it's sometimes desirable to control the order of major components in the alarm  config, but order-of-creation really isn't practical in the long run because you'll always have to re-import the complete setup.
What's usually more than good enough is simply adding an initial number:

01 Site Utilities
02 Front End
03 Linac
04 Ring
05 Target

PR #2865, opened 2023-11-12
Fix/workaround for unwanted ScrollPane blurring

georgweiss:
I should add that observed blur effect may  depend on the host's graphical capabilities. For instance, I cannot see it on my Mac display, but moving the window to an external monitor the blur becomes quite obvious.


kasemir (approved):

kasemir:
Thanks. Also add to Embedded Display?

georgweiss:
Ah, my bad. Yes, will add it to EmbeddedDisplay.

Issue #2856, opened 2023-10-30
Main Phoebus window wants to be on top....mostly


DuncanAtISIS:
When opening a dialog (ie from a script, using FileUtil.saveFileDialog) or popped out phoebus display (from an action), the main phoebus window (more often than not) immediately pushes the new window behind itself.

DuncanAtISIS:
Running on MS Windows 10

georgweiss:
Able to reproduce on Windows, but not Mac.
Looks like this is connected to rendering of the tooltip on the action button. Can you please confirm by testing the following:

1. Move mouse pointer away from the button quickly once button is pressed.
2. Set the tooltip of the action button to empty string

DuncanAtISIS:
Thanks for looking at this.  I tried those 2 tests and the results seem to confirm that it's related to tooltip rendering - the popped out display stayed on top.

georgweiss:
OK, thanks for testing. Not sure if this is a JavaFX bug. In any case I'll try to come with a fix.

DuncanAtISIS:
Thanks :)

georgweiss:
Turns out that the behavior described in this ticket is reproducible on Mac OS in a simple JavaFX application. So my suspicion about JavaFX may be incorrect.

georgweiss:
Also tested on Ubuntu (20) with Gnome: behaves like on Mac, i.e. new window is *not* pushed to background when mouse hovers over action button and tooltip is rendered.

georgweiss:
Fixed in #2868

PR #2868, opened 2023-11-13
CSSTUDIO-2070 Workaround for issue where tooltips give window the focus

abrahamwolk:
This PR contains a workaround for the bug reported in [1], where displaying a tooltip causes the window that contains the widget whose tooltip is displayed to receive the focus. This behavior has been reproduced under Windows and Mac OS, but not under Linux.

A workaround is not optimal when compared with a fix, but it seems unlikely that there will be a fix in JavaFX considering that there is a bug-report [2] with the status "Unresolved" from 2012 that seemingly describes the same problem that this PR works around. 

The workaround that I have implemented consists of disabling tooltips on windows that do not currently have the focus. This is achieved by installing event handlers on the `focusedProperty()` of windows. The event handlers install the tooltips of the nodes in the window when the window receives focus and they uninstall the tooltips of the nodes in the window when the window loses focus.

This functionality is centralized in the new class `TooltipHelper`: in this PR, calls to `Tooltip.install()`, `Control.setTooltip()` and `Tab.setTooltip()` have been replaced by calls to the new functions `TooltipHelper.installTooltip()`, and `TooltipHelper.setTooltip()`. (The method `TooltipHelper.setTooltip()` is overloaded to work on instances of both `Control` and `Tab`).

This PR also removes the reliance of the class `TooltipSupport` on the JavaFX implementation detail that the key `TOOLTIP_PROP_KEY = "javafx.scene.control.Tooltip` can be used to lookup the tooltip of a node.

As I see it, one main drawback of the way that the workaround is structured is that we cannot enforce using the compiler that tooltips that are added in the future to Phoebus must be added through these new helper functions; it would have been nice (but I cannot see a way to achieve this) to hide the methods `Tooltip.install()`, `Control.setTooltip()` and `Tab.setTooltip()` in the Phoebus project so that tooltips can only be installed through the functions `TooltipHelper.installTooltip()`, and `TooltipHelper.setTooltip()`.

Finally, I would like to point out another addition to the codebase from this PR: the new function `JavaFXHelper.deferUntilJavaFXPropertyHasValue()` can be used to write code that depends on possibly uninitialized JavaFX properties. How to use it can be seen in the functions `TooltipHelper.installTooltip()`, and `TooltipHelper.setTooltip()`.

[1] Phoebus GitHub Issue https://github.com/ControlSystemStudio/phoebus/issues/2856
[2] JavaFX Bug Report https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8088624


georgweiss:
I should add that I have not been able to reproduce on Ubuntu 20.04 running Gnome.

kasemir:
This is a pretty heavy workaround for something that we don't directly control; it's pretty close to implementing our own tooltip feature.
Have you looked at a simpler way to handle this?
Would it be possible to add a JFX mouse event _filter_ (not handler!) to the window, something that simply suppresses pretty much any mouse events for unfocused windows? Maybe that can prevent tool tips because the mouse events are suppressed before they ever reach the tool tip mechanism? That would be just one addition to the "stage" code instead of having to patch every tooltip installation.

abrahamwolk:
> This is a pretty heavy workaround for something that we don't directly control; it's pretty close to implementing our own tooltip feature. Have you looked at a simpler way to handle this? Would it be possible to add a JFX mouse event _filter_ (not handler!) to the window, something that simply suppresses pretty much any mouse events for unfocused windows? Maybe that can prevent tool tips because the mouse events are suppressed before they ever reach the tool tip mechanism? That would be just one addition to the "stage" code instead of having to patch every tooltip installation.

That is a good idea, I think. I will see if I can implement this.

abrahamwolk:
@kasemir: I have now re-implemented the workaround according to your idea. The result is *much* better. Thanks for the suggestion!

kasemir:
Does that actually work?? Instead of changing 82 files, simply this??

```
stage.addEventFilter(MouseEvent.MOUSE_MOVED, mouseEvent ->
            if (!stage.focusedProperty().get())
                mouseEvent.consume();
```

That'd be the simplification of the century

abrahamwolk:
> Does that actually work?? Instead of changing 82 files, simply this??
> 
> ```
> stage.addEventFilter(MouseEvent.MOUSE_MOVED, mouseEvent ->
>             if (!stage.focusedProperty().get())
>                 mouseEvent.consume();
> ```
> 

Yes, it seems to work.

> That'd be the simplification of the century

:)

This way of solving it just didn't occur to me. Thank you, again!

kasemir (approved):
Tremendous idea, brilliant!

PR #2870, opened 2023-11-14
Save restore ks

shroffk:


shroffk:
The conversion is in keeping with spring conventions that roles are represented as all the authorities converted to uppercase with a prefix of ROLE_ ( I can't find the documentation right now... it was a painful discovery )

georgweiss:
Ah, yes, the infamous Spring upper case sorcery... Forgot that one.

PR #2875, opened 2023-11-17
fix the bug in the creation of the userpassword credential for git push

shroffk:
@jeonghanlee 
fixed a bug in how the username/password credentials were created.

Tested with pushing alarm config changes to remote repo hosted on github using https and username/password


shroffk:
Note since github no longer supports passwords the above test was using username and github tokens.

PR #2877, opened 2023-11-20
CSSTUDIO-2012 Add "Line Spacing" option to the Widget Properties of the Label widget

abrahamwolk:
This PR adds the option "Line Spacing" to the Label widget.

By setting a value (note that the value can also be negative!), it is possible to set the line spacing between lines for Label widgets containing two or more lines of text.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -240,6 +240,7 @@ public class Messages
                          WidgetProperties_LevelLoLo,
                          WidgetProperties_LimitsFromPV,
                          WidgetProperties_LineColor,
+                         WidgetProperties_LineSpacing,
                          WidgetProperties_LineWidth,
                          WidgetProperties_Locale,
                          WidgetProperties_Macros,
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/LabelWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/LabelWidget.java
@@ -28,11 +28,14 @@ import org.csstudio.display.builder.model.WidgetCategory;
 import org.csstudio.display.builder.model.WidgetConfigurator;
 import org.csstudio.display.builder.model.WidgetDescriptor;
 import org.csstudio.display.builder.model.WidgetProperty;
+import org.csstudio.display.builder.model.WidgetPropertyCategory;
+import org.csstudio.display.builder.model.WidgetPropertyDescriptor;
 import org.csstudio.display.builder.model.persist.ModelReader;
 import org.csstudio.display.builder.model.persist.NamedWidgetColors;
 import org.csstudio.display.builder.model.persist.NamedWidgetFonts;
 import org.csstudio.display.builder.model.persist.WidgetColorService;
 import org.csstudio.display.builder.model.persist.WidgetFontService;
+import org.csstudio.display.builder.model.properties.CommonWidgetProperties;
 import org.csstudio.display.builder.model.properties.HorizontalAlignment;
 import org.csstudio.display.builder.model.properties.RotationStep;
 import org.csstudio.display.builder.model.properties.VerticalAlignment;
@@ -104,11 +107,16 @@ public class LabelWidget extends MacroWidget
     private volatile WidgetProperty<WidgetColor> background;
     private volatile WidgetProperty<Boolean> transparent;
     private volatile WidgetProperty<WidgetFont> font;
+    private volatile WidgetProperty<Integer> lineSpacing;
     private volatile WidgetProperty<HorizontalAlignment> horizontal_alignment;
     private volatile WidgetProperty<VerticalAlignment> vertical_alignment;
     private volatile WidgetProperty<RotationStep> rotation_step;
     private volatile WidgetProperty<Boolean> auto_size;
     private volatile WidgetProperty<Boolean> wrap_words;
+
+    /** 'line_spacing' property */
+    public static final WidgetPropertyDescriptor<Integer> propLineSpacing =
+            CommonWidgetProperties.newIntegerPropertyDescriptor(WidgetPropertyCategory.DISPLAY, "line_spacing", Messages.WidgetProperties_LineSpacing);
 
     /** Constructor */
     public LabelWidget()
@@ -122,6 +130,7 @@ public class LabelWidget extends MacroWidget
         super.defineProperties(properties);
         properties.add(text = propText.createProperty(this, Messages.LabelWidget_Text));
         properties.add(font = propFont.createProperty(this, WidgetFontService.get(NamedWidgetFonts.DEFAULT)));
+        properties.add(lineSpacing = propLineSpacing.createProperty(this, 0));
         properties.add(foreground = propForegroundColor.createProperty(this, WidgetColorService.getColor(NamedWidgetColors.TEXT)));
         properties.add(background = propBackgroundColor.createProperty(this, WidgetColorService.getColor(NamedWidgetColors.BACKGROUND)));
         properties.add(transparent = propTransparent.createProperty(this, true));
@@ -150,6 +159,12 @@ public class LabelWidget extends MacroWidget
     public WidgetProperty<WidgetFont> propFont()
     {
         return font;
+    }
+
+    /** @return 'line_spacing' property */
+    public WidgetProperty<Integer> propLineSpacing()
+    {
+        return lineSpacing;
     }
 
     /** @return 'foreground_color' property */
--- a/app/display/model/src/main/resources/org/csstudio/display/builder/model/messages.properties
+++ b/app/display/model/src/main/resources/org/csstudio/display/builder/model/messages.properties
@@ -223,6 +223,7 @@ WidgetProperties_LevelLow=Level Low
 WidgetProperties_LevelLoLo=Level LoLo
 WidgetProperties_LimitsFromPV=Limits from PV
 WidgetProperties_LineColor=Line Color
+WidgetProperties_LineSpacing=Line Spacing
 WidgetProperties_LineWidth=Line Width
 WidgetProperties_Locale=Locale
 WidgetProperties_Macros=Macros
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java
@@ -71,6 +71,7 @@ public class LabelRepresentation extends RegionBaseRepresentation<Label, LabelWi
         model_widget.propVerticalAlignment().addUntypedPropertyListener(styleChangedListener);
         model_widget.propRotationStep().addUntypedPropertyListener(styleChangedListener);
         model_widget.propWrapWords().addUntypedPropertyListener(styleChangedListener);
+        model_widget.propLineSpacing().addUntypedPropertyListener(styleChangedListener);
 
         // Changing the text might require a resize,
         // so handle those properties together.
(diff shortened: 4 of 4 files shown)


kasemir:
Well, I would generally really prefer not to waste time on detailed adjustments like line spacing.
Other EPICS display managers like EDM, MEDM, EDD/DM, DM2K didn't offer line spacing adjustments.
If you do add it, what should the units be?
Pixel? Relative to what? Font points? How about the "1.0", "2.0" factor used in MS Office?
Do you plan to implement this in the next tool?
Web Runtime https://github.com/ornl-epics/dbwr/?
Do you really have a need for adjusting the line spacing on all your labels, or is this actually about a very specific case where an unfortunate font implementation results in overlaps, so it's really a broken font??

abrahamwolk:
> Well, I would generally really prefer not to waste time on detailed adjustments like line spacing.

I disagree with this assessment. As I see it, Phoebus is primarily a tool for creating and and displaying GUIs for EPICS and EPICS related services. I think implementing functionality that can improve the UIs that are created increases the usefulness of Phoebus.

> Other EPICS display managers like EDM, MEDM, EDD/DM, DM2K didn't offer line spacing adjustments.

It has not been my impression that a requirement for added functionality to Phoebus is that it has previously been implemented in other display managers. Why would it be a requirement in this case?

> If you do add it, what should the units be? Pixel? Relative to what? Font points? How about the "1.0", "2.0" factor used in MS Office?

The unit is `pixels`. As far as I know, `pixels` is an absolute unit.

> Do you plan to implement this in the next tool? Web Runtime https://github.com/ornl-epics/dbwr/?

No, I have not planned to implement this for https://github.com/ornl-epics/dbwr/.

> Do you really have a need for adjusting the line spacing on all your labels, or is this actually about a very specific case where an unfortunate font implementation results in overlaps, so it's really a broken font??

The font we use is not broken, and it is not the case that there are overlaps between the lines.

The motivation for adding this feature is that we have multi-line labels that would benefit from having their line spacing adjusted.

kasemir:
>  As I see it, Phoebus is primarily a tool for creating and displaying GUIs for EPICS and EPICS related services. 

Great to hear that you like CS-Studio/Phoebus. I agree that it's a good tool, but it's not the only one. See https://controlssoftware.sns.ornl.gov/training/2022_USPAS/Presentations/03%20User%20Interfaces.pdf for a simplified history. Every 10 years there's been a new display manager, and in recent years this has accelerated (BOY -> Display Builder on Eclipse -> Display Builder on Phoebus, also web runtime).

There's the fundamentally different approach of making every display a "program", using tcl/tk, perl/tk, Borland Delphy, Matlab, Visual Basic, C#, Python/Qt, ... In that approach, you have full control over everything. It's easy to compute an FFT inside the display, add some automation. But each display is its own program. Maybe split the widget layout out into a QtDesigner file, but it's not like a *.bob file.

With display managers there's ideally no programming, just the display file. Anybody can just "Open in editor" to look at a display design, tweak as desired, save/run. I'd say most EPICS sites prefer the display manager approach over custom programs. Unfortunately, display technologies change (Xt, Motif, GTK, Qt, SWT, JavaFX, ...). Right now we like JavaFX, it looked promising when we started, was part of the (Oracle) JDK. Now it's a separate project and who knows how it'll fare. Is there a tiobe index, https://www.tiobe.com/tiobe-index/, for computer graphics libraries? Seems to me the rest of the world is concentrating on HTML5 and couldn't care less about JFX. So in 5..10 years we might have to look for something else again.

With display files, we have a decent track record of translating from EDM and MEDM to *.bob files, but that gets harder with each new feature that we add, and if you're already unwilling to support line spacing in the web runtime, it looks like you don't consider this a key feature anyway, so maybe just skip it?

abrahamwolk:
>Great to hear that you like CS-Studio/Phoebus. I agree that it's a good tool, but it's not the only one.

I did not assert any value judgment in my previous comment.

> See https://controlssoftware.sns.ornl.gov/training/2022_USPAS/Presentations/03%20User%20Interfaces.pdf for a simplified history. Every 10 years there's been a new display manager, and in recent years this has accelerated (BOY -> Display Builder on Eclipse -> Display Builder on Phoebus, also web runtime).

> There's the fundamentally different approach of making every display a "program", using tcl/tk, perl/tk, Borland Delphy, Matlab, Visual Basic, C#, Python/Qt, ... In that approach, you have full control over everything. It's easy to compute an FFT inside the display, add some automation. But each display is its own program. Maybe split the widget layout out into a QtDesigner file, but it's not like a *.bob file.

> With display managers there's ideally no programming, just the display file. Anybody can just "Open in editor" to look at a display design, tweak as desired, save/run. I'd say most EPICS sites prefer the display manager approach over custom programs. Unfortunately, display technologies change (Xt, Motif, GTK, Qt, SWT, JavaFX, ...). Right now we like JavaFX, it looked promising when we started, was part of the (Oracle) JDK. Now it's a separate project and who knows how it'll fare. Is there a tiobe index, https://www.tiobe.com/tiobe-index/, for computer graphics libraries? Seems to me the rest of the world is concentrating on HTML5 and couldn't care less about JFX. So in 5..10 years we might have to look for something else again.

> With display files, we have a decent track record of translating from EDM and MEDM to *.bob files, but that gets harder with each new feature that we add,

I understand your argument as follows:
 1. The probability that JavaFX is not supported in 5-10 years is not low.
 2. Due to point 1, the probability that a successor to Phoebus will have to be built in 5-10 years is not low.
 3. If we add a feature to *Display Runtime*, then it will have to be added to the successor of Phoebus, which will result in greater development cost for the successor of Phoebus.

For the purpose of this discussion, let's assume points 1 and 2. Then, doesn't *every* feature (i.e., not just the ones associated with Display Runtime) added to Phoebus come with this cost? E.g., if we add a feature to the Data Browser, most likely that feature will have to be re-implemented if a compatible successor to Phoebus is implemented.

In other words, wouldn't the following stronger version of conclusion 3 be motivated?

 3. If we add a feature to *Phoebus*, then it will have to be added to the successor of Phoebus, which will result in greater development cost for the successor of Phoebus.

> and if you're already unwilling to support line spacing in the web runtime, it looks like you don't consider this a key feature anyway, so maybe just skip it?

I am working on Phoebus, not on Display Builder Web Runtime. Display Builder Web Runtime is a separate project.

shroffk:
I think there are 2 valid points here:

1. we can't stop all development on Phoebus out of concern for future support in a new visualization solution
2. new features we add increase the scope of what we must support... not just in a future tools but what we need to support **_now_**

I think the hard bit is how to we balance both issues...

For services, most core modules, a lot of the applications model and logic code the above issue is not that grave... the fact that a lot of the code (at least the interfaces) comes from the old eclipse CSS informs us that the longevity of this part of Phoebus is quite secure.

The issue of how to balance these two considerations is particularly acute with DisplayBuilder... where past experiences have lead to the adoption of a more conservative approach.
As an example, autosizing embedded screens in linking containers is a perfectly reasonable feature for DB.... however, in CSS when used with multiple levels of linking containers, this very reasonable feature manifested itself as a mess of blinking screens and slow opi load time... finally at NSLS2 we updated our screens to not use it. In the end it was not that bad, users just had to pay attention of the sizes instead of relying on the autoresizing and almost all screens were updated to work without this feature with very little effort.
While I understand that line spacing will mostly not have such drastic consequences... it might be useful to understand the scope of the requirement ( 1 screen, 10 screens, 100 screens, all...) is there no other way to achieve the outcome.
Additionally, while DBWR is a separate project, planning for adding the feature in DBWR tends to be a good exercise in evaluating how we could address future migration issues associated with features.


So... coming back to the PR specifically.
@abrahamwolk what is the impact of this missing feature at ESS?

abrahamwolk:
> I think there are 2 valid points here:
> 
>     1. we can't stop all development on Phoebus out of concern for future support in a new visualization solution
> 
>     2. new features we add increase the scope of what we must support... not just in a future tools but what we need to support **_now_**

I think point 2 is not clear:
 - Who does "we" refer to?
 - What does "what we must support" refer to? Does it include only Phoebus, or does it also include DBWR?

> I think the hard bit is how to we balance both issues...
> 
> For services, most core modules, a lot of the applications model and logic code the above issue is not that grave... the fact that a lot of the code (at least the interfaces) comes from the old eclipse CSS informs us that the longevity of this part of Phoebus is quite secure.
> 
> The issue of how to balance these two considerations is particularly acute with DisplayBuilder... where past experiences have lead to the adoption of a more conservative approach. As an example, autosizing embedded screens in linking containers is a perfectly reasonable feature for DB.... however, in CSS when used with multiple levels of linking containers, this very reasonable feature manifested itself as a mess of blinking screens and slow opi load time... finally at NSLS2 we updated our screens to not use it. In the end it was not that bad, users just had to pay attention of the sizes instead of relying on the autoresizing and almost all screens were updated to work without this feature with very little effort. While I understand that line spacing will mostly not have such drastic consequences... it might be useful to understand the scope of the requirement ( 1 screen, 10 screens, 100 screens, all...) is there no other way to achieve the outcome. Additionally, while DBWR is a separate project, planning for adding the feature in DBWR tends to be a good exercise in evaluating how we could address future migration issues associated with features.
> 
> So... coming back to the PR specifically. @abrahamwolk what is the impact of this missing feature at ESS?

The impact is small, as the labels can be separated into multiple labels that can be placed manually.

Since it is not easy to reach a consensus here that it is a good feature to add to Phoebus, and because the impact for us is small, I close this PR.

kasemir:
> What does "we" .. and  "what we must support" .. refer to .. 

Good question. "We" is a collection of many "I", so one view of the elephant:

I've worked on at least 5 different EPICS display tools.  SNS currently operates with a mixture of EDM, Display Builder (Phoebus), Web-EDM and DBWR. EDM for the bulk of the accelerator, display builder for instruments/experiments and everything "new", Web-EDM and DBWR to view things remotely. It's OK for the web view to be read-only. Sometimes simplified/smaller displays are specifically created to better support the web view, but NOT having a way to show things on a phone has become a deal breaker.

With a new site, you have the luxury of picking one new great display tool. The difficulties arise when you then need to maintain that for 20+ years. You either "freeze" your OS so the tool still runs while the world pulls the rug from under your feet, or you periodically transition to new tools. For the latter, having two types of text inputs, three types of sliders, alarm indication via border color, background color, font color, blinking, ... complicates matters. In the relatively near history we had sites where Eclipse/BOY was significantly customized to local desires, and now they're stuck with that.

Keeping compatibility for the display tool in mind does limit you a little in the addition of new features. The display builder has the same basic widgets as EDM, MEDM, .. There are still huge advantages because you can use the context menu to send a PV to probe, PV tree, PV table, Data Browser with access to history.  Save/Restore is available within the same product. You can open displays from alarms and submit screenshots to a logbook. The image widget understands PVA images, there's support for PVA in general, IPv6 and a first glimpse at encrypted communication. EDM has the one advantage that it's light-weight and runs many displays without CPU load issues, but otherwise it's a Linux-only standalone tool, while the display builder can seamlessly open most of the EDM displays and offer a better ecosystem.

My idea for high priorities and what we must support:

 * Fix bugs, both functional and related to performance
 * Keep phoebus on a recent version of JavaFX and JDK. VirtualThreads might help with CPU load, value objects might help with memory. Each VType and data browser sample is now a class/object when it could just be a 'struct' type value.
 * Keep phoebus running on current Windows, Mac (for now Intel and aarch64), Linux
 * Stay up to date with PV Access (primarily as a client)
 * Significant display builder widgets and their features need to be mirrored in DBWR
 * Alarm system, save/restore, ... need to stay with reasonably current versions of kafka, elastic, ...

So that's in large part about stability and maintenance.
There's room for site-specific additions (we have some machine protection bypass tool that shows what bypasses are requested, possible, active; important for operation here and nicely integrated to CSS, but it won't be of use outside the SNS infrastructure).
There's also room for new tools that are of general interest (save/restore is a large component that was added in recent years).

Non-priorities to me are additions to the very basic Label and textupdate widgets. These should be as lightweight as possible. Instead of adding features we could ideally remove some, because compared to EDM each label eats up room in the JavaFX scene graph, preventing us from running as many displays as EDM.

abrahamwolk:
@kasemir: Thanks for the explanation, I think I understand your position better now.

Issue #2879, opened 2023-11-20
User-friendly exception trace in ExceptionDetailsErrorDialog


kasemir:
The stack trace in the ExceptionDetailsErrorDialog
https://github.com/ControlSystemStudio/phoebus/blob/master/core/ui/src/main/java/org/phoebus/ui/dialog/ExceptionDetailsErrorDialog.java
 can be extremely useful to developers and administrators, but it tends to confuse end users.

A stack trace often has this form:
```
java.lang.Exception: Cannot do this or that
 at org.phoebus...
 at ...
 at ...
 at java.base/...
Caused by: ..NetException: Cannot connect to host 196.165.43.5
 at java.base/...
```

To the knowledgable this provides what-failed-where and consequences.
To end users, it says "Java error" and the "Caused By" is totally lost.

Ideas:
Somehow highlight the message of the exception and the caused-by message, so it looks more like

java.lang.Exception: **Cannot do this or that**
 at org.phoebus...
 at ...
 at ...
 at java.base/...
**Caused by: ..NetException: Cannot connect to host 196.165.43.5**

Always expand the details, since they're by default collapsed?
Or show just the messages for the exception and all caused-by when collapsed, somewhat like this:

java.lang.Exception: **Cannot do this or that**
**Caused by: ..NetException: Cannot connect to host 196.165.43.5**

and opening the "detail" will show the complete stack trace

shroffk:
I agree with the need to highlight the core issues

The ExceptionDetailsErrorDialog has a title and a message field.
Maybe we can use one of those to capture the highlighted information? Basically instead of highlighting the information and expecting users to pay attention to the highlight we simply extract that info and display it separately

![image](https://github.com/ControlSystemStudio/phoebus/assets/2111304/44980cb0-0756-4a95-b690-258ddf9de45b)

```
    /** Open dialog that shows detail of error
     *
     *  <p>May be called from non-UI thread
     *
     *  @param title Title
     *  @param exception Exception
     */
    public static void openError(final String title, final Exception exception)
    {
        StringBuilder message = new StringBuilder();
        message.append(exception.getMessage()).append(System.lineSeparator());
        Throwable cause = exception.getCause();
        while(cause != null) {
            message.append(cause.getMessage()).append(System.lineSeparator());
            cause = cause.getCause();
        }
        openError(null, title, message.toString(), exception);
    }
```

kasemir:
YES, I like that.
[...]

PR #2881, opened 2023-11-20
#2879 error dialog with the message text extracted from the stacktrace

shroffk:
#2879

### DIFF
--- a/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
+++ b/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTableController.java
@@ -559,7 +559,7 @@ public class AlarmLogTableController {
                         }
                         alarmInfo.show();
                     }),
-                    (url, ex) -> ExceptionDetailsErrorDialog.openError("Alarm Log Info Error", ex.getMessage(), ex)
+                    (url, ex) -> ExceptionDetailsErrorDialog.openError("Alarm Log Info Error", ex)
             );
 
         });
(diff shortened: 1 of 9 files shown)


kasemir (approved):

kasemir:
The reformatting is good.
One more thing: For some exceptions, `getMessage()` returns `null`.
That's part of why we should always `Logger.getLogger(..).log(Level.XXX, "Trouble at mill!", the_exception)` and not `...the_exception.getMessage()`. The exception has the full stack trace, line number etc., while `getMessage()`  could just be `null`.
For the dialog, the "Detail" section will always have the full information, which is great.
For exceptions that hold a message, this new end-user summary of message.. cause[1] ... is even better.
Not sure if there's a good way to improve on "null Cause[1] null Cause[2] null".
If getMessage returns null, print the exception class name? "InvalidArgumentException Cause[1] ..."??

shroffk:
When the message is null then the class name is printed (broadly useful... some exception class names are quite self explanatory)

image


Also,

I think we should have a flag `append_ex_stacktrace_messages`  to allow user to add their own **first line** of message followed by the above abbreviated stack trace messages

Unable to retrieve log entries <-- user message

Olog search failed                   <-- first ex message
[1] Connection timeout           <-- summary of stack trace messages
[2] No host with name....

kasemir:
Excellent! Users will now look forward to seeing errors because they're so exquisitely formatted!!!
We should add some online help: If you don't see any errors and miss the dialog, here's how you can cause some trouble...

kasemir:
This is really good! No more need to ask users to open the details and scroll to the bottom because that's where the actual cause can be found..

shroffk:
@kasemir I ready to merge this... any objections :)

kasemir:
I'm happy about it.

PR #2882, opened 2023-11-21
Add ability to restore PVs from server

jembishop:
Hi there, 

This is a PR which is part of the larger project to enable snapshot comparision and restoration via the archiver appliance. 
This PR adds the ability of the save-restore server to restore the PV values, so that in the future the client does not need this logic. I use the `CompletableFuture` api to do this cleanly, and concurrently write the non null PVS in the provided snapshot items with a timeout of 5 seconds. To enable this I had to change the signature of the `PVAClient.write` method to be `CompletableFuture` rather than `Future`, it looks like that should have been there in the first place anyway.

I also wanted to use the Java record syntax for defining `RestoreResult` so I upgraded the java version to 17 in the `pom.xml`. Hopefully this is OK. 

The API part can either accept a list of snapshot items directly, or grab the snapshot items from a specified node.

I have basically copy pasted the code to convert a `VType` to an `Object` from the save-restore client code. This could be possibly made more reusable by factoring this code out. However it seems to me that the `getValue` method should probably be an interface method of `VType`, which seems more idiomatic, let me know if you want this change.

I haven't added tests yet, I will add them after hearing your feedback.

Thanks, would greatly appreciate your review.


georgweiss:
Thanks for the effort. I have two comments at the time being:

1. Can you please make the change in ```PVAChannel.write``` (not ```PVAClient```) a separate pull request? This change is not linked to save&restore.
2. Using Java17 is actually not OK at this time. We foresee to step up soon, but cannot really say when. We can of course develop and test this feature under Java17 and wait to merge.

georgweiss:
@jembishop, sorry for not being very clear about it, but could you make the PR for the change in ```PVAChannel``` from a different branch? This I think would eliminate potential confusion.

jembishop:
Hi Georg, 

Thanks for your feedback. I have raised https://github.com/ControlSystemStudio/phoebus/pull/2883/files to address the issue of moving the `PVAClient` methods to use `CompletableFuture`. 

I have also removed the Java 17 dependency and refactored the code to not use the record syntax. It would be nice to upgrade at some point in the future, because this syntax reduces alot of boilerplate.

georgweiss:
Agree, I'm looking forward to Java17 too. We need it anyway to be able to go from JavaFX 19 -> 20.

shroffk:
about the jdk dependency...
the recent save and restore branch also means we cannot use jdk11.0.2... maybe we should take a poll of who still needs jdk11 and make a move to jdk17 depending on the results

georgweiss:
@jembishop, I will turn my attention to this next week.

jembishop:
@georgweiss Thanks, I am currently working on using the archiver appliance to generate the snapshot items on the save-restore server, which will be the next prerequisite step.

georgweiss:
Sorry for delayed feed-back...
There are a few minor things that we need to update, but let's not dwell on those details for now.

More importantly: there is a need for the service to access the EPICS address list, in which EPICS gateways and IOC addresses can be listed. For the Phoebus application there is a setting ```org.phoebus.pv.pva/epics_pva_addr_list``` (or ```org.phoebus.pv.ca/addr_list``` for Channel Access) the PVA module needs. Please check how this is done on the Phoebus client side. Same (I think) goes for setting ```org.phoebus.pv/default```, which probably defaults to ```ca```.

I tried the endpoint ```/restore/node``` using ```curl```, and that works fine.

jembishop:
Thanks for your feedback Georg, 

I am currently on holiday til Jan 7th, so I will address these comments properly after I’m back.

 In terms of your review comments, I was wondering if you could elaborate on the first point, I thought that the PVASettings code should automatically take care of things like setting of the address list?

The next step on this PR will be to add the unit tests as you mention, and also update the related API documentation, now that it looks like you are happy with the overall implementation.

georgweiss:
@jembishop, I have most probably caused some confusion...

In ```SnapshotRestorer``` you have chosen to connect to PVs through ```PVAClient.getChannel```, while the original code in the save&restore client uses ```PVPool.getPV()```. Your approach will/may indeed use the system environment in terms of EPICS addresses, so Phoebus preferences are not needed.

However, I think it makes sense to prefer ```PVPool.getPV()``` for multiple reasons, e.g. support for channel access and PV pooling. This approach will then need the preferences settings I mentioned.

jembishop:
Hi, I have re-implemented the PV restoring using the `PVPool` class [here](https://github.com/ControlSystemStudio/phoebus/blob/d751e16bb3fa900e57b81ac49e42084bc27658db/services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/epics/SnapshotRestorer.java) as you suggested.

This reuses some code from the launcher to set the preferences according to the `settings.ini` file, which should be included for configuration now.

I am working on the unit tests right now, after that I think this should be good to merge once you have reviewed it.

georgweiss:
@jembishop, I will evaluate this, but in regards to a merge... Since we have just merged access control in save&restore and I'd like to give this some runtime with my users before we take the next step and move restore to the service. Also, the move will also require some changes on the client side. 

So, all in all, a merge I think will need to wait for a little bit.

jembishop:
I had another question, I have basically copied pasted this [function](https://github.com/ControlSystemStudio/phoebus/blob/17b0a7e7752d088f0a2bd9035908374eb3fe2f88/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/Utilities.java#L291), for converting a `VType` to a Java type. However this utility function seems very useful in general. Should it be moved to the core-pva library do you think?

georgweiss:
@jembishop, agree, but maybe move the entire class to ```core-vtype```?

jembishop:
Hi 

I have added in 2 pretty basic tests. Is this sufficient, or would you like the endpoints tested as well?

Also should I update the save-restore api docs, or should we wait till the archiver portion and the UI portion is done?

I have pre-emptively merged in https://github.com/ControlSystemStudio/phoebus/pull/2912.

georgweiss:
Agree, the archiver stuff may also affect APIs, so let's wait with the API doc update.

shroffk:
For the archiver portion I had a few ideas on how best to address them

I was thinking of creating a new datasource in core-pv which would retrieve the data from the archiver
like
sim://noise(-10,10,1)

aa://pv_name(time_instant,...)
aa://pv_name(start_time, end_time,..)
etc...

I am not sure what the exact retrieval operations should be supported here...hence I think we can discuss this at the next monthly meeting and finalize an API I can start implementing.

jembishop:
@georgweiss Do you think the two tests I have for the SnapshotRestorer class are sufficient, or should I add some endpoint tests?

georgweiss:
Let me check...

georgweiss:
@jembishop, I'm sorry to say, but still getting build errors. Test classes cannot use ```org.junit.Assert``` as we've moved to Junit5. You instead need ```org.junit.jupiter.api.Assertions```.

As for test cases: I'd like a test case verifying that the new endpoints are available. See for instance ```NodeControllerTest``` on how to set up such a test using Spring's ```MockMvc```.

kasemir:
Not sure how this PR combines with #2912. In here, the VEnum vs  VEnumArray is still inconsistent, and VDisconnectedData misses the corresponding alarm severity. Would merging this revert parts of 2912?

jembishop:
In theory the merge to master should get those updated changes, but I have merged to master just now just in case.

georgweiss:
@jembishop, very sorry, we're struggling with deployment of save&restore with LDAP integration. Not as straight forward as one might wish. Also detected a few bugs related to authorization, so those I think should be fixed first.

jembishop:
Do you think this is ready to merge now, if your LDAP issues have been solved?

shroffk:
Glad to hear I am not the only one :)
How are your users responding to moving from the "no authentication/no authorization" model to the "need to log in to restore"

georgweiss:
@jembishop, can I please ask for one more week of patience? Hoping to get feedback on the current code base and all recent changes during this week.

jembishop:
@shroffk we haven’t actually got round to properly testing the LDAP for save restore at the moment, we had some other priorities I’m afraid. We will let you know if there are any issues!

jembishop:
@georgweiss No problem, take as much time as you need! Just checking in 🙂.

jembishop:
Could I kindly check up on this? Did you manage to solve your LDAP issues?

georgweiss:
A final question: what changes are made on the client side?

jembishop:
Thanks for getting back to me Georg. There are no changes made on the client side in this PR, this is just a server side change. However as I mentioned earlier, the documentation for the save-restore api is not updated. Do you want me to update the documentation?

georgweiss:
Yes, why not. But if it helps you we can merge and update documentation separately.

jembishop:
I added the documentation in the most recent commit.

georgweiss (approved):

georgweiss:
There is an issue in ```SnapshotRestorer```:
```
 pv = PVPool.getPV(pvName);
 future = pv.asyncWrite(rawValue);
``` 
The write operation may/will fail as the PV may not be connected yet. I think you will need to call ```pv.onValueEvent().subscribe()``` and wait for the subscription handler to get a callback before write (or read) can be called. That's how I think it works.

jembishop:
Thanks Georg, I will make that change and get back to you in a separate PR.

jembishop:
Hi Georg.
I was thinking of implementing this like:
```java
pv.onValueEvent().timeout(5, TimeUnit.SECONDS).subscribe(value ->
    {
        if (!PV.isDisconnected(value))
            connect.countDown();
    });
connect.await();
future = pv.asyncWrite(rawValue);

```
Like in the example.
Its a bit awkward as you really want to convert this `io.reactivex.rxjava3.disposables.Disposable`  thing into a `CompletableFuture` to avoid blocking but it doesn't look like it supports that interoperability.

georgweiss:
What if PV never connects? Is the subscription handler called anyway? If not, ```await()``` will block.

jembishop:
Yes it will block until it hits the timeout. In which case an exception should be raised. So in the event you have multiple PVs which are not connected you will lose some concurrency.

georgweiss:
In the current implementation on the client side, a thread pool is used to avoid waiting for the timeout. Of course, if some PV does not connect, there will in practice be some wait time as restore of all other PVs is in most cases faster than the timeout.

jembishop:
That sounds like an acceptable option too. Its a bit unfortunate that we have to use a threadpool when we have a async futures interface to avoid spawning threads, but perhaps that is the best option in this case.

georgweiss:
@jembishop, where are we on this?

jembishop:
Hi Georg

I have recently left the ISIS Muon and Neutron source for a new position, and I don't have access to an easy dev setup to work on Phoebus right now. It might be faster if you were to refactor the code in the manner discussed, it looks quite similar to the existing client side code. I probably won't be getting much time to contribute to Phoebus seriously in the future. Sorry about this, I wish all the best for the project.

PR #2883, opened 2023-11-21
PVAClient methods should implement `CompletableFuture`

jembishop:
This allows much more ergonomic handling of the futures using the `CompletableFuture` java api. It implements `Future` already so its fully backwards compatible.

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/PVAChannel.java
+++ b/core/pva/src/main/java/org/epics/pva/client/PVAChannel.java
@@ -11,7 +11,6 @@ import static org.epics.pva.PVASettings.logger;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -116,7 +115,7 @@ public class PVAChannel extends SearchRequest.Channel implements AutoCloseable
     }
 
     /** Wait for channel to connect
-     *  @return {@link Future} to await connection.
+     *  @return {@link CompletableFuture} to await connection.
      *          <code>true</code> on success,
      *          {@link TimeoutException} on timeout.
      */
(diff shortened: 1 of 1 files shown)


kasemir:
Well, this is unfortunate. As a generic low-level PVAccess library, we should use the basic interfaces. Take `PVAStructure` which presents its elements as a `List`. We don't expose if it's internally kept as a linked list, array list, wrapped as `Collections.unmodifiableList` and so on.
For the PVA channel where we want to handle operations that might take some time, `Future` is the basic interface. `CompletableFuture` is an internal implementation detail. We certainly don't want a user to `complete(..)` the future on us. We tell the user when we're done, not the other way around.
But `CompletableFuture` has all these nice API additions (`then...`, `when...`) that make using futures fun. Methinks those should have been on the `Future` or on a separate `FutureTools` helper, but at the time nobody asked me.

So unless we fix the standard Java library, exposing the `CompletableFuture` certainly makes our library easier to use. But we should  add some comment:
 `CompletableFuture` is really an implementation detail. It's exposed to make the API more convenient, but note that user code must not call `complete(value)` nor `completeExceptionally(error)` on any of the returned  `CompletableFuture` references.

jembishop:
Interesting, I'm not really a Java person so I can't comment on the design aspect. Here is the code which I would ideally like this for https://github.com/ControlSystemStudio/phoebus/blob/1a1564a5c1400c2c9753c496bbb8374bcb8a40e7/services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/epics/SnapshotRestorer.java

Probably should be some canonical code example to inform users now to read a PV. I had a look at https://github.com/ControlSystemStudio/phoebus/blob/master/core/pva/src/main/java/org/epics/pva/client/PVAClientMain.java#L112 but it looks like it doesn't read the values concurrently, which is not what I want.

kasemir:
Your example shows why the API added to the CompletableFuture is nice.
Still, when we hand out CompletableFuture references, user code can call `complete(value)` or `completeExceptionally(error)` and then wonder why it doesn't work as expected.

Still, let's go ahead with this, but please add something like this to the overall PVAChannel javadoc:
```
 *
 * Note that several methods return a CompletableFuture.
 * This has been done because at this time the Futures used internally are indeed CompletableFutures
 * and this type offers an extensive API for composition and chaining of futures.
 * But note that user code must never call 'complete(..)' nor 'completeExceptionally()'
 * on the provided CompletableFutures.
```

kasemir (approved):

Issue #2884, opened 2023-11-21
updated architecture diagram


shroffk:
Creating a common folder which hosts a google diagram representing the architecture of Phoebus and other Phoebus services
https://drive.google.com/drive/folders/1EJ-m4sLv6DkQla_3A46eLEGFj-mjwKlH?usp=sharing

I am hoping that this common editable resource should make is possible for anyone to fix this and easily use in presentations etc...

Updating the read the doc's with the new architecture diagram ( the big difference is the inclusion of services )

![image](https://github.com/ControlSystemStudio/phoebus/assets/2111304/a2def15a-c5d0-4d66-b506-f8a3822c7d3b)

PR #2885, opened 2023-11-21
#2884 adding an updated phoebus architecture diagram

shroffk:


kasemir:
In the image, Alarm Display is shown twice, under applications (OK) but also under services.

"Java 11*" stands out because of the version number and star.
In the image, I'd simply say "Java", just like it just says "JavaFX".

In the text, we can then explain:
Need at least Java 8 because of all the new things that it introduced (lambdas, streams, Instant,..).
Thought we need at least Java 9 because of modules, but haven't actually used modules, yet.
Generally planning to require the current "long term release" version to support recent releases of our open-source dependencies.
At the time of this writing that is Java 11.

kasemir:
Looking good

PR #2894, opened 2023-12-14
Added file path recommendation in help doc

georgweiss:


kasemir (approved):
Yes.
Also note https://github.com/ControlSystemStudio/phoebus/wiki/Display-Builder-Compatibility#paths :
We try to turn the result from the file selector into a relative path, but the file selector tends to resolve symlinks. So you may have setup this:


/home/displays/main.opi
/home/displays/motors  -> symlink to /home/epics/motors/V1.2.3/displays
/home/displays/pumps  -> symlink to /home/epics/pumps/V4.5.6/displays

In a display /home/displays/pumps/some_pump.bob you want to link to simply"../motors/some_motor.bob", a relative link that will work no matter how you adjust the symlink for each version update of the motor displays.

 Using the file selector, that might put "../motors/V1.2.3/displays/some_motor.bob" into the link, which is then locked to that version of the motor displays, because the file selector returned the path with resolved symlinks. Such cases require manually entering the link...

PR #2895, opened 2023-12-15
prepare for epics core java 7.0.10 release

shroffk:


shroffk:
still waiting on the maven central repo to be updated...I see that jca and the parent is updated but don't see the other bit released a few hours ago

georgweiss (approved):

PR #2901, opened 2024-01-09
implement completable future on PV class

jembishop:
Hello

Similarly to my previous PR, this implements the `CompletableFuture` rather than the `Future` class on the `PV` method return values. 

This allows use of the useful `CompletableFuture` API methods. 

One additional question, the code for `asyncWrite` looks a bit strange, it doesn't seem to wait for the result of the write, and just returns null. Its probably a case of me not knowing java very well, but I would be interested in the reasoning there. 

Thanks

### DIFF
--- a/core/pv/src/main/java/org/phoebus/pv/PV.java
+++ b/core/pv/src/main/java/org/phoebus/pv/PV.java
@@ -10,7 +10,6 @@ package org.phoebus.pv;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -236,17 +235,17 @@ public class PV
 
     /** Issue a read request
      *
-     *  <p>{@link Future} allows waiting for
+     *  <p>{@link CompletableFuture} allows waiting for
      *  and obtaining the result, or its <code>get()</code>
      *  calls will provide an error.
      *
      *  <p>As a side effect, registered listeners will
      *  also receive the value obtained by this call.
      *
-     *  @return {@link Future} for obtaining the result or Exception
+     *  @return {@link CompletableFuture} for obtaining the result or Exception
      *  @exception Exception on error
      */
-    public Future<VType> asyncRead() throws Exception
+    public CompletableFuture<VType> asyncRead() throws Exception
     {
         // Default: Return last known value
         return CompletableFuture.completedFuture(last_value);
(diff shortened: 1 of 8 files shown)


kasemir:
As for asyncWrite, the default implementation in PV indeed returns "Done" right away.
So PVs that don't really support that idea (loc://, sim://, ..) use that.

Pls that do support it replaces asyncWrite with something that only completes once the server reports completion, see https://github.com/ControlSystemStudio/phoebus/blob/master/core/pv/src/main/java/org/phoebus/pv/ca/JCA_PV.java or https://github.com/ControlSystemStudio/phoebus/blob/master/core/pv/src/main/java/org/phoebus/pv/pva/PVA_PV.java

kasemir:
.. and that's where you now get a build error, because JCA and PVA PVs which override the Async methods now also need to be declared with CompletableFuture, which I think will just be a text replacement because they do use that already internally, methinks.

jembishop:
Ah, good point thanks for the explanation! I'll make that change.

jembishop:
Thanks, for your feedback. I implemented the change on all inherited classes. I had to change the class hierarchy so now `PVRequester` inherits from `CompletableFuture` to avoid compilation errors, hope that is ok.

kasemir:
I'm not worried about changes to `opva`. That's using the original PVA library, the one that's no longer maintained and I don't think anybody uses it with phoebus, but we keep it around as long as it doesn't cause problems.
The changes to `ca` and `pva` look benign, just exposing the `CompletableFuture` that we already have at hand anyway.

As mentioned in the previous PR, this is all a bit unfortunate. `Future` is the basic class, like `List`. Ideally, we'd use the basic interfaces, not exposing and then relying on `CompletableFuture` or `ArrayList` or `LinkedList` or `CopyOnWriteArrayList` that we happen to use in the implementation. But  `CompletableFuture` has all those nice helpers which the `Future` lacks. It's as if `List` had been designed without key features.
I guess the alternative to these updates would be casting the `Future` to `CompletableFuture` in case the added features are desired, and we happen to know it is really a  `CompletableFuture` anyway. But that seems worse than locking us into  `CompletableFuture`.

shroffk:
> But CompletableFuture has all those nice helpers which the Future lacks. 

Ya, not sure how what should be a strategy with all the other places where the interfaces return `Future` 

> I guess the alternative to these updates would be casting the Future to CompletableFuture in case the added features are desired, and we happen to know it is really a CompletableFuture anyway. But that seems worse than locking us into CompletableFuture.

I agree, this would be terrible... using Future with the missing feature or just exposing CompletableFuture both sound preferable

kasemir:
Maybe I'm missing some wrapper that's already there?

Since Future lacks "then..(..)", is there a something like "new BasicFutureWrapper(future).then...(..)"?

PR #2902, opened 2024-01-12
Save&restore support for NTTable

georgweiss:


core/pva/src/main/java/org/epics/pva/client/PutRequest.java
```diff
@@ -226,8 +236,11 @@ else if (subcmd == PVAHeader.CMD_SUB_DESTROY)
             // Indicate completion now that server confirmed PUT
             complete(null);
         }
+        /*
```
kasemir, line 239: Do we indeed need to disable the exception?

core/pva/src/main/java/org/epics/pva/data/PVAData.java
```diff
@@ -22,7 +22,7 @@
 public abstract class PVAData
 {
     /** Name */
-    protected final String name;
+    protected String name;
```
kasemir, line 25: Does the name need to change, no longer final?

georgweiss:
Let me check...

georgweiss:
@kasemir , changes reverted. Thanks for pointing out.

kasemir (approved):

PR #2905, opened 2024-01-16
Save & restore: support search on PV name(s)

georgweiss:
In save&restore search view user may specify comma separated list of PV names to use in query.
Input field supports PV name completion backed by channel finder.
Search result will contain only configuration nodes.
Search will consider both setpoint PV and readback PV.
PV names must match exactly, wildcard not supported.
Documentation updated.


shroffk (approved):

shroffk:
So, I will close my implementation... which differed in 2 ways

1. it was searching both configs and snapshots... but I guess if you locate the configuration locating the snapshot is easy. In someway I guess your implementation is better since you have narrower search results.

2. I was using match instead of term queries. 
```
{
  "query": {
    "match": {
      "pvList": {
        "query": "BR:[RRR]_string_{00005}"
      }
    }
  }
}
```
I do not observe any functional difference between the two, so we can stick to the term queries which fit better with the strict PV name mapping.
In the future if we want to support a bit more flexible searching then we might want to consider match ( however the usefulness is limited by the flattened mapping )

PR #2912, opened 2024-01-22
Move Utilities.toRawValue to core-vtype

jembishop:
Hi there, in the process of doing another pull request I found that it would be a good idea to refactor the `Utilities.toRawValue` method in the save-and-restore ui code into the shared `vtype` library. This is because converting a `VType` to a  native java object was useful in my usecase and sounds like a generally useful thing to want to do. 

I also thought it would be best to take the associated `VDisconnectedData` type into the `vtype` library as well, given it shows up in the `Utilities.toRawValue` tests. 

Another minor change when making the modifications to the tests was to use wildcard import for the V types as they now pretty much import all the classes into the file.

### DIFF
--- a/core/vtype/src/main/java/org/phoebus/core/vtypes/VTypeHelper.java
+++ b/core/vtype/src/main/java/org/phoebus/core/vtypes/VTypeHelper.java
@@ -506,4 +506,49 @@ public class VTypeHelper {
         }
         return false;
     }
+
+    /**
+     * Extracts the raw value from the given data object. The raw value is either one of the primitive wrappers or some
+     * kind of a list type if the value is an {@link Array}.
+     *
+     * @param type the value to extract the raw data from
+     * @return the raw data
+     */
+    public static Object toObject(VType type) {
+        if (type == null) {
+            return null;
+        }
+        if (type instanceof VNumberArray) {
+            if (type instanceof VIntArray || type instanceof VUIntArray) {
+                return VTypeHelper.toIntegers(type);
+            } else if (type instanceof VDoubleArray) {
+                return VTypeHelper.toDoubles(type);
+            } else if (type instanceof VFloatArray) {
+                return VTypeHelper.toFloats(type);
+            } else if (type instanceof VLongArray || type instanceof VULongArray) {
+                return VTypeHelper.toLongs(type);
+            } else if (type instanceof VShortArray || type instanceof VUShortArray) {
+                return VTypeHelper.toShorts(type);
+            } else if (type instanceof VByteArray || type instanceof VUByteArray) {
+                return VTypeHelper.toBytes(type);
+            }
+        } else if (type instanceof VEnumArray) {
+            List<String> data = ((VEnumArray) type).getData();
+            return data.toArray(new String[data.size()]);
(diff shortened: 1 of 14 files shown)


core/vtype/src/main/java/org/phoebus/core/vtypes/VTypeHelper.java
```diff
@@ -506,4 +506,49 @@ public static boolean toBoolean(VType vtype) {
...
+            } else if (type instanceof VFloatArray) {
+                return VTypeHelper.toFloats(type);
+            } else if (type instanceof VLongArray || type instanceof VULongArray) {
+                return VTypeHelper.toLongs(type);
+            } else if (type instanceof VShortArray || type instanceof VUShortArray) {
+                return VTypeHelper.toShorts(type);
+            } else if (type instanceof VByteArray || type instanceof VUByteArray) {
+                return VTypeHelper.toBytes(type);
+            }
+        } else if (type instanceof VEnumArray) {
```
kasemir, line 535: It's inconsistent to convert a VEnumArray to List<String>, but a single enum is converted to its index.
With an enum having both a numeric and a string value, it's of course a good question which one to pick, and I'm not sure how you want to decide that right now, but the result should be consistent.

jembishop: @georgweiss I imagine in both cases it should be an int?

kasemir: Int has the advantage that you can plot it over time. But you never know if 0/1 represents OK/Fault or Off/On or Left/Right etc. Still, turning it into a number is likely best.

georgweiss (commented):
Agree, enum array should be converted to list of ints.

core/vtype/src/main/java/org/phoebus/core/vtypes/VTypeHelper.java
jembishop: Would be nice to have a feature in the plot which could automatically display the enum value.

jembishop:
Enum array is now list of ints.

jembishop:
@kasemir I think this is ready to merge now. The returned datatype for the enumarray is now `int []`.

kasemir:
The `VDisconnectedData` should carry a disconnected alarm state to be compatible with the current expected behavior, see https://github.com/ControlSystemStudio/phoebus/blob/3f6be6cf10b77dda5b22cf465955799d350ea98d/core/pv/src/main/java/org/phoebus/pv/PV.java#L211-L222

kasemir (approved):
Looks good, thanks!

georgweiss:
@jembishop, I'm afraid this PR was approved a bit too quickly. I suspect your development branch was not quite up-to-date, and I did not notice.
Thing is code to handle VTable PVs in save&restore evaporated in the merge. Can you please check this?

jembishop:
@georgweiss  I didn't realize that sorry. I will raise another PR to fix this.

jembishop:
@georgweiss  This is fixed here https://github.com/ControlSystemStudio/phoebus/pull/2937

PR #2923, opened 2024-01-29
CSSTUDIO-2067 Add auto-completion for the layout name when saving a layout

abrahamwolk:
This PR adds auto-completion when saving a layout.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
+++ b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
@@ -601,6 +601,11 @@ public class PhoebusApplication extends Application {
         return menuBar;
     }
 
+    private List<String> listOfLayouts = new LinkedList<>();
+    protected List<String> getListOfLayouts() {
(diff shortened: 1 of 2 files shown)


georgweiss (approved):

kasemir (approved):
Interesting how you prefer LinkedList over ArrayList. The list of layouts tends to be small; 10, maybe 20? With ArrayList, that amounts to 1 or 2 memory allocations, and iterations over the array are then as fast as they could be. With LinkedList, each time you "add" an element, a new Node is allocated and linked. Iterations run along the links, slower than array iteration.
With the overall small number of elements it doesn't matter much, but as a rule of thumb I'd always start with an ArrayList. If there are many additions and thus re-allocations, maybe set an appropriate initial array capacity. LinkedList only makes sense if there are many addictions and removals **within** the list, like the middle of the list, but here we only append to the end.

Otherwise, happy to see that the AutocompleteMenu and Proposal* API works nicely for this use case as well, something that wasn't foreseen at the time.

Issue #2922, opened 2024-01-27
v4.7.3 save and restore url setting


tynanford:
After upgrading the phoebus engine and the save and restore service to v4.7.3 there seem to be issues getting the phoebus client to connect to the service API.

Setting this variable - ```org.phoebus.applications.saveandrestore/jmasar.service.url=http://mydomain/save-restore``` in the settings.ini file doesn't impact the save restore client. And in the debugger I see it is always set to the default http://localhost:8080/save-restore.

https://github.com/ControlSystemStudio/phoebus/blob/v4.7.3/app/save-and-restore/app/src/main/resources/save_and_restore_preferences.properties

I think I was able to trace the issue back to this PR https://github.com/ControlSystemStudio/phoebus/pull/2816

georgweiss:
Due to refactoring of the code the settings is now:

```org.phoebus.applications.saveandrestore.**client**/jmasar.service.url=http://mydomain/save-restore```

Sorry for the inconvenience.

tynanford:
No worries, thanks for the help. I created a PR https://github.com/ControlSystemStudio/phoebus/pull/2927 to split out these `client` package settings so they show up correctly on the documentation pages

PR #2927, opened 2024-01-29
Move saveandrestore.client preference settings into separate file

tynanford:
https://github.com/ControlSystemStudio/phoebus/issues/2922

Tools like the web documentation script and create_settings_template.py rely on the `# Package org...` string in the .properties file to determine which package each setting belongs to, so this is an attempt to get those working again for saveandrestore.client settings

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreJerseyClient.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreJerseyClient.java
@@ -64,7 +64,7 @@ public class SaveAndRestoreJerseyClient implements org.phoebus.applications.save
         mapper.registerModule(new JavaTimeModule());
         mapper.setSerializationInclusion(Include.NON_NULL);
 
-        PreferencesReader preferencesReader = new PreferencesReader(SaveAndRestoreClient.class, "/save_and_restore_preferences.properties");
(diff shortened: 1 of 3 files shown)


shroffk (approved):

katysaintin:
Hello, since this commit, the preferences defined in save_and_restore_preferences.properties are not generated in the preferences list in the documentation https://control-system-studio.readthedocs.io/en/latest/preference_properties.html#saveandrestore .

This list is  generated by https://github.com/ControlSystemStudio/phoebus/blob/master/docs/source/conf.py , I suppose that the script does not manage 2 preferences files in the same package.

tynanford:
Hi @katysaintin , I can take a look at this

katysaintin:
> Hi @katysaintin , I can take a look at this
Hello @tynanford , thanks, I found the fix for this problem.
It is solved in this pull request https://github.com/ControlSystemStudio/phoebus/pull/3322
In fact , the package name in app/save-and-restore/util/src/main/resources/save_and_restore_util_preferences.properties was wrong. It has to be "org.phoebus.saveandrestore.util" and not "org.phoebus.applications.saveandrestore" . Because both have the same package name, the script conf.py take in account the last one it processes.
See my comment in https://github.com/ControlSystemStudio/phoebus/pull/3322#issuecomment-2711135548

Thank you again for your help :) .
Katy

tynanford:
Ah okay, thank you for fixing that!

PR #2930, opened 2024-01-30
Automatic purge of alarm log indices

georgweiss:
Triggered by issues with "overflow" of Elasticsearch alarm log indices, I've added a Spring Component that may purge indices automatically at regular intervals.

By default the Component is not even instantiated, i.e. automatic purge is disabled. To enable, the service must be started by setting ```retention_period_days``` to a value larger or equal to 100. This number also defines the purge policy, e.g. only indices where last inserted message time is ```now - retention_period_days``` are considered. The lower limit of 100 days is there to avoid accidental removal of non-obsolete/active indices.

A cron expression is used to determine when to run the purge process, default is midnight every Sunday.


kasemir (approved):

shroffk:
Ah, 
This is a good idea...manually managing the indices would have resulted in this error at some point of the other.

Instead of using a different temporal system ( days ) we could use the same system we have for creating new indices.

```
# A flag indicating if the service should use index names based on date/time generated periodically based on date_span_units
use_dated_index_names=true

# The units of the indices date span: Days (D), Weeks(W), Months(M), Years(Y).
date_span_units=M

# how many indices to retain (e.g. if you have selected date_span_units as M and set the retain_indices_count to 6, then indices 
# older than 6 months will be deleted
retain_indices_count=6
```

georgweiss:
@shroffk, the alarm logger is a Spring Boot application, but does not take full advantage of the framework. For instance, the Elasticsearch client is not exposed as a bean, which would be desirable to facilitate unit testing.

Any reason to not update the service to be more aligned with the common practices in Spring Boot?

georgweiss:
@shroffk, updated as requested.

shroffk (approved):

PR #2931, opened 2024-02-01
Simplifying preferences handling in save-and-restore

georgweiss:
Using AnnotatedPreferences API for all settings.


shroffk:
Looks good...
Just to confirm this should be completely backward compatibility ( I did not notice anything in the code that looked like any of the existing preferences would need to change )

shroffk (approved):

PR #2936, opened 2024-02-04
Fix in LDAP authorization for save&restore

georgweiss:
This PR fixes a ClassCastException not easily detectable unless running save&restore service with LDAP integration. 

Will add suitable integration tests soon.


georgweiss:
@shroffk, sorry for IntelliJ's auto formatting....

PR #2943, opened 2024-02-08
CSSTUDIO-2037 Target the current DockPane when opening applications from the context menu

abrahamwolk:
This PR changes the behavior when right-clicking on a PV in an OPI and, e.g., opening a DataBrowser when using SplitPanes: currently, the DataBrowser would be launched in the top-left-most Pane of the window. This PR changes the behavior so that the DataBrowser is launched in the same pane as the OPI itself.


kasemir (approved):
Unclear if this is the final solution.

It should work for the context menu of displays, where it will now force the 'active' stage to be the one that holds the widget.

But it removes the more generic code that was supposed to support any "parent_node", not just display builder widgets.

Still, if this gives more reliable results from widget context menus, which I'd say is the most frequent use case, then overall this would still be an improvement

shroffk:
https://github.com/ControlSystemStudio/phoebus/pull/2947

Issue #2939, opened 2024-02-06
Allow blocking or explicitly selecting PV sources


smarsching:
At the moment, the choice of data source implementations for certain protocols is hard-coded. For example, the `org.phoebus.pv.ca.JCA_PVFactory` will always be registered for PVs with the type `ca`.

I am in the process of preparing a custom product of Phoebus, and due to having experienced issues with JCA in the past, we want to use [EPICS Jackie](https://oss.aquenos.com/epics/jackie/) instead of JCA. Some time ago, we did the same for a custom product of CSS. There, this was easy because the relevant bundle could simply be excluded from the built and be replaced with a different bundle.

In Phoebus, however, the CA support that is based on JCA is part of the `core-pv` module. So, in order to replace `JCA_PVFactory` with a different implementation we would have to maintain a fork of `core-pv`, which is bothersome in the long run.

I see three ways how this could be avoided:

1. The contents of the `org.phoebus.pv.ca` package could be moved from the `core-pv` to a separate module (e.g. `core-pv-ca`).
2. A `block_pv_sources` preference could be added to the `org.phoebus.pv.PVPool` class. This preference would specify a comma-separated list of classes that should be ignored when discovering `PVFactory` implementations.
3. A `pv_sources_map` preference could be added to the `org.phoebus.pv.PVPool` class. This prerence would specify a comma-separated list of type to class mappings (e.g. `ca:com.example.MySource,pva:com.example.OtherSource`) which would override the discovery mechanism for the specified types.

Option 1 upsides:

- Clear separation of interface and implementation.
- Code which is not used does not have to be distributed with a product. Neither do their dependencies.

Option 1 downsides:

- In order to make this truly modular and to allow the distributor to choose which data sources shall be provided, this would also have to be done for all the other data sources that are part of the `core-pv` module. So, several new modules would have to be added.
- The build logic of existing products would have to be modified to explicitly depend on the desired modules because unlike the `core-pv` module, this modules would not be pulled in as transitive dependencies. However, I think that this problem could be alleviated by adding another module that depends on all these new modules and adding this new module as a dependency of all modules that currently only depend on `core-pv`. Distributors wanting to exclude certain PV sources could exclude this dependency in their build configuration and add explicit dependencies to the desired modules instead.

Option 2 upsides:

- Very simple, localized change. Only touches a few lines of code in `PVPool`.
- Blocked data-source classes will not have to be loaded at all, so dependencies of these classes do not have to be included in the product.

[...]

PR #2948, opened 2024-02-12
Split core-pv module into several modules

smarsching:
As discussed in #2939, this PR splits the `core-pv` module into several modules. The core classes for the PV support layer and the “internal” sources remain in `core-pv` and the control-system specific classes are split off into five new modules:

- core-pv-ca
- core-pv-mqtt
- core-pv-opva
- core-pv-pva
- core-pv-tango

The example Phoebus product now has explicit dependencies on all these modules, so that it offers the same range of PV sources that it offered before this change was made.

As a side effect of this change, a few small changes had to be made in other parts of the code base:

**app/databrowser**

- A dependency on `org.epics:epics-util` was added because this module is not included trasitively any longer.
- A new class `ChannelAccessStatusUtil` was added, so that the class `gov.aps.jca.dbr.Status` from the JCA library is not needed any longer. The JCA library was only used to translated numeric CA status IDs to string. Adding a dependency on the full library (which due to the Channel Access PV souce now being optional is not included transitively any longer) just in order to translate 22 possible integer values to strings seemed a bit excessive, so adding a simple utility class doing the translation seemed like the better option.

**app/display/representation-javafx** 

This module needs Apache Commons Lang 3. Before, this dependency was included transitively, now it is declared explicitly.


kasemir (approved):
Seems fine.
Importantly, while the preference files moved, their package names remained the same, so no changes to preference settings.
Once merged, I can update the ant build files.

shroffk:
you will need to add the core-pv-xx to the services. 
alarm-server
save-restore
archive-engine
scan-server

I would include ca and pva

shroffk:
At the very least they will have to be testing dependencies.

smarsching:
@shroffk As far as I can tell, the save-restore service does not seem to depend on the PV layer at all (it does not depend on `core-pv`, neither directly nor transitively), so I do not think that dependencies have to be added there.

For the other servies (alarm-server, archive-engine, scan-server), I added all the PV modules as dependencies. At the moment, the modules for these services represent the “product”, so I think that they should include all the supported PV sources.

If you agree, I would still like to split these services into a “code” and a “product” part. The existing modules would become the product and the Java code would be moved into a “-core” module for each service (e.g. `alarm-server-core`). This way, users could easily built their own products for these services, adding or removing dependencies as needed. However, I think that this should be done in a separate PR because it is not closely related to the changes made in this PR.

shroffk:
@smarsching you are right currently there is no core-pv required by save and restore. There is a PR which will soon be opened to add this feature of performing the restore operation from service... at which point we will need to add `core-pv` and the datasource implementations as pom dependencies.

In the long run we can explore the splitting of the service code into "core" and "product" packaging... we are doing this for services like Phoebus-Olog and ChannelFinder which allow sites to plug in processors and notifiers via SPI. Currenlty, this would be overkill for these services.
Let us start with including the broadly required DS (ca, pva, sim, loc)

smarsching:
@shroffk As requested I removed the support for `mqtt`, `opva`, and `tga` / `tgc` from the services, only leaving `ca`, `pva`, and the internal ones (`sim`, `eq`, `loc`, etc.).

shroffk:
@georgweiss @kasemir @tynanford @tanviash 
I am fine with merging this. If you have repackaged any of the services and need to support the additional datasources like tango/mqtt you will have to ensure they are explicitly added.

georgweiss:
No objections.

tynanford:
Fine with me. This might require an update to PVWS pom.xml file?

kasemir:
> This might require an update to PVWS pom.xml file?

Does that mean we first need a "release" to get the new core-pv-* published for the PVWS build to find them?

tynanford:
oh yeah, nevermind. It doesn't matter until we release this and update PVWS core-pv above 4.7.3. Something to remember down the road

kasemir:
Still, if you have time it would be great to check a local build of the core-pv-*SNAPSHOTs and use that with PVWS, which ideally should work, but who knows.

shroffk:
As you mention, we just need to keep this mind when do a release. We will update the dependencies of PVWS then.

tynanford:
> Still, if you have time it would be great to check a local build of the core-pv-*SNAPSHOTs and use that with PVWS, which ideally should work, but who knows.

PVWS seems to work fine! At least with the pva, ca, sim, and sys PV types.

Also even in the startup logs for PVWS you can see things change based on what's included in the pom. For instance, only including core-pv-ca in pom shows this:

`pvws  | 2024-02-21 20:09:56 INFO [pvws.ws] Supported PV types: [loc, sim, sys, eq, ca]`

While including core-pv-ca, core-pv-pva, and core-pv-tango shows:

`pvws  | 2024-02-21 20:07:14 INFO [pvws.ws] Supported PV types: [loc, tgc, sim, sys, eq, ca, tga]`

kasemir:
Thanks for checking that out!

PR #2953, opened 2024-02-21
upgrade Kafka client and stream dep to 3.6.x

shroffk:


georgweiss:
@shroffk, does this assume any specific Kafka server version?

shroffk:
A few findings

Using JDK17 and Kafka 2.x I had issues with the server failing to resolve the address when setting up the listeners.
I could not get the old kafka server working with JDK17 ( debian 12 and rhel 8-9 )

kasemir:
>  failing to resolve the address when setting up the listeners

Is this about the alarm clients, or about the link between Kafka and zookeeper?
For the latter I sometimes have to configure the listeners as "localhost" for Kafka to find zookeeper, see https://github.com/ControlSystemStudio/phoebus/blob/master/app/alarm/Readme.md:
```
...define the host name in  config/server.properties. For tests, you can use localhost:

listeners=PLAINTEXT://localhost:9092
advertised.host.name = localhost
advertised.listeners=PLAINTEXT://localhost:9092
```

shroffk:
This is about the kafka connection to zookeeper.

`WARN Session 0x0 for server localhost/<unresolved>:2181`

The preferences are set with localhost and the exact same configuration work fine if you use JDK11.

kasemir:
What if you enter IP address 127.0.0.1?

shroffk:
tried 127.0.0.1 and the static IP of the server...can't remember the exact error but that failed to.
Not sure if it worth following up since moving to Kafka 3.x resolves the issue.

I just wanted to make a note in case someone else tries to move to JDK17 without upgrading to kafka3

Also, when you upgraded from kafka 2-3 did you just wipe out the old installation and start fresh or did you try to migrate the kafaka/data and logs?

kasemir:
Haven't updated in a while but would always start over and import the XML config.

shroffk:
> Haven't updated in a while but would always start over and import the XML config.

ok good... that is exactly what I landed up doing.

kasemir (approved):

georgweiss (commented):
I think we have discussed this, but this would imply a need to upgrade Kafka as well? What about data migration?

shroffk:
You will need kafka 3.X. the instruction have for awhile now pointed to kafka 3.3

Both @kasemir and I migrate by simply exporting the alarm configuration and then importing it into the new services.

georgweiss:
@shroffk, sorry to bring this up again...
I've managed to connect to Kafka 2.x using the updated 3.x Kafka libs. It seems backwards compatible, even though I do see  a stack trace when switching between configurations (connection shutdown?):
```
2024-05-07 10:59:06 SEVERE [org.apache.kafka.clients.producer.internals.Sender] [Producer clientId=producer-2] Error in kafka producer I/O thread while aborting transaction: 
java.lang.IllegalStateException: Transactional method invoked on a non-transactional producer.
        at org.apache.kafka.clients.producer.internals.TransactionManager.ensureTransactional(TransactionManager.java:1084)
        at org.apache.kafka.clients.producer.internals.TransactionManager.handleCachedTransactionRequestResult(TransactionManager.java:1196)
        at org.apache.kafka.clients.producer.internals.TransactionManager.beginAbort(TransactionManager.java:326)
        at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:273)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

So, my question remains whether Kafka should be updated before new release of Phoebus with the 3.x libs. Comments in this issue ticket state both backwards compatibility as well as need to update.

shroffk:
> So, my question remains whether Kafka should be updated before new release of Phoebus with the 3.x libs.

Short answer, Yes. We need to be using Kafka 3 with the libraries.

Long answer, I observed different behaviour when switching to different versions of the the JDK ( 11-17 )

georgweiss:
OK, thanks.

PR #2954, opened 2024-02-21
Add EPICS Jackie PV source

smarsching:
This PR adds a PV source that uses [EPICS Jackie](https://oss.aquenos.com/epics/jackie/), an alternative Java implementation of the Channel Access protocol. The new PV source uses the `jackie://` prefix and can thus coexist with the existing `ca://` PV source.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -9,6 +9,7 @@
     <module>pva</module>
     <module>pv</module>
     <module>pv-ca</module>
+    <module>pv-jackie</module>
     <module>pv-mqtt</module>
     <module>pv-opva</module>
     <module>pv-pva</module>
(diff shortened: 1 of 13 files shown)


kasemir (approved):

kasemir:
Does EPICS jackie come with a public git repo? On https://oss.aquenos.com/epics/jackie/ I see downloads of source snapshots, but nothing like a git repo for issue tracking, pull requests etc.

smarsching:
The Git repo is at https://gitlab.aquenos.com/oss/epics/epics-jackie/.

kasemir:
Perfect, thanks!

smarsching:
I didn’t know about that one. It seems like it was started around the same time that I started working EPICS Jackie, so I guess that’s how I missed it.

Due to its GPL license, it would not be possible to integrate it into Phoebus though (GPL and EPL are incompatible as far as I know).

kasemir:
Not sure about user base. Last commit was 4 years ago

PR #2947, opened 2024-02-09
Add utility class for helping create a setFocus runnable for given Node

shroffk:
@

PR #2955, opened 2024-02-22
Issue #2947 use the existing utility class for manipulating focus

shroffk:
Not sure how I missed the existing `FocusUtil`
This PR merges the 2 utility classes into 1


kasemir:
Excellent!

I also like your compromise regarding the bracket alignment.

Your preference would be
```
public static Runnable setFocusOn(final Node node){
  Window window = ...
}
```

Mine would be
```
public static Runnable setFocusOn(final Node node)
{
  Window window = ...
}
```

By double-bracketing, we're both happy:
```
public static Runnable setFocusOn(final Node node){
      {
          Window window = ...
      }
}
```

shroffk:
while I would like to take credit I think this is the IDE trying to be in both our good books

PR #2961, opened 2024-02-27
Add JSON archive reader

smarsching:
This PR adds support for a JSON-based archive reder.

This archive reader can read archived data from any HTTP server implementing the [JSON archive access protocol](https://oss.aquenos.com/cassandra-pv-archiver/docs/3.2.6/manual/html/apb.html).

The [Cassandra PV Archiver](https://oss.aquenos.com/cassandra-pv-archiver/) and the [JSON Archive Proxy](https://oss.aquenos.com/epics/json-archive-proxy/) implement the server side of this protocol.

### DIFF
--- a/app/databrowser-json/pom.xml
+++ b/app/databrowser-json/pom.xml
@@ -1,0 +1,60 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.phoebus</groupId>
+    <artifactId>app</artifactId>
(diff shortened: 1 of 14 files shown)


kasemir (approved):
Not 100% convinced by the data source prefix. `json://` is quite generic and suggests that it can read anything JSON, including some data dumped into a file in JSON format.
Something like "Jason Archive Data Source" `jads://` would avoid that. On the other hand, `json://` clearly looks better and since that text is only used in the context of an archive data source preference setting, it might actually be the better choice.

shroffk:
We use `pbraw` (protobuf raw) for the archiver appliance... not sure I like the prefix not being related to the source and instead related to the form of the data, but that seems to be the pattern here.

Personally, I think the prefix being something that clearly informs users about where the data is coming from is better...but...not willing to die on that hill.

kasemir:
You're right, it's already quirky. "jdbc:" doesn't mean it can read any type of relational database, "pbraw:" doesn't mean it can read any google protocol buffer. So "json:" is fine.

PR #2971, opened 2024-03-11
CSSTUDIO-2305 Factor out the Phoebus Web Browser widget into a separate module

abrahamwolk:
This PR factors out the WebBrowser widget into its own module.

The motivation for this PR is to allow building Phoebus without the WebBrowser widget included. (By default, the WebBrowser widget is still included, but the module created by this PR can easily be excluded.)

I have two questions:
 1. The class `WidgetTransfer` implements drag'n'drop functionality, and the corresponding functionality for the WebBrowser widget is removed by this PR. I am not familiar with the functionality implemented in `WidgetTransfer`; is this functionality important, and can it be supported when the WebBrowser widget is implemented in its own module?
 2. I have added lines to `build.xml` files for the Ant build system, but I have been unable to verify that the changes work using Ant. The reason is that the master-branch doesn't build for me using Ant. Could someone using Ant verify that the changes work using Ant also?

### DIFF
--- a/phoebus-product/pom.xml
+++ b/phoebus-product/pom.xml
@@ -153,6 +153,11 @@
         </dependency>
         <dependency>
             <groupId>org.phoebus</groupId>
+            <artifactId>app-display-webbrowser</artifactId>
+            <version>4.7.4-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.phoebus</groupId>
             <artifactId>app-display-editor</artifactId>
             <version>4.7.4-SNAPSHOT</version>
         </dependency>


kasemir:
I agree that the web widget can be problematic. We certainly don't want to run complete web applications in there, better use a designated web browser for that. But it can be useful to show control system guidance which is easier formatted on web pages than trying to create them with labels.
This is a good use case for the `hidden_widget_types` setting of the editor, https://control-system-studio.readthedocs.io/en/latest/preference_properties.html#display-builder-editor
That way, displays created with all the standard widgets will "work" because the widgets are supported in the runtime, but each site can hide them from the editor to discourage the creation of new local displays with those widgets.

kasemir:
> I am not familiar with the functionality implemented in WidgetTransfer

The idea is that we allow drag/drop of plain text into the editor, see basic display builder overview https://controlssoftware.sns.ornl.gov/training/2022_USPAS/Presentations/05%20Display%20Builder.pdf page 19.

With plain text we obviously don't know if that's just text, or maybe a list of PVs, or??
So for the most part we open a dialog to let the user decide:

Is it PV names that you want to display in a text update, bar, ...? Then select one of those widgets.
Is it just text? Then select the Label widget.
Does it look like the path or URL for an image? Then create a picture widget.
Is it a URL that should be shown in a web browser? Then select the web widget.

Support for the web browser widget was originally added by Claudio for ESS, https://github.com/kasemir/org.csstudio.display.builder/blame/master/org.csstudio.display.builder.editor/src/org/csstudio/display/builder/editor/util/WidgetTransfer.java, so it's only fair that ESS takes it back out

shroffk:
I think we should make the `webbrowser` an option dependency of the common product.
This way if a site specific product want to support the widget they will have to explicitly add it to their product.

overall, I think this is a good re organization

kasemir:
If you want to make it optional, you either need to

a) remove support for it from the WidgetTransfer,  so you can no longer drop a URL and have that open in the web widget
b) use introspection in the WidgetTransfer to support the web widget when it's available, and not when it's not
c) come up with yet another SPI for contributing WidgetTransfer support

shroffk (commented):
Why remove this... I thought we were thinking of making this an optional dependency

shroffk:
`optional` dependencies will be included in the common product but they will not automatically be added to the site specific products. To include this feature in the ESS or NSLS2 product we will have to explicitly redefine the dependency.

I guess if we all agree we don't want this then we could remove it from the common product too... in the meanwhile the optional dependency is a compromise to keep the common product as broad as possible while still allowing site specific products to be built on top of the common product without pulling in *all* dependencies

kasemir:
> Why should the common product be as broad as possible

It should showcase what's possible. It can't include site-specific settings like CA gateways, email servers, archive data sources, or the SNS MPS Bypass tool, but it should show all the generic widgets, and maybe come preconfigured to look at archived data from "pbraw://localhost/....", so if you start the simple AA example, it'll connect to that.

So it certainly should include the web widget. Since the example displays include the web widget, I'd still say that even local products should include the web widget, because your product looks silly if it can't run the generic examples. You might also get 3rd party displays that include the web widget just to show a nice looking online-help page. To meet the understandable desire of discouraging the creation of new local displays that relay on the web display, there's the existing config setting for hiding widgets in the editor, so local pages cannot add the web widget, while everything else still works.
Bottom line, I'd scratch this PR as too-complex for very little if no benefit.

Issue #2968, opened 2024-03-05
Alarm annunciation


shroffk:
I need to make the alarm annunciator a bit more configurable.
Instead of speaking out the alarm I need to play notification sound/s...

The simplest solution would be to add some preferences and extend the Annunciator to have a `AudioClip/MediaPlayer` to play some sound file set in the prefrences.

Another way to do this would be the use of SPI so that the `Annunciator` becomes an implementation of an interface with a `speak(AnnunciatorMessage message)` method that can be implemented to satisfy different use cases. 

any thoughts?

kasemir:
The current implementation based on FreeTTS is clearly not optimal.
FreeTTS is somehow using the AWT event pipeline, whereas CS-Studio used SWT and now JFX, which can create all sorts of hangups. FreeTTS has not been updated for a long time (14 years ago??, https://en.wikipedia.org/wiki/FreeTTS). I could only get the "kevin16" voice going, all the others were worse or dysfunctional.
Having an update, maybe a pluggable mechanism, would be good.

In the Eclipse implementation, wasn't there a way to select between pluggable implementations. Vaguely remember one was sending the text as UDP packets, because one site had a tool which turned those into speech.

Or, since we are using Kafka, you could simply not run the annunciator in the GUI but write your own annunciator service that subscribes to the Kafka 'talk' topic and turns that into speech or any other type of sound.

shroffk:
> Having an update

I was looking into this, the best/newest solutions I could find were calling the google translator or and AWS service which require user accounts.

> maybe a pluggable mechanism, would be good.

Ok, maybe it is worth use SPI then. 

> Or, since we are using Kafka, you could simply not run the annunciator in the GUI but write your own annunciator service that subscribes to the Kafka 'talk' topic and turns that into speech or any other type of sound.

Sure, we can argue for the annunciator to be a stand alone service... in fact we can expand the scope to a notifier where the stand alone service has a pluggable interface for you to annunciate, play a sound, make a call, send a text, call the fire dept.

kasemir:
>  solutions I could find were calling the google translator or and AWS service

Right, you clearly want something local in the control room, not call out to a cloud service.
On the Mac, you get a "say" command line tool that sounds very good. For Linux, "festival -tts" is also better than the FreeTTS voice. I'd rather not launch them as external processes from with the Linux GUI, though, but leave that to for example a python-based background service.

tynanford:
[...]

PR #2973, opened 2024-03-12
#2968 annunciator

shroffk:
Work on making the annunciator plugable


app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/annunciator/Annunciator.java
```diff
@@ -1,51 +1,20 @@
...
-        // Define the voices directory.
-        System.setProperty("freetts.voices", "com.sun.speech.freetts.en.us.cmu_us_kal.KevinVoiceDirectory");
-        voiceManager = VoiceManager.getInstance();
-        voice = voiceManager.getVoice(voice_name);
-        voice.allocate();
-    }
+    /**
+     * Initialize the Annunciator resources.
+     */
+    default void initialize(){}
```
kasemir, line 8: In the new Annunciator SPI interface, I don't see how `initialize()` is ever called. If it's not needed, remove? Implementations can initialize in their constructor?

shroffk: > Implementations can initialize in their constructor?

Ya, that is what the freetts implementation does too... I just thought for the sake so  symmetry to have `initialize` to mirror the `shutdown`. I can remove this.

app/alarm/freetts-annunciator/src/main/java/org/phoebus/applications/alarm/freetts/annunciator/FreeTTSAnnunciator.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Annunciator class. Uses freeTTS to annunciate passed messages.
+ * @author Evan Smith, Kunal Shroff
+ */
+@SuppressWarnings("nls")
+public class FreeTTSAnnunciator implements Annunciator
+{
+    private final VoiceManager  voiceManager;
+    private final Voice         voice;
+    private static final String voice_name = "kevin16";
```
tynanford, line 23: I know @kasemir said only the kevin16 voice worked well but I'm wondering if it is worth making this a runtime preference setting and default it to kevin16 ?

kasemir: As far as I remember, the other voices not only don't work well, they're useless. Really wouldn't bother.

kasemir: Like one is only for numbers, something like that

shroffk:
@tynanford can you try the updated version

tynanford:
@shroffk Yes had a chance to test today, looking good! When I build with `app-alarm-audio-annunciator` in pom.xml I can confirm the timeout works for me, different alarm levels play different sounds, annunicator mute/test work fine, nag period works as expected

When I just build from this branch though, the product pom.xml contains both app-alarm-freetts-annunciator and app-alarm-audio-annunciator. And this produces behavior where both kevin16 and the audio clip are playing at the same time. How do we had this with other pluggable systems in phoebus? I think it could confuse new users who just clone from github and build phoebus

shroffk:
@tynanford you are right about the duplicate annunciators.
I removed the audio one and kept the freetts one as an *optional* dependency of the common product.

People using the common product will get the freetts one... but site specific products will have the ability to explicitly include the annunciator they want to use... in the ALS-U case you can use the audio one. Others will have to include the freetts one.

tynanford (approved):

tynanford:
That makes sense to me, looks good. Added Han as a reviewer as well

tynanford:
well looks like I can't add Han as a "reviewer" so @jeonghanlee , any issues before it is merged?

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/annunciator/AnnunciatorController.java
```diff
@@ -130,7 +137,9 @@ private void processMessages()
                     {   // Annunciate if marked as stand out.
                         addToTable.accept(message);
                         if (! muted)
-                            annunciator.speak(message.message);
+                        {
+                            loader.stream().forEach(annunciatorProvider -> annunciatorProvider.get().speak(message));
```
kasemir, line 141: Without stepping into this in the debugger, it's unclear if this will load all Annunciator implementations and 'get()', that is, construct them, for each annunciated messages. Would be bad for the FreeTTS annunciator to allocate a voice or the AudioAnnunciator to load all sounds each time one message is handled.
Maybe fetch all annunciators once at startup into a `List<Annunciator>` and then just call `speak()` on each?

kasemir: Set a breakpoint in the constructor of the FreeTTS or Audio annunciator constructor, or add a printout there, and check if it's called just once or each time a message is handled.

shroffk:
Sorry, I have fixed the SPI contributions being constructed each time.

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/annunciator/AnnunciatorController.java
shroffk: Sorry, I have fixed the SPI contributions being constructed each time.

shroffk:
I am ready to merge this... at the monthly meeting I mentioned how after this PR is merged site specific products will have to explicitly include the annunciator they want to use.

PR #2975, opened 2024-03-14
Create only one instance of save&restore Jersey client

georgweiss:


shroffk:
someday we will be able to use the new java11 http client or maybe the spring one

shroffk (approved):

georgweiss:
Do we dare add Spring dependency into Phoebus?

PR #2978, opened 2024-03-18
Credentials dialog

georgweiss:
This adds a custom dialog into the core-ui module.

The intention/use case is when a client may need to prompt user for credentials, e.g. logbook or save&restore. If credentials are already cached, the dialog is not rendered. Otherwise the dialog will delegate the authentication to client provided authentication provider and save credentials to cache upon success.

Typical client call would look like

```
if(!CredentialsDialog.prompt(new SaveAndRestoreAuthenticationProvider())){
    return;
}
// proceed normally
``` 

Personally I would prefer to support an AspectJ pattern here such that a method in need of calling the ```CredentialsDialog``` could be annotated like so:
```
@CredentialsPrompt(SaveAndRestoreAuthenticationProvider.class)
```

But that would require a change to the build process. Not sure it would be accepted.


kasemir:
> intention/use case is when a client may need to prompt user for credentials, e.g. logbook or save&restore.
> If credentials are already cached, the dialog is not rendered.

We have two scenarios: Personal setup on user's computer, and shared setup in control room.
For personal setup it's fine to cache/save the credentials so user is only prompted once.
For shared setup, user must be prompted each time.

For the logbook, that's configurable via `org.phoebus.logbook.ui/save_credentials`. Does save/restore have a similar preference? Or should that type of preference be in `core-ui`?

georgweiss:
Good point, totally missed that.
Question is if credentials cache setting should be "global", or per service.

shroffk:
> Or should that type of preference be in core-ui?

I prefer the idea of having this in the `core-ui`, most of the times you will want to cache/not cache all the credentials... there are no cases I can think of where you would want to cache the user credentials for one service and not for another.

kasemir:
Agree, the `save_credentials` option can be global.
But details are a mess. You do need to separately save the credentials for making logbook entries, save/restore, ..., each application that uses credentials, because they could be different.
Ideally, some site-wide setup of course makes it that that the end user will have the same account/password for all those services, but you still get one dialog for the first logbook entry, another for the first save/restore action, ..., one for each application, even though they all use the same password.

georgweiss:
And at the same time the different scopes are still different in nature. A group account will work for some save&restore cases, but would be quite useless for the logbook (although technically allowed).

georgweiss:
I will add the credentials cache setting into core-ui, and remove it from logbook.

Agreed?

PR #2995, opened 2024-04-10
CSSTUDIO-2231 Add flag `-select_settings` for selecting a Phoebus configuration file on startup

abrahamwolk:
This PR adds the flag `-select_settings` which allows for selecting a Phoebus configuration file on startup of Phoebus.

When Phoebus is started with the option `-select_settings <directory>`, a dialog will open that allows one to select a configuration file in `directory` when starting Phoebus. Any file in `<directory>` that ends with either `.ini` or `.xml` can be selected.

`-select_settings` is not allowed to be used in conjunction with the option `-settings` in order to avoid confusion.

**Testing request:** Phoebus seems to support settings in the form of XML-files. Unfortunately, I do not have a configuration file in XML-format to test with. Therefore the loading of configuration files in the XML format in this PR has not been tested. Could someone test this functionality with a configuration file in XML format and confirm that it works?

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
+++ b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
@@ -335,7 +335,7 @@ public class PhoebusApplication extends Application {
             if (indexOfFlag < 0) {
                 throw new RuntimeException("Error, this should never happen!");
             }
-            if (application_parameters.size() >= indexOfFlag) {
+            if (application_parameters.size() > indexOfFlag) {
                 String iniFilesLocation_String = application_parameters.get(indexOfFlag + 1);
                 File iniFilesLocation_File = new File(iniFilesLocation_String);
                 if (iniFilesLocation_File.isDirectory()) {


georgweiss (commented):
Strings shown to user should be externalized to messages.properties (e.g. PhoebusApplication).

Failure to start (e.g. incorrect configuration file, unsupported combination of program arguments) should trigger a dialog with suitable error message. Otherwise user will not know why Phoebus did not start, and console/stdout/stderr might not be available.

kasemir:
> -select_settings is not allowed to be used in conjunction with the option -settings in order to avoid confusion.

So far it's fine to list multiple `-settings`, they are accumulative. We use that in our launcher scripts, always loading global `-settings` but then adding something like "beam line settings" and finally "user settings", if they exist.
So the launcher script adds one or two more `-settings` to the command line.

Similarly, it should be OK to run like this if we support `-select_settings`:
```
phoebus.sh -settings /home/controls/global.ini -settings /home/controls/beamline.ini -select_settings /home/controls/options`
```

So you have several base settings which add up, and finally some selectable option. So `-select_settings` should NOT prohibit additional `-settings`.

georgweiss:
Multiple dialogs is not ideal.
What about allowing multiple selections in the dialog?

georgweiss:
> then two or more configurations could be chosen from <directory1>, for example.

That would not be different than supporting multiple -settings. Supporting multiple -select_settings to support multiple configuration directories is maybe a degree of non-essential flexibility.

georgweiss:
The message string 
>TheArgumentIsNotADirectory="the argument '{0}' is not a directory!"
should probably not be double quoted... 
Also, single quote around {0} fails the formatting:
Screenshot 2024-04-11 at 13 00 18

kasemir:
Mind you, I am questioning the need for `-select_settings`. Typical end users should not have to pick anything. The launcher script should figure it out for them and then use `-settings` to start CSS. If your users do need to pick between a few settings, you can already create the appropriate number of desktop icons or start menu links. If you really need a selection dialog, you can do that in for example python/Qt from the launcher script. For what it's worth, on our beam lines we tend to have two entries in the Linux desktop menu to start CSS: One to just start it as the automatically logged-in generic beam line user, and another to first change to a different user. Such change-user-ID gymnastics need to be handled in a launcher script, can't do that within the JRE once CSS is running. 
Bottom line, how the settings are determined is a site-specific matter and can already be handled in launcher scripts. A `-select_settings` option doesn't add any functionality that's not already possible by other means, so there's no show stopper here that needs to be fixed.

georgweiss (approved):

kasemir:
I'd consider it superfluous, but it's not breaking anything, so go ahead.

PR #2996, opened 2024-04-17
Optimize Apache Batik dependencies

georgweiss:
Instead of listing various Apache Batik Modules, this PR instead pulls in batik-all. While this may inflate the overall footprint a wee bit, it simplifies pom files. Moreover, since Batik pulls in the xml-apis:xml-apis artefact, compilers (e.g. ajc) may be in trouble as that contains APIs duplicated from standard java module. Excluding xml-apis becomes easier when referencing only one single Batik artefact.


kasemir:
I'd like to avoid ajc, and when it comes to maven, I'm a victim, not expert, so can't really review this.

georgweiss:
This PR does not pull in ajc.

kasemir:
You mention that this PR will cause trouble for ajc because it pulls in packages that clash with standard java modules. Again, I can't say much about how to best use maven.

georgweiss:
This PR simplifies/improves Maven dependency management with respect to the APIs contained in xml-apis.

PR #2998, opened 2024-04-23
CSSTUDIO-2071 & CSSTUDIO-2081: Add options to copy & append PV names (with descriptions) to Display Runtime & to `WidgetInfoDialog`

abrahamwolk:
This PR implements part of the functionality described in https://github.com/ControlSystemStudio/phoebus/discussions/2972. In particular, it implements points 1 and 2 described in that discussion, i.e.:
1. _Extended support for copying PV names by right-clicking in the Display Runtime_
2. _Support for copying PV names in the Widget Info window_

The description of 2 in the linked discussion is not exactly what has been implemented: checkboxes have been added to the Widget Info dialog (and it is possible to right-click and select "Select All" and "De-Select All"), and only one button "Copy..." has been added.  The "Copy..." button opens a dialog, in which one then copy the selected PV names, either with or without descriptions.

~~I am working on a PR for point 3 in the linked discussion also. (I.e., 3. _Modified and extended functionality to paste several PV names (with an optional display name) into the Data Browser_.)~~
**EDIT:** The PR https://github.com/ControlSystemStudio/phoebus/pull/3000 implements point 3 in the linked discussion also. (I.e., 3. _Modified and extended functionality to paste several PV names (with an optional display name) into the Data Browser_.)

I would be very happy to discuss the functionality, the implementation, or suggestions for improvements.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/ContextMenuAppendPvToClipboard.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/ContextMenuAppendPvToClipboard.java
@@ -1,4 +1,4 @@
-package org.phoebus.applications.probe;
+package org.csstudio.display.builder.runtime;
 
 import javafx.scene.image.Image;
 import javafx.scene.input.Clipboard;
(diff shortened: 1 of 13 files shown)


georgweiss (commented):
Occurrences of ```"\n"``` should maybe be replaced by ```System.lineSeparator()```.

In the widget info dialog I'd prefer "select all" and "deselect all" to be handled by a checkbox in the Selected table header instead of a context menu. This would be more in line with common practice.

I would also suggest another button "Copy PV names" in the widget info dialog as I assume this would be the most common use case. It would save the user from interacting with another dialog.

In the Copy PV Name(s) dialog I find the repeated use of plural (s) a bit disturbing. Since upon launch of the dialog the PV selection is known, one could define separate strings for singular and plural.

PR #2999, opened 2024-04-23
Improve app menu on macOS

smarsching:
This PR fixes a problem with app menu on macOS. The app menu is the leftmost menu with a bold label in the macOS menu bar.

In particular, it resolves a problem where the “Hide CS-Studio” menu item (shortcut ⌘H) does not hide the application and instead shows a standard about dialog (not even the “real” about dialog for Phoebus).

This change does not affect other platforms. I considered moving this logic into a separate module and only including this module when building for macOS. However, doing though would mean that an extension point would have to be added to the `core-ui` module, so that new module can add this functionality.

This would make the code much more complicated and when considering that the external library that adds the necessary function only adds about 130 KB to the total distribution size, I think that carrying this library around on other platforms even though it does not do anything on them, is acceptable.


shroffk:
argh!!!
Is there a possibility that we could address this with an update to javaFX.
Should we open a ticket with javaFX

smarsching:
@shroffk Yeah, I know, it’s really annoying.

While looking for a solution, I found a [discussion from nine years ago](https://openjfx-dev.openjdk.java.narkive.com/GNXgQkvm/menubar-in-osx-to-set-about-preferences-quit), that linked to two related issues:

- https://bugs.openjdk.org/browse/JDK-8091517
- https://bugs.openjdk.org/browse/JDK-8088585

Considering that both this issues have not been addressed in about 10 years, does not make me very optimistic that proper support for the macOS menu bar has a very high priority for the JavaFX project and is going to be addressed any time soon…

shroffk:
> However, not being able to hide the application is a major issue. I think that the impact of this is comparable to not being able to minimize the application window on Windows.

I think I begrudgingly agree that this might be a valid reason to address this.

I would vote for keeping the PR as it is, personally I feel like the complexity of managing many OS specific dependencies is not worth it.

georgweiss:
In my experience almost every single Mac OS update brings ever so small changes to application behavior. For instance, just discovered that after update to Mac OS 14.1.1 the Phoebus splash screen does not work, with the (probable) side effect that main Phoebus window is not put to front if user launches from Finder.

My point is that third-party libs to overcome Mac OS quirks might break on next Mac OS update.

smarsching:
> In my experience almost every single Mac OS update brings ever so small changes to application behavior. For instance, just discovered that after update to Mac OS 14.1.1 the Phoebus splash screen does not work, with the (probable) side effect that main Phoebus window is not put to front if user launches from Finder.

@georgweiss The splash screen is working for me on macOS 14.4.1, but you are right that the Phoebus app is not automatically brought into the foreground any longer, so I think that the two issues are unrelated.

Regarding the splash screen, it did not appear on a different system running macOS 13.6.6, but I could fix that after running Phoebus one with the `-splash` parameter. After that, the splash screen was also shown again when starting Phoebus without any parameters (e.g. from the Finder). So, the issue with the splash screen rather looks like an issue with Phoebus’s preference system than a macOS-related bug.

> My point is that third-party libs to overcome Mac OS quirks might break on next Mac OS update.

Yes, making sure that there is a good user experience across different platforms is a process that never ends. I do not know any widget toolkit that succeeds in “hiding” away all the differences between platforms. This is why virtually all Java applications that provide a good UX across platforms have some platform-specific code. This is true for stuff like IntelliJ IDEA (using Swing), Eclipse (using SWT), NetBeans, etc.

So, for me the question is not whether we need platform-specific code but rather how such code can be implemented in the cleanest way possible. I see two possible paths:

1. If the Phoebus core application is supposed to properly support all three major platforms (Linux, macOS, Windows), we maintain the necessary platform-specific code in the Phoebus source tree.
2. If the Phoebus core application is supposed to only provide the minimal core functionality that is needed by everyone, ensuring a good UX is the job of the respective maintainers of the Phoebus products. In that case, the Phoebus core should provide proper extension points to enable product maintainers to add the code that is relevant to the platforms that they want to support.

The point about build time and number of dependencies that @kasemir made is relevant in this context.

However, I that this argument also applies to many other aspects of Phoebus. For example, the build would be quicker and there would be fewer dependencies if support for different archive sources, PV sources, logbook systems, etc. was not part of the core repository.

So, we either want to keep one big repository for everything (the approach also used by the Linux kernel), which will invariably lead to more dependencies and longer build times as more functions are added. Or we could switch to a concept that maintains different components in separate repositories, shortening build times and reducing the number of dependencies for each component, but having other downsides.

To be clear, I am very much in favor of the current approach of maintaining everything in a single place because it is so much easier than trying to maintain a proper API for each component without breaking backwards-compatibility. I just think that this goal cannot be achieved while also minimizing build times and the number of dependencies.

In summary, I do not expect people who do not care about the UX on macOS to put work into improving it, but we have to figure out a way for those who do care to make the necessary changes, be it in Phoebus itself or in separately maintained extension modules.

georgweiss:
Thanks @smarsching for spending more time than me investigating this.

Now that we are on Java 17 I so no reason not to go for JavaFX 21.

shroffk:
> Now that we are on Java 17 I see no reason not to go for JavaFX 21.

+1 on moving to JavaFX 21

kasemir:
Fixing a JavaFX issue by simply updating to a more recent JavaFX version is perfect, and brings this specific Mac OS issue to an ideal conclusion.

Still, some key points of the discussion need to continue, maybe you can do that at the upcoming codeathon.

EPICS has been around for 30+ years. Display tools only last ~10 years:

198x: edd/dm
199x: medm
200x: edm
2009: Eclipse-based CS-Studio w/ BOY
2016: Start of phoebus
2018: Phoebus-based CSS w/ Display builder presented at Chicago EPICS meeting
2019: Web runtime
There were other approaches based on tcl/tk, Visual Basic, Borland Delphi, .. that didn't find broad acceptance and fizzled out. There are several python/qt based approaches that have caught on.

Phoebus is in its 8th year. Will it be dead in 2 more?
Opinions will vary, but what makes a tool successful?
With EPICS base you can typically get the sources as one *.tgz or one "git clone", then "make", and you're done. The only dependency that you had to prepare was the deadline library, and that's recently become optional resp. auto-detected.
edd/dm, medm, edm used to be that way, but by now their graphics libraries are no longer bundled in default linux distros. Plus even when you do get it all built, the result shows its age.
I don't need to detail our issues with the Eclipse-based CS-Studio.
Phoebus looked great back when JFX was included in Oracle JDK 9. I was worried when it got split out, but this very ticket suggests that JFX is alive and remains a good choice. We do support a basic "git clone; mvn run", but I think we reached the point where nobody understands the dependencies. We simply hope that maven somehow resolves them.

shroffk:
Yes,
This does sound like a perfect topic for discussion at the codeathon

Just to ensure that I keep my maven fan club membership active... We have had 
successful CI builds for years now and the expansion of dependencies has not broken anything. I have checked the issue tickets and my email and except for 1-2 problems, primarily related to the use to the wrong JDK, it does not seem like the broader community is struggling with builds.

We can pull out the services, core, and each app into individual repo which would make the dependency tree smaller but at the cost of having to deal with the complexity for multiple repos.

Let us have this discussion in person though (preferably with some beer) and figure out our main concerns and how to address them moving forward :)

shroffk:
@smarsching can I close this PR and we can open a new one with javaFX 21

PR #3000, opened 2024-04-24
CSSTUDIO-2072 Modified and extended functionality to paste several PV names (with an optional display name) into the Data Browser

abrahamwolk:
This PR implements point 3 described in the discussion https://github.com/ControlSystemStudio/phoebus/discussions/2972, i.e. _3. Modified and extended functionality to paste several PV names (with an optional display name) into the Data Browser._ (The PR https://github.com/ControlSystemStudio/phoebus/pull/2998 implements points 1 and 2 of that discussion.)

In particular, this PR adds a new context-menu item "Add PV(s) from the Clipboard" to the Data Browser that allows one to paste a collection of newline-separated PV names, optionally together with associated Display Names, into the Data Browser:
contextMenu

The collection of pasted PV names can either be added to one axis only that they share, or they can be given individual axes; the following is a screenshot of the dialog that is opened by clicking on "Add PV(s) from the Clipboard":
screenshot

**EDIT (30 July 2024):** The PR has now been updated to implement the proposal described in https://github.com/ControlSystemStudio/phoebus/pull/3057#issuecomment-2255833668.

The the original parsing mechanism (that existed before this PR) is now used when parsing PV names, with one modification only: when encountering a single quote character ', the parser will continue to the matching single quote character ', in order to allow for spaces etc in display names; this is how double quotes " and brackets () were handled already.

PV names that start and end with single quote characters are interpreted as the display name of the preceding PV name (if it exists).

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/Messages.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/Messages.java
@@ -65,6 +65,7 @@ public class Messages
                          CursorValueTT,
                          DataBrowser,
                          DataBrowserMenuPath,
+                         DefaultDisplayName,
                          DeleteAxis,
                          DeleteAxisWarningFmt,
                          DeleteItem,
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/AddPVDialog.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/AddPVDialog.java
@@ -253,7 +253,7 @@ public class AddPVDialog extends Dialog<Boolean>
     {
         nameAndDisplayNames.get(i).getKey().setText(pvName);
         nameAndDisplayNames.get(i).getValue().setText("");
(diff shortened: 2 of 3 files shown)


georgweiss:
General comment: this cannot be fully tested unless #2998 is merged first.

In ```DroppedPVNameParser```, should there be another case on top of ```\r\n```, i.e. ```\n```, or even ```System.lineSeparator()```?

```AddPVsFromTheClipboardMenuItem``` has hard coded message strings.

shroffk:
I am adding this to the discussion topics for Sept 16th

georgweiss (commented):
I realize this is more work, but the AddPVDialog would in my view merit layout management in fxml.

georgweiss:
I suggested FXML as a means of "continuous improvement" since this ticket is changing the layout and thus offers an opportunity to migrate away from layout managed in code.

shroffk:
I think we should aim to close this PR soon and push all improvements to future PR's

I believe this is ready to merge, any objections?

georgweiss (approved):

PR #3001, opened 2024-04-25
Improve Elasticsearch connection in alarm logging service

smarsching:
This PR introduces three changes to the alarm logging service:

1. As an alternative to specifying `es_host` and `es_port`, it is now possible to specify `es_urls`. In `es_urls` a scheme can be specified, so it is possible to use HTTPS instead of HTTP. It is also possible to specify multiple URLs belonging to the same Elasticsearch cluster, thus enabling automatic failover when one of the nodes is not available.
2. There are new configuration options `es_auth_header`, `es_auth_username`, and `es_auth_password`. `es_auth_username` and `es_auth_password` can be used to enable HTTP Basic authentication. As an alternative, the `es_auth_header` option can be used to directly set the contents auf the HTTP Authorization header, thus enabling arbitrary authentication schemes.
3. A race condition in `ElasticClientHelper.getInstance()` has been fixed. This race condition could have the effect that contrary to expectation, more than one `ElasticClientHelper` was created. It could also result in an uninitialized instance of `ElasticClientHelper` being returned.

The first two changes are very similar to a [PR](https://github.com/ChannelFinder/ChannelFinderService/pull/127) that was recently merged into the Channel Finder service.

### DIFF
--- a/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/AlarmLoggingService.java
+++ b/services/alarm-logger/src/main/java/org/phoebus/alarm/logging/AlarmLoggingService.java
@@ -54,6 +54,9 @@ public class AlarmLoggingService {
         System.out.println("-es_port  9200                           - elastic server port");
         System.out.println("-es_urls  http://localhost:9200          - comma-separated list of elastic server URLs");
         System.out.println("-es_sniff  false                         - elastic server sniff feature");
+        System.out.println("-es_auth_header header                   - elastic authorization header");
+        System.out.println("-es_auth_username username               - elastic username");
+        System.out.println("-es_auth_password password               - elastic password");
         System.out.println("-bootstrap.servers localhost:9092        - Kafka server address");
         System.out.println("-kafka_properties /opt/client.properties - Properties file to load kafka client settings from");
         System.out.println("-properties /opt/alarm_logger.properties - Properties file to be used (instead of command line arguments)");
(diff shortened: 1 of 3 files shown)


georgweiss (approved):
Looks good, but I'd like to keep the default values for ```es_host``` and ```es_port``` in the properties file as I find it useful when running the service locally with local Elasticsearch.

smarsching:
@georgweiss I set them to the empty string in the properties file, so that it is easier to detect whether both the URL and the hostname / port has been set (and log an appropriate warning). The setting of `http://localhost:9200` still is the effective default, when none of the properties are set.

I could change this and instead detect in the code when more than one of these properties has a non-default value, but I think that this would make the code slightly less readable and it does not really provide any benefits for the user. Would you still prefer that solution?

shroffk:
> I could change this and instead detect in the code when more than one of these properties has a non-default value, but I think that this would make the code slightly less readable and it does not really provide any benefits for the user. Would you still prefer that solution?

I think this would be messy and confusing.

A part of me dislikes have multiple properties controlling the same attributes...but backwards compatibility.

@georgweiss can we mark the old host and port as deprecated and remove them in a future release?

shroffk (approved):

georgweiss (approved):

georgweiss:
Sure, go ahead and merge

GDH-ISIS:
This update has certainly improved the connection between the alarm logging service and elasticsearch. Many thanks.

Issue #2945, opened 2024-02-09
Clicking twice on an "Action button" in "Open Display" mode open a blank window


lcaouen:
When clicking twice on an "Action button" in "Open Display" mode, a blank window is opened

![image](https://github.com/ControlSystemStudio/phoebus/assets/3267663/ba0b0f7a-3ffe-4514-bcb0-ec0c5177e30e)

When having a look at the source code, it's because the application_name given to the getDockItemWithInput function in org.phoebus.ui.docking.DockStage has some macros with values not replaced.
If you use the project attached, you will see that the application_name has some values not replaced in the P macro because 
```$(Dev)``` and ```$(Idx)``` are defined deeper in the embedded widgets.
However, when the window is created, the name stored in the dockItem List is correct and the ```$(Dev)``` and ```$(Idx)``` are replaced in the P macros.
So when searching if the windows already exists, the name does not match.
I think the application_name given in the application_name given to the getDockItemWithInput should be evaluated again before sending it to the getDockItemWithInput function and not just when the parent tab is created.

![image](https://github.com/ControlSystemStudio/phoebus/assets/3267663/8680e40c-f09f-48f9-8802-739b1d4fc072)

To reproduce the bug:
* unzip the file page_blanche.zip
* open page_blanche\Scl\main\main_CM1.bob
* clic on "Solenoids power supplies"
* clic on the ... button
* Keep the new window open
* clic again on the ... button and a blank page appears !

[page_blanche.zip](https://github.com/ControlSystemStudio/phoebus/files/14220626/page_blanche.zip)

PR #3004, opened 2024-04-30
Macro values are not replaced correctly

lcaouen:
This is a solution to solve the issue https://github.com/ControlSystemStudio/phoebus/issues/2945
This is not the best solution as I don't know Phoebus code so much.
Maybe we can find a better way to do it.


kasemir (commented):
To replace macros, use this which handles both `$(..)` and `${...}` and supports escaped dollar signs:
```
strBuffer = MacroHandler. replace(macros, strBuffer);
```

kasemir (approved):

PR #3005, opened 2024-04-30
Add 'Open Application' action

georgweiss:
This PR adds a new type of action for widgets: Open Application.
The idea is to be able to use (for instance) an Action Button to quickly navigate to an application, and optionally provide an input URI to configure the view of the target application.

In a sense this is a generic version of Open Display or Open File. The use case motivating this addition is the ability to open the Save And Restore application to show a particular view.

In the action details view user may choose from existing applications (rendered from display name) in order to save users from looking up a unique application identifier:

Screenshot 2024-04-30 at 16 12 54

Note: the icon for this action is to be treated as temporary until I find a better one.


kasemir:
I get the use case of opening save/restore.
I have the use case of opening the file browser.

Still, this expands on what's typically done within the display tool.
It won't be handled by the web runtime.
It won't be handled by the Diamond Web UI presented at the EPICS meeting last week.

These use cases are great examples for a short script doing something that's not usually possible within the display tool.
Maybe such scripts could be even shorter by adding a helper to the ScriptUtil, so we end up with for example
```
from org.csstudio.display.builder.runtime.script import ScriptUtil
ScriptUtil.openApplication("App Name", uri)
``` 

So instead of changing 10+ files and adding new elements to the display file which move away from a generic display format that's acceptable to other tools (web runtime, Diamond Web UI), just one new "openApplication" method in the ScriptUtil handles the use cases.

georgweiss:
Not sure I understand how a ```ScriptUtil``` method would make support for launching any Phoebus application through an action easier to implement and maintain for a web runtime. Where would the script be executed or evaluated, JS client or server?

kasemir:
Opening a file browser or save/restore GUI from the display runtime is only possible for the CSS desktop GUI.
It's not possible from the web runtime, and it's unclear what might be supported by the next display tool.
Scripts are understood to be non-portable. You can right now add a script to your displays to open a file browser, save/restore GUI, scan monitor, ... and by using a script it's understood that this is beyond the expected display runtime feature set. It might be a good example for using a script, and it would be fine to make it a little easier by adding `ScriptUtil.openApplication("App Name", uri)`, but at the same time it's a good idea to keep this on the script level and not elevate it to an action type that's then expected to be implemented by all users of the display file format.

PR #3007, opened 2024-05-03
Upgrade to Java FX 21

smarsching:
This PR is very simple: It replaces the dependency on JavaFX 17 with Java FX 21 (the newest revision 21.0.3, to be exact).

This upgrade resolves the two issues discussed in #2999.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
     <epics.version>7.0.10</epics.version>
     <epics.util.version>1.0.7</epics.util.version>
     <vtype.version>1.0.7</vtype.version>
-    <openjfx.version>19</openjfx.version>
+    <openjfx.version>21.0.3</openjfx.version>
     <jackson.version>2.12.3</jackson.version>
     <batik.version>1.17</batik.version>
     <mockito.version>2.23.4</mockito.version>


smarsching:
So, I build our Phoebus product based on this branch and macOS ARM / x86_64, Linux x86_64 and Windows x86_64 all seem to work fine.

However, Linux AArch64 is broken now, because there are no artifacts of JavaFX 21.0.3 for that architecture in Maven Central.

After reading https://gluonhq.com/services/javafx-support/ I believe that this is by intention: LTS releases for less common platforms (Linux on ARM apparently being considered one) are only provided for customers with a support subscription.

The latestest public release of JavaFX 21 including support for Linux AArch64 is 21.0.1, but according to the [release notes](https://gluonhq.com/products/javafx/openjfx-21-release-notes/), this release does not seem to include all the bugfixes that we want.

As an alternative, we could use JavaFX 22.0.1, which should include the fix and is available for Linux AArch64, but 22 is not a LTS release, so I am not too happy about this option.

So, the question is: Do we need support for Linux on ARM? Is anyone using it. I only included in our CI/CD pipeline because I thought that it might become interesting at some point, but not supporting it is not a blocker for me, and if I absolutely had to support it, I guess I could [build JavaFX from source](https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX).

Any opinions on this?

kasemir:
I'm on an apple silicon Mac, and have been using the AArch64 version.
Last time I tried, the JavaFX libs for intel actually worked fine, but I think that's because all Macs here get the "rosetta" compatibility layer installed.
I've had a user who couldn't run that on a new Mac without rosetta, so I added the AArch64 version to the list of published products on https://controlssoftware.sns.ornl.gov/css_phoebus/
I think that means we should plan to support Mac-AArch64, because there are Mac users and that's what Mac sells right now.
Linux AArch64 might indeed be less common, I've never had a need to build that.

smarsching:
Mac AArch64 works fine with JavaFX 21.0.3, I tested it. This issue only applies to Linux AArch64. I added support for it in our CI/CD pipeline because I thought it might be fun to run Phoebus on a Raspberry Pi, but unlike Mac AArch64 I do not think that support for Linux AArch64 is critical. If someone absolutely needed it, they could build JavaFX from source or decide to use an older revision of JavaFX 21 for their build.

smarsching:
There is just a warning:

> WARNING: A command line option tried to select the GTK 2 library, which was removed from JavaFX.
> WARNING: The GTK 3 library will be used instead.

I guess it would fail if GTK 3 wasn’t present, but you would probably have to use a very old Linux distribution for GTK 3 not being available…

kasemir:
Looks good to me!

Issue #1848, opened 2021-03-31
Data browser trace issues


kasemir:
Add trace "A", then trace "B" -> You have traces "A", "B".
Delete trace "A". Undo. -> You now have traces "B", "A". Correct color, same axis assignments, but order in properties has changed.

Add trace "A", then trace "B". Add annotation to trace "A". Hide trace "A".
Add trace "C". Show trace "A" again -> Annotation on trace "A" is gone.

The two issues are likely related. Adding/removing traces can change the order of traces. Annotations are associated with traces via the index 0, 1, .. of the trace in the list of traces, and as that order changes, the association is lost.

Suggested fix is to internally hold annotations as members of the trace. This change can be limited to the data browser model, not affecting the *.plt file nor the RTPlot.

thelarsjohansson:
@kasemir 
Hi!

Similar / same behavior has been found when .plt file is being opened by Phoebus, There is IndexOutOfBoundsException since 

  <annotations>
    <annotation>
      <pv>-1</pv>

pv is referred to with index -1. It's not yet clear as to how files became like this but it may be for reasons mentioned in text above.

What's the status of this issue?

Kind regards
Lars Johansson (ESS)

kasemir:
Haven't found time to look into it, un-assigned myself since I won't get to it in the near future

thelarsjohansson:
It might be multiple issues behind behavior discussed here

- delete of trace with annotation does not adjust item index for remaining annotations
      remaining annotations might refer to wrong or non-existing pv

- undo, after delete of trace with annotation, does not bring back annotation

- annotation item index with value -1
      -1 is not explicitly set but might come from indexOf operation, to find index for pv, returning -1 for index not found

thelarsjohansson:
I am currently testing solution candidate for 1st issue.

It may be that 3rd issue mentioned, annotation item index -1, is related to 1st issue. 

If annotation, that is left after delete of other pv with annotation, refers to non-existing pv and annotations then are updated ( e.g. after annotation is moved in view ), then update tries to find non-existing item index for annotation ( list.indexOf() method ) and item index now ends up as -1.

Fix for 1st issue may solve 3rd issue.

thelarsjohansson:
Above mentioned 3 issues fixed. Proceeding with hide/show traces with annotations.

thelarsjohansson:
[...]

PR #3013, opened 2024-05-07
CSSTUDIO-2043 recalculate annotation item index

thelarsjohansson:
Fix issues related to annotations for traces in data browser
- delete of trace with annotation does not adjust item index for remaining annotations
- undo, after delete of trace with annotation, does not bring back annotation
- annotation item index with value -1

Related to 
- https://github.com/ControlSystemStudio/phoebus/issues/1848


kasemir (approved):
Looks good. Maybe add some comment to the mapModelItemAnnotations:
Annotations, unfortunately, keep the index of their model item instead of a reference to the model item. As model items are added/removed, their index can change.
This map tracks annotations by model item so we can later restore the correct index.

georgweiss (approved):

PR #3014, opened 2024-05-07
Make toolbar entries configurable

### LINKED ISSUE #2888: Configurable toolbar
The toolbar allows additions via SPI https://github.com/ControlSystemStudio/phoebus/blob/master/core/ui/src/main/java/org/phoebus/ui/spi/ToolbarEntry.java

The  ToolbarEntryService that reads the SPI entries, https://github.com/ControlSystemStudio/phoebus/blob/master/core/ui/src/main/java/org/phoebus/ui/application/ToolbarEntryService.java, currently lists them all, not sure about their order.

Proposal is to make this configurable via preference settings.

Some new setting `toolbar_entries=entry,entry,entry` with a default value of `*` or maybe empty would have the ToolbarEntryService list all available entries as it currently does.
But by setting `toolbar_entries=File Browser, ...` the entries shown as well as their order can be configured.

Note that using the existing SPI information, the entries would be selected by the name they provide via `ToolbarEntry#getName()`. For the file browser, that's the localized name, so the name for the same app might differ between sites. Likely no problem because the setting is site-specific anyway, and the default of `*` or empty would work for all.

kasemir:
Introduces preference setting for listing toolbar entries.

Format: Comma-separated list of entries.

The special entry "*" adds all remaining available toolbar entries,
basically what used to happen, without any control over their order.

An entry starting with "!" removes that item from the available entries.
The order of the initial buttons "Home, Top Resources, Layouts"
cannot be changed because they're somewhat special, for example using a drop-down
instead of a button, and being added in PhoebusApplication.
But they can still be suppressed by adding "!", for example "Home, !Top Resources, !Layouts"

Examples:

Default buttons, then all remaining available items:

    Home, Top Resources, Layouts, *

Default buttons, then assert that File Browser comes next:

    Home, Top Resources, Layouts, File Browser, *

Only Home and File Browser:

    Home, !Top Resources, !Layouts, File Browser

Fixes #2888

### DIFF
--- a/app/scan/ui/src/main/java/org/csstudio/scan/ui/monitor/ScanMonitorMenuEntry.java
+++ b/app/scan/ui/src/main/java/org/csstudio/scan/ui/monitor/ScanMonitorMenuEntry.java
@@ -11,6 +11,7 @@ import org.csstudio.scan.ScanSystem;
 import org.phoebus.framework.workbench.ApplicationService;
 import org.phoebus.ui.javafx.ImageCache;
 import org.phoebus.ui.spi.MenuEntry;
+import org.phoebus.ui.spi.ToolbarEntry;
 
 import javafx.scene.image.Image;
 
@@ -18,7 +19,7 @@ import javafx.scene.image.Image;
  *  @author Kay Kasemir
  */
 @SuppressWarnings("nls")
-public class ScanMonitorMenuEntry implements MenuEntry
+public class ScanMonitorMenuEntry implements MenuEntry, ToolbarEntry
 {
     @Override
     public String getName()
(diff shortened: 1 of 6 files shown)


georgweiss:
Not sure if this has been clarified, but when the display name for an app is localized, the corresponding entry in ```org.phoebus.ui/toolbar_entries``` must match a localized override.

georgweiss:
For the File Browser, users with French locale will need to put ```Navigateur de fichiers``` instead of ```File Browser``` in the settings file. 
My point is that the documentation may need to clarify this.

kasemir:
.. or, while you're at it, you could fix this.

For example, add a method `ToolbarEntry.getID()` that returns an  ID that's fixed, not localized like the name. Implement it for all existing toolbar contributions to return what's now the English name. Then use that ID in the `ToolbarEntryService`.

That way, those who have currently configured `org.phoebus.ui/toolbar_entries` with the English name don't have to change their settings. Users in other locales will also be able to use the English name. Even if we change the English localization, the ID will remain, so no need to then adjust `org.phoebus.ui/toolbar_entries`.

PR #3017, opened 2024-05-15
Proposal for actions re-factoring

georgweiss:
**NOTE** This is in draft/proposal state.

In order to make actions (Action Button, context menu) more generic, this PR proposes two SPI interfaces and an implementation (Open Display) to demonstrate and discuss feasibility. The objective is decoupling and ability to add optional actions.

Naming and renaming of classes is preliminary.

```PluggableActionInfo``` is an SPI interface dealing with all aspects of an action, except execution.
```ActionHandler``` is the SPI interface performing the action in runtime.

### CODE REVIEW

**Comment by kasemir** [comment]
This might be a good topic for the codeathon.

**Comment by shroffk** [comment]
Yes,
In general I am leaning towards having something like this to support actions like the open/restore a save and restore snapshot ( A use case that has come up with my users too. 

I will add this to the agenda/project list.

**Comment by georgweiss** [comment]
Some major PR here: 

- ActionInfo is an interface implementations can provide over SPI to add custom actions for widgets.
- ActionHandler implementations (again over SPI) providing runtime behavior. 
- Created a separate module for existing/legacy actions. This was necessary to avoid a dependency mess (e.g. circular dependencies).

TODO (?): not sure if and how (M)EDM screens support actions, so conversion code is not necessarily tested well enough. I'd be happy to do this if such files are added here.

**Comment by shroffk** [comment]
Could you use these as an test case?

https://github.com/areaDetector/ADCore/tree/master/ADApp/op

I am also attaching some of our converted screens (they should be useful for a rough test...) 
[edl_conv.tar.gz](https://github.com/user-attachments/files/16131890/edl_conv.tar.gz)

**Comment by shroffk** [comment]
Unfortunately, I was able to find some conversion issues with EDM --> BOB
I will work on fixing the conversion and add the fix to this branch

**Comment by shroffk** [comment]
The following conversion is failing 

```
        <script file="EmbeddedPy">
          <text><![CDATA[from org.csstudio.display.builder.runtime.script import ScriptUtil
ScriptUtil.closeDisplay(widget)
]]></text>
        </script>
```

kasemir:
For the record, EDM and I think also MEDM have a designated "close display button" widget. The converter could ignore that, because we already allow users to close displays by closing the 'tab' or 'window'. No other program known to mankind has a big fat "Close" button.
But to make it look the same as in EDM, that EDM button gets converted into an action button with such a script to close the display.

georgweiss:
@shroffk, thanks for testing. Is there anything you need me to test/verify?

shroffk:
@georgweiss no.. I think I am going to remove the problematic "close" button.
I agree with Kay. We should refrain from just "converting" everything... especially if it does not make sense in the new tools.

shroffk:
Argh!!!
@georgweiss sorry I think I messed up this PR. Is it ok if I open a new PR

shroffk:
I have creates a new PR #3097 based on this

PR #3022, opened 2024-05-21
MEDM converter: Fix patching of bool expression

### LINKED ISSUE #3016: bug? medm converter does not convert a boolean expression correctly..
There might be a bug in .adl file converter...

This part of code
https://github.com/epics-modules/motor/blob/master/motorApp/op/adl/motorx.adl
```
                calc="A!=0||B==7"
		chan="$(P)$(M).SET"
		chanB="$(P)$(M).STAT"
```
got converted to 
https://github.com/epics-modules/motor/blob/master/motorApp/op/bob/autoconvert/motorx.bob
```
      <rule name="vis_calc" prop_id="visible" out_exp="false">
        <exp bool_exp="!(pv0!==0||pv1==7)">
          <value>false</value>
        </exp>
        <pv_name>$(P)$(M).SET</pv_name>
        <pv_name>$(P)$(M).STAT</pv_name>
```
and cause syntax error
==============================================
java.lang.Exception: Cannot compile rule: rectangle:rectangle #40:vis_calc.rule.py
.
.
Caused by: SyntaxError: ("no viable alternative at input '=='", ('rectangle:rectangle #40:vis_calc.rule.py', 27, 18, '    if  not (pv0!===0 or pv1==7):\n'))
==============================================

should be "pv0!=0", not "pv0!==0".

Many display in motor module got affected.

kasemir:
Fixes #3016

### DIFF
--- a/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/AbstractADL2Model.java
+++ b/app/display/convert-medm/src/main/java/org/csstudio/opibuilder/adl2boy/translator/AbstractADL2Model.java
@@ -150,12 +150,13 @@ public abstract class AbstractADL2Model<WM extends Widget>
                             widgetModel);
                 else if (dynAttr.get_vis().equals("calc"))
                 {
-                    // Make widget visible, then add rule to make invisible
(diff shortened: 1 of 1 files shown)


prjemian:
@kasemir Thanks!

PR #3044, opened 2024-06-13
Ux analytics plugin

daykin:


kasemir:
#3046 added the essential hooks into the display runtime.

The data collection and analysis are somewhat separate.
Most end-users who download phoebus to run on their office PC don't need it.
It meant to enabled in a control room, requires something like mongo for storage etc.

Still, it's currently not much code, so once a Readme.md is added that shows what to prepare, how to enable it, and maybe one example for looking at the data, maybe we do include it in this repo?

shroffk:
@daykin could you create a readme.md using the material you presented during the codeathon?

daykin:
Yes, I will add something once development is a bit more stable. Still hammering out the strategy for "How was this page opened" (From file browser? Menu entry? Action button?)

daykin:
To do list:

- [ ] Use Credential manager instead of hand-rolled thingy
- [ ] Write tests
- [ ] Eliminate separate pathways for action button vs other opening sources?
- [ ] Heat mapping

shroffk:
@daykin 
during the next developers meeting would you like to take up the cause for this PR :)

PR #3046, opened 2024-06-13
Expose JFXRepresentation's ToolkitListener in DIsplayRuntimeInstance

daykin:
Two additions that allow for plugins to track the runtime state. 

1. Add a FutureTask called representation_init so others who hear about the RuntimeInstance from e.g. a ListChangeListener can wait until the display is loaded with something interesting before dealing with it. An interested client would call DisplayRuntimeInstance.getRepresentation_init().get() to block until the instance is ready.
2. Add methods to add and remove ToolkitListeners to the instance's underlying JFXRepresentation.

For consideration: The side-effect of (2) is that anyone with a valid reference to a (perhaps critical) toolkit listener can mess with it from an additional place besides JFXRepresentation. However, JFXRepresentation.addListener/removeListener are already public, so this isn't exactly bypassing any built-in separation of concerns.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java
@@ -17,6 +17,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.FutureTask;
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


kasemir:
A slightly more canonic phrasing for a future or future task that returns nothing might be
`FutureTask<Void>` instead of `FutureTask<Object>`. With `Object`, you wonder what it is.

daykin:
> A slightly more canonic phrasing for a future or future task that returns nothing might be `FutureTask<Void>` instead of `FutureTask<Object>`. With `Object`, you wonder what it is.

Changed to FutureTask<Void>, and undid whitespace changes

kasemir (approved):

Issue #3049, opened 2024-06-14
[codeathon]Add AdvancedConverter to convert massively and recursively CSS OPI to Phoebus BOB files


katysaintin:
Add AdvancedConverter to convert massively and recursively CSS OPI to Phoebus BOB files
- [x] Manage a opi folder in addition to a list of a opi
- [x] Recursive conversion and reproduce the same tree structure of the original
- [x] Copy pictures files in the output folder if the output is different of opi folder, respecting the tree structure
- [x] Modify the import in script files (embedded scripts too) to use Phoebus "org.csstudio.display.builder" import 
- [x] Copy scripts files to phoebus_files in order to allow opi files use the original script file that use "org.csstudio.opibuilder" import
- [x] Update the references in generated bob file for Linking Container (replace by bob file), for phoebus scripts .
- [x] Create corresponding unit test
- [ ] Create a app to call AdvancedConverter 
- [x] Add a documentation
- [ ] Fixing other issues unexisting properties and widgets and so on see https://github.com/ControlSystemStudio/phoebus-converter

PR #3050, opened 2024-06-14
Add AdvancedConverter see issue #3049

katysaintin:
[codeathon]Add AdvancedConverter to convert massively and recursively CSS OPI to Phoebus BOB files #3049
This is an additionnal class that extends Converter class, and add corresponding Unit Tests

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
@@ -15,6 +15,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 
+import javafx.application.Platform;
 import org.csstudio.display.builder.model.DirtyFlag;
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.UntypedWidgetPropertyListener;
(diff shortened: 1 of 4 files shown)


kasemir:
Maybe add some info about the AdvancedConverter to the documentation?
All the display converter info is currently in https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/convert-medm/doc/index.rst

kasemir:
You use org.apache.commons.io.FileUtils to copy, move, read a file to string, write string to file.
That’s almost as easy with java.nio.file.Files.copy, move, Files.readString(Path.of(“…”)), Files.writeString(Path.of("..."), text).

The only difference is really the added `Path.of(...)`

shroffk:
@katysaintin can we replace the file copy using java nio?
This would help us remove the dependency to apache commons

katysaintin:
> @katysaintin can we replace the file copy using java nio? This would help us remove the dependency to apache commons

Yes of course,
I have to modify my code in order to remove apache commons.io dependency.
This dependency was already brought by core-ui .
So that why I thought that I can use it.

I will update my code and add rst documentation.
As @kasemir request to me.

Katy

shroffk:
@katysaintin @lrioult Why was this PR closed?
generally this would be a feature I was hoping to include in our Phoebus

katysaintin:
> @katysaintin @lrioult Why was this PR closed? generally this would be a feature I was hoping to include in our Phoebus

Hello @shroffk , I have ask to Léo to make a new Pullrequest , taking in account the request of Kay .
Removing commons.io dependency and adding a rst documentation.
So Now O think it's OK, I will reopen my pull request.

Sorry for this deletion, I will fix that in few minutes

PR #3052, opened 2024-06-17
Refactor guava

shroffk:
Based on the discussion at the codeathon I have started removing the use of google.guava

The last remaining requirement is from the `app-databrowser-json` which uses

```
import com.google.common.primitives.ImmutableDoubleArray;
import com.google.common.primitives.ImmutableIntArray;
import com.google.common.primitives.ImmutableLongArray;
```


shroffk:
I plan to push the app-databrowser-json fix as a separate PR to make testing easier.

kasemir (approved):

PR #3054, opened 2024-06-18
Replacing use of guava ImmutableLists with Collections.unmodifiableList

shroffk:
Based on what I understand from the code I tried to replace the google lists with the collections solution.
I don't have a way to test this and hence would appreciate some help review/testing of this

### DIFF
--- a/app/databrowser-json/pom.xml
+++ b/app/databrowser-json/pom.xml
@@ -40,12 +40,6 @@
     </dependency>
 
     <dependency>
-      <groupId>com.google.guava</groupId>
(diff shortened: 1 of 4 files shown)


kasemir:
I've neither created nor used the json data reader, but what you did to replace ImmutableLists with Collections.unmodifiableList looks good to me, thanks!

kasemir (approved):

kasemir:
.. apart from test failures which still depend on guava

kasemir:
👍

kasemir (approved):

PR #3055, opened 2024-06-18
Basic listener so interested parties can see who opened a given resource

daykin:
As part of the 'UX Analytics' project, we can see from a 'push' perspective where an 'OpenDisplay' action leads. but I got to thinking this weekend that I'd also like to know how a display appeared if it wasn't from another OpenDisplay action - e.g. FileBrowser, Top Resources menu, context menu, or the back/forward buttons. 

My best guess at accomplishing this is a static 'service' to notify zero or more registered listeners. I figure that as part of core-ui, it introduces no new dependencies.

If this is OK, notifying interested parties requires a one-line invocation of the static ResourceOpenedNotifierService.NotifyListeners(resourceName, fromWho) in:

1. FileBrowserController (When display is opened)
2. core.ui,PhoebusApplication.createTopResources (attached to menu entries, when they are clicked)
3. display.builder.runtime.app.DisplayRuntimeInstance (When tab is restored)
4. display.builder.runtime.app.ContextMenuSupport? (When display is opened from right-click menu)

Is this a reasonable way to do it, or maybe there's something better?

### DIFF
--- a/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitListener.java
+++ b/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitListener.java
@@ -44,4 +44,7 @@ public interface ToolkitListener
      *  @param value The value
      */
     default public void handleWrite(Widget widget, Object value) {};
+
+    default public void handleMethodCalled(Object... user_args) {};
+
 }
--- a/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitListener.java
+++ b/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitListener.java
@@ -45,6 +45,11 @@ public interface ToolkitListener
      */
     default public void handleWrite(Widget widget, Object value) {};
 
+    /**
+     * A method was called from the UI that other listeners might be interested in.
+     * @param user_args Zero or more objects relevant to what was called.
+     *      Case-specific Implementations should expect and check these.
+     */
     default public void handleMethodCalled(Object... user_args) {};
 
 }
--- a/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitRepresentation.java
+++ b/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitRepresentation.java
@@ -689,6 +689,20 @@ abstract public class ToolkitRepresentation<TWP extends Object, TW> implements E
         }
     }
 
+    public void fireMethodCall(Object... user_args) {
+        for (final ToolkitListener listener : listeners)
+        {
+            try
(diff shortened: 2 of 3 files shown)


kasemir:
The interface and service look benign, but I'm wary of having everybody then use that to fire the updates, as you already mention the FileBrowserController, TopResources, restored tab, ..., potential new navigator, and all that with, in most cases, nobody listening.

Since you have the basic hook for detecting when a display is opened from the UX analytics side, you could right there look at the thread info and parse the stack to see how you got there, somewhat like

```
System.out.println("Display opened from:");
for (StackTraceElement level : Thread.currentThread().getStackTrace())
      System.out.println(level.getClassName() + "." + level.getMethodName());
```

In there you can then look for `PhoebusApplication.restoreState` to detect if it's restored during the initial startup etc.
That way there are zero changes to any other apps.

daykin:
Oh nice, I didn't know such a thing existed. That is probably a better way to do it.

kasemir:
It might be tedious to look for various patterns in the call stack. It's the "big data" type approach where you look for something useful in the haystack. Still, please give it a try because it would localize such changes.

daykin:
The current hook into the DisplayRuntimeInstance is a ToolkitListener implementation, which can only see actions, PV Writes and clicks. Would it be a good idea to add an (optional) "handleDisplayOpen" that would accept either the raw or filtered stack trace?

kasemir:
If you had a hook into DisplayRuntimeInstance.trackCurrentModel() , I think you'd be called whenever something is opened, and in your callback you could then check the stack for how you were called: Initial memento, action, layout change, ...

kasemir:
That makes sense because the process is something like this:
User pushes a button -> we load display in background job -> eventually the display is rendered on the UI thread

So you'd need an earlier hook that shows on the stack trace how it's getting there, not the last step which always looks the same. Still, I hope there's a way to find such a hook in the display and not having to update many other places.

daykin:
I get something helpful from calls to `loadDisplayFile`, now the challenge is making sure someone's listening before that happens.

daykin:
I like this better. It only modifies the display runtime module, and avoids firing tons of events that nobody cares about. Or, at least, if they *are* fired (as of now, just in the relatively infrequent case of a display being opened), they noop by default.

kasemir (approved):
The change to the display runtime & representation is indeed minimal, so if that suffices for your UX analysis, that's great!

daykin:
Couple of changes: Only fire the event if the display model is actually (timely) loaded, Made getDisplayInfo() public, grab application stack trace so the event (actually fired from a background thread) carries the punchline (where the event came from) with it.

kasemir (approved):

PR #3057, opened 2024-06-20
Update the widget info to allow simpler selection of PV's

shroffk:
Based on some of use case discussion we had during the codeathon. I have updated the WidgetInfo in the display builder to simplify the process of selecting a set of PV's from an screen or a widget

image

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
@@ -18,7 +18,14 @@ import java.util.List;
 import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
+import javafx.event.EventHandler;
+import javafx.scene.control.ContextMenu;
+import javafx.scene.control.MenuItem;
+import javafx.scene.control.SelectionMode;
+import javafx.scene.control.cell.PropertyValueFactory;
+import javafx.scene.input.ContextMenuEvent;
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.Widget;
 import org.csstudio.display.builder.model.WidgetDescriptor;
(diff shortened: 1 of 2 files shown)


kasemir (approved):
I can't complain, looks good

shroffk:
I am hoping this is an alternate solutions which can address some of the points from the https://github.com/ControlSystemStudio/phoebus/pull/2998 

We might have to add some more features, but this was a simple change which would address some of the use cases which lead to the above PR

kasemir:
1) copy & append

I'd prefer to stay with just the "copy". Keeps the context menu smaller. If you add "append", there'll always be somebody who prefers a different concatenation. 


2) paste PV names into the data browser

Sure, there could be a "Paste PV" context menu entry which ends up calling what you now get when you drag/drop something like "One, Two    Three", making a simple attempt to split the text into PV names:

Screenshot 2024-07-23 at 8 39 44 AM

kasemir:
How do you want to parse PV names and display names from the pasted text?

kasemir:
I thought comma already separates PV names.
Is `Temperature, Pressure` a list with two PV names, or is it PV `Temperature` with description "Pressure"?

kasemir:
> .. do not currently support filters ..

We do support them. In fact we're future-proof in a sense because we simply pass _anything_, as entered, as a PV name.

Run `softIocPVA -d app/display/model/src/main/resources/examples/pvaccess/example.db` and then show both `counter01.{"dbnd":{"abs":5.0}}` and `counter01` in probe or data browser, and you'll see the different update rates:

Screenshot 2024-07-25 at 9 08 50 AM

Works with `pva://...` as well. It does work because the filtering is done on the IOC end. We simply pass that whole PV name on as given. We don't try to understand what rsrv, qsrv, or qsrv2 on the IOC, or pcaspy, p4p, pvaPy, ... in a similar looking CA or PVA server, or other protocols like `mqtt://..`, `sim://...` might have in mind when they look at a PV name.

This does mean if you have an archive engine configured to sample `counter01`, we'll only find archived data when you try to plot `counter01`. If you entered `counter01.{"dbnd":{"abs":5.0}}`, we'll send that as a name to the archive data server which will likely return nothing, because it doesn't match `counter01`.

Trying to parse filters would be tricky. You probably don't want to understand the complete filter syntax, since that would require updates whenever something new is added on the IOC end. So you simply remove anything between `{` and the closing '}', right? Well, there are sites which use something like `BR:C001-MG:1{QDP:D}Fld-SP` as PV names, the `{QDP:D}` is not a filter, you must not strip that out.

kasemir:
>  the current functionality to split PV names

Yes, it would split on spaces or commata. It's not claiming to be perfect. That's in part why it offers what we think we got as PVs in the "Add PV" dialog, so end user can then copy/paste/edit the names back together.

 > Are you sure names like BR:C001-MG:1{QDP:D}Fld-SP are in use?

@shroffk, that's from your channel finder examples...

shroffk:
> Are you sure names like BR:C001-MG:1{QDP:D}Fld-SP are in use?

yes, these are very much in use and this pattern is applicable to 100,000s of PV's here

kasemir:
`pv-name-1 'display-name-1' pv-name-2` looks like the most practical approach.
In any case, it should result in the "Add PV" dialog so user can check if it makes sense, maybe tweak name, description, axes, .., then "OK" (or "Cancel")

kasemir:
.. "Add PV" dialog which needs to be extended to include display name ..

PR #3062, opened 2024-06-27
Add the archive readers to the common product

shroffk:
Adding the archive readers to the common product itself...
site specific products built using the common product will not have to be updated


georgweiss (approved):
Seems harmless. Only pulls in core stuff and pbraw, right?

PR #3063, opened 2024-06-28
CSSTUDIO-2472 Bugfix: Fix freezes of Display Runtime when rapidly changing between embedded displays

abrahamwolk:
This PR is intended to solve a bug in Phoebus where Display Runtime may freeze when rapidly changing between embedded displays.

The following bug has been observed, and this PR intends to fix it: an Embedded Display widget contains a complex OPI. The OPI contained in the embedded display is determined by a local PV that the embedded display subscribes to. When quickly switching (with a frequency around 3 times/second), the Display Runtime often freezes.

It is only possible to close the Display Runtime instance in question by twice attempting to close it.

This PR implements the following changes:

1. `EmbeddedDisplayRepresentation.updatePendingDisplay()` and `EmbeddedDisplayRepresentation.dispose()` have both been modified so that they are executed on the JavaFX Application Thread. Empirically, it seems that these two methods could interfere with each other, and cause `NullPointerException` to be thrown.
2. `JFXBaseRepresentation.dispose()`, `XYPlotRepresentation.dispose()`, and `WidgetRepresentation.dispose()` have been modified _not_ to set certain state to `null`, since these modifications also seemed to lead to `NullPointerException` being thrown in some cases.

I do, unfortunately, not have a full explanation for why these changes solve the observed issue. However, it seems that concurrent modification of state is the cause of the issue, and intuitively it seems that these changes can prevent concurrent modification of the variables in question.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/EmbeddedDisplayRepresentation.java
@@ -15,6 +15,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 
+import javafx.application.Platform;
 import org.csstudio.display.builder.model.DirtyFlag;
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.UntypedWidgetPropertyListener;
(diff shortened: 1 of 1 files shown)


kasemir:
Not sure if this doesn't worsen the situation.

The atomic reference and statements like `pending_display_and_group.getAndSet(..)` are in there to handle concurrency, to assert that we have a known order of what's pending.
The atomic reference logic was specifically introduced to handle the situation where we want to show content A, B, C, but representing A took so long that by the time we perform the "next" update, we can simply skip B and show C right away.
With your update, wouldn't we show A, then B, then C, all on the UI thread, and overall be slower?
If you do remove the concurrency and perform all work on the UI thread, you should be able to remove the atomic reference logic, simplify that code, update the comments on how displays A, B, C requested in quick succession are now handled.

It also looks like too many `xx = null` statements are removed. When the XYPlot is disposed, its `plot` is disposed, and `plot = null` is good to have in there to assert that we detect accidental access to a plot that is no more. Same with the `null` assignments in `WidgetRepresentation.destroy()`

You may well be on to some concurrency error, but I'm not in favor of leaving atomic references and associated logic which is no longer needed, and removing those set-disposed-stuff-to-null safeguards because that just seems to make it harder to spot the real issue in the long run.

kasemir:
> the current PR should be conservative

I agree and suggest we merge.

As for setting unused references to null, this was generally done to prevent access to objects that have been disposed and should no longer be accessed. We obviously disagree about that practice. Specific comments like "Speedup GC" tend to go back to the time when I was frequently running the code under jprofiler to look for CPU bottlenecks and memory usage. After for example closing a display, related objects can linger around for a long time, and explicitly setting references to null would result in a cleaner memory view.

PR #3067, opened 2024-07-04
Override default EPICS protocol for save&restore service. Updated doc.

georgweiss:
Updated doc to clarify how to start the service to override default protocol and how to configure address list.


shroffk (approved):
I guess this is a hack...but don't know of better solutions right now

PR #3076, opened 2024-07-15
Allow adding a permissive hostname verifier to the olog rest client

shroffk:
Maybe we might want to explore adding this to the other rest clients too... but if we are switching to java httpclient (https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/HttpClient.html) then we might have to redo this solution again


georgweiss (commented):
Honestly I'd like to start by ditching Jersey...

PR #3079, opened 2024-07-17
Fix AlarmStateInitializer to store most recent kafka records from the alarm state topic

### LINKED ISSUE #3078: Alarm acknowledgements do not persist between alarm server reboots
When an alarm server is brought down completely and then restarted, acknowledgements of alarms are lost.

1. Have an alarm server running, with a PV in an alarm state
2. Acknowledge the alarm from the Phoebus UI
3. Restart the alarm server
4. The alarm is no longer acknowledged

PR to follow..

jbellister-slac:
Fixes #3078 

Change the end index of the substring() call to 6 so that it will be possible for `type` to equal the `AlarmSystem.STATE_PREFIX` in the equality check.

After I made this change, confirmed that acknowledgements do persist between alarm server reboots.

Although the change is minor this does cause this code block to be hit that wasn't possible before. Reading through the code it looks ok to me, and testing showed things seemed to be working as expected.

The eventual change this leads to outside this class is that initial_states will have something to pass to the `AlarmServerPV` here:

https://github.com/ControlSystemStudio/phoebus/blob/6a6c5be6ae17e578e4fae0faf14c84ff81a8d856/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/ServerModel.java#L394

### DIFF
--- a/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmStateInitializer.java
+++ b/services/alarm-server/src/main/java/org/phoebus/applications/alarm/server/AlarmStateInitializer.java
@@ -94,14 +94,8 @@ public class AlarmStateInitializer
         final ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(100));
         for (final ConsumerRecord<String, String> record : records)
         {
-            if (record.key().length() < 6)
-            {
-                logger.log(Level.WARNING, "Invalid key, expecting type:path, got " + record.key());
-                continue;
-            }
-            final String type = record.key().substring(0, 6);
             // Only handle state updates
-            if (type.equals(AlarmSystem.STATE_PREFIX))
+            if (record.key().startsWith(AlarmSystem.STATE_PREFIX))
             {
                 final String path = record.key().substring(6);
                 final String node_config = record.value();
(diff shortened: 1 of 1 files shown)


kasemir (commented):
What you propose looks better, but while you're at it, how about this, which is overall shorter
```
for (final ConsumerRecord<String, String> record : records)
{
    // Only handle state updates
    if (record.key().startsWith(AlarmSystem.STATE_PREFIX))
    {
      final String path = record.key().substring(6);
      ...
    }
}
```

jbellister-slac:
Thanks! Agreed that it is easier to read this way. Change made and confirmed it still works as expected.

kasemir:
Also, one might wonder why "6" in here: 

```
if (record.key().startsWith(AlarmSystem.STATE_PREFIX))
{
    final String path = record.key().substring(6);
```

It's because `AlarmSystem.STATE_PREFIX` happens to contain "state:" and we want to get what's after that.
It'll break if we ever changed the value of the state prefix to say "status:", and now we'd have to skip 7 chars.

This would be safer:
```
if (record.key().startsWith(AlarmSystem.STATE_PREFIX))
{
    final String path = record.key().substring(AlarmSystem.STATE_PREFIX.length());
```

... but now one could argue that just "6" is faster than calling `String.length()`.
It would certainly hurt in C where `strlen(some_string)` actually counts all the chars for each call.
With Java strings, I guess it only fetches one number, so quite fast and certainly safer than hoping the length of `AlarmSystem.STATE_PREFIX` is and remains 6.

kasemir (approved):
Thanks, that looks better.
And if it actually works, that'd be perfect!

kasemir:
Thanks for looking into this!

Issue #3075, opened 2024-07-15
Fix Macro readthedocs


shroffk:
I believe the documentation is incorrect here

https://control-system-studio.readthedocs.io/en/latest/app/display/editor/doc/macros.html#system-macros

The macros **_cannot_** be overwritten by symbol macros

shroffk:
@kasemir I can go ahead and push a fix but wanted your review to verify my understanding

kasemir:
That would be perfect, thanks!

PR #3087, opened 2024-07-23
#3075 fix system macros documentation

shroffk:
small but rather important correction in the documentation


kasemir:
How did "Symbol" in there anyway?  Symbol widget?   Or was the idea to talk about "system properties" and how they correspond to "system macros"?
Maybe just delete that, so it becomes:

```
The following macros are predefined:

 * $(DID): The unique ID for each display. ...
 * $(DNAME): The name of the display.
```

kasemir (approved):
Ideal

PR #3092, opened 2024-07-25
A new formula function for creating a VImage

shroffk:
A simple formula for creating a VImage.

In the past we had many many scripts to handle channel access images
=imageOf(array_data, height, width, ...)

### DIFF
--- a/core/formula/src/main/java/org/csstudio/apputil/formula/areadetector/ImageOfFunction.java
+++ b/core/formula/src/main/java/org/csstudio/apputil/formula/areadetector/ImageOfFunction.java
@@ -91,8 +91,12 @@ public class ImageOfFunction implements FormulaFunction
             throw new Exception("Failed to parse the image type from  " + args[4]);
         }
         final VNumberArray data;
+        final Alarm alarm;
+        final Time time;
         if (VTypeHelper.isNumericArray(value)) {
             data = (VNumberArray) value;
+            alarm = data.getAlarm();
+            time = data.getTime();
         } else {
             throw new Exception("Failed to parse the image data array.");
         }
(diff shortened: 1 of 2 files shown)


kasemir (approved):

core/formula/src/main/java/org/csstudio/apputil/formula/areadetector/ImageOfFunction.java
```diff
@@ -0,0 +1,132 @@
...
+
+        int xoffset = Double.valueOf(VTypeHelper.toDouble(args[5])).intValue();
+        int yoffset = Double.valueOf(VTypeHelper.toDouble(args[6])).intValue();
+
+        boolean xreversed = Boolean.parseBoolean(VTypeHelper.toString(args[7]));
+        boolean yreversed = Boolean.parseBoolean(VTypeHelper.toString(args[8]));
+
+
+        Alarm alarm = Alarm.none();
+        Time time = Time.now();
```
kasemir, line 108: Maybe there's a way to copy the time stamp from the image data array? That time stamp might be important.

kasemir: same with alarm info.  Admittedly, for the image widget it doesn't much matter, we plot the data and ignore the time stamp. This is more about future usage of `=imageOf(....`

shroffk: done

kasemir:
Bummer... Should these be removed?
```
import org.epics.pva.data.PVAByteArray;
import org.epics.pva.data.PVADoubleArray;
import org.epics.pva.data.PVAFloatArray;
import org.epics.pva.data.PVAIntArray;
import org.epics.pva.data.PVALongArray;
import org.epics.pva.data.PVAShortArray;
```

They're not used inside the new function, which makes sense because the function deals with VType data, not the underlying PVA... types.
So no need to import `org.epics.pva.data....`.
In fact must not do that, because now I get a build error:
```
[javac] /opt/jenkins-..../phoebus/core/formula/src/main/java/org/csstudio/apputil/formula/areadetector/ImageOfFunction.java:11: error: package org.epics.pva.data does not exist
    [javac] import org.epics.pva.data.PVAByteArray;
```

PR #3096, opened 2024-07-30
CSSTUDIO-2372 Add Display Name to the information displayed when using the cross-hair cursor in the Data Browser.

abrahamwolk:
This PR adds the Display Name to the information that is displayed when using the cross-hair cursor in the Data Browser.

The motivation for this feature is that it can be helpful to see which PV a given value is associated with, without having to consult the "Traces" tab, especially when a large number of PVs are displayed in the Data Browser.

There is a GitHub discussion of this feature here: https://github.com/ControlSystemStudio/phoebus/discussions/3095

Screenshot of the functionality:
screenshot


shroffk (approved):

thelarsjohansson (approved):
Looks good!

PR #3104, opened 2024-08-02
Adding order to the Actions as they appear in the actions edit dialog

shroffk:


kasemir:
Ah, I guess because it's a comparable object?

shroffk:
> Ah, I guess because it's a comparable object?

yes that was the goal

Issue #3108, opened 2024-08-07
backward compatibility with macros setting via scripts


shroffk:
The old `DataUtil.createMacrosInput` API used to provide a map which is not supported in the `MacrosWidgetProperty.setValueFromObject`

solution 1. change the DataUtil to return a macro object instead of a map  
```
    /**Create a MacrosInput, which can be used as the macros input for a container widget or display.
     * New macro can be added or replaced by
     * <code>MacrosInput.put(String macroName, String macroValue);</code>
     * @param include_parent_macros If parent macros should be included.
     * @return a new created {@link Macros}.
     */
    public final static Macros createMacrosInput(boolean include_parent_macros)
    {
        return new Macros();
    }
}
```
but this would require **all** scripts to be updates to replace the .put() with .add()

solution 2. add support to accept `if (value instanceOf Map)`  
```
    /** @param value Must be {@link org.csstudio.display.builder.model.spi.ActionInfo} array(!), not List */
    @Override
    public void setValueFromObject(final Object value) throws Exception
    {
        if (value instanceof Macros)
            setValue((Macros) value);
        else
            throw new Exception("Need Macros, got " + value);
    }
```

What is the reason for not supporting List/Maps here, this solution would not require existing scripts to be updated

kasemir:
Macros can have order.
`A=1,B=2,T=$(A),A=$(B),B=$(T)` and you end up with 2 for `$(A)` and ` for `$(B)`.
In a map, you'd only have one setting per key.

So a Map cannot hold the full detail that's possible with `Macros`.
But we should still allow `setValueFromObject` to read a `Map` in addition to `Macros`.

I'm even OK with `setValueFromObject` allowing a `String` of the form `name=value,name=value,...` because that would be the easiest way to get one or two macros defined from within a script.

shroffk:
Ok, 
Thank you, this makes a lot of sense.

I will add the support for the map and the string to the setValueFromObject

PR #3109, opened 2024-08-08
expanding support for types of values set-able on macro properties #3108

shroffk:


kasemir (commented):
`instance Map` just checks if it's a Map.
We cannot know for sure that it's a `Map<String, String>`.
Might be a map of objects.

All that could still be handled via something like
```
private static Macros fromMap(Map<Object, Object> names_and_values)
...
names_and_values.entrySet().forEach(
   e -> macros.add(Objects.toString(e.getKey()),
                   Objects.toString(e.getValue())));
```

shroffk:
Actually, I was planning on just throwing an exception.
`Objects.toString` might not result in anything very useful if it isn't a String.

but.. I am fine with your suggestion too

kasemir:
When called from Jython, I'm not exactly sure what you get.
Could be some `Map<PyString, PyString>` type.

Since you proposed handling maps from a script, I assume you have a test script to try this, so check if casting to `Map<String, String>` works.
Throwing an exception for other types is fine as well.

shroffk:
The existing scripts I have are javascripts ( and the more I work on this the more I believe we should not be using scripting )

However the good thing is that the scripts uses `DataUtil.createMacrosInput` which gives a Map<String, String>

I tested with a Map<Object, Object> and we basically would have a cast except when trying to access a key or value that was not a String.

I have decided to go with your suggestion since it can potentially support things like a Map<String, Integer> with example being 
Cell=22

shroffk:
Note:
This should be another example for why to avoid scripts

PR #3111, opened 2024-08-16
fixed orphan periodicSearch instances after Alarm Log Table tab is cl…

jeonghanlee:
Hi Kunal, @shroffk 

As I mentioned to you last week when I was at BNL, this is the solution on that subject.

Currently, the last - there are many things we are still studying, but we believe this is the critical issue on the alarm services performance is the orphaned process of periodicSearch (as a single thread) executed by the creation of the AlarmLogTable UI. If users "close" and "open" AlarmLogTable UI several times, the number of orphaned processes (threads) will increase continuously at 10-second intervals. So there are many threads that have random (user behavior) intervals. This causes unpredictable stability problems for the Phoebus UI, alarm services and OS resources. 

And the solution which is in the pull request is what @kasemir suggested at https://github.com/ControlSystemStudio/phoebus/blob/d6206dd0db8fc800f3a4bddd577e6235240ffc16/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java#L88

We've also seen several class uses that work incorrectly, but we're not including them in this pull request to keep our request as simple as possible.

However, I do recommend updating the following area to follow best practices that we know well.

```bash
$ grep -r setOnClosed
app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:        properties_tab.setOnClosed(event ->
app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:            search_tab.setOnClosed(event -> autoMinimizeLeft());
app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:            export_tab.setOnClosed(evt -> autoMinimizeBottom());
app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:            inspect_tab.setOnClosed(evt -> autoMinimizeBottom());
app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:            waveform_tab.setOnClosed(evt -> autoMinimizeBottom());
app/logbook/ui/src/main/java/org/phoebus/logbook/ui/LogEntryCalender.java:        tab.setOnClosed(event -> {
app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryTable.java:            tab.setOnClosed(event -> controller.shutdown());
app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCalender.java:            tab.setOnClosed(event -> {
```

Best,

### DIFF
--- a/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTable.java
+++ b/app/alarm/logging-ui/src/main/java/org/phoebus/applications/alarm/logging/ui/AlarmLogTable.java
@@ -48,7 +48,7 @@ public class AlarmLogTable implements AppInstance {
             });
             tab = new DockItem(this, loader.load());
             controller = loader.getController();
-            tab.setOnClosed(event -> {
+            tab.addClosedNotification(() -> {
                 controller.shutdown();
             });
             if (resource != null) {


kasemir:
> .. the solution .. is .. what @kasemir suggested ..

Always sane advice !

> .. Instead of `setOnClosed`, use {@link DockItem#addClosedNotification}. 
> .. recommend updating the following area ..
> app/databrowser/src/main/java/org/csstudio/trends/databrowser3/ui/Perspective.java:       XXXX.setOnClosed(

For all top-level DockItems of an App, yes, call `addClosedNotification`, don't replace the `OnClosed` handler provided by the `DockItem`.
But those GUI elements inside the data browser 'perspecive' are plain Tabs. They're no `DockItem`, they're tabs down within the scene graph of a `DockItem`. So there it's fine to call the internal tabs' `setOnClosed`, there is no `Tab#addClosedNotification`

jeonghanlee:
@kasemir Thank you for your comments, we thought they could not be replaced, but it is always better to ask experts to review them more thoroughly. That was my hidden intention. 

Still, I am a very conservative person who wants to replace it, because we - people - always follow what they want to do. Oh, this is just my wish. We don't have to do anything. I am also wondering that Java does not have a private which is C++. Or I don't know how Java works.

Thanks!

kasemir:
>  I am also wondering that Java does not have a private which is C++

This would be the same in C++

It's like this.
The JavaFX `Tab` has a `public setOnClosed(handler)` method.
For a plain Tab that you have somewhere inside a pane it's fine to use that.

Our top-level GUI element for each display runtime, editor, data browser, ... panel, the `DockItem`, is based on a `Tab`, with added functionality to show the app name or display name in the title, there's a context menu, you can drag the tabs around.
That `DockItem` installs an onClosed handler to clean up. But since it's derived from the `Tab`, the `public setOnClosed(handler)` method is still visible, you can't make something private that was visible in a base class.

To  we're relying on documentation to say "Yes, it's a Tab, but please ignore that and treat it as a black box. Specifically, please don't call setOnClosed."

In principle, we could completely implement our `DockItem` by for example copying the `Tab` code, but not offering a `pubic setOnClosed`. That would mean we have to revisit this with each JavaFX update to get potentially useful updates from the Tab code, so copy that over into our DockItem. At this point, we're not doing that.

Bottom line, Marty's axiom still holds: He who implements OOC classes considers them based on divine inspiration. The next person who looks at that simply thinks they suck.

kasemir (approved):

shroffk (approved):

shroffk:
I will merge this PR and open a new one fixing the log applications
Thank you @jeonghanlee @Sangil-Lee for fixing the alarm log table...In the past 2 codeathons we have always had a discussion item on improving the alarm log table, but we always ran out of time. I am hoping we can address it at the upcoming meeting and you will be able to share your other findings

PR #3115, opened 2024-08-20
CSSTUDIO-2555 Bugfix: edit the source of log entries (if available), not the description of log entries.

abrahamwolk:
This PR implements a change to the Edit window of the logbook application: the result of `getSource()` is used (if available) instead of the result of `getDescription()`. As a consequence, editing formatted logs is easier: formatting (such as headings, etc.) is not lost when editing a log entry, and does not have to be recreated again.

To provide a concrete example, consider the following log-entry:
```
# Test 1
## Test 1.1
### Test 1.1.1
```

Currently (without this PR), when editing this log-entry, it would appear as
```
Test 1
Test 1.1
Test 1.1.1
```
and the formatting has to be recreated.


shroffk (approved):

shroffk:
Ah,
This makes perfect sense. When editing we should be working with the source.

PR #3121, opened 2024-08-28
CSSTUDIO-2371 Open links to logbook-entries in the Logbook application and add "Back"- and "Forward"-buttons

abrahamwolk:
This pull request adds the following two features to the Logbook application:

1. Links to other logbook entries are opened in the Logbook application itself, instead of opening them in an external browser. To determine whether a link is a link to another logbook entry, it is checked whether the link starts with `LogbookUIPreferences.web_client_root_URL`. If it does, the log entry ID is parsed from the link, and the logbook entry with the ID is loaded using `LogClient.getLog()`.
2. "Back" and "Forward" buttons are added to the Logbook application, in order to navigate back and forward, respectively. I have highlighted the "Back" and "Forward" buttons with a red rectangle in the following screenshot of the functionality:
screenshot


shroffk (approved):
LGTM

Issue #3127, opened 2024-09-03
Yet another way to specify settings file?


georgweiss:
With the recent changes where ability to import settings has been removed, it is not possible to build an installer with ```jpackage``` using the ```$HOME``` symbol to specify a location of a settings file. At least not on Mac OS as ```$HOME``` is not expanded but rather used as-is.

At the same time, instructing users to maintain their settings in a file like:
```/Applications/CSS-Phoebus.app/Contents/app/settings.ini``` feels a bit awkward.

Suggesting to update the implementation of ```-settings /path/to/foo.ini``` to look for ```foo.ini``` in user's home directory if it turns out that ```/path/to/foo.ini``` is incorrect.

Alternatively look for ```settings.ini``` in user's home directory if it cannot be found in the installation directory.

kasemir:
When a complete path is given as in `-settings /path/to/foo.ini`, we should not second-guess. An absolute path is pretty well defined.

But what about looking in the home directory for the case of `-settings foo.ini`?
Or in the "current" directory and the "home" folder? Might want to define the order: First current, then home?

georgweiss:
> But what about looking in the home directory for the case of ```-settings foo.ini```?

Actually that was my intention, just phrased it a bit wrong. Order "current" before "home" is fine.

georgweiss:
#3128

PR #3128, opened 2024-09-03
Offer additional ways of finding settings file for the -settings argument

georgweiss:
#3127


kasemir (approved):

georgweiss:
@kasemir, did some changes: realized absolute path and file in current dir are one and the same for ```new File(location)```. Also updated doc.

kasemir (approved):

PR #3134, opened 2024-09-09
Sar login

shroffk:
TODO: need to still add the login prompt in the viewer/action buttons


georgweiss (approved):
While this should help eliminating the confusion around disabled menu items, I'd still prefer the strategy we discussed at some point, i.e. to not disable menu items and launch the credentials dialog when needed. The only case where this is not desirable is for folder creation, so in that case a login context menu item makes perfect sense.

That said, launching the credentials dialog "when needed" is a perfect example of the power of aspects. For instance, any method where credentials must be checked could be annotated with a custom aspect annotation, which in turn will invoke the necessary business logic. That way the business logic stays in one place, and code update is limited to adding a method annotation. Caveat: I have not tested this for the use case at hand, so this remains to be proven.

Issue #3143, opened 2024-09-24
rename 'save snapshot' in context menu


kathryn-baker:
Following the discussion around save-and-restore at the EPICS Collaboration meeting, would it be possible to rename the 'Save Snapshot' option in the context menu to something like 'Save **Screenshot**'? I think this might help reduce confusion for anyone new to Phoebus and save-and-restore as this 'snapshot' is taking a snapshot of the screen rather than a snapshot of the PV values. 

<img width="185" alt="image" src="https://github.com/user-attachments/assets/175c6ea7-5df6-4ea4-9311-f965988f8862">

georgweiss:
Agree completely

kasemir:
Yes. Actually, "Save Screenshot ...", ending with "..." to indicate that it's not click&done. You'll be prompted for a file name etc.

Just like "Save" (no "...") simply saves, while "Save As..." (with "...") will take a little longer.
Similarly, "Create Log..." and "Send Email ..." need "..." because they have some follow-up interaction.

kasemir:
Thank you for volunteering to clean that up!

PR #3144, opened 2024-09-24
context menu to use screenshot instead of snapshot

kathryn-baker:
Addresses issue #3143 to rename snapshot to screenshot in context menu.

Currently this just replaces the text in the context menu but doesn't change the variable name, is this sufficient? 

I think I can see where I'd need to update the variable names as well but as I'm not a Java developer and those changes would be much more likely to break something it may require a more rigorous review / testing!


kasemir:
If you renamed the variable, you'd also have to update the French i18n property file and who knows maybe some site already has a local i18n property file, so I'd leave the name as found.

shroffk:
@kathryn-baker sorry I did not wait for the french fix

PR #3148, opened 2024-09-26
Ditch jersey client

georgweiss:
This removes the Jersey dependency from save&restore.

Also added an integration test to facilitate verification.


shroffk:
Omg!! this is brilliant.... image this will make @kasemir very happy to be able to remove a large chuck of dependencies

At first glance this looks fairly straight forward.

I need to test a few edge cases... like some of the other REST client do a more relaxed host name verification. We would need to figure out how to support that here.

Next stop ChannelFinder, Alarm logger, and Olog clients will ditch jersey too :)

georgweiss:
Indeed, will turn to Olog next since there are other improvements I'd like to do while I am at it.

shroffk:
Phew!!! 
I am glad that the recent SAR changes did not cause conflict with this branch

shroffk (approved):

PR #3153, opened 2024-10-01
Reproducible builds

shroffk:
With each packaging of the Phoebus jar there is now a `git.properties` file which includes the commit id associated with this build

Additionally, I use the git commit time to mark the `project.build.outputTimestamp` which can be used to compare builds as follows

```
$ mvn clean verify artifact:compare -DskipTests
```

```
[INFO] Saved aggregate info on build to C:\git\cs-studio\phoebus\services\save-and-restore\target\service-save-and-restore-4.7.4-SNAPSHOT.buildinfo
[INFO] Aggregate buildinfo copied to C:\git\cs-studio\phoebus\target\parent-4.7.4-SNAPSHOT.buildinfo
[INFO] Checking against reference build from central...
[INFO] Reference buildinfo file not found: it will be generated from downloaded reference artifacts
[INFO] Reference build java.version: 17 (from MANIFEST.MF Build-Jdk-Spec)
[INFO] Reference build os.name: Windows (from pom.properties newline)
[INFO] Minimal buildinfo generated from downloaded artifacts: C:\git\cs-studio\phoebus\target\reference\service-save-and-restore-4.7.4-SNAPSHOT.buildinfo
[INFO] Reproducible Build output summary: 183 files ok
[INFO] Reproducible Build output comparison saved to C:\git\cs-studio\phoebus\services\save-and-restore\target\service-save-and-restore-4.7.4-SNAPSHOT.buildcompare
[INFO] Aggregate buildcompare copied to C:\git\cs-studio\phoebus\target\parent-4.7.4-SNAPSHOT.buildcompare
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,6 +92,8 @@
     <maven.compiler.target>17</maven.compiler.target>
     <activemq.client.version>5.18.4</activemq.client.version>
     <commons.compress.version>1.26.1</commons.compress.version>
+    <!--suppress UnresolvedMavenProperty -->
+    <project.build.outputTimestamp>${git.commit.time}</project.build.outputTimestamp>
   </properties>
   <build>
     <plugins>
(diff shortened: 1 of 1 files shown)


kasemir (approved):

kasemir:
Looks like io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1 requires Maven version 3.6.3, failed with maven 3.3.9.
Updated to maven 3.9.9, but that requires changes in the syntax of some locally used antrun tasks. Maven 3.9.1 seems to be a usable version.

shroffk:
I will add a note in the readme about the minimum maven requirement ( I tested with 3.8.X )

Is there someplace else that I should fix the maven version too?

georgweiss:
Running:
```$ mvn clean verify artifact:compare -DskipTests``` gives me:
Failed to execute goal org.apache.maven.plugins:maven-artifact-plugin:3.5.1:compare (default-cli) on project parent: Build artifacts are different from reference -> [Help 1]

PR #3155, opened 2024-10-03
work on getting all the embedded screens from a root screen

shroffk:
Not necessarily very happy with the structure of the code and am quite happy for input


georgweiss (approved):
While this change add an API to spot all embedded displays, there is still some processing to perform. Patching pv_name in all files is not sufficient, the parent file referencing the embedded display must be updated to reference the patched bob file.

georgweiss:
I'm afraid that with a non-trivial structure of bob files referencing resources in other parts of the file system, only a deep copy of substantial parts of the file system would be needed. Not being able to access resources will make the display look broken and that in my view defeats the purpose of "replaying" a display.

We've discussed background screen recording as an alternative, in particular as stakeholders claim the use case is about quite recent events in the OPI rather than replaying something in a more distant past.

For the time being we can put this on hold.

PR #3161, opened 2024-10-09
SAR

shroffk:
Still work in progress

I have tested that the restore works from client and service.
There are preferences to allow us to configure so that everything works as it used to.

I still can't get some of the DAO tests ( but I think this is a problem in my end / elastic setup )


georgweiss (commented):
Tests look good, but restore from client does not take advantage of
```SnapshotUtil.restore()```.

shroffk:
Fixed the app so that the client uses the same utility class for client side restore too.

shroffk:
@georgweiss this is ready for now

There are a few TODOs but we can open separate PR's
- preference/properties reading in the utility which functions correctly both in Phoebus and in the Springboot service.

georgweiss (approved):

PR #3162, opened 2024-10-10
CSSTUDIO-2731 Fix bugs in handling SplitPanes when splitting left/right or top/bottom.

abrahamwolk:
This pull request contains fixes for the case of a `SplitPane` when splitting `DockPane`s into left/right or top/bottom.

The case of a `SplitPane` only occurs when the ESS Phoebus Navigator is loaded. The changes should therefore not affect other sites than ESS.


georgweiss (commented):
In the comments you reference the "Navigator application", but that is an ESS internal module.

georgweiss (approved):

PR #3165, opened 2024-10-14
Add unique id to ToolbarEntry

georgweiss:
Noted during testing: the wildcard * does not actually include all ```ToolbarEntry``` implementations as not all of them declare the SPI implementation on META-INF/services. Is this an omission or intentional?


georgweiss:
Well, if both ogbook/olog and logbook/ui are included in the product then there will be other issues to deal with?

What is the use of the interface implementations if they are not detected by ```ServiceLoader```?

kasemir:
Are  logbook/olog and logbook/ui mutually exclusive? Then they can use the same ID.

I don't understand the second comment. There's a `ToolbarEntry` interface. The `ToolbarEntryService` uses SPI to locate all "available" entries. SPI requires you to declare your contributions in `META-INF/services`. Just implementing the interface doesn't suffice, must also list it in `META-INF/services`.

georgweiss:
They should be mutually exclusive yes, but if both included by mistake then users will for sure notice.
```ToolbarEntryService``` will never be able to locate (for instance) ```LoggingConfigurationToolbarEntry``` as there is no entry in any META-INF/services file.

kasemir:
Correct, and `LoggingConfigurationToolbarEntry` indeed doesn't show in the toolbar. If you want it to show up, you need to list it in `META-INF/services`, and then either list it specifically in the preference setting, or implicitly via `*`

georgweiss:
My point is that ```LoggingConfigurationToolbarEntry``` is redundant if there is no ```META-INF/services``` reference.

kasemir:
I still don't understand. Redundant to me means it's an unnecessary extra, implementing a functionality that's already provided by other means.

I'd say it's incomplete. The `LoggingConfigurationToolbarEntry` has been provided, but it's missing the `META-INF/services` entry to become functional.

kasemir:
Tweaking the log settings is not something that end users routinely need to do. It's only necessary to help with debugging issues.  If there was a toolbar button for it, I would exclude it from the preference setting to keep the tool bar compact.

But that's why we have the preference setting, so I'd leave the decision to whoever's maintaining the logging configuration tool. @shroffk ?

shroffk:
the logging configuration is a debugging tool and I don't think it merits inclusion in the toolbar ( keeping it clutter free helps usability )

I can remove the unused toolbar entry implementation for this app

georgweiss:
Logging config is but an example, there are others, e.g. Probe, PVTree...

kasemir:
Probe and PVTree: I would not include them in our preference settings here. Probe and PVTree are most useful when opened from the context menu of a display to peek into something in more detail. Opening them from the menu and then having to type a PV name is rare, so really no need to add to our toolbar.
But maybe you want a "development" version of CSS where you use them very frequently? So I'm OK with completing the SPI support, since I have the preference setting to control the end result

georgweiss:
@kasemir, not sure why you assume I want a "development version" of CSS.
What I do want is get rid of code that cannot be used as-is, or make it usable.

kasemir:
The eclipse-based CSS had a toolbar button for probe. Was that just there because probe was the very first tool we implemented, and in doing so we played with all the options: Menu bar, toolbar, context menu? Or was that actually useful when you start to develop some display, so your very first step is to view some PV in probe, then figure out how to handle it in the display?
My leaning is that I don't want probe and PV tree in the toolbar, but you need to decide for yourself.

georgweiss:
Things is I cannot add them event if I want unless I add the ```META-INF/services``` files. Will do that for Probe and PVTree.

shroffk:
I think we can break this PR into 2 bits
1. have tool bar entries have a unique Id, this is accomplished here
2. figure out a way to contribute things to the toolbar in a away that is easily configurable site to site, product to product.

I think it is Ok if we have a few SPI implementations which are not included by default (I understand that it might mean they are never used). I had hoped that site specific products should be able to add them using the services file. So someone who packages the product and understand the needs can add them while building the product... I might be missing something here so very much open to better understanding the use cases here :)

kasemir:
> I had hoped that site specific products should be able to add them using the services file.

Well, that can be used to show/hide. 
We should not encourage changing the source code for a local product, like patching `app/filebrowser/....../META-INF/services`. But I guess you could remove all the `META-INF/services` files from the common repository and have each site put those into their `product/META-INF/services`.

Still, I find preference settings easier to handle, and in this case, the toolbar preference can not only show/hide but also control the order of toolbar buttons.

georgweiss:
> I think it is Ok if we have a few SPI implementations which are not included by default (I understand that it might mean they are never used). I had hoped that site specific products should be able to add them using the services file. So someone who packages the product and understand the needs can add them while building the product... I might be missing something here so very much open to better understanding the use cases here :)

Configurable toolbar as implemented by @kasemir is in my view precisely the option needed to support per-site customization. So I see no harm in adding ```META-INF/services``` files to the common code.

shroffk:
Ok,
I think both of you make sense and I take back my initial suggestion.
In addition to the preferences being easier I think the ordering is an important limitation associated with SPI contributions

georgweiss:
@kasemir , @shroffk , OK to merge?

kasemir:
Fine with me

PR #3174, opened 2024-10-24
Option to customize logbook app name reinstated

georgweiss:
Fixes complaints mentioned in #3163 
@tynanford, this makes it possible to customize the app name again using ```org.phoebus.logbook.olog.ui/log_entry_table_display_name```


kasemir (approved):

tynanford:
Awesome, thanks!

PR #3186, opened 2024-11-12
Support the Standalone Alarm Logging service

jeonghanlee:
Hi Kunal @shroffk and Kay @kasemir 

Based on our mode of operation, we want to run the alarm logging service (alarm-logger) as a standalone service.

For the elasticsearch folder relocation with the limited requirement, we've verified that the Pheobus alarm log table and the standalone alarm logging service work well without printing our so many error messages regarding the communication to the alarm server (Kafka). This could be a potential use for independent alarm studies without interfering with a production environment. 

Please let us know what you think about this PR.

P.S. I @jeonghanlee  updated the README to reflect the minimal documentation description of the service. It looks like a lot of the documentation will be updated to reflect what we have changed since a couple of years ago. We can discuss them when we have the meeting together later this year or next year.

### DIFF
--- a/services/alarm-logger/README.md
+++ b/services/alarm-logger/README.md
@@ -1,6 +1,7 @@
-# Alarm Logging
+# Alarm Logging Service
 
-Logging alarm state and other messages to an elastic back end.
+The alarm logging service (aka alarm-logger) records all alarm messages to create an archive of all alarm state changes and the associated actions.
+This is an elasticsearch back end.
 
 ## Dependencies ##
 1. Elasticsearch version 8.x OS specific release can be found here:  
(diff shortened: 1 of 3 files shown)


kasemir:
Fine with me.
For what it's worth, my view of the alarm logger is as a tool that forwards messages from a running alarm system to elastic.
You might be able to replace it with Logstash, input = Kafka, output = elastic.
I then look at the alarm stats in Kibana.
The 'standalone' alarm logger option is mostly for the alarm history table, but you can get that same information from Kibana as well.

jeonghanlee:
> Fine with me. For what it's worth, my view of the alarm logger is as a tool that forwards messages from a running alarm system to elastic. You might be able to replace it with Logstash, input = Kafka, output = elastic. I then look at the alarm stats in Kibana. The 'standalone' alarm logger option is mostly for the alarm history table, but you can get that same information from Kibana as well.

@kasemir I agree with you about the main purpose of the alarm logging service, yes, we can do that through Kibana and something else. However, I would like to keep it at least in some basic way within Phoebus and its service environment. Phoebus Alarm Log Table is not in good shape either, but still it can help users to find and search what Alarm Logging Service can do without any external dependencies.

PR #3189, opened 2024-11-13
doc: show full preference names

mdavidsaver:
As a follow on to #3187.  Prepend package names when generating `preference_properties.html` so they appear complete, because clearly users will copy+paste without reading the page preamble.

I couldn't immediately see a way to compute the package name from the preferences file name, so I opt to parse out the `# Package ...` line present in (almost) all files.  This parsing is simplistic, looking at the first non-space character.  It also can not fix preference names which appear in comments.

Did I guess the correct package for `app/databrowser-json/src/main/resources/archive_reader_json_preferences.properties`?

### DIFF
--- a/app/databrowser-json/src/main/resources/archive_reader_json_preferences.properties
+++ b/app/databrowser-json/src/main/resources/archive_reader_json_preferences.properties
@@ -1,3 +1,7 @@
+# ---------------------------------------
+# Package org.phoebus.archive.reader.json
+# ---------------------------------------
+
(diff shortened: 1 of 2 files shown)


kasemir (approved):
Good idea.

kasemir:
> Did I guess the correct package for app/databrowser-json/src/main/resources/archive_reader_json_preferences.properties?

The actual package is `org.phoebus.archive.reader.json`.

mdavidsaver:
> The actual package is `org.phoebus.archive.reader.json`.

Updated.

PR #3190, opened 2024-11-13
Make DESC field use in databrowser non-default

georgweiss:
This PR is a follow-up on #3135.

My users have expressed dissatisfaction over the changed default behavior. Underlying issue is in my view that IOC developers may or may not consider the DESC field to be something for a high level client. There is also a bug in the current implementation: if DESC is not set, then the description is set to an empty string, not to ```null```.

I have added a preference setting to control the default policy: ```pv-name``` or ```pv-desc``` where ```pv-name``` is the default. To use DESC user will need to add ```org.csstudio.trends.databrowser3/value_axis_label_policy=pv-desc``` to the settings file.

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PVItem.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PVItem.java
@@ -398,11 +398,23 @@ public class PVItem extends ModelItem
         setUnits(display.getUnit());
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


kasemir (approved):
Yes, this is much better for backwards compatibility because as you noticed DESC tends to not be set

```
# pv-name: classic policy. Default if this property is not set or set to an unsupported value.
```

georgweiss:
Yes, oftenwise DESC is not set, or set to same value across multiple PVs.

PR #3191, opened 2024-11-13
CSSTUDIO-2167 Support for "decorations" in the Logbook application

abrahamwolk:
At ESS, we have implemented a "decoration" feature into the Logbook application.

Using this feature, it is possible to enter PV Names of VEnum-type into the "Advanced Search" part of the Logbook UI, and the logbook entries are then decorated by the status of the specified VEnum PVs using color coding: each possible Enum-value is represented using one color. A graphical symbol next to the logentry displays the value of the PV.

The intention is to be able to see, at a glance, what state the corresponding PV was in when a log entry was made and to see where the PV changed its value.

Since this feature is very likely to be of interest to ESS only, the feature itself has been implemented in a ESS-specific module. However, for the feature to function, there needs to be support for adding the decoration in the main Phoebus source code. This pull request adds this functionality to Phoebus.

In particular, this pull request adds the SPI interface `org.phoebus.logbook.olog.ui.spi.Decoration`, which has the following signature:

```
public interface Decoration {
    public Node getDecorationInputNode();

    public void setLogEntries(List<LogEntry> logEntries);

    public void setRefreshLogEntryTableView(Runnable refreshLogEntryTableView);

    public Node createDecorationForLogEntryCell(LogEntry logEntry);
}
```

A new decoration (displayed next to the search results in the Logbook application) is added by implementing these four methods:
 - `createDecorationForLogEntryCell()` computes a decoration to be displayed next to one line of the search results (corresponding to the `logEntry` that is its argument) in the Logbook application.
 - `setRefreshLogEntryTableView()` receives as an argument the `Runnable refreshLogEntryTableView` which refreshes the `LogEntryTableView` that displays the search results. This can be used by the implementation of the decoration to update the `LogEntryTableView`. In our case, the data that the decoration depends on needs to be fetched from the archiver, and when the data has arrived and the correct decoration has been determined, `refreshLogEntryTableView()` is called in order to refresh the UI.
 - `setLogEntries()` is called by the Logbook application when it has received a new set of log entries to be displayed. (E.g., when a search query has returned, or when the user clicks to see the next page of the results list.)
 - `getDecorationInputNode()` specifies a JavaFX node that is displayed under "Advanced Search" in the Logbook UI, and where the user can specify input for a given decoration. In our case, this is implemented by textfields where the user can specify PVs of type VEnum.

To make the use-case more concrete, the following is a screenshot of some results in the result list of the Logbook application, decorated using a decoration that has been implemented using this interface:

screenshot

### DIFF
--- a/app/logbook/olog/ui/src/main/resources/org/phoebus/logbook/olog/ui/LogEntryCell.fxml
+++ b/app/logbook/olog/ui/src/main/resources/org/phoebus/logbook/olog/ui/LogEntryCell.fxml
@@ -15,6 +15,9 @@ summary of the log entry and its metadata.
     <HBox>
         <HBox fx:id="decorationNodes" fillHeight="true" />
         <VBox fx:id="logEntryCell">
+            <padding>
+                <Insets bottom="3.0" left="3.0" right="3.0" top="3.0"/>
+            </padding>
             <HBox.hgrow><Priority fx:value="ALWAYS"/></HBox.hgrow>
             <GridPane>
                 <columnConstraints>


app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCellController.java
```diff
@@ -142,21 +99,20 @@ public void initialize() {
...
-            Text textNode = new Text(text);
-            textNode.setFill(Color.BLACK);
+    public void setDecorations(List<Decoration> decorations) {
+        this.decorations = decorations;
+    }
 
-            StackPane stack = new StackPane();
-            stack.getChildren().addAll(rectangle, textNode);
+    @FXML
+    public void refresh() {
```
georgweiss, line 109: Why is this annotated with @FXML? The method is not referenced in the fxml file.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/spi/Decoration.java
```diff
@@ -1,7 +1,16 @@
 package org.phoebus.logbook.olog.ui.spi;
 
 import javafx.scene.Node;
+import org.phoebus.logbook.LogEntry;
+
+import java.util.List;
 
 public interface Decoration {
-    public Node decorationInputNode();
```
georgweiss: Methods need not be public (may generate lint warnings)

georgweiss (commented):
A few comments inlined.
In my view any UI visualization of decorations configuration does not belong in the advanced search view since it is not related to how search is performed in Elasticsearch.

georgweiss (approved):

PR #3195, opened 2024-11-19
CSSTUDIO-2572: Add a "Jump to Log Entry" TextField to the Logbook application

abrahamwolk:
This pull request adds a "Jump to Log Entry" textfield to the Logbook application:
screenshot

By writing a Log Entry ID into this textfield, one can jump to the corresponding log entry.

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryTable.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryTable.java
@@ -173,11 +173,10 @@ public class LogEntryTable implements AppInstance {
         }
 
         protected boolean loadLogEntryWithID(Long id) {
-            goForwardActions.clear();
-            addGoBackAction();
-
             try {
                 LogEntry logEntry = controller.client.getLog(id);
+                goForwardActions.clear();
+                addGoBackAction();
                 gotoLogEntry(logEntry);
                 return true;
             }


kasemir:
How likely it is for an end user to know the ID of a log entry? A plain text search might be more relevant, where entries with that text in the author, title, body will be highlighted, with next/previous buttons to jump between log entries that contain the word.

Otherwise fine.

georgweiss:
It is already possible to share links: see Copy URL button, which is also available in the web client.

georgweiss (commented):
Log entry ids are always numerical, so the input field should accept only digits.
Also, the cursor is aligned in the middle and after uses presses <ENTER> it jumps back to the start of the entered value.

georgweiss:
Using this new feature and then Back button. Eventually the history is empty, which is indicated by a disabled Back button. At this point I think the input field should be cleared too.
Screenshot 2024-11-19 at 14 19 27

georgweiss:
>If the parsing of the log entry ID fails (e.g., the input contains non-numeric characters), the input is discarded

In my view disallowing invalid input is nicer. That would also eliminate the need to handle the invalid input.

And yes, I think right-aligning cursor is more intuitive. And the entered chars should be left-aligned, in my opinion.

georgweiss:
When reaching (in my example) the "last" id 451, the Forward button will take user to 451.

georgweiss (approved):

PR #3196, opened 2024-11-19
Add a unit test to check if all the translation in messages_{LOCALE}.properties files are synchronized with the default one

katysaintin:
As it mentioned in the last CS-meeting with @georgweiss 
Function **checkMessageFilesDifferences** added in org.phoebus.framework.nls.NLS class in core-framework module.
It lists the differences between all the existing messages_{LOCALE}.properties and messages.properties file.
If there is a missing Key in the messages_{LOCALE}.properties or if a key should be removed.

A unit test in org.phoebus.framework.nls.NLSMessagesTest is added .
checkAllMessagesResources() lists all the differences in all the Phoebus project.
Using this unit test, all the messages_fr.properties are updated now.


georgweiss (commented):
I'd prefer the static methods in class ```NLS``` (e.g. ```checkMessageFileDifferences```) to be in the unit test class since they are not used outside that test.
When I run the test class it fails on one test. If differences are detected it's OK to print that, but a unit test failure will fail a build of Phoebus. I think we should avoid it in this case.

katysaintin:
> I'd prefer the static methods in class `NLS` (e.g. `checkMessageFileDifferences`) to be in the unit test class since they are not used outside that test. When I run the test class it fails on one test. If differences are detected it's OK to print that, but a unit test failure will fail a build of Phoebus. I think we should avoid it in this case.

OK, no problem, I understand, what I can do is : 
- move the code in a static method in NLS
- modify the unit test , just in printing the differences but, consider that it is not a failure.

Katy

georgweiss:
Now I am a bit confused... the static methods used in the test class should not be in ```NLS.java``` but in ```NLSMessagesTest.java```

katysaintin:
Ah OK, I understand the invert, sorry, it's my bad. I will move all the code in NLSMessagesTest.java 
but we'll get there eventually

katysaintin:
It should be OK now, NLS.java is revert to the original version.
Only NLSMessagesTest.java changes, without considering differences in messages.properties as a failure.
Update all the messages_fr.properties according to messages.properties

georgweiss (approved):

georgweiss:
Interesting to see that with a change in Locale, ```SEVERE``` becomes ```SCHWERWIEGEND``` in German...

PR #3197, opened 2024-11-22
CSSTUDIO-2749 Add configuration option 'org.csstudio.display.builder.representation/svg_rendering_resolution_factor'

abrahamwolk:
This pull request adds the configuration option `org.csstudio.display.builder.representation/svg_rendering_resolution_factor` to Phoebus.

Setting this option to a value greater than `0.0` scales the resolution that SVGs are rendered at both horizontally and vertically by that factor, at the expense of significant memory consumption: the memory consumption of rendered SVGs scales as the square of `org.csstudio.display.builder.representation/svg_rendering_resolution_factor`: when `org.csstudio.display.builder.representation/svg_rendering_resolution_factor=1.9`, then the memory consumption will be (approximately) 3.61 times larger than the memory consumption when `org.csstudio.display.builder.representation/svg_rendering_resolution_factor=1.0`.

The motivation for the functionality is to enable sharper rendering of SVGs when zooming in.

By default,  `org.csstudio.display.builder.representation/svg_rendering_resolution_factor` is set to `1.0`.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/SVGHelper.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/SVGHelper.java
@@ -60,7 +60,7 @@ public class SVGHelper {
             svg_rendering_resolution_factor = Preferences.svg_rendering_resolution_factor;
         }
         else {
-            logger.log(Level.WARNING, "The option 'org.csstudio.display.builder.representation/svg_rendering_resolution_factor' is set to an invalid value. Setting svg_rendering_resolution_factor to 1.");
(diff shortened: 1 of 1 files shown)


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/SVGHelper.java
```diff
@@ -53,11 +54,20 @@ public class SVGHelper {
      */
     public static Image loadSVG(String imageFileName, double width, double height) {
         String cachedSVGFileName = imageFileName + "_" + width + "_" + height;
+
+        double svg_rendering_resolution_factor;
+        if (!Double.isNaN(Preferences.svg_rendering_resolution_factor) && Preferences.svg_rendering_resolution_factor > 0) {
```
kasemir, line 59: Would it make sense to enforce a minimum value of 1.0 instead of 0.0?

georgweiss:
ESS makes extensive (extensive!) use of SVGs. I would prefer the scaling factor to be applied only upon zoom actions.

PR #3201, opened 2024-11-25
Adding documentation on the alarm config editor

georgweiss:
Since users were asking for how disable alarm works...


kasemir:
I think the more convenient way is via the context menu in the alarm tree.
It allows you to enable/disable an item and all that's below within the hierarchy.

kasemir:
Yes on all

georgweiss:
OK, will add some content on the context menu options.

kasemir:
```
**NOTE 2:** An alarm configured as enabled cannot be disabled using the checkbox in the editor dialog. Instead, user needs to
 specify a disable-until date/time.
```

Is that true? Does the checkbox to enable/disable an alarm in the config no longer have any effect?
Are we now forced to always use disable-until-time?
That would be a bug.

shroffk:
I agree that this would be a bug.
If I remember correctly, "disable" using the check box (without the dialog) will disable the alarm until a default configured duration. 
This duration can be -1 for disable forever or 3600s (default disable for 1 hour)

If this is not already the case then I can add this functionality

georgweiss:
At this point I'm merely interested in the documentation to be able to fence off support questions. 
Agree that the UI is a bit confusing as it stands. I would assume that in most cases alarms are disabled from the context menu, and not from the config dialog.

georgweiss:
@shroffk, no need for you to step in, I can take care of the changes. 
We need to agree however on the expected behavior: I suggest using the checkbox in the dialog should do the same as the context menu item. And let's also agree that "forever" actually means: until enabled or XML file imported where item is enabled.

kasemir:
Yes, the checkbox in the dialog should enable/disable the channel exactly like the context menu item.
That's how it used to be before the temporary disablement, the disable-with-auto-renablement was added.
The temporary disablement option is only available via the dialog where one needs to enter a date/time.

georgweiss:
Will fix the checkbox issue.

PR #3206, opened 2024-12-06
Save and Restore: Add authentication for Elasticsearch.

mgonzal1:


shroffk:
looks good to me
Could we add some comments in the save and restore applications.properties which explains how to use an authenticated client to elastic.

shroffk (approved):

PR #3209, opened 2024-12-12
Fix alarm disable bug + add documentation

georgweiss:
Additional testing appreciated.


shroffk (approved):

shroffk:
Associated with PR https://github.com/ControlSystemStudio/phoebus/pull/3201

shroffk:
Behavior is as described in the issue ticket.

I do believe that we might have to work a bit to come up with a way to consistently implement the "disable with timeout" behaviour in a backward compatible way.

georgweiss:
>...to consistently implement the "disable with timeout" behaviour in a backward compatible way

Not sure what this means...

tynanford:
thanks for this fix!

PR #3211, opened 2024-12-12
Set latching to true by default

### LINKED ISSUE #3210: XML and json readers have Inconsistent latching assigned when undefined
According to https://control-system-studio.readthedocs.io/en/latest/app/alarm/ui/doc/index.html, alarms should latch by default. In https://github.com/pcdshub/pcds-nalms/issues/39, we noticed that if we did not set the latching value in an xml file, it would not latch by default. 

I can see latching set to true by default in
https://github.com/ControlSystemStudio/phoebus/blob/f54907903721eb1c7992b54c7d2680604ca934ec/app/alarm/model/src/main/java/org/phoebus/applications/alarm/model/json/JsonModelReader.java#L238
but it is set to false by default in 
https://github.com/ControlSystemStudio/phoebus/blob/f54907903721eb1c7992b54c7d2680604ca934ec/app/alarm/model/src/main/java/org/phoebus/applications/alarm/model/xml/XmlModelReader.java#L286

I think both of these should be true to be consistent with each other and match the documentation.

KaushikMalapati:
Meant to close #3210.

### DIFF
--- a/app/alarm/model/src/main/java/org/phoebus/applications/alarm/model/xml/XmlModelReader.java
+++ b/app/alarm/model/src/main/java/org/phoebus/applications/alarm/model/xml/XmlModelReader.java
@@ -283,7 +283,7 @@ public class XmlModelReader
         }
 
 
-        pv.setLatching(XMLUtil.getChildBoolean(node, TAG_LATCHING).orElse(false));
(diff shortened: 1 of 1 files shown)


shroffk:
I think these changes make sense.

kasemir (approved):
Looks good

PR #3216, opened 2024-12-18
Fix saveandrestore take-snapshot put api endpoint

### LINKED ISSUE #3217: Save and restore take-snapshot PUT not working (broken endpoint)
I am trying to access the save and restore API from a external python script. The `take-snapshot` endpoint seems to be a good way of taking a snapshot of a predefined configuration. There seems to be a typo in the code which makes this endpoint unusable.

I have made a pull request for the small fix. After that change I could take a snapshot and restore it.

Maybe the error would have been caught if there was also a test for the put methode of the API in `services/save-and-restore/src/test/java/org/phoebus/service/saveandrestore/web/controllers/TakeSnapshotControllerTest.java`.

JonasDevelops:
Closes #3217


georgweiss:
Thanks for fixing.

PR #3218, opened 2024-12-19
CSSTUDIO-2487 Add documentation when building the Phoebus package in 'phoebus-product' using maven

abrahamwolk:
This pull request adds the Phoebus documentation (if it has been built, by for example running `mvn -Dsphinx -Nverify` in the `phoebus` source directory) to the Phoebus package that is created when building `phoebus-product` using maven.

The intention is to facilitate the inclusion of the documentation when deploying the built Phoebus package.

If the documentation has not been built, then it is not included.

### DIFF
--- a/phoebus-product/src/assembly/package.xml
+++ b/phoebus-product/src/assembly/package.xml
@@ -27,6 +27,15 @@
       </includes>
     </fileSet>
     <fileSet>
+      <directory>${project.parent.basedir}/docs/build/html</directory>
+      <outputDirectory>/doc</outputDirectory>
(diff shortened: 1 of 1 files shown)


georgweiss (commented):
While this seems to do the right thing - i.e. to include out put of ```docs/build/html``` - I'm not sure if a ```mvn clean``` should delete the contents of ```docs/build```. 

@shroffk, can you share your input? This PR is about being able to include the output of docs build into the tar ball in ```phoebus-product/target```.

georgweiss (approved):

PR #3219, opened 2024-12-20
Switch to native HttpClient in Olog ES

georgweiss:
The notion of request filters supported in Jersey seems to be missing from the native APIs. In the case of the Olog client this means:

- Permissive host name settings is implemented through call to ```System.getProperties().setProperty()```. Unfortunately I do not have a setup to verify that this works.
- The debugging filter would be replaced by JVM options, see [jdk.httpclient.HttpClient.log](https://docs.oracle.com/en/java/javase/22/docs/api/java.net.http/module-summary.html). Sadly that does not provide the same functionality: request/response body not logged even at highest debug level settings.
- Authentication filter based on username/password from properties file I have removed as I (erroneously?) see it as legacy and not needed.


shroffk (approved):
Looks good

I think I might have to add some more flexible hostname verification

PR #3228, opened 2025-01-09
Fix issue in app-pvtable for PV that contains dot in the name the last part of the name is removed

katysaintin:
Fix issue in app-pvtable for PV that contains dot in the name (should happen with a custom datasource)
Replace last field only if it is a known EPICS fields .
All EPICS fields are in UpperCase andhas a maximum length of 4 characters ex : VAL EGU SCAN ...
Else .DESC is added to the name for example : MyPVName.Information will be MyPVName.Information.DESC for description
and MyPVName.Information.EGU will be MyPVName.Information.DESC for description too .

Improvement on PVTable , the description cell will display the DESC pv name in tooltip behind the description value.
DescriptionPvTable


app/pvtable/src/main/java/org/phoebus/applications/pvtable/model/PVTableItem.java
```diff
@@ -148,12 +152,23 @@ private void createPVs(final String name)
...
+            desc_name = name + DOT + DESC_FIELD;
+            if(!name.endsWith(DOT + DESC_FIELD) && name.contains(DOT)) {
+                // If name already includes a field
+                // It can be a EPICS fields such as .VAL .EGU ... 
+                // EPICS fields are always in Upper Case
+                // EPICS fields are 4 characters length max
+                final int sep = name.lastIndexOf('.');
+                String fieldVal = name.substring(sep + 1);
+                //System.out.println("fieldVal=" + fieldVal);
+                //Test if it in uppercase and max length 4
```
kasemir, line 165: EPICS field names are neither limited to 4 characters nor need they be uppercase.

In very early EPICS releases, field names were internally treated as 4-byte numbers, using a number comparison instead of a proper string comparison to locate fields because the number comp can be performed in a single machine code instruction, whereas string comparisons check character-by-character until you reach the end of the string. For a long time now, EPICS record field names are treated as strings without a 4 character limitation. Check the EPICS base include files, they use `char *` for field names, not `char[4]` or the like. See also motor record with its `IGSET` field, https://epics.anl.gov/bcda/synApps/motor/motorRecord.html

 If you create your own record, you can add fields with lower-case field names, there's no rule against that.

katysaintin: OK , thanks for your answer, but what can we do our custom datasource ?
Actually, my pv named muscade://PLC_1.AlwaysFALSE is searching for a PLC_1.DESC vor the description .. so it's not works. 
Can I add a test of datasource as it done for excluding "sim:" and "loc:" datasource from this automatically replacement ?

It's sound good for you if I add a test as  : 
```
if (name.startsWith("ca://") {
  // automatically replace the last field by DESC 
}
```
?

katysaintin: Probably it is not really clean to add a test on our custom datasource as it done for "sim:" and "loc:" actually.
What do you prefer between 

```
if (name.startsWith("muscade://") {
  // do my own treatment
}
```

or 

```
if (name.startsWith("ca://") {
  // keep the old treatment
}
```

or improve PV class in adding a method getDescription()

kasemir: "ca://" could be the default. So you need to do something like this:

```
// Maybe do a better check with regular expression, looking for `[a-z]+://` ?
boolean name_has_no_prefix = name.indexOf("://) < 0;

if (name.startsWith("ca://")   ||    (name_has_no_prefix && PVPool.default_type.equals("ca")) )
{
    // For Channel Access, try to create separate ".DESC" channel
}
```

katysaintin: Ok let I try to submit a new proposition for this fix.

katysaintin: I finish the modification using DisplayProvider interface .
1. First I get the pv type using   PVPool.TypedName.analyze(name) method (that supports the default datasource) in my case the default datasource is muscade
2. Test if the PV or the value is a DisplayProvider first to get the description from Display
3. If no description is defined by this way create a .DESC PV only for channel access => get rid of test on sim and loc

I also add a feature in PVTable to create a PVTable from a PV list copy in the clipboard and using a TextArea for the last row of PVTable.

app/pvtable/src/main/java/org/phoebus/applications/pvtable/model/PVTableItem.java
```diff
@@ -175,6 +196,24 @@ private void createPVs(final String name)
...
         }
     }
+    
+    private void updateDescription() {
+        if(desc_value == null) {
+            //update description from value or pv
+            VType currentValue = getValue();
+            if(currentValue != null) {
+                PV thePV = pv.get();
+                Display display =  thePV instanceof DisplayProvider ? ((DisplayProvider) thePV).getDisplay() : null;
```
kasemir, line 206: Not sure how this will work. A value (VType) can be a DisplayProvider. A PV will never be a DisplayProvider.

The following check of `currentValue instanceof DisplayProvider` makes sense, but `thePV instanceof DisplayProvider` should be removed

app/pvtable/src/main/java/org/phoebus/applications/pvtable/model/PVTableItem.java
kasemir: That approach sounds good!

app/pvtable/src/main/java/org/phoebus/applications/pvtable/model/PVTableItem.java
katysaintin: Please , it is the only way I found to manage EnumVType for Boolean value for my own datasource Muscade. EnumVType is not implemented DisplayProvider , and the description is only defined for Number .. while a DESC field is a common field for EPICS, so I don't know how to considere the description for bo or bi ? Have you got a suggestion for me to solve my problem. Furthermore this test is not impacting the functionnement. Thank you for the suggestion, else could you approve my pull request .
Thank you for your understanding.
Katy

kasemir: @shroffk:  Looks like the description was added to the wrong part of the VType and/or EPICS normative types. `DisplayProvider` started out with units and the min/max values for display, control, alarms. Those only apply to numeric data, not strings, enumerated, booleans. Then the description was added because PVAccess happens to provide it, but you're correct, in principle there's a description for bool/enum/string, not just numerics.

Alright, the PV Table cannot fix what's broken in channel access/pv access/normative types/vtypes. Still, maybe add a comment to the `thePV instanceof DisplayProvider`:
```
// DisplayProvider is an optional interface for VType values,
// not PVs, but the Muscade datasource happens to implement
// DisplayProvider for enum and bool PVs, so check for that here 
```

kasemir: Yes, a `DescriptionProvider` would be a better approach.
Agree that this would be a painful change that also extends into the normative types/PVAccess, because it has the same idea for "display" information.
Let's defer that by about 5 years.

katysaintin: Hello @kasemir and @shroffk ,

I'm totally agree with you about DescriptionProvider interface addition in core-pva .
I think it is the proper way to solve my issue.
And I also agree with that the impact on epics vtype clients could be problematic.

Thank you for your help, I'm OK with the comment , for clarifying the situation.
Have a good day and thank you again.

Katy

shroffk:
I opened a Issue ticket with epicsCoreJava
https://github.com/epics-base/epicsCoreJava/issues/141

@kasemir In the meanwhile, would it be OK if we added changes with the above comments.... maybe even add a TODO so that we know we should refactor this in the future based on the outcome of the above Issue?

kasemir:
Yes, I think this PR is fine when there's a comment on `thePV instanceof DisplayProvider`, explaining why that interface that's meant for VType data could appear here, something like
```
// DisplayProvider is an optional interface for VType values,
// not PVs, but the Muscade datasource happens to implement
// DisplayProvider for enum and bool PVs, so check for that here 
```

app/pvtable/src/main/java/org/phoebus/applications/pvtable/model/PVTableItem.java
katysaintin: > @shroffk: Looks like the description was added to the wrong part of the VType and/or EPICS normative types. `DisplayProvider` started out with units and the min/max values for display, control, alarms. Those only apply to numeric data, not strings, enumerated, booleans. Then the description was added because PVAccess happens to provide it, but you're correct, in principle there's a description for bool/enum/string, not just numerics.
> 
> Alright, the PV Table cannot fix what's broken in channel access/pv access/normative types/vtypes. Still, maybe add a comment to the `thePV instanceof DisplayProvider`:
> 
> ```
> // DisplayProvider is an optional interface for VType values,
> // not PVs, but the Muscade datasource happens to implement
> // DisplayProvider for enum and bool PVs, so check for that here 
> ```

@shroffk & @kasemir 
I add the comment you ve ask in previous comment.
Is ok for you , in waiting for VType's modifiations ?

Thank you again for your help :)

kasemir (approved):

Issue #3236, opened 2025-01-15
Olog : Intermittent connection


GDH-ISIS:
The latest build of the windows client (15/01/2025) appears to give very intermittent connection with the following message. On the rare occasion, it does connect.

![image](https://github.com/user-attachments/assets/3cca6461-51db-4e53-844f-d2147808fa15)

In the past, we have had a good connection to the olog service (on docker). Our Windows 10 client connected to the olog docker service without problems. I previous windows client install still does connect without issue. The olog web client still works.

GDH-ISIS:
Error log
2025-01-15 16:22:16 INFO [org.phoebus.logbook.olog.ui.LogEntryTableViewController] Single search: desc=*&start=12 hours&end=now
2025-01-15 16:22:16 WARNING [org.phoebus.olog.es.api.OlogHttpClient] failed to retrieve log entries
java.io.IOException: selector manager closed
	at java.net.http/jdk.internal.net.http.HttpClientImpl.send(Unknown Source)
	at java.net.http/jdk.internal.net.http.HttpClientFacade.send(Unknown Source)
	at org.phoebus.olog.es.api.OlogHttpClient.findLogs(OlogHttpClient.java:249)
	at org.phoebus.olog.es.api.OlogHttpClient.search(OlogHttpClient.java:317)
	at org.phoebus.logbook.olog.ui.LogbookSearchJob.lambda$getRunnable$0(LogbookSearchJob.java:59)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.io.IOException: selector manager closed
	at java.net.http/jdk.internal.net.http.HttpClientImpl$SelectorManager.selectorClosedException(Unknown Source)
	at java.net.http/jdk.internal.net.http.HttpClientImpl$SelectorManager.register(Unknown Source)
	at java.net.http/jdk.internal.net.http.HttpClientImpl.registerEvent(Unknown Source)
[...]

PR #3238, opened 2025-01-16
Fix Olog HttpClient timeout and URL encoding

georgweiss:
```URLEncoder.encode()``` converts space char to ```+```, which does not work if the string is a path variable.
Suggest to change the API for fetching attachments to use the (MongoDB) id instead of original file name, but this will require a change in the service.

All in all the native ```HttpClient``` is a bit of a challenge to work with... Or: the Jersey client makes a better job keeping the developer happy.

Also optimizes instantiation of LogEntryTable app to avoid multiple calls to create a OlogHttpClient.

See #3236


shroffk (approved):

shroffk:
I expected the `app.getClient()` to always return the same instance if the client had been instantiated along the lines of a singleton pattern
This is more explicit though

PR #3240, opened 2025-01-17
Reinstating all channel finder client methods

georgweiss:
@kasemir, would you mind having a go on this one to confirm it works with your client?


kasemir (approved):
Everything that was missing is back, looks good.

Since there's a `close()` call, maybe `ChannelFinderClient` should be based on `AutoCloseable` to enforce proper cleanup. That would be a change from the original API but seems worth it

georgweiss:
Feel free to hack away :-)

kasemir:
Well, made ChannelFinderClient AutoCloseable and it still builds without errors, so I suggest to go with that.

PR #3241, opened 2025-01-17
ant and eclipse settings

kasemir:
#3239 made core-formula depend on core-util.
This adapts ant and eclipse settings


shroffk:
sorry I forgot about this

kasemir:
Don't worry, it's easy to update when using ant, but hard to guess what's needed when you don't.

PR #3243, opened 2025-01-19
Update of the translation into French

XavSPM:
I propose various updates and additions for the French translation


shroffk:
@katysaintin can you review this PR please :)
My french is "pas si bon"

katysaintin:
Ok @shroffk it's seems to be OK. I also check the synchronization between message.properties it is OK too.

PR #3245, opened 2025-01-28
CSSTUDIO-2939 Add property "SVG Rendering Resolution Factor" to the Picture and Symbol widgets

abrahamwolk:
This pull request adds the widget property "SVG Rendering Resolution Factor" to the Picture and Symbol widgets:

screenshot

The functionality enables a sharper image when zooming in, at the expense of a quadratic increase in memory consumption. E.g., by setting the SVG rendering resolution factor to `2.0`, an SVG can be rendered at twice the width and twice the height, but the memory required to represent the image grows by a factor of four.

This functionality is similar to the functionality implemented in https://github.com/ControlSystemStudio/phoebus/pull/3197, except that the rendering resolution factor is now set per-widget.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Preferences.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Preferences.java
@@ -37,6 +37,9 @@ public class Preferences
     @Preference(name="macros") private static String macro_spec;
     /** Preference setting */
     @Preference public static boolean enable_saved_on_comments;
+    /** Preference setting */
+    @Preference public static boolean enable_svg_rendering_resolution_factor;
+
     private static Macros macros;
 
     static
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/PictureWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/PictureWidget.java
@@ -15,6 +15,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.csstudio.display.builder.model.Messages;
+import org.csstudio.display.builder.model.Preferences;
 import org.csstudio.display.builder.model.Version;
 import org.csstudio.display.builder.model.Widget;
 import org.csstudio.display.builder.model.WidgetCategory;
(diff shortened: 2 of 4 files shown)


kasemir:
This is all quite unfortunate. The S in SVG stands for Scalable. Draw once, view in any size.
I guess there was no obvious way to directly display SVG in a JavaFX widget, so Batik was dragged in to convert SVG into a pixel image. Now the end user is expected to configure some internal scaling factor of that conversion? On each picture and symbol that uses SVG?!
Can't this be hidden? As suggested in #3197,  make that a preference that the end user doesn't see, and only apply it when zoomed? Or automatically determine this scaling factor based on the zoom factor?
Or maybe reconsider how SVG is handled. Can https://docs.oracle.com/javase/8/javafx/api/javafx/scene/shape/SVGPath.html be used to directly render SVG?

kasemir:
> complicated

Yes.
If we could at least hide that property unless the image is an SVG, but we don't have an easy way to show/hide properties in the editor, so that would be its own side project.

Since this only applies to SVG, should we hack this into the image file name?
If your image is `/path/to/image.svg`, it's rendered with a scaling factor of 1, but you can enter an image name of `/path/to/image.svg[x2.5]` to get a scaling of 2.5? Yes, the file dialog wouldn't understand that, so whenever you use the file dialog to select an image, you then need to manually enter the "...[x2.5]" to the file name.

How big is the actual problem? What do your screens look like when you don't have that scaling factor? How about simply not using SVG in the displays? Batch-convert all your SVGs into PNGs of sufficient size (on Linux, use ImageMagic) and then use those in the display.

georgweiss (commented):
Due to the memory consumption impact I'd propose to add some kind of information to alert the user of the consequences of boosting the scaling factor. Maybe a tooltip, or an addition to one of the example OPIs.

kasemir:
> Would it be OK for you if the widget property is hidden in the UI by default, but can be enabled by setting a special option in phoebus.ini?

I like that.

georgweiss (commented):
What if a display file is created with the scaling factor and then opened in Phoebus where the property is disabled?

georgweiss (approved):

Issue #3226, opened 2025-01-08
logbook/olog/ui duplicate dependency


jodreyer:
In the pom.xml file of app/logbook/olog/ui the definition of the core-ui dependency is duplicated.
This results in a warning of maven about future incompatibilities.<br>
Removing one of the two entries resolves this warning.<br>
Observed with maven 3.9.9 and Java 17.0.13 from openjdk on Linux.

jodreyer:
The Entry in lines 18-22 has to be removed. Removing the one in lines 43-47 results in a failing build.

georgweiss:
Sure, the duplicated dependency should be removed. However, the build should not fail. Do you have a console output to share here?

katysaintin:
@lcaouen , it perhaps the same problem for us. Could you check that please ?

shroffk:
hmm... the build being sensitive to only one of the dependencies is interesting.

I have pushed a cleaned up version of the pom.xml @jodreyer can you test it out.

Also,
Intellij sometimes fails to properly resolve dependencies and then tried to be helpful and add the maven dependency again creating 2 copies of it... that could have been the cause of this.

jodreyer:
@shroffk I tested the core-ui-dep branch and it builds correctly on my system.

PR #3254, opened 2025-02-03
Fix olog ui dependency #3226

shroffk:


shroffk:
@jodreyer tested and verified this

georgweiss (approved):

Issue #3263, opened 2025-02-05
Horizontal and Vertical profiles from Images


shroffk:
In CS Studio the intensity plot allowed us to export the x and y profiles from the intensity plot. While this is not available in Phoebus I have tried to add a few formula functions to allow us the recreate this functionality

![Image](https://github.com/user-attachments/assets/0ebf6249-bc7a-40a3-8a79-a6c5a9929209)

shroffk:
I created a set of formula functions

2 which get the horizontal and vertical profile from a VImage
2 which get the horizontal and vertical profile from array data ( this is the image data coming from older IOCs which are not using pva)

PR #3264, opened 2025-02-05
Image profiles

shroffk:
#3263


kasemir (approved):
Very good idea! Much preferred over adding options to the image widget, and also more flexible

PR #3271, opened 2025-02-11
Consistent Alarm naming convention between Alarm Table and Alarm Log Table

jeonghanlee:
@shroffk and @kasemir 

To catch up our previous PR https://github.com/ControlSystemStudio/phoebus/pull/3242
Here is the new proposal to keep the consistency naming convention between Alarm Server and Alarm Logger services.

We introduced the "middle" layer to translate their inconsistency with elasticsearch index compatibilities. 

Here is the table which we've developed so far. 

Name | Proposed Name | Area
-- | -- | --
PV | PV | Alarm Table
Description | Description | Alarm Table
PV Severity | PV Severity | Alarm Table
PV Status | PV Message | Alarm Table
Alarm Severity | Alarm Severity | Alarm Table
Alarm Status | Alarm Message | Alarm Table
Alarm Time | Alarm Time | Alarm Table
Alarm Value | Alarm Value | Alarm Table
  |   |  
Config | Config | Alarm Log Table
PV | PV | Alarm Log Table
Severity | Alarm Severity | Alarm Log Table
Message | Alarm Message | Alarm Log Table
Time | Alarm Time | Alarm Log Table
Message Time | Alarm Log Time | Alarm Log Table
Current Severtiy | PV Severity | Alarm Log Table
Current Message | PV Message | Alarm Log Table
User | User | Alarm Log Table
Host | Host | Alarm Log Table
Value | Alarm Value |  
Time Delta | Time Delta | Alarm Log Table
Command | Command | Alarm Log Table
  |   |  
PV | PV | Alarm Log Table Search
Severity | Alarm Severity | Alarm Log Table Search
Message | Alarm Message | Alarm Log Table Search
Current Severity | PV Severity | Alarm Log Table Search
Current Message | PV Message | Alarm Log Table Search
User | User | Alarm Log Table Search
Time | Alarm Log Time | Alarm Log Table Search
Start Time | Start Time | Alarm Log Table Search
End Time | End Time | Alarm Log Table Search
Command | Command | Alarm Log Table Search

Here is the screen capture on Phoebus pages.

pas1
pas2
pas3

We changed the Alarm Log log output to `WARNING` since they printed too many query messages in every 10 secs. And we also changed the default `8080` to `9000` to relieve `tomcat 8080` port.

Please let me know what you think.

@shroffk we also have to work together to update the alarm documentation to reflect these changes. 

Thanks,


georgweiss:
While I realize that some deployment scenarios need to define a port different than 8080, I'd prefer to keep it as-is since 8080 is "standard" for Spring Boot applications. An alternative port can be specified on command line, e.g.
```-Dserver.port=9000```

jeonghanlee:
@georgweiss  Yes, you are right. We can customize anything we want during compilation and runtime. However, from my point of view, why not define our own default for all alarm services in general, and it is time to define its service port more clearly. 8080` is too common for Tomcat, and `save-and-restore` also uses `8080`.

georgweiss:
@jeonghanlee, if we set up a scheme for custom ports for different services, the developer(s) of any new service added to the project will need to be aware of this scheme, and also be able to look up documentation around how such a scheme works.

I therefore recommend sticking to conventions.

jeonghanlee:
@georgweiss I accepted to convert it to 8080, not because your logic is correct but because I am tired. 

I will commit later, so ignore the port number. And please review what implementation is truly.

jeonghanlee:
@kasemir, @shroffk, and @georgweiss I removed all unnecessary changes. And please focus on what truly implementation is.

jeonghanlee:
> @jeonghanlee, if we set up a scheme for custom ports for different services, the developer(s) of any new service added to the project will need to be aware of this scheme, and also be able to look up documentation around how such a scheme works.
> 
> I therefore recommend sticking to conventions.

Where do I find the conventions which you mentioned? Is this your convention, or any community conventions to cover all phoebus and related middle ware services? Do I miss something?

georgweiss:
@jeonghanlee, "convention" may be not be the correct word here. What I'm referring to is the default port 8080 used by Spring Boot services.

Deployment scenarios of the Phoebus middleware will vary between sites. I think it is safe to assume that each site has its own tool and custom scheme of configuring the services in terms of port or even default context root.

jeonghanlee:
I reverted my changes, so we can focus the subject which I would like to discuss to make the consistent naming among the Phoebus Alarm System.

georgweiss (commented):
This looks OK to me save for one thing: 
Changes to the ```org.phoebus.applications.alarm/alarm_table_columns``` preference will be an issue for someone who does not know that the column names have changed. As there is code to map between old and new strings, could not the same be done for this preference? For instance, if ```PV Status``` is encountered in the preference, then it could be mapped to ```PV Message```.

georgweiss:
My concern is that if a user lists ```PV Status``` in a local settings file, that column will not show up. User may not be aware of the change, nor the documentation. That's why I propose some mapping code to handle that.

kasemir:
The changed column name is unfortunate for the `org.phoebus.applications.alarm/alarm_table_columns` setting.
Maybe log an error when "Status" is found?
```SEVERE ... "alarm_table_columns entries with 'Status' must now use 'Message'"```

Users will notice pretty soon when the previous PV Status column is missing, and the error message in the log will then explain what to do?

jeonghanlee:
> My concern is that if a user lists `PV Status` in a local settings file, that column will not show up. User may not be aware of the change, nor the documentation. That's why I propose some mapping code to handle that.

Based on my understanding, "user setting" is not transparent to its sources, that was my own struggle to implement this. There are too many "hard-code" names in preference file, and sources codes. Literally, it is not good. 

I would like to let users to understand what they uses clearly. That is the reason why I will work on the documentation instead of "hiding" changes within codes. It also is much easy to change their local configuration file (one line).

jeonghanlee:
@shroffk and @kasemir do you have any comments on this PR? Or do you want to meet us to discuss further?

shroffk:
So, 
As Han mentioned our goal is to standardize some of the names across the various alarm apps and services and at the same time update a few to make them more understandable.

I agree that the incompatibility with the existing preferences is unfortunate.
However, we need to come up with a process of supporting changes ( in some cases ones which are not fully backward compatible )

In this case I think I like Kay's suggestion of adding a clear error log message with the suggested solution itself is a good compromise. ( We should remember to remove that too... since we can't fill up our code based with compatibility warning :) )

I am fine with merging this with Kay's recommendation before the 5 release or we can do this after the 5 release.

georgweiss (approved):

georgweiss:
I do not object to a merge

jeonghanlee:
Great! Now we can move the forward to finalize our production version. Thank you again.

PR #3273, opened 2025-02-12
update to new sonatype central

shroffk:


shroffk:
I think we need to merge this now and start trying the release with the new sonotype plugin and publishing repo

kasemir (approved):
Sigh, nexus-staging-maven-plugin only made it to version 1.6.8 before it was replaced by central-publishing-maven-plugin version 0.7.0.

shroffk:
we can look into publishing on github at some point... in the meanwhile the goal is to ensure that other than the 15 lines in the main pom.xml the rest of the project remains quite insulated from this publishing platform details.

PR #3286, opened 2025-02-21
CSSTUDIO-3017 Fix three issues with configuration window in Alarm Tree application

abrahamwolk:
This pull request fixes three issues related to the configuration window for PVs in the Alarm Tree application in Phoebus:

1. When pressing the key "enter" in the date picker for the option "Disable until" in the configuration window for PVs, the entered value is currently not committed. I.e., if a point in time is entered for automatic re-enabling is entered and the user presses "enter", the PV is disabled _without the automatic re-enabling_; only when the user clicks the "OK" button the entered value is committed. This pull requests adds an event handler that commits the entered value when pressing enter if it is well-formed and doesn't commit it otherwise. (See the commit https://github.com/ControlSystemStudio/phoebus/commit/068bb1d9cb5ef2766f615fc56044d7f117f1cd4b)
screenshot

3. When the point in time that is entered for the option "disable until" lies in the past, currently the PV is disabled _without automatic re-enabling_. This pull request changes this so that if the point in time lies in the past _or is the present moment_, then an alert is displayed and the input is not accepted. (See the commit https://github.com/ControlSystemStudio/phoebus/commit/8be15905c65c7df0d2f232ed9eb5b1ec5d7aedca)

4. When double-clicking on leaf (representing a PV) in the Alarm Tree application, the configuration window opens for the PV in question. On the other hand, when one double clicks on an internal node of the tree (representing alarm components), then a RuntimeException is thrown and printed to the console. This pull request adds a check for whether a leaf or a node was clicked on, and only opens the configuration window when a leaf was clicked on. If a node was clicked on, the configuration window is not opened. (See the commit https://github.com/ControlSystemStudio/phoebus/commit/965b4867b124994742ea492e2ca2e3c69df7e819)

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/ItemConfigDialog.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/ItemConfigDialog.java
@@ -40,6 +40,7 @@ import org.phoebus.applications.alarm.client.AlarmClientLeaf;
 import org.phoebus.applications.alarm.client.AlarmClientNode;
 import org.phoebus.applications.alarm.model.AlarmTreeItem;
 import org.phoebus.applications.alarm.ui.tree.datetimepicker.DateTimePicker;
+import org.phoebus.ui.dialog.DialogHelper;
 import org.phoebus.ui.dialog.ExceptionDetailsErrorDialog;
 import org.phoebus.util.time.SecondsParser;
 import org.phoebus.util.time.TimeParser;
@@ -336,6 +337,7 @@ class ItemConfigDialog extends Dialog<Boolean> {
                 prompt.setTitle("'Disable until' is set to a point in time in the past");
                 prompt.setHeaderText("'Disable until' is set to a point in time in the past");
                 prompt.setContentText("The option 'disable until' must be set to a point in time in the future.");
+                DialogHelper.positionDialog(prompt, enabled_date_picker, 0, 0);
                 prompt.showAndWait();
 
                 event.consume();


kasemir:
On point 3:
Not crashing for double-clicks on intermediate nodes is of course better than crashing. But can you figure out what happened there, and bring it back to opening the config for either one? 
A double-click on any item in the alarm tree used to open the config dialog. The dialog looks different for an intermediate node vs. a PV node, but in any case you got the dialog.

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/ItemConfigDialog.java
```diff
@@ -287,29 +313,43 @@ private void validateAndStore(final AlarmClient model, final AlarmTreeItem<?> it
...
+                    final LocalDateTime update_date = LocalDateTime.now().plus(amount);
+                    validEnableDate = pv.setEnabledDate(update_date);
+                } else {
+                    pv.setEnabled(itemEnabled.get());
+                    validEnableDate = true;
+                }
+            }
+
+            if (!validEnableDate) {
+                Alert prompt = new Alert(Alert.AlertType.INFORMATION);
```
kasemir, line 335: Opening an alert while already in a dialog isn't ideal.
With more than one monitor, we might not know where the alert pops up.
With dialogs, we tend to use the dialog helper to position the dialog close to the button etc. which triggered it.
.. still, this is likely good enough and not worth a lot of extra code

kasemir (approved):

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/ItemConfigDialog.java
georgweiss: Frequently used pattern: decorate the widget holding malformed data, e.g. red border plus error message in the input field.
But that is more coding for sure...

georgweiss (approved):

PR #3291, opened 2025-02-25
Create save-and-restore-docker-image.yml

georgweiss:


shroffk (approved):

shroffk:
The PR is great.

I am wondering if there is a way to trigger this workflow only when changes are made to `services/save-and-restore` or `app/save-restore`

georgweiss:
@shroffk, I will investigate if that is possible.

PR #3303, opened 2025-03-01
Updated error handling for alarm AudioAnnunciator plugin

tynanford:
We noticed with the AudioAnnunciator that eventually the annunicator stops playing all sounds (including the "Test" button). Closing and re-opening the application fixed the issue. This was the error message:

```
Exception in thread "Annunciator" java.lang.NullPointerException: Cannot invoke "org.phoebus.applications.alarm.model.SeverityLevel.ordinal()" because "message.severity" is null
        at org.phoebus.applications.alarm.audio.annunciator.AudioAnnunciator.speak(AudioAnnunciator.java:65)
        at org.phoebus.applications.alarm.ui.annunciator.AnnunciatorController.lambda$processMessages$2(AnnunciatorController.java:163)
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
        at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
        at org.phoebus.applications.alarm.ui.annunciator.AnnunciatorController.processMessages(AnnunciatorController.java:162)
        at java.base/java.lang.Thread.run(Thread.java:840)
```

I'm not sure yet why the severity was null but this PR adds null checks and some logging so the application doesn't crash on error

### DIFF
--- a/app/alarm/audio-annunciator/src/main/java/org/phoebus/applications/alarm/audio/annunciator/AudioAnnunciator.java
+++ b/app/alarm/audio-annunciator/src/main/java/org/phoebus/applications/alarm/audio/annunciator/AudioAnnunciator.java
@@ -15,6 +15,8 @@ import org.phoebus.applications.alarm.ui.annunciator.Annunciator;
 import org.phoebus.applications.alarm.ui.annunciator.AnnunciatorMessage;
 
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Annunciator class. Uses Audio files to annunciate passed messages.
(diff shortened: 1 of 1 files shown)


shroffk:
That is very strange...a null check should not hurt though

kasemir:
Might be a good idea to log the `AnnunciatorMessage message`.
That way, you see what type of message contains the `null`.
Maybe it's not a normal alarm message but "Alarm Server Disconnected" type of information.

tynanford:
That is a good idea, just submitted another PR for that

tynanford:
To follow up on this. Here is a null severity message with the new logging:

```
2025-03-06 10:15:27 WARNING [org.phoebus.applications.alarm.audio.annunciator.AudioAnnunciator] Received AnnunciatorMessage with null severity: 2025-03-06 10:15:27.799 null There are 4 new messages. Playing default alarm sound
```

So this is from the "flurry" message

kasemir:
Ah! Well, looks like a null severity needs to be expected, not an error

tynanford:
Yes... I guess playing the default sound makes most sense. Or adding another audio option for flurry alerts

PR #3305, opened 2025-03-03
Make number of steps in loop command predictable

DenSinH:
In the previous situation, the loop command was executed as 
```java
for (double value = start; value <= end; value += step)
```
which made the number of steps in the loop highly unpredictable due to finite precision errors. It seems that there was left some possibility to change the loop step and end "on the fly", though I don't see how the values could ever change given a `LoopCommand` instance. 

Making the number of steps in the loop command easily computable would make it a lot easier to make 2D plots from scans executed with the scan server, as the image dimensions can be computed ahead of time (without using an O(n) algorithm, i.e. just doing the for loop).

Perhaps I missed something in why/how the loop bounds can be changed on the fly, unless one somehow inherits and re-implements `LoopCommand` with some overrides, but this seems like a strange edge case to take into account.

Perhaps you prefer a floor or round on the `getNumSteps`, I just figured the closest behavior to the "old" implementation would be this, because of the <= comparison.

### DIFF
--- a/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
+++ b/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
@@ -138,8 +138,9 @@ public class LoopCommandImpl extends ScanCommandImpl<LoopCommand>
     {
         final SimulatedDevice device = context.getDevice(context.getMacros().resolveMacros(command.getDeviceName()));
 
-        final double start = getLoopStart();
         final double step  = getLoopStep();
+        // step is < 0 means we are stepping down
+        double start = step < 0 ? getLoopEnd() : getLoopStart();
         int num_steps = getNumSteps();
         for (int i = 0; i < num_steps; i++)
             simulateStep(context, device, start + i * step);
--- a/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
+++ b/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
@@ -84,7 +84,7 @@ public class LoopCommandImpl extends ScanCommandImpl<LoopCommand>
     @Override
     public long getWorkUnits()
     {
-        final long iterations = 1 + Math.round(Math.abs((command.getEnd() - command.getStart()) / command.getStepSize()));
+        final long iterations = getNumSteps();
         long body_units = 0;
         for (ScanCommandImpl<?> command : implementation)
             body_units += command.getWorkUnits();
@@ -129,7 +129,7 @@ public class LoopCommandImpl extends ScanCommandImpl<LoopCommand>
     }
 
     public int getNumSteps() {
-        return (int)Math.ceil(Math.abs(((getLoopEnd() - getLoopStart()) / getLoopStep()))) + 1;
+        return (int)Math.ceil(Math.abs(((command.getEnd() - command.getStart()) / command.getStepSize()))) + 1;
     }
 
     /** {@inheritDoc} */
--- a/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
+++ b/services/scan-server/src/main/java/org/csstudio/scan/server/command/LoopCommandImpl.java
@@ -204,9 +204,11 @@ public class LoopCommandImpl extends ScanCommandImpl<LoopCommand>
         else
             condition = null;
 
-        double start = getLoopStart();
         double step  = getLoopStep();
+        // step is < 0 means we are stepping down
+        double start = step < 0 ? getLoopEnd() : getLoopStart();
         int num_steps = getNumSteps();
+
         for (int i = 0; i < num_steps; i++)
             executeStep(context, device, condition, readback, start + i * step);
     }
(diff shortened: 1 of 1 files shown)


DenSinH:
Wouldn't you just be able to abort the scan? I believe this still saves any data that was obtained before aborting

kasemir:
Maybe you don't want to abort. You ramped some temperature over a wide range, which takes hours. You notice that the range covered so far includes the signal of interest, so you want to end the loop and otherwise move on, not abort.

DenSinH:
Another option might be to do some sort of "dual" approach, where we calculate the number of steps predictably, but deviate if the step size / loop end changed during iteration.

kasemir:
More important than the somewhat hidden feature of being able to modify the loop in the editor is the "alternating scan".
You always loop start to end, but for step<0 we want to go end to start.

See https://controlssoftware.sns.ornl.gov/css_pyscanclient/html/commands.html#loop and note the `reverse` flag in the loop code.

kasemir:
That should work. Please test it with the example from the doc,
```
Loop('xpos', 0, 5, 1, [ Loop('ypos', 0, 5, -1 ])
```

DenSinH:
Not sure how to launch the Jython interpreter, but creating and submitting this scan:
```xml
<commands>
  <loop>
    <device>loc://xpos</device>
    <start>0.0</start>
    <end>5.0</end>
    <step>1.0</step>
    <tolerance>0.1</tolerance>
    <body>
      <loop>
        <device>loc://ypos</device>
        <start>0.0</start>
        <end>5.0</end>
        <step>-1.0</step>
        <tolerance>0.1</tolerance>
        <body>
          <log>
            <devices>
              <device>loc://xpos</device>
              <device>loc://ypos</device>
            </devices>
          </log>
        </body>
      </loop>
    </body>
  </loop>
</commands>
```
yielded this data with the "old" implementation:
```csv
# Data for scan ID 1635
Time	loc://xpos	loc://ypos
2025-03-03 16:50:06.089	0.0	5.0
2025-03-03 16:50:06.109	0.0	4.0
2025-03-03 16:50:06.112	0.0	3.0
2025-03-03 16:50:06.114	0.0	2.0
2025-03-03 16:50:06.116	0.0	1.0
2025-03-03 16:50:06.120	0.0	0.0
2025-03-03 16:50:06.123	1.0	0.0
2025-03-03 16:50:06.126	1.0	1.0
2025-03-03 16:50:06.130	1.0	2.0
2025-03-03 16:50:06.132	1.0	3.0
2025-03-03 16:50:06.135	1.0	4.0
2025-03-03 16:50:06.136	1.0	5.0
2025-03-03 16:50:06.139	2.0	5.0
2025-03-03 16:50:06.142	2.0	4.0
2025-03-03 16:50:06.144	2.0	3.0
2025-03-03 16:50:06.145	2.0	2.0
2025-03-03 16:50:06.147	2.0	1.0
2025-03-03 16:50:06.149	2.0	0.0
2025-03-03 16:50:06.151	3.0	0.0
2025-03-03 16:50:06.154	3.0	1.0
2025-03-03 16:50:06.156	3.0	2.0
2025-03-03 16:50:06.158	3.0	3.0
2025-03-03 16:50:06.159	3.0	4.0
2025-03-03 16:50:06.161	3.0	5.0
2025-03-03 16:50:06.163	4.0	5.0
2025-03-03 16:50:06.165	4.0	4.0
2025-03-03 16:50:06.167	4.0	3.0
2025-03-03 16:50:06.168	4.0	2.0
2025-03-03 16:50:06.170	4.0	1.0
2025-03-03 16:50:06.172	4.0	0.0
2025-03-03 16:50:06.174	5.0	0.0
2025-03-03 16:50:06.176	5.0	1.0
2025-03-03 16:50:06.178	5.0	2.0
2025-03-03 16:50:06.179	5.0	3.0
2025-03-03 16:50:06.181	5.0	4.0
2025-03-03 16:50:06.182	5.0	5.0
```
and this with the "new" implementation:
```
Time	loc://xpos	loc://ypos
2025-03-04 08:47:59.836	0.0	5.0
2025-03-04 08:47:59.845	0.0	4.0
2025-03-04 08:47:59.847	0.0	3.0
2025-03-04 08:47:59.850	0.0	2.0
2025-03-04 08:47:59.852	0.0	1.0
2025-03-04 08:47:59.858	1.0	5.0
2025-03-04 08:47:59.861	1.0	4.0
2025-03-04 08:47:59.863	1.0	3.0
2025-03-04 08:47:59.864	1.0	2.0
2025-03-04 08:47:59.866	1.0	1.0
2025-03-04 08:47:59.869	2.0	5.0
2025-03-04 08:47:59.871	2.0	4.0
2025-03-04 08:47:59.873	2.0	3.0
2025-03-04 08:47:59.875	2.0	2.0
2025-03-04 08:47:59.877	2.0	1.0
2025-03-04 08:47:59.880	3.0	5.0
2025-03-04 08:47:59.882	3.0	4.0
2025-03-04 08:47:59.885	3.0	3.0
2025-03-04 08:47:59.886	3.0	2.0
2025-03-04 08:47:59.888	3.0	1.0
2025-03-04 08:47:59.890	4.0	5.0
2025-03-04 08:47:59.894	4.0	4.0
2025-03-04 08:47:59.896	4.0	3.0
2025-03-04 08:47:59.897	4.0	2.0
2025-03-04 08:47:59.899	4.0	1.0

```
(I broke something in my build so I had to come back to it this morning). I guess that we will always need to do an extra step.

kasemir:
Thanks for testing that loop example! The plot from the documentation is a power point drawing, might well have been wrong about the initial direction of the inner loop. Still, the end result is now the same as before, so that's good.

Issue #3308, opened 2025-03-04
core/pv-mqtt: MQTT client can't reconnect after String to Double conversion parsing error


flef:
Hello,

I noticed while using PV MQTT that there are issues with the MQTT client disconnecting and then being unable to reconnect. This particularly happens when a PV has a String value but is defined as a VDouble (either by default or explicitly). This causes a type conversion error.

Upon investigating the code, I observed that the AtomicBoolean responsible for monitoring the MQTT client's connection status is out of sync with the actual client connection. After a conversion error, the MQTT client gets disconnected, but the AtomicBoolean remains in the connected state. Therefore, to reconnect the MQTT client, the only solution is to close and reopen Phoebus.

see: https://github.com/ControlSystemStudio/phoebus/blob/master/core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PVConn.java#L37

I am proposing a corrective patch in an upcoming Pull Request.

Best regards,
Florian.

PR #3309, opened 2025-03-04
fix: Enhance MQTT client connection status check.

flef:
Regarding issue core/pv-mqtt: MQTT client can't reconnect after String to Double conversion parsing error #3308, here is the patch to improve the MQTT client's connection status check.

### DIFF
--- a/core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PV.java
+++ b/core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PV.java
@@ -205,9 +205,8 @@ public class MQTT_PV extends PV
         }
         catch (Exception ex)
         {
-            logger.log(Level.SEVERE, "Could not parse message: '" + new_value + "' to " + getName());
-
-            throw new Exception("Failed to parse message", ex);
+            notifyListenersOfDisconnect();
+            logger.log(Level.WARNING, "Could not parse message: '" + new_value + "' to " + getName());
         }
     }
 
--- a/core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PV.java
+++ b/core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PV.java
@@ -177,8 +177,7 @@ public class MQTT_PV extends PV
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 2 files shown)


core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PV.java
```diff
@@ -205,9 +205,8 @@ public void messageArrived(String topic, MqttMessage msg) throws Exception
         }
         catch (Exception ex)
         {
-            logger.log(Level.SEVERE, "Could not parse message: '" + new_value + "' to " + getName());
-
-            throw new Exception("Failed to parse message", ex);
+            notifyListenersOfDisconnect();
+            logger.log(Level.WARNING, "Could not parse message: '" + new_value + "' to " + getName());
```
kasemir, line 209: Only logging the issue, not throwing an exception, may be the better approach here.
But since the original `Exception ex` is no longer wrapped and passed up, add it to the log so we see the underlying cause in the log:
```
logger.log(Level.WARNING, "Could not parse message: '" + new_value + "' to " + getName(), ex);
```

core/pv-mqtt/src/main/java/org/phoebus/pv/mqtt/MQTT_PVConn.java
```diff
@@ -158,16 +155,15 @@ public void publishTopic(String topicStr, String pubMsg, int pubQoS, boolean ret
         }
     }
 
-    private void disconnect()
+    private synchronized void disconnect()
```
kasemir, line 158: Back in 78a698b3a7c86c81b90f88698ef5ef21e66f4d9f we removed synchronized sections for some reason and tried to handle it with atomic flags.

Usage of MQTT seems limited, so if the sync. causes deadlock, it might take some time to get into that situation. Maybe it's fine?

flef: I did not experience any deadlock or significant latency after several days of testing, even when irregularly disconnecting the MQTT broker. The MQTT client reconnects properly.

kasemir: Thanks for testing it a few days!

PR #3310, opened 2025-03-05
CSSTUDIO-2964 Alarm Tree application: Add indicators for the existence of disabled leaves

abrahamwolk:
This pull request adds indicators for the existence of disabled leaves to the Alarm Tree application.

The following screenshots demonstrate the implemented functionality. The idea is that one should not have to expand the alarm tree in order to see that there exist disabled alarms in an alarm component. The indicators are currently always in grey text. The existence of a timer is indicated.

The existence of at least one disabled alarm is indicated as follows:
screenshot1
The same alarm tree expanded:
screenshot2
The existence of a timer is indicated as follows:
screenshot3
When all leaves of "Alarm Component 1" are disabled, this is indicated as follows ("Alarm Component 1.3" contains no leaves, and hence it is not greyed out):
screenshot4

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
@@ -111,17 +111,7 @@ class AlarmTreeViewCell extends TreeCell<AlarmTreeItem<?>>
                 } else {
                     if (leaf.getEnabled().enabled_date != null) {
                         LocalDateTime enabledDate = leaf.getEnabled().enabled_date;
-                        String enabledDateString;
-
-                        LocalDateTime now = LocalDateTime.now();
-                        if (!(now.getDayOfMonth() == enabledDate.getDayOfMonth() &&
-                                now.getMonthValue() == enabledDate.getMonthValue() &&
-                                now.getYear() == enabledDate.getYear())) {
-                            enabledDateString = DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(enabledDate);
-                        }
-                        else {
-                            enabledDateString = DateTimeFormatter.ISO_LOCAL_TIME.format(enabledDate);
-                        }
+                        String enabledDateString = DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(enabledDate);
                         disabledTimerIndicator.setText("(" + Messages.disabledUntil + " " + enabledDateString + ")");
                     } else {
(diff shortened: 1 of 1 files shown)


app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
```diff
@@ -97,7 +111,20 @@ protected void updateItem(final AlarmTreeItem<?> item, final boolean empty)
                 }
                 else
                 {
-                    text.append(" (disabled)");
+                    if (leaf.getEnabled().enabled_date != null) {
+                        LocalDateTime enabledDate = leaf.getEnabled().enabled_date;
+                        String stringToAppend = padWithLeadingZero(enabledDate.getHour()) + ":" + padWithLeadingZero(enabledDate.getMinute()) + ":" + padWithLeadingZero(enabledDate.getSecond());
+
```
georgweiss, line 114: Would not ```SimpleDateFormat``` be a simpler approach here?

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
```diff
@@ -97,7 +111,20 @@ protected void updateItem(final AlarmTreeItem<?> item, final boolean empty)
...
+                        LocalDateTime now = LocalDateTime.now();
+                        if (!(now.getDayOfMonth() == enabledDate.getDayOfMonth() &&
+                              now.getMonthValue() == enabledDate.getMonthValue() &&
+                              now.getYear() == enabledDate.getYear())) {
+                              String paddedMonthNumber = padWithLeadingZero(enabledDate.getMonthValue());
+                              String paddedDayNumber = padWithLeadingZero(enabledDate.getDayOfMonth());
+                            stringToAppend = enabledDate.getYear() + "-" + paddedMonthNumber + "-" + paddedDayNumber + "T" + stringToAppend;
+                        }
```
georgweiss, line 122: Same here: native Date/Time APIs could be a simpler option.

georgweiss:
Do not see the need to differentiate formatting of "disabled until" date/time based on whether its the current date or not, see screen shot. If someone is looking at the tree around midnight, what happens then when a future date becomes current date?

Screenshot 2025-03-05 at 13 35 10

georgweiss:
I'd claim that differences in how date/time is rendered may also be define as cognitive load, in particular as after midnight the date/time is not re-rendered and potentially adds confusion.

kasemir (approved):

georgweiss (approved):

PR #3312, opened 2025-03-05
Docker files for alarm server

georgweiss:
After som trials and errors.... Docker for the alarm service comes with some limitations. I have created three files to cover the use cases:

- Run only Zookeeper and Kafka. Saves user from downloading and configuring Kafka.
- Run the entire stack to import a configuration, only for default topic "Accelerator"
- Run the entire stack in "normal mode", only for default topic "Accelerator"

Peer testing and feed-back appreciated.

### DIFF
--- a/.github/workflows/alarm-server-docker-image.yml
+++ b/.github/workflows/alarm-server-docker-image.yml
@@ -1,4 +1,4 @@
-name: Save-And-Restore Docker Image CI
+name: Alarm Server Docker Image CI
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


shroffk:
Is there a reason you closed this PR?

georgweiss:
Yes, I realized there might be a better strategy:

1. Docker compose bundling Kafka/Zookeeper, Elastic and MongoDB. This would then be common for all middleware services.
2. Docker compose for the individual middleware services.

Will create a new PR when I have verified this scheme.

shroffk:
yup that makes perfect sense...
docker compose for middlelayer backends... this would make networking much easier

PR #3319, opened 2025-03-07
Add 'pick_on_bounds' property to symbolwidgets for clicking intricate symbols

DenSinH:
We have some intricate symbols with transparency, but clicking them can be quite annoying, as the mouse might hover over a small transparent region of the image, which makes it not register the hover state / makes it unclickable. We added a property named after the corresponding JavaFX ImageView property "pick on bounds", which makes the entire image bounds clickable, regardless of transparency. For backwards compatibility, it is false by default.

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/Messages.java
@@ -273,6 +273,7 @@ public class Messages
                          WidgetProperties_Rotation,
                          WidgetProperties_Rules,
                          WidgetProperties_RunActionsOnMouseClick,
+                         WidgetProperties_PickOnBounds,
                          WidgetProperties_Running,
                          WidgetProperties_ScaleFactor,
                          WidgetProperties_ScaleFormat,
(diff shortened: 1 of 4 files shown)


kasemir:
Our site doesn't rely on clickable symbols, in fact I'd discourage that (https://github.com/ControlSystemStudio/phoebus/wiki/Display-Builder-Compatibility#actions-action-buttons).

But if you're stuck with displays full of transparent clickable symbols, this is what you'll have to do, short of updating all symbols to have for example a faint gray background ... as we had to do with the CSS icon because at least on Mac OS X it had the same problem: Click on the "C" icon in the "Doc" and nothing happens because it's mostly transparent...

Still, do we really need a new property for this, or can we simply turn this into a one-line change, making all symbols always fully clickable?

```
imageView.setPickOnBounds(true);
```

DenSinH:
Fine by me, I figured we would do it this way just to not break the "default" behavior for other users.

kasemir (commented):
I'd rather avoid a new property for this. Make it a one-line change, all symbols always fully clickable:
```
imageView.setPickOnBounds(true);
```

Or make it a preference setting...

```
... class JFXPreferences
{
    /** Are symbol widgets fully clickable? */
    @Preference public static boolean pick_on_bounds;
```

.. with corresponding info and default
```
# Are symbol widgets fully clickable,
# including their transparent pixels?
pick_on_bounds=true
```

.. which is then again applied to all symbol widgets:

```
imageView.setPickOnBounds(JFXPreferences.pick_on_bounds);
```

shroffk:
I added the recommended solution in PR #3424

PR #3325, opened 2025-03-13
use Maven CI friendly versions

minijackson:
This is a proposition to centralize the Phoebus version in the root `pom.xml`, in the `revision` property. This is an official Maven workflow, specified here: https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html

This enables downstream users to make their own versions, by specifying `-Drevision=4.7.4-42-SNAPSHOT` in the Maven command-line.

This also should make the Maven release plugin unnecessary, since the version is specified in a central place, according to: https://stackoverflow.com/a/67012045

Tested with the command `mvn package -DskipTests -Drevision=4.7.4-42-SNAPSHOT`

### DIFF
--- a/app/3d-viewer/pom.xml
+++ b/app/3d-viewer/pom.xml
@@ -3,19 +3,19 @@
   <parent>
     <groupId>org.phoebus</groupId>
     <artifactId>app</artifactId>
-    <version>4.7.4-SNAPSHOT</version>
+    <version>${revision}</version>
   </parent>
   <artifactId>app-3d-viewer</artifactId>
   <dependencies>
     <dependency>
       <groupId>org.phoebus</groupId>
       <artifactId>core-ui</artifactId>
-      <version>4.7.4-SNAPSHOT</version>
+      <version>${revision}</version>
     </dependency>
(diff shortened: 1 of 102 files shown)


georgweiss:
@katysaintin, we use the release plugin to also push build artifacts to Maven Central. How does this PR affect that step?

katysaintin:
I will ask to @minijackson , test a release in his fork before. @minijackson could you test a release with this modification in order to check if there is no unexpected effect side ?

minijackson:
I'm in training this week, so I'll probably won't have the time to check it until next week, but according to the StackOverflow answer, it should make the release plugin not needed anymore.

But I don't know how you use the release plugin, so I'll need your help to figure out if that's the case.

georgweiss:
Things is that the release plugin works together with the sonatype staging plugin such that build artifacts are published to Maven Central. Consequently I am a bit concerned about removing the release plugin.

shroffk:
@minijackson 
I am going to close this since we have had a few releases and it might be easier to recreate this than resolve the conflicts

PR #3329, opened 2025-03-14
Add new formula function - arrayTotal

tynanford:
There is an existing `arraySum` function which adds/subtracts a fixed offset to each element in an array (and returns an array).

The new function `arrayTotal` will return a number which is the sum of the elements in the array

image


kasemir (approved):
In hindsight, the original function should have been called `arrayOffset`, because `arraySum` sounds like what you're doing here. Well, too late..

tynanford:
Yes... well such is life :)

PR #3338, opened 2025-03-24
CSSTUDIO-3113 PVAClient: Accept TCP connections on a separate thread.

abrahamwolk:
We have encountered an issue where with an IOC that accepts UDP but drops TCP packets: when loading an OPI that contains a PV on the IOC in question, any PVs that have not been connected to yet show as being disconnected.

It appears that the reason is that the code that connects over TCP after receiving a UDP response to a search query gets stuck waiting for the TCP connection to be established. It appears that this code runs on a single thread, and that as a consequence the TCP connections to subsequent IOCs are not established.

This pull request calls the code to establish the TCP connection on a separate thread. Please note that I do not know this code base, and that this pull request is meant as a starting point for a discussion and perhaps as inspiration for a real fix. In particular, I don't know which parts of the code are thread-safe and in which way. This has to be considered carefully.

I can reproduce the issue by the following steps:
1. Running a few soft IOCs locally
2. Creating an OPI with Text Update widgets displaying them
3. Configuring the firewall to drop TCP packets on the port of one of the soft IOCs
4. Refreshing the OPI.

@kasemir, I would like to ask whether you think that this looks like a plausible fix for the error that we have observed?

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/PVAClient.java
+++ b/core/pva/src/main/java/org/epics/pva/client/PVAClient.java
@@ -250,32 +250,35 @@ public class PVAClient implements AutoCloseable
         channel.setState(ClientChannelState.FOUND);
         logger.log(Level.FINE, () -> "Reply for " + channel + " from " + (tls ? "TLS " : "TCP ") + server + " " + guid);
 
-        final ClientTCPHandler tcp = tcp_handlers.computeIfAbsent(server, addr ->
-        {
-            try
+        Thread thread = new Thread(() -> {
+            final ClientTCPHandler tcp = tcp_handlers.computeIfAbsent(server, addr ->
             {
-                return new ClientTCPHandler(this, addr, guid, tls);
+                try
+                {
+                    return new ClientTCPHandler(this, addr, guid, tls);
+                }
+                catch (Exception ex)
+                {
+                    logger.log(Level.WARNING, "Cannot connect to TCP " + addr, ex);
+                }
+                return null;
+            });
+            // In case of connection errors, tcp will be null
+            if (tcp == null)
+            {   // Cannot connect to server on provided port? Likely a server or firewall problem.
+                // On the next search, that same server might reply and then we fail the same way on connect.
+                // Still, no way around re-registering the search so we succeed once the server is fixed.
+                search.register(channel, false /* not "now" but eventually */);
             }
-            catch (Exception ex)
+            else
             {
-                logger.log(Level.WARNING, "Cannot connect to TCP " + addr, ex);
-            }
-            return null;
-        });
-        // In case of connection errors, tcp will be null
-        if (tcp == null)
-        {   // Cannot connect to server on provided port? Likely a server or firewall problem.
-            // On the next search, that same server might reply and then we fail the same way on connect.
-            // Still, no way around re-registering the search so we succeed once the server is fixed.
-            search.register(channel, false /* not "now" but eventually */);
-        }
-        else
-        {
-            if (tcp.updateGuid(guid))
-                logger.log(Level.FINE, "Search-only TCP handler received GUID, now " + tcp);
+                if (tcp.updateGuid(guid))
+                    logger.log(Level.FINE, "Search-only TCP handler received GUID, now " + tcp);
 
-            channel.registerWithServer(tcp);
-        }
+                channel.registerWithServer(tcp);
+            }
+        });
+        thread.start();
     }
 
     /** Called by {@link ClientTCPHandler} when connection is lost or closed because unused


kasemir:
Looks like the same problem we had with channel access, https://github.com/epics-base/jca/issues/36

Thoughts from that:

When you have an IOC that replies to a search with "Talk to me via TCP on address:port" but then doesn't respond on that TCP address:port, you have a problem. Can't operate like that, need to fix the IOC (file handle resources, ...) and/or the network (firewall). Still, on the client side we need something better than hanging forever.

In your tests, does it really hang forever? Or do you see a long connection timeout (minutes...)?

For CA, we decided that failing right away is bad because then you enter a tight loop of search, get UDP response, fail to connect via TCP, search, ...
Dropping the channel from the search means that you'll never connect, even after the IOC gets restarted.
So, for CA, the decision was to implement a configurable TCP connection timeout, https://github.com/epics-base/jca/pull/78/files 
I suggested a short default for the timeout, like the 1 second timeout used in the `caget` command line, but the value of 120 secs was chosen to match the original problem. Good to have a fix that looks exactly like the problem?!

Moving the TCP connection to a new thread wouldn't be my first choice. Creates additional threads that we otherwise don't need, plus should still have a connection timeout (with a default of ~5 seconds?).

Can you try that? Add a connection timeout, ~5 sec. The screens would then connect somewhat sluggishly, not perfect, but then checking the log will show the connection errors and provide good pointers to the underlying problem, which IOC is failing to allow connections.

kasemir:
For example, try this:

```
--- a/core/pva/src/main/java/org/epics/pva/common/SecureSockets.java
+++ b/core/pva/src/main/java/org/epics/pva/common/SecureSockets.java
@@ -158,12 +158,17 @@ public class SecureSockets
     public static Socket createClientSocket(final InetSocketAddress address, final boolean tls) throws Exception
     {
         initialize();
+        final int timeout_ms = 2000; // TODO get from a new PVASettings.EPICS_PVA_...
         if (! tls)
-            return new Socket(address.getAddress(), address.getPort());
-
+        {
+               Socket s = new Socket();
+               s.connect(address,  timeout_ms);
+            return s;
+        }
         if (tls_client_sockets == null)
             throw new Exception("TLS is not supported. Configure EPICS_PVA_TLS_KEYCHAIN");
-        final SSLSocket socket = (SSLSocket) tls_client_sockets.createSocket(address.getAddress(), address.getPort());
+        final SSLSocket socket = (SSLSocket) tls_client_sockets.createSocket();
+        socket.connect(address, timeout_ms);
         socket.setEnabledProtocols(PROTOCOLS);
         // Handshake starts when first writing, but that might delay SSL errors, so force handshake before we use the socket
         socket.startHandshake();

```

georgweiss:
Before dismissing threading...

The issue reported triggering this investigation was an OPI with a large number of PVs, some of which were running on the crippled IOC. At this point there was no clue in the OPI nor in the log how to identify the IOC. Connecting on separate threads would in this case have a better chance to reveal which PVs were actually responsive when the OPI was reloaded to trigger new connection attempts.

Agree that threading comes with potential issues, but those can be managed. As for resources... we could consider a common thread pool to support the connection process. Granted, the thread pool would be exhausted if the number of non-responsive PVs is lager than the pool size.

kasemir:
>  ...diff .. tried running with it, and the observed behavior is similar to the behavior I described above

I think we need to understand this before adding threads and really making it more complicated.

With a connection timeout of about 2 seconds, I would accept that some connections to "good" IOCs are delayed but only by about those 2 seconds.
Maybe the issue here is this:
 
 1) Sending UDP search
 2) Getting UDP reply
 3) Attempting TCP connection
 4) TCP times out after 2 seconds
 5) Re-schedule the search

.. and then we enter the loop right away, so overall we're constantly in that 2 second wait-for-timeout that blocks everything else.

Can you increase the log level to the point where you see all the search and related details to check if that's what's happening, and if not, what is happening?

If the scenario is like that, we need to modify this part
```
                search.register(channel, false /* not "now" but eventually */);
```
to delay the search. Register the search such that the `SearchedChannel`'s `search_period` is already at say half the `MAX_SEARCH_PERIOD`.

kasemir:
The search bucket business is complicated. I've added about 40 lines of comments to explain it, but when I re-read that now I'm not 100% sure.

> the PV that times out is represented multiple times

This comment suggests that we are indeed not removing a channel from all buckets for some reason, which might explain what you see:
```
    public PVAChannel unregister(final int channel_id)
    {
...
            // NOT removing `searched` from all `search_buckets`.
            // .. reasons...
```

Maybe that's part of the problem? For failed searches, should we remove the channel from all buckets, then schedule it with a long delay?
Not sure if the longest delay means
`bucket = (bucket - 1)  % search_buckets.size();` or if that requires a "-2", see comments about MAX_SEARCH_PERIOD

shroffk:
So if I understand correctly, right now we will only do step 1. 

> Implement a removal of searched from search_buckets in unregister().

Should this be completed before we create the Phoebus 5 release or is this a unique enough situation that we should not hold up the release.

kasemir:
I'm for doing two things:
Add a configurable timeout to the connection,
and removal of searched from search_buckets in unregister().

Goal is to not hang for a long time.
A short connection delay from those ~2 sec timeouts is acceptable at times where your network is messed up.

georgweiss:
Conclusion from ESS point of view should be that it is rare. We've used pva extensively and this was identified only very recently when an IOC git into a deadlock state.

shroffk:
I feel like I have less that an incomplete understanding of the code ( but that hasn't stopped me from having an opinion)
A lot of our IOC's have hundreds of PV's which are all usually displayed together... when we open one of those screens and the IOC is not behaving correctly have a few more threads might not significantly mitigate the problem.

Could we have a single thread setup which works with 2 queues (buckets).
The priority queue in which all new PV connection request start of... if they fail because the name resolution did not result in a reply or the TCP connection failed even after a successful name resolution then we move those PV's into the secondary queue.
We always try to connect to all the PV's in the primary queue first and only then move to the secondary queue.
Maybe the schedule with a long delay achieves the same.

kasemir:
Step 1, the socket connection timeout, is the most important step to keep us from blocking for a long time.
Once that's merged, the next step should be twofold:
In `PVAClient.handleSearchResponse`, move the code from `tcp_handlers.computeIfAbsent..` on into a thread, as @abrahamwolk suggested in this PR. Instead of plain threads, lightweight threads do make sense, except they're a new concept. We have for now allowed building core-pva with JDK 8, because that's what Matlab required. Maybe that no longer matters?
The other part of the second step would be in `ChannelSearch.unregister`. It really needs to remove the channel from all buckets, because otherwise, after a TCP connection failure, the re-search will happen to frequently:

```
public PVAChannel unregister(final int channel_id)
{
    final SearchedChannel searched = searched_channels.remove(channel_id);
    if (searched != null)
    {
        logger.log(Level.FINE, () -> "Unregister search for " + searched.channel.getName() + " " + channel_id);
        synchronized (search_buckets)
        {   // Remove from all buckets!!
            for (LinkedList<SearchedChannel> bucket : search_buckets)
                bucket.remove(searched);
        } 
        return searched.channel;
    }
    return null;
}
```

kasemir:
No. The search messages are not a problem. They are efficiently merged such that one UDP search contains all the channels that are scheduled for being searched in that time slot. The creation of a new TCP socket connection is usually no problem, either, but moving that to a (LW) thread avoids the occasional delays.

Issue #3340, opened 2025-03-25
Failure to convert some OPI actions to BOB using the converter


rjwills28:
I have found a few bugs relating to the conversion of OPI files to BOB format. Note that all 3 of those listed below **worked in version 4.7.3**. It is only in the version I've built from GitHub master where it fails indicating that these bugs have been introduced since the last release.

**ISSUES**

1. This relates to the actions that are attached to say an ActionButton or any other widget. If you have an OPI widget with an action to **EXECUTE_CMD** or **EXECUTE_PYTHONSCRIPT**/**EXECUTE_JAVASCRIPT** they fail to be parsed and so are removed. I see the following error:

    ```
    2025-03-21 09:37:15 SEVERE [org.csstudio.display.builder.model] Error reading widget Widget '' (action_button) property <actions>, line 91
    java.lang.RuntimeException: No action implementation matching type 'EXECUTE_CMD' found.

    2025-03-21 09:37:15 SEVERE [org.csstudio.display.builder.model] Error reading widget Widget '' (action_button) property <actions>, line 144
    java.lang.RuntimeException: No action implementation matching type 'EXECUTE_PYTHONSCRIPT' found.

    2025-03-21 09:37:15 SEVERE [org.csstudio.display.builder.model] Error reading widget Widget '' (action_button) property <actions>, line 198
    java.lang.RuntimeException: No action implementation matching type 'EXECUTE_JAVASCRIPT' found.
    ```

    The problem here I think relates to the comparison of the 'type' attribute. I needed to make one change to the ExecuteCommandAction.java and 2 changes to the ExecuteScriptAction.java to get these to parse correctly.

2. Once I'd fixed the above I no longer saw the exceptions and could open the converted BOB file to see that the actions were indeed there, however in the case of an embedded EXECUTE_PYTHONSCRIPT/EXECUTE_JAVASCRIPT the actual EmbeddedScript was missing again. The error it throw is rather obscure but digging down I found this is because there is also a bug in the `writeToXML()` method meaning it fails to detect that these should be embedded scripts. 

    ```
   2025-03-24 10:03:20 WARNING [org.csstudio.display.builder.runtime] Script for action failed to compile
    Display '/home/user/CSS/opi_actions.bob', Widget 'Action Button_3' (action_button), Execute Script
     java.io.FileNotFoundException: /home/user/CSS (Is a directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
    ```

[...]

PR #3341, opened 2025-03-25
Fix conversion of OPI actions to execute scripts/commands into BOB

rjwills28:
I found a few issues during the conversion of OPI to BOB when the widgets contained an action to EXECUTE_CMD, EXECUTE_PYTHONSCRIPT or EXECUTE_JAVASCRIPT. These are detailed in https://github.com/ControlSystemStudio/phoebus/issues/3340.

This PR fixes the 3 issues detailed in the issue ticket above and also adds 2 Java unit tests to test this functionality to ensure we don't miss an regression to converter in the future.


georgweiss:
+1 for the unit tests

PR #3347, opened 2025-04-02
Fix an error in databrowser when preference org.csstudio.trends.databrowser3/archives is empty

katysaintin:
When org.csstudio.trends.databrowser3/archives is empty databrowser3
still create a ArchiveDataSource with an empty URL and throws an error


kasemir (approved):

katysaintin:
Thanks @kasemir for your approvment

PR #3350, opened 2025-04-03
Update to java 21

jacomago:
Added this to everything, could reduce down to only core.

### DIFF
--- a/.github/workflows/alarm-logger-docker-image.yml
+++ b/.github/workflows/alarm-logger-docker-image.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Maven and Java Action
         uses: s4u/setup-maven-action@v1.18.0
         with:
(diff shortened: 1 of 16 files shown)


kasemir:
Also update java version in https://github.com/ControlSystemStudio/phoebus/blob/master/.github/workflows/build.yml and associates?

jacomago:
Looks like minimum version of spring boot that supports java 21 is 3.3. I'll split this up to only cover the non spring boot things.

georgweiss:
Hmm, why should Spring 2.7 not work on Java 21? The opposite I can understand...

georgweiss:
Thanks @jacomago ,  stepping up Spring Boot version must be a separate task...

shroffk:
closing this for now and merging the effort into the migration to JDK 25

PR #3352, opened 2025-04-03
Update spring boot services to build using java 17

jacomago:


jacomago:
Not sure why there isn't a setting for save and restore. Maybe it inherits from parent?

georgweiss (approved):

Issue #3366, opened 2025-04-14
Feature: Configurable Runtime Zoom Factor


UrbanBobek:
Hi all,

we implemeted a way to set a zoom factor for the runtime screens in the `settings.ini` file. 

The motivation behind this was to have a way to make GUI sizes look consistent across different monitor sizes and resolutions, but it can also serve as a way for operators to costumize GUI size for their preference. 

Is this someting you would like to see in the master branch? 

Cheers,
Urban

georgweiss:
@UrbanBobek, are you able to provide an example?

UrbanBobek:
You can specify the zoom factor in the settings file and all screens that are opened have the default 100% zoom multiplied by this factor:
```
##################################################
# Package org.csstudio.display.builder.runtime.app
##################################################

org.csstudio.display.builder.runtime.app/default_zoom_factor=1.23
```

![Image](https://github.com/user-attachments/assets/f11e2a23-8cf4-4d14-aa6d-5d9f5b901377)

I can open a PR if that will help you with understand the functionalty better.

georgweiss:
So this would in practice override the default zoom factor of 1.0.

I cannot see why this should not be supported.

PR #3369, opened 2025-04-15
Implement user defined runtime zoom factor

UrbanBobek:
Add a mechanism to apply a custom zoom factor for the runtime window. 

The desired zoom factor can be set in settings.ini as:
```
org.csstudio.display.builder.runtime.app/default_zoom_factor=1.23
```

Refers to #3366

### DIFF
--- a/app/display/runtime/src/main/resources/zoom_preferences.properties
+++ b/app/display/runtime/src/main/resources/zoom_preferences.properties
@@ -1,6 +1,0 @@
-# ----------------------
-# Package org.csstudio.display.builder.runtime.app
-# ----------------------
-
-# Default zoom factor (percentage) of display runtime window
-default_zoom_factor=100
--- a/app/display/runtime/src/main/resources/zoom_preferences.properties
+++ b/app/display/runtime/src/main/resources/zoom_preferences.properties
@@ -1,0 +1,6 @@
+# ----------------------
+# Package org.csstudio.display.builder.runtime.app
+# ----------------------
+
+# Defualt zoom factor(double) of display runtime window
+default_zoom_factor=1.00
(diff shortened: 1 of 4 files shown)


app/display/runtime/src/main/resources/zoom_preferences.properties
```diff
@@ -0,0 +1,6 @@
+# ----------------------
+# Package org.csstudio.display.builder.runtime.app
+# ----------------------
+
+# Defualt zoom factor(double) of display runtime window
```
kasemir, line 5: I'd fix the "Defualt" typo. You mention that it's a double, but it's not absolutely clear if the value is in units of percent or pixel or ...
Since the user interface is in terms of percentage ("123 %"),  I suggest using percentages here as well:

```
# Default zoom factor (percentage) of display runtime window
default_zoom_factor=100.0
```

UrbanBobek: Thanks for the comments. I agree that using percentages make mores sense. I pushed the changes.

kasemir (commented):
Sorry, I missed that the first time around:

Can't the new `default_zoom_factor` be added to the existing
https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/resources/display_runtime_preferences.properties
and associated 
https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/Preferences.java ?

Seems wrong to start a new preference file for what's just another runtime preference

UrbanBobek:
> Sorry, I missed that the first time around:
> 
> Can't the new `default_zoom_factor` be added to the existing https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/resources/display_runtime_preferences.properties and associated https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/Preferences.java ?
> 
> Seems wrong to start a new preference file for what's just another runtime preference

Good point, I didn't consider this during implementation. Fixed.

PR #3371, opened 2025-04-16
Save&Restore search on "Description/Comment" seems to be case sensitive

jasonbrudvikess:
It seems there were several issues connected to searching on the  "Description/Comment" and "Node Name"  fields.

For the "Description/Comment"  field:
- Searches containing upper case letters failed, even if the decription being looked for had upper case letters and matched
- Searches with lower case letters succeeded
- Searches for sub-words failed
- Searches containing multiple words failed

While for the "Node Name" field there was no case sensitivity issue, but it had the following issues in common with the  "Description/Comment"  field:
- Searches for sub-words failed
- Searches containing multiple words failed

This branch fixes those issues.

Perhaps I have been overly verbose with my logging output - that can be reduced.

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/search/SearchAndFilterViewController.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/search/SearchAndFilterViewController.java
@@ -493,13 +493,13 @@ public class SearchAndFilterViewController extends SaveAndRestoreBaseController 
      */
     private String buildQueryString() {
         Map<String, String> map = new HashMap<>();
-        if (nodeNameProperty.get() != null && !nodeNameProperty.get().isEmpty()) {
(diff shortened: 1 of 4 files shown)


georgweiss (changes requested):
This is client-side changes only. My preference would be to update search implementation on service side.

I will propose an alternative in a f2f meeting.

jasonbrudvikess:
> This is client-side changes only. My preference would be to update search implementation on service side.
> 
> I will propose an alternative in a f2f meeting.

Ok, that sounds fine.

shroffk:
I agree with Georg, that we should make the searches case insensitive on the server side.

Thanks for the work @jasonbrudvikess

georgweiss (approved):

shroffk:
I thought we were going to add this feature to the service.
I will create an Issue ticket.

shroffk:
Yes I just noticed that :)

PR #3376, opened 2025-04-22
CSSTUDIO-3196 Change the semantics of colors defined to be equal to other colors

abrahamwolk:
Consider the color definitions
 ```
GREEN = 61, 216, 61
OK = GREEN
 ```
Currently, this will result in Phoebus having the following two colors defined:
 ```
GREEN = 61, 216, 61
OK = 61, 216, 61
```
The intent of this pull request is to change the semantics so that the color definitions instead results in _only_ the color
```
GREEN = 61, 216, 61
```
being defined, while any reference to the color "OK" is replaced with a reference to the color "GREEN" instead. (I.e., "OK" points to the the color `GREEN` instead of being defined as an identical color.)

The motivation for this change is that widgets in Phoebus have hard-coded default colors, and setting them all to default values results in identical colors being defined that differ only in their name. With this change, it is possible to instead "redirect" the default colors to other colors.

### DIFF
--- a/app/display/model/src/main/resources/examples/color.def
+++ b/app/display/model/src/main/resources/examples/color.def
@@ -1,11 +1,23 @@
 # Named colors
 #
 # Format:
-#   NameOfColor = red, green, blue [, alpha ] | PreviouslyDefinedNameOfColor
+#   NameOfColor = red, green, blue [, alpha ] | PreviouslyDefinedNameOfColor | alias(PreviouslyDefinedNameOfColor)
 # with values in 0..255 range.
 #
 # Whenever possible, use named colors in displays
 # instead of arbitrary red/green/blue values.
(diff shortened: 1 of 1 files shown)


georgweiss (approved):

kasemir:
When I open the color selection dialog, do I still see "OK" and "GREEN" listed in there?

For an LED that displays a good/bad state, I want to pick "OK".
For a rectangle that represents a tank full of green slime, I want to pick "GREEN".
Right now they might give me the same color, and it's great to be able to define "OK=GREEN" so OK is indeed the exact same color as GREEN.
But "GREEN" must always remain green, while "OK" could over time be defined as bright green, dull green, dark red or gray like a fault LED that's off because we're OK and so on.

PR #3380, opened 2025-04-23
docs: switch to Read the Docs theme

### LINKED ISSUE #3160: Documentation pointing to /latest/, but /stable/ documentation seems better
Hi, I just want to point that the default link to the docs points to the /latest/ documentation, which, at least in my opinion, is much worse than the one in /stable/. I don't know if this is the expected behaviour, or if it has been discussed before, and if the /latest/ one is the one that wants to be displayed to new users like myself, but if it is not, consider changing it. 

Thank you!

minijackson:
Fixes #3160.

This switches the documentation from the "default" theme to the "Read the Docs" theme, which provides a sidebar that is consistent across all pages, which makes it easier to navigate, and easier to grasp the structure of the documentation.

There are other Sphinx themes, which you can see here: https://sphinx-themes.org/
but the RTD theme is very popular, and well designed, in my opinion.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -91,7 +91,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -8,10 +8,9 @@ Welcome to the CS-Studio/Phoebus Documentation!
 
 .. image:: splash.png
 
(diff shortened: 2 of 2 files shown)


minijackson:
Note: not in the sphinx-themes.org directory, but there's also the [Shibuya](https://shibuya.lepture.com/writing/) theme that looks really good.

shroffk:
I have enabled the build on PR

minijackson:
I had to do a git push to trigger the RTD build.

I've also used the `:caption:` `toctree` argument, so that it's shown on the sidebar, and removed the appendinx from the index, because it didn't contain anything.

jacomago:
Looks way better to me

shroffk:
I agree... like mentioned in the issue ticket this will bring back the missing index.
@jacomago @minijackson can I merge this or are there other changes too?

minijackson:
@shroffk I personally think merging as-is should be fine

PR #3387, opened 2025-05-03
Extend widget info dialog to show archived status

georgweiss:
NOTE: at this point this is more of food for thought than PR.

Based on user request: accessing information on whether PVs in an OPI are archived or not is currently a bit cumbersome. This idea adds a feature where user may click a button to invoke a job querying the archiver data source for latest value. Presence of a result is interpreted as "Yes", absence as "No". Initially nothing is shown in the dedicated column.

The extra column and the button are suppressed if site specific product does not include the archiver data source.

See screen shots.

Potential discussion topics:

- Using the archiver data source implies one request per PV. As far as I understand the archiver supports a batch request endpoint that would be far more efficient to use.
- The added column shows plain text. Icons would improve readability...
- Only thing indicating ongoing archiver communication is the new button: retrieval is not done as long as button is disabled.
- Is Yes/No sufficient? Is there any value in showing the last archived value and - maybe more interesting - time stamp of last archived value?

At ESS we have also discussed a possible enhancement: by means of the Channel Finder APIs one could also add a button for those items not being archived: "Start Archiving". 

A few refinements remain, e.g. localize string resources.

Screenshot 2025-05-03 at 09 43 24
Screenshot 2025-05-03 at 09 43 34


kasemir:
> accessing information on whether PVs in an OPI are archived or not is currently a bit cumbersome

For a PV of interest, you can simply invoke "Data Browser" from the context menu and then see the data.
This new option requires you to press "Get Archiver Status", wait for a "Yes", which still doesn't tell you what time range and granularity might be available, and finally you likely show those PVs of interest in the data browser, no simplification at all.

Implementation wise, I like how it uses the `archive:` data source, no new direct dependencies between data browser/archive client/display, and it even auto-enables based on `archive:` being supported or not.

So, whatever, ..., go ahead

kasemir (approved):

georgweiss:
> you can simply invoke "Data Browser"

is not considered simple enough. In particular, if you quickly want a list of non-archived PVs the Data Browser trick will not do.

Agree though that a simple yes/no based on ```archive:``` data source returning something may not be sufficient. The PV might exist in the archiver, but may be paused.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
```diff
@@ -523,22 +600,70 @@ private Tab createWidgetStats(final DisplayModel model)
...
+                    pv.setArchiverValue(archivedStatus);
+                }
+            } finally {
+                getFromArchiverInProgress.setValue(false);
+            }
+        });
+    }
+
+    private ArchivedStatus getFromArchiver(String pvName) {
+        if (pvName.toLowerCase().startsWith("loc://") || pvName.toLowerCase().startsWith("sim://")) {
```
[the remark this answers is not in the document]
georgweiss: Good point, ``sys://``` should be added. 
As for formulas... even if PV names are extracted, the list shows only one yes/no value, so let's assume this is an unsupported corner case.
If ```mqtt://``` and related cannot be archived, then the archiver data source module should not be added to the product, and consequently the dialog UI will not show the associated elements.

shroffk (approved):

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/WidgetInfoDialog.java
```diff
@@ -523,22 +599,80 @@ private Tab createWidgetStats(final DisplayModel model)
...
+            return ArchivedStatus.NOT_APPLICABLE;
+        }
+        // Check if pv name is prefixed with a scheme, e.g. pva://, ca://...
+        int indexSchemeSeparator = pvName.indexOf("://");
+        if (indexSchemeSeparator > 0 && pvName.length() > indexSchemeSeparator) {
+            pvName = pvName.substring(indexSchemeSeparator + 1);
+        }
+        // Prepend "archiver://"
+        pvName = "archive://" + pvName + "(" + TimestampFormats.SECONDS_FORMAT.format(Instant.now()) + ")";
+        try {
```
kasemir, line 661: This pattern...
```
try
{
    PV pv = PVPool.getPV(pvName);
    ...
    PVPool.releasePV(pv);
}
catch (Exception e)
{
   ...
}
```
leaks PVs if there's an exception before the PV can be released.

Something like this would always release the PV:
```
PV pv = PVPool.getPV(pvName);
try
{
    ...
}
catch (Exception e)
{
   ...
}
PVPool.releasePV(pv);
```

georgweiss: Thanks @kasemir , will update.

abrahamwolk: I think that a `finally`-block may be even better: https://docs.oracle.com/javase/tutorial/essential/exceptions/finally.html (E.g., in case the `try { ... } catch (Exception e) { ... }` throws another exception, or contains a `return`-statement, a `continue`-statement, or similar.)

PR #3393, opened 2025-05-12
Fix elog client exceptions issue

wanglin86769:
This PR is trying to fix https://github.com/ControlSystemStudio/phoebus/issues/3198 with the following modifications,
1. Add search() method to ElogClient.java
2. Support pagination log request
3. Support sort by date
4. Fix a typo which which causes Text be sent for both text and subject

### DIFF
--- a/app/logbook/elog/src/main/java/org/phoebus/elog/api/ElogApi.java
+++ b/app/logbook/elog/src/main/java/org/phoebus/elog/api/ElogApi.java
@@ -90,7 +90,7 @@ public class ElogApi {
 
     String encoded_Text = new String(attributes.get("Text").getBytes(), StandardCharsets.ISO_8859_1); // charset hell - this seems stupid but works
     attributes.put("Text", encoded_Text);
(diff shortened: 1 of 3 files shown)


andrea-celentano:
Dear @wanglin86769, thanks for this update. I tried to recompile the latest version of phoebus, and I confirm it works properly regarding the integration with elog, both in terms of new message posting to the logbook and previous message display from the logbook.

The only minor concern I observe regards the formatting of some messages retrieved from the logbook.
I report below an example. The original message looks like:

image

And it is rendered on phoebus as:

image

Thanks,
Andrea

wanglin86769:
Dear @andrea-celentano , thanks for reporting. It seems that the log description is displayed on a JavaFx `WebView` tag in the following source file, which by default collapses whitespaces and newlines to single whitespace.
`phoebus\app\logbook\olog\ui\src\main\resources\org\phoebus\logbook\olog\ui\SingleLogEntryDisplay.fxml`

I have created a PR https://github.com/ControlSystemStudio/phoebus/pull/3413 to preserve whitespaces as follows,

image

image

Hope the modification does not affect the behavior of Olog, because the Phoebus Elog client actually uses the Olog UI.

Issue #3407, opened 2025-05-21
Missing context menu for some control widgets


rjwills28:
I have discovered a bug in some of the recent updates to Phoebus. If you right click on a **Boolean Button** widget (or a **Check Box** or a **Slide Button** widget) you no longer get a context menu and I see the following exception in the log:

```
WARNING [org.csstudio.display.builder.representation] Context menu failure for Widget 'Boolean Button' (bool_button)
java.lang.ClassCastException: class org.csstudio.display.builder.model.properties.ConfirmDialog cannot be cast to class java.lang.Boolean (org.csstudio.display.builder.model.properties.ConfirmDialog is in unnamed module of loader 'app'; java.lang.Boolean is in module java.base of loader 'bootstrap')
	at org.csstudio.display.builder.runtime.app.ContextMenuSupport.fillMenu(ContextMenuSupport.java:138)
	at org.csstudio.display.builder.runtime.app.ContextMenuSupport$1.handleContextMenu(ContextMenuSupport.java:95)
	at org.csstudio.display.builder.representation.ToolkitRepresentation.fireContextMenu(ToolkitRepresentation.java:666)
	at org.csstudio.display.builder.representation.javafx.widgets.JFXBaseRepresentation.lambda$createComponents$1(JFXBaseRepresentation.java:140)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
```

I had a look at the change made to [app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/ContextMenuSupport.java](https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/ContextMenuSupport.java) and can see that it is assuming that `prompt.get().getValue()` is a boolean. This is the case for most widgets where the property 'Confirmation Dialog' is true/false (e.g. a Choice Button, Action Button), but is not that case for the 3 widgets highlighted (like the Boolean Button) where 'Confirmation Dialog' is actually an option (e.g. 'None', 'On both', 'On set', 'On clear'). As indicated by the exception, it can't convert this to a Boolean and so fails.

rjwills28:
I have implemented a fix for this and can open a PR shortly for review.

PR #3408, opened 2025-05-21
Fix missing context menu for certain controls widgets

rjwills28:
Any widget that uses an 'option' for their Confirmation Dialog (e.g. None, On both, on set, etc) is throwing an exception when you right click to display the context menu. This is because this property value cannot be converted to a Boolean. 

See ISSUE [3407](https://github.com/ControlSystemStudio/phoebus/issues/3407) for full details.

I have fixed this by also obtaining the `propConfirmDialogOptions` property and checking whether it is of type `ConfirmDialog`. If it is then I use logic to convert the enum to a boolean. If it is not then we are in the original case where the Confirmation Dialog is just true/false so we can obtain the value from that. A boolean cannot be converted to a `ConfirmDialog` enum and a `ConfirmDialog` enum cannot be converted to a boolean so I think the only way to check this is to grab both property values.


kasemir (approved):
Thanks!

PR #3412, opened 2025-05-22
Fix Write Action command => manage default type and Enum PV Type

katysaintin:
Fix Action WritePV on ButtonWidget.
- default datasource was not take in account (for a custom default datasource as Muscade) action doesn't work
- Write 1 on a Enum PV (such as a bo record) doesn't work, because it sent the wrong type

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/WidgetRuntime.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/WidgetRuntime.java
@@ -26,6 +26,7 @@ import org.csstudio.display.actions.WritePVAction;
 import org.epics.vtype.VType;
 import org.phoebus.framework.macros.MacroHandler;
 import org.phoebus.framework.macros.MacroValueProvider;
+import org.phoebus.pv.PVPool.TypedName;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -235,7 +236,9 @@ public class WidgetRuntime<MW extends Widget> {
                     final String pv_name = ((WritePVAction) action).getPV();
                     try {
                         final String expanded = MacroHandler.replace(widget.getMacrosOrProperties(), pv_name);
-                        final RuntimePV pv = PVFactory.getPV(expanded);
+                        // Manage default datasource if not ca
+                        final TypedName type_name = TypedName.analyze(expanded);
+                        final RuntimePV pv = PVFactory.getPV(type_name.toString());
                         action_pvs.add(pv);
                         addPV(pv, true);
                     } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


shroffk:
No rush
I am going to hold all PR's until after 5.0.2 release so please enjoy your vacation

katysaintin:
> No rush I am going to hold all PR's until after 5.0.2 release so please enjoy your vacation

Please find the settings.ini and the Display to reproduce the bug.
[settings.txt](https://github.com/user-attachments/files/21414413/settings.txt)
in settings.ini the default pv datasource set to loc factory (in my case it is set to muscade).
`org.phoebus.pv/default=loc`

And a Display to reproduce the bug 
[BugAction.bob.txt](https://github.com/user-attachments/files/21414456/BugAction.bob.txt)

In the display , Write 1 on loc://my_pv works fine in TextEntry and Boolean Button . without precising the loc://
But in Action Button, the command doesn't works, because  the action doesn't check the default datasource and do not 
concat loc://$(pv_name) and it searching for a $pv_name that does not exist in PVPool .

It will occur also if you set `org.phoebus.pv/default=pva' if the defaut EPICS protocol is pv access.
All the Writing Actions will failed, because , Action will not add pva://pvname .

katysaintin:
Hello Kunal, could we approve this PR ?
Thanks to the Unit Test, we don't need and extra settings file anymore .
It will be useful for my following PR , I fix pv-tango and test it for real . 

Thank you for your help .
Katy

georgweiss:
@katysaintin, sorry for the delay, will check this.

georgweiss:
@katysaintin, I am a bit confused. Actually with default datasource ca and pva I can write to a bo PV from the action button if PV name is taken from widget PV name field (i.e. not specified explicitly in action dialog editor other than $(pv_name)).

katysaintin:
> @katysaintin, I am a bit confused. Actually with default datasource ca and pva I can write to a bo PV from the action button if PV name is taken from widget PV name field (i.e. not specified explicitly in action dialog editor other than $(pv_name)).

In fact, if in your settings you set a default PV in modifying the configuration in this way.
org.phoebus.pv/default=muscade
or 
org.phoebus.pv/default=tango

To reproduce the bug in setting a default datasource different from ca or pva, such as local 
you can run the following UnitTest app/display/runtime/src/test/java/org/csstudio/display/builder/runtime/test/WidgetRuntimeTest.java

This test, set loc as a default type.
So when you set on your ActionButton a pv_name=my_pv 
It's mean that the complete name of the PV will be pv_name=loc://my_pv

The problem is for muscade or tango default type,
when you not specifiy the type, PVAction will try to write with the pv-ca factory instead of pv-muscade or pv-tango factory.

To test my correction , you juste have to run the Unit test, without my modification and with.
Normally, the test will failed without my modification.

I hope is it clear.

thank you for your help.

Katy

georgweiss:
@katysaintin, I'm fine with this, but I'd like @shroffk to share his take on this.

shroffk (approved):

shroffk:
We can merge this after a quick review during the next developers meeting

shroffk:
I can confirm that the resource leak exits before the changes of this PR
I will merge this and then open a new issue ticket to address the resource leave assocaited with loc:// PVs

PR #3413, opened 2025-05-23
Preserve whitespaces on detail log webview

wanglin86769:
For PSI Elog, if log is submitted as plain text and retrieved using Phoebus client, whitespaces and newlines in description will be collapsed as single whitespace. 

For example, 
333   | 444   | 555
ccd   | ddd   | eee

will be displayed as,
333 | 444 | 555 ccd | ddd | eee

This PR adds one line CSS style in order to preserve the whitespaces.

### DIFF
--- a/app/logbook/olog/ui/src/main/resources/detail_log_webview.css
+++ b/app/logbook/olog/ui/src/main/resources/detail_log_webview.css
@@ -42,6 +42,7 @@ blockquote{
 
 p{
     margin-top: 5px;
+    white-space: pre-wrap;
 }
 
 .olog-table {


georgweiss:
I'm afraid this change has a side effect.
Olog uses Markdown for the log entry body. Trailing white space in Markdown is treated as described here: https://commonmark.org/help/tutorial/03-paragraphs.html.
So if a line ends in a single white space, the content that follows is put on same line as in the example provided in this issue ticket. While the proposed change would in a sense make it easier for users (no need to use two space chars to break line),
existing entries would be affected. 
There is also the web browser client for Olog to consider, which would need to adopt the same change.

@shroffk, what is your take on this?

wanglin86769:
Thanks for the comment. If this PR is not appropriate, probably there is another potential solution as shown in PR https://github.com/ControlSystemStudio/phoebus/pull/3414 .

shroffk:
Thanks, I think the wrapping the elog text is better. I think we should continue to stick with the commonmark specification so that we will be able to use third party libraries and solutions.

@wanglin86769 could we also revert the changes from #3413... sorry for the additional work

shroffk:
I wonder if we should look into a way to allow site specific css contributions

PR #3414, opened 2025-05-23
Wrap Elog plain text inside <p> tag

wanglin86769:
It seems that the previous PR https://github.com/ControlSystemStudio/phoebus/pull/3413 affects the behavior of Olog, and the approach is not appropriate. So this PR takes a different approach.

For Elog, there are three kind of encodings as follows,

**plain**
image

**ELCode**
image

**HTML**
image

For `HTML` encoding, the newline is controlled by the `<p>` tag and whitespace by `&nbsp;`, so there is no problem when displayed on Olog UI.

For `plain` encoding, this PR wraps the Elog text inside `<p style="white-space: pre-wrap;"></p>` tag in order to preserve whitespaces.

### DIFF
--- a/app/logbook/elog/src/main/java/org/phoebus/elog/api/ElogApi.java
+++ b/app/logbook/elog/src/main/java/org/phoebus/elog/api/ElogApi.java
@@ -198,7 +198,13 @@ public class ElogApi {
         attributes.put( data[0], data[1] );
       }
     }
-    attributes.put( "Text", String.join("\n", returned_msg.subList( delimiter_idx + 1, returned_msg.size() ) ) );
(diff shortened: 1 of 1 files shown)


wanglin86769:
I just observed that when editing a log entry with plain text encoding, the `<p>` tag for wrapping also shows up as follows, which is not expected.

Probably this PR is not a good solution either and I am closing it.

image

georgweiss:
Actually embedding HTML in Markdown is OK, and that's why you see the <p> tags in the editor. That said, the "editor" is a plain TextArea component, so it won't process either HTML or Markdown.

PR #3416, opened 2025-05-26
fix Reproducible Builds issue

hboutemy:
storing too much details on Git status is causing issues: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/phoebus/README.md

see diffoscope https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/phoebus/parent-4.7.4.diffoscope

letting timestamp updated by release plugin is perfect

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,8 +92,6 @@
     <maven.compiler.target>17</maven.compiler.target>
     <activemq.client.version>5.18.4</activemq.client.version>
     <commons.compress.version>1.26.1</commons.compress.version>
-    <!--suppress UnresolvedMavenProperty -->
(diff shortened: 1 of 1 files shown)


shroffk:
Thank you for a potential fix.

We do need to include the git properties file but I see how we might be including too much info.
If I exclude all properties except the `git.commit.id` and `git.commit.id.abbrev` I believe the checks should be fine.

+1 on removing this since I see we have the property defined twice.
```
<project.build.outputTimestamp>${git.commit.time}</project.build.outputTimestamp>
```

shroffk:
implementing the fixes with some changes in PR #3431

PR #3417, opened 2025-05-28
Updates to deal with EPICS connection timeout

georgweiss:
At ESS we are seeing failures to take snapshot (performed by service) or restore (when performed by service). Potential culprit is the threading strategy which may trigger an avalanche of broadcasts.

This PR updates the threading strategy to instead use a fixed size thread pool. 

Also in this PR: default connection timeout (5000ms) may be overridden in ```application.properties``` or on command line with
```-Dconnection.timeout=10000```


georgweiss:
@shroffk, to be honset, the change in threading strategy is not based in any specific debug data. I just thought it would be better to limit broadcasts rather that firing them in parallell as quickly as possible for *all* PVs in the configuration. 

I suspect the connection timeouts may be caused by a congested network due to other clients and activities. We've seen that large waveforms with frequent updates are quite popular at times.

I will use a 10s timeout at next deployment. Since save&restore currently has been awarded a lead role in many of our control room activities, I'm sure I'll hear about issues should they resurface.

PR #3420, opened 2025-06-03
Add message if take snapshot or restore fails to connect to PVs

georgweiss:
In save&restore the service (or client) may fail to connect to PVs. In case the list of PVs exceeds the number of visible rows in the UI, users may not be aware of such failures.

This PR adds a UI element highlighting connection failures:
Screenshot 2025-06-02 at 13 59 53

Click on the message launches a dialog showing list of PVs that failed to connect.

Additionally a few minor changes/fixes and updated documentation (e.g. new screen shots).


georgweiss:
Will come up with a better UI strategy to indicate connection failures.

PR #3427, opened 2025-06-09
feature: add formula evaluation on button PV Write Actions

flef:
Hi Everyone, 

Thanks for the latest release.

We would like to be able to evaluate formula in Button Write PV's Action. 

I propose a small patch to handle it.

Regards, 
Flef.

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/actionhandlers/WritePVActionHandler.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/actionhandlers/WritePVActionHandler.java
@@ -1,8 +1,11 @@
 /*
  * Copyright (C) 2024 European Spallation Source ERIC.
+ * Copyright (C) 2025 Thales.
  */
 
 package org.csstudio.display.builder.runtime.app.actionhandlers;
+
+import static org.csstudio.display.builder.model.ModelPlugin.logger;
 
 import org.csstudio.display.builder.model.Widget;
 import org.csstudio.display.builder.model.spi.ActionInfo;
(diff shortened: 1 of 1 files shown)


app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/actionhandlers/WritePVActionHandler.java
```diff
@@ -54,8 +67,39 @@ public void handleAction(Widget sourceWidget, ActionInfo pluggableActionInfo) {
...
+     * @param formulaExpression The complete formula expression including =` and `
+     * @return The evaluated result as a string
+     */
+    private String evaluateFormulaPV(final String formulaExpression)
+    {
+        try
+        {
+            final PV formulaPV = PVPool.getPV(formulaExpression);
+
+            final VType value = formulaPV.read();
```
kasemir, line 81: `read` returns the currently known value of the PV.
Right after `PVPool.getPV`, that's very likely `null`.
https://github.com/ControlSystemStudio/phoebus/blob/68c455d87fedf444550a176def7d294bfcc02ab1/core/pv/src/main/java/org/phoebus/pv/PV.java#L231
`read` will not hang and wait for a value to arrive.
You're expected to subscribe to updates.

Check the test examples for ways to await a certain number of values, like https://github.com/ControlSystemStudio/phoebus/blob/68c455d87fedf444550a176def7d294bfcc02ab1/core/pv/src/test/java/org/phoebus/pv/FormulaTest.java#L60 or https://github.com/ControlSystemStudio/phoebus/blob/68c455d87fedf444550a176def7d294bfcc02ab1/core/pv/src/test/java/org/phoebus/pv/ReactivePVTest.java#L57

You would obviously NOT want to block and wait like that in the button-press handler, because then pushing the button will freeze the GUI until the formula has received a value, which could be _never_.
So you'd have to spawn a thread to handle the button push...

With a formula, you might get lucky because the inputs of that formula are already shown somewhere else on the display. Those PVs have already been created, they connected and received a value. When the formula then gets evaluated, those values are available right away, but you cannot guarantee that, and requiring those who want to use a formula in the button to also show all the formula inputs separately would be odd.

flef: Thank you for this very comprehensive response.

I would like to write PVs based on user inputs:

Currently, I proceed as follows:
- I create widgets with local variables to capture user inputs.
- I create a button with a "WritePV" action, the "WritePV" button action writes a '1' to a local PV (for example: loc://my_button_trigger).
- I create a script on the same button that is triggered by loc://my_button_trigger.
- I retrieve the value of all my PVs in the script and write it to the target PV.
- I also have to manage initialization of the trigger (to 0, when the display is loaded to not trigger unintentionally my script)

I find this method particularly complicated.

Perhaps there is a better way?

If there is no better approach, I will work on this PR, however, I think the user experience will not be optimal. It seems odd to wait for PV updates on a button, risking triggering actions long after pressing the button. 

Do you have any recommendation ?

kasemir: Yes, a script would work because as the script is initialized, it will create all those PVs, they have time to connect.
In that script, remember to only act when the trigger value is for example 1 and reset it to 0.
Somewhat like this:
```
trigger = PVUtil.getInt(pvs[0])
if trigger:
    ... do your thing
    pvs[0].write(0)
```

Or put that logic into an IOC, be it a "normal" IOC with record logic or a "python" IOC based on pcas or p4p. Either way, you'll end up with PVs to capture user inputs, and a "do it" PV for the button. That way, you can access it from any display, and you won't have to re-do your script in 5..10 years when your site transitions to a new display tool.

PR #3429, opened 2025-06-09
PVA: Handle client's connection to server in receive thread

kasemir:
Solves the original issue of #3348, #3338 :

The ClientTCPHandler now performs the TCP connection at the start of its receiver thread.
We no longer hang in the TCP connection setup when the ClientTCPHandler is created.
Nor do we need to create a new short-lived thread for the ClientTCPHandler.


kasemir:
Next steps:

Fully unregistered searches as in #3348 

Then support PVA proto version 3 with new replySrcPort flag, https://github.com/epics-docs/epics-docs/pull/134

PR #3431, opened 2025-06-12
fix reproducible builds based on suggestions from #3416

shroffk:
based on #3416 

keeping just a small subset of the git properties which do not impact the reproducible builds


hboutemy (approved):
very good approach, thanks

Issue #3384, opened 2025-04-25
Could we have a dedicated 'Close Display' action?


rjwills28:
Many of our existing CS-Studio OPI screen have dedicated 'EXIT' buttons as part of the UI design that run a script that calls close on the display. The same can be done in Phoebus but we were wondering whether instead we could add a dedicated 'Close display' action (much like the 'Open Display' action) to avoid the complication of having to use scripts.

I appreciate that the screens already have the 'x' on the top right of the display but many of users still use the EXIT button and we believe good UI design does involve redundancy.

Note I would be happy to look into implementing this if the community thinks it would be useful to have.

rjwills28:
As further motivation to have a dedicated action to close a display, I have come across a case where if I disable the property 'Always show tabs' I can end up with a layout where I can't actually close a particular display without re-enabling the 'Always show tabs', e.g.

![image](https://github.com/user-attachments/assets/fb6a3136-f1a0-4970-a8cb-e0825b76dbf0)

In this case it would be useful to have an 'EXIT' ActionButton.

rjwills28:
Yes - thanks!

PR #3437, opened 2025-06-18
Add a 'Close Display' action

rjwills28:
I looked into adding a dedicated 'close display' action that can be used in ActionButtons. This is following my request in https://github.com/ControlSystemStudio/phoebus/issues/3384.

As I explain in that issue, it is nice to have some redundancy in screens and so offer the user multiple ways to close the display, i.e. in addition to the 'x' at the top of the display/window. Note, that as also explained in that issue, this isn't always an option if you have 'Always show tabs' disabled and you are using split windows. 

This is a simple action that just closes the current display. 

Please let me know what you think?


rjwills28:
I have checked that all resources are tidied up by this new action to close a display and I can confirm that the method I call to close the display `DockItemRepresentation.closeWindow(...)`: https://github.com/ControlSystemStudio/phoebus/blob/09b7db292297249e4f33fc6ddcd16c56fc0ea3bd/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DockItemRepresentation.java#L146
calls `DockItem.prepareToClose()`: https://github.com/ControlSystemStudio/phoebus/blob/09b7db292297249e4f33fc6ddcd16c56fc0ea3bd/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DockItemRepresentation.java#L155
which performs checks to make sure that it is OK to closer the display:
https://github.com/ControlSystemStudio/phoebus/blob/09b7db292297249e4f33fc6ddcd16c56fc0ea3bd/core/ui/src/main/java/org/phoebus/ui/docking/DockItem.java#L672

This is the same method called when you click the cross on the top of a display, therefore I conclude that resources are cleaned up in the same way.

I have also used the PV List app in Phoebus to confirm that when you close a display using the new action in an action button, the connections to the PVs are closed and the PV List shows that no PVs are in use.

shroffk:
@rjwills28 thank you for double checking

PR #3442, opened 2025-06-23
feat: Add Widget, dropdown menu with autocomplete from predefined list

flef:


georgweiss:
@flef , not entirely clear to me how this is supposed to work.
I guess the idea is to write a PV value selected from either of the items in the dropdown, and I suppose the items can be populated from some other (enum?) PV. How does one specify items if not pulled from a PV?

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/AutoCompleteRepresentation.java
```diff
@@ -0,0 +1,597 @@
...
+                } catch (NumberFormatException e) {
+                    VType convertedValue = VType.toVType(value);
+                    mappedValue = (convertedValue != null) ? convertedValue : value;
+                }
+            }
+
+            toolkit.fireWrite(model_widget, mappedValue);
+        } catch (Exception e) {
+            System.err.println("Error writing to PV: " + e.getMessage());
+            e.printStackTrace();
```
kasemir: Instead of println and e.printStackTrace, use logger:
```
import static org.csstudio.display.builder.representation.ToolkitRepresentation.logger;
...
logger.log(Level.WARNING, "Error writing to PV", e);
```

kasemir: Also, I assume the most likely use case is with string PVs, right? Enum is already handled with combo box widget. Is String handled via the fall-through code? Not enumerated, parsing as double or int fails, so fall back to `VType.toVType(value)`? Maybe add a comment there:
```
// This will handle String PVs, which are expected to be most likely
```

flef: Thank you for your reviews, I have made the suggested improvements.

georgweiss:
I have a usability concern:
If I type a (string) value not present in suggestion list and press <ENTER>, the value is written to the PV. If on the other hand I type a value that is present in the list, pressing <ENTER> will not write the value. Instead I need to select form the dropdown to trigger a write operation.

This raises the question: should values other than those offered in the list be accepted?

It would also be of value to add an example (or extend existing) in the ```app/display/model/src/main/resources/examples``` folder.

flef:
I'll push example.

georgweiss:
Thanks, but I still find it a bit confusing that pressing the \<ENTER\> key when typing a value present in the list does **not** write to PV.

shroffk:
Thanks florian
Do you think you will be able to attend the developers meeting on the 14th. We had a few questions about your work and its potential to replace/augment text update and combobox

kasemir:
> It could eventually replace the ComboBox

I like that idea! What are the remaining differences?

Combo: Enter list of "Items" or select "Items from PV"

New widget: "Autocomplete Item" contains "Option1\nOption2\nOption3"

The combo approach seems better for the typical use cases: Get options from PV, fall back to a handful of "Items".
It doesn't scale to 10,000 possible values. There, one long "Autocomplete Item" string, or even an external file seems better.

kasemir:
As for InfluxDB, we had been working on support for that, see presentation under http://www.rri.kyoto-u.ac.jp/EPICS/program.html

It was faster than an RDB, but we eventually decided to not use it because

 * It's implemented in go, a somewhat uncommon language

 * The data is transferred in JSON, and the actual implementation resulted in lost detail for small/large floating point values, which was a problem for our vacuum readings which are something like "1e-20" which turned into "0". Reported, https://github.com/influxdata/influxdb-java/issues/276, but closed as “wont-fix” which was at the time quite frustrating.

 * While influx supports auto-aggregation, for example replacing older samples with the 1-hour-average to reduce the data size, how it does that exactly by storing the data under a different name wasn't convenient. The data retrieval would basically need to look in several different places: Is there recent data? Is there data reduced to the 1 hour avg? Is there data reduced to the daily avg?

 * Compared to RDB systems with a track record of having been available for decades and keeping data for that long, influx is still quite new. Cool for system logs or your Arduino fishtank controller data from last month. But control system data going back to 2006??

flef:
Thank you for the information, I am interested in the branch that supports InfluxDB. I wasn’t aware of this rounding issue. That is indeed unfortunate. I will check if this bug is still present in InfluxDB versions 2 and 3.
I think it might be interesting for me to present our usage context, as I feel that our use case is a bit different. On our side, we collect telemetry data at relatively high frequency (more than 5,000 values at over 250 Hz) over relatively short periods (from 1 hour to a maximum of 2 days).

flef:
@kasemir @georgweiss , can one of you send me an invitation for the meeting please ?

PR #3443, opened 2025-06-23
Expand macros in OpenNodeActionHandler

DenSinH:
The title says it all really, we want to be able to pass SaR node IDs as macros, this seems like the most general way of doing so.


georgweiss:
@shroffk, FYI: the action handler editor UI does not support macros.

PR #3444, opened 2025-06-23
Add explicit check for response code when downloading attachments for http olog. 

Peyton-Hill-CLS:
Currently, the olog http client does not check the response code from the server when downloading attachments, this results in the saved images sometimes containing non image data. This can happen if attachments are deleted from the olog database for example. This pr is an attempt to improve the error handling and logging of this.


georgweiss:
@Peyton-Hill-CLS, thanks for adding check and logging.

PR #3447, opened 2025-06-27
Add 'in' operator for scan server scripts

DenSinH:
We want to use the scan server to run some procedures, in which we want to wait for a state (a discrete value) to be any of a specific set of values (or not any of a specific set of values). Since this was not really possible with the scan server as it is now, we decided to implement it. There may be better ways of doing this though, as I am currently not super happy about the comma separated value. Perhaps someone here has a better idea, but I couldn't think of a nicer way of doing this, without implementing a completely new command or causing big changes to the current wait command implementation. Feedback is welcome, but perhaps this is already fine as it is.

### DIFF
--- a/app/scan/model/src/main/java/org/csstudio/scan/command/Comparison.java
+++ b/app/scan/model/src/main/java/org/csstudio/scan/command/Comparison.java
@@ -43,7 +43,13 @@ public enum Comparison
     INCREASE_BY("to increase by"),
 
     /** Value has decreased by some amount */
-    DECREASE_BY("to decrease by");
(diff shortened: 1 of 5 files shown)


DenSinH:
Okay so I added string list parsing. I saw that the `StringOrDouble.parse(...)` method does some 'string parsing' (i.e. removing starting and ending quotes) which will break in input string like `"ERROR", "INVALID"` as it will strip the outer quotes... Just checking if the comparison expects a list will fix this, so it works now.

kasemir:
Thanks!

PR #3449, opened 2025-06-30
PVA: Track searches by CID

### LINKED ISSUE #3446: Display of PV valueAlarm fields intermittent
When a bob file is created to display a PV via pvaccess (e.g. pva://root:ai1), Phoebus client connects to the PV and updates. However, when asked to display the PV alarm or warning limit fields, the widgets do not display the field in question reliably.

Points to note

- Example PV/field     pva://root:ai1/valueAlarm/highAlarmLimit
- Phoebus client built to commit 49e516e063f9a9c0132fdf1542a025bbdc5b1763
- Connection appears to work sometimes (Re-load Display works intermittently - feels possibly like an immediate timeout)
- Client on Windows operating system
- Java 21, Phoebus 5.0.1 SNAPSHOT
- Using standard example EPICS base IOC running in docker. Simple bob file attached.
- Using a pvaccess gateway, but I think this is fine due to below.
- Older Phoebus client works.
- Phoebus client built to commit 98f3e5906a37d9b3a28f28c52ccd4789e60fa4c1 does appear to work.

[test.zip](https://github.com/user-attachments/files/20927588/test.zip)

kasemir:
Fixes #3446 

In #3430, PVA searches were properly cleaned up once a reply arrived... by name.

When using PVs "pva://x/field1" and "pva://x/field2", these would both search for "x", then use different "field(...)" specs in the subscription. Since searches were cleaned by name, a search might get cleared before it ever went out.

This update tracks searches by client channel ID, so only the search that received a reply gets cleared.

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/ChannelSearch.java
+++ b/core/pva/src/main/java/org/epics/pva/client/ChannelSearch.java
@@ -101,19 +101,31 @@ class ChannelSearch
             // to handle it
         }
 
-        // Hash by channel name
+        // Searches are identified by CID because we might have the
+        // same channel name in different searches when we try to access
+        // the same channel with different field(...) qualifiers.
+        // One could try to optimize this by fetching "field()" (everything)
+        // and then picking the subelements in the client,
+        // but in case there's only a single PV for
+        //     pva://GigaBytePV/substruct/double_field
+        // we'd want to use "field(substruct.double_field)"
+        // and avoid fetching the complete structure.
+        // ... unless there is later a PV "pva://GigaBytePV",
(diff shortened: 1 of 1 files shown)


shroffk:
This scenario was fun to read about

kasemir:
> This scenario was fun to read about

To me, network protocol issues like these always start out as a horror scenario. You need several PVs to duplicate the problem, then increase the log level or capture Wireshark traces. Either way, you get many pages of protocol dumps and wonder where to start.
In the end, it was a very simple fix and mostly comments to explain what's happening, quite satisfying.

PR #3457, opened 2025-07-04
Add check for HEIC file attachments

georgweiss:
HEIC may be the default iOS file format for images. We have seen users attaching such files to Olog, but since HEIC is not supported in JavaFX, and only in Safari browser, it makes sense to disallow HEIC attachments.

This PR adds a check on file extension. If HEIC file is encountered, the attach file(s) operation aborts with an error dialog:

Screenshot 2025-07-04 at 14 19 31

On the Olog service side a similar check will be added. 

HEIC to JPG conversion has been evaluated as an option. Only to find that it is not: needs either third party app (ImageMagic) or a closed source lib at $1600 per/year.

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/Messages.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/Messages.java
@@ -40,6 +40,7 @@ public class Messages
             GroupingFailed,
             GroupSelectedEntries,
             JumpToLogEntry,
+            HeicFilesNotSupported,
             Level,
             Logbook,
             LogbookNotSupported,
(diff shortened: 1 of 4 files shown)


georgweiss:
Improved this a bit, for a good reason. HEIC image files from IOS converted to JPEG fail to render unless one uses Safari, or maybe some image file extension on OS level. In order to detect these as files converted from a HEIC file, we need to employ Apache Tika. This PR hence pulls in a new 3rd party lib (~750 kb), but ensures that image files that fail to render in Olog clients (web browsers included, though not Safari) cannot be uploaded.

shroffk:
is there a open source library available to help up provide support for this file format?

georgweiss:
For the sake of completeness: I found https://github.com/openize-com/openize-heic-java, which _appears_ to be a Java-based heic/heif converter. However, pull from Maven fails despite having an entry in Maven Central. 

Pull the source code to evaluate, but documentation is a disappointment, and my attempts to convert a file either failed or hanged. Overall quality seems inadequate for our use case.

PR #3474, opened 2025-07-23
CSSTUDIO-3337 Restore behavior of widgets after PR to `epicsCoreJava`

abrahamwolk:
The background for this pull request is described in https://github.com/epics-base/epicsCoreJava/pull/143.

This pull request:

1. Updates the widgets `Text Update`, `Text Entry`, and `Spinner` to display `<PV-Name>` when a value is received that has alarm status equal to `Alarm.disconnected()`. (This restores the behavior of the widgets prior to the changes in the pull request  https://github.com/epics-base/epicsCoreJava/pull/143.)
2. Updates the tool tip-functionality for widgets so that the alarm ranges are not displayed when a value with alarm status equal to `Alarm.disconnected()` is received.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/RegionBaseRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/RegionBaseRepresentation.java
@@ -266,7 +266,7 @@ abstract public class RegionBaseRepresentation<JFX extends Region, MW extends Vi
                 if (alarm != null  &&  alarm.getSeverity() != AlarmSeverity.NONE)
                     // Have alarm info
                     severity = alarm.getSeverity();
-                else if (value instanceof VType)
+                else if (value != null)
                     // VType that doesn't provide alarm, always OK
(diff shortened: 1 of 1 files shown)


kasemir (approved):

shroffk:
While I understand this is a good fix.
@abrahamwolk I would keep the epics core issue open since I feel like you and Kay have raised valid points about the confusion around undefined and invalid.

at the very least we might want to document some of that conversation.

kasemir:
Since this change I get 
```
java.lang.NullPointerException:
Cannot invoke
"org.csstudio.display.builder.model.WidgetProperty.getValue()"
because "this.value_prop" is null
	at org.csstudio.display.builder.representation.javafx.widgets.RegionBaseRepresentation.connectionOrValueChanged(RegionBaseRepresentation.java:258)
```

In RegionBaseRepresentation.java:258, there's this:
```
final VType value = value_prop.getValue();
```

`value_prop` may be null.  Before this PR, it was

```
if (value_prop != null  && ...
       final Object value = value_prop.getValue();
```

and something similar is still found a few lines later, but the initial access to `value_prop` without null check needs to be fixed.

Issue #3458, opened 2025-07-07
Duplicate displays get open if already open from a memento


rjwills28:
We open a memento which, in this example has a 'Main' display in one pane and a 'Topup properties' display in another pane. From the 'Main' display you can also click an action button to display the 'Topup properties' display. As this display is already open one would expect it to just put the 'Topup properties' display into focus, however instead it opens the 'Topup properties' display again so we have 2, see screenshot below:

![Image](https://github.com/user-attachments/assets/450f0fa7-53e6-44bc-85e9-c26ab85b4609)

I looked into the src code and found that it does check whether the screen already exists based on the URI, see https://github.com/ControlSystemStudio/phoebus/blob/f67cf148f9150e72cfe44f9c35409ef0d3cac531/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeApplication.java#L82

With some debugging I found that the screen loaded from the memento has the _URI: /home/bob/topup/topup_properties.bob_, while the one opened from the action button has _URI: /home/bob/topup/topup_properties.bob?EXAMPLE_MACRO=Value+from+Preferences&LCID=WDf39eba2&TEST=true_. Because of this, Phoebus does not detect that this screen is already open.

My question is why do these example macros get appended to the screen? Note, that this screen has no macros attached to it.
- The `EXAMPLE_MACRO` is being added here: https://github.com/ControlSystemStudio/phoebus/blob/f67cf148f9150e72cfe44f9c35409ef0d3cac531/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/DisplayRuntimeInstance.java#L415 
from some prefined value in Preferences: https://github.com/ControlSystemStudio/phoebus/blob/f67cf148f9150e72cfe44f9c35409ef0d3cac531/app/display/model/src/main/resources/display_model_preferences.properties#L35 
- The LCID looks to be coming from https://github.com/ControlSystemStudio/phoebus/blob/f67cf148f9150e72cfe44f9c35409ef0d3cac531/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/EmbeddedDisplayWidget.java#L316.

kasemir:
FRIB requested it for their legacy displays, see #1405, #1407
Not sure if that's still necessary, it's been 5 years.
I'm fine with removing it.

PR #3475, opened 2025-07-23
Issue-3458: fix opening duplicate displays due to arbitrary macros added to URI

rjwills28:
PR to fix duplicate displays opening because some default macros get added to the URI.

See https://github.com/ControlSystemStudio/phoebus/issues/3458 for full details.

I have remove the legacy LCID macro that was being added to the URI when opening embedded screens (now the method in the parent class will just be called). If the same screen was already open (e.g. when starting Phoebus loading a memento), Phoebus could not identify that the screen was already open as the URIs did not match (due to the opening display having this LCID macro attached).

In addition, if the `org.csstudio.display.builder.model/macros` property is not overwritten in settings.ini, the default macros added to the URI in the above case are `<EXAMPLE_MACRO>Value from Preferences</EXAMPLE_MACRO><TEST>true</TEST>`. I propose that the default in this case should be to add no macros. The user can then use the mentioned property to specify these if needed.

I have tested and this fixes the problem of duplicate screens opening.

### DIFF
--- a/app/display/model/src/main/resources/display_model_preferences.properties
+++ b/app/display/model/src/main/resources/display_model_preferences.properties
@@ -31,7 +31,8 @@ font_files=examples:font.def
 # * Must start with character
 # * May then contain characters or numbers
 # * May also contain underscores
-# 
+# E.g.
+#  macros=<EXAMPLE_MACRO>Value from Preferences</EXAMPLE_MACRO><TEST>true</TEST>
 macros=
 
 


app/display/model/src/main/resources/display_model_preferences.properties
```diff
@@ -32,7 +32,7 @@ font_files=examples:font.def
 # * May then contain characters or numbers
 # * May also contain underscores
 # 
-macros=<EXAMPLE_MACRO>Value from Preferences</EXAMPLE_MACRO><TEST>true</TEST>
```
kasemir, line 35: Can we keep an example in here? User needs to know if the syntax is `EXAMPLE_MACRO=Xyz` or `<EXAMPLE_MACRO>Xyz</EXAMPLE_MACRO`.

How about this:
```
#macros=<EXAMPLE_MACRO>Value from Preferences</EXAMPLE_MACRO><TEST>true</TEST>
macros=
```

```

kasemir: Removing LCID is good, and making the default for macros nothing instead of that EXAMPLE_MACRO and TEST is also good, just keep the example as a comment

kasemir (approved):
👍

app/display/model/src/main/resources/display_model_preferences.properties
rjwills28: Good idea! I have added a comment with an example following your suggestion.

rjwills28:
Thanks for catching this @georgweiss and sorry I missed it! PR with updated test case: https://github.com/ControlSystemStudio/phoebus/pull/3516

PR #3484, opened 2025-07-31
Migration

jacksoes:
-➕ Add necessary dependencies from old PVWS client.

-🚗Move foundational classes from old PVWS client such as PVWSdata modeling, modular utils classes and main SessionHandler client class.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -23,6 +23,7 @@
     <module>launcher</module>
     <module>vtype</module>
     <module>websocket</module>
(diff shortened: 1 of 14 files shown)


kasemir:
What's the status of this?
Reading from a PVWS might be low performance, but could be a very practical approach when you can't use UDP searches or get access to CA or PVA ports.
Still, PVA should be the way forward since it's about to support IPv6 and security.

PR #3486, opened 2025-08-01
CSSTUDIO-2954 Add support for informative tooltips for widget properties

abrahamwolk:
This pull request adds support for informative and hopefully helpful tooltips for widget properties in the Display Editor. Example:

informativeTooltip

The motivation for this functionality is to increase the user friendliness of the Display Editor. In this pull request, I have added some informative tooltips to the (non-runtime-) properties defined in the classes `Widget`, `VisibleWidget` and `PVWidget`. Of course, it would be nice to add tooltips to more properties in the future.

The implementation adds a field `informativeTooltip` to the class `WidgetProperty`, together with a setter and a getter. The informative tooltip is set by calling the setter, typically directly after a property has been created. Logic has been added to the class `PropertyPanelSection` to detect and add the informative tooltip if it exists.

Of course, it would be possible to extend the list of parameters of the `abstract` method `WidgetPropertyDescriptor.createProperty()` to include a parameter of the form `Optional<String> informativeTooltip`. However, there are 70 implementations of this method, and therefore I chose this approach instead, where `WidgetProperty.setInformativeTooltip()` must be called separately after a property has been created.


kasemir:
Would it be more practical to not change the WidgetProperty nor any Widget?

Instead, in the editor, implement `possiblyInformativeTooltip` to look into some map of property name to info tooltip.
It would be initialized from a file like this:

```
# File editor property info
x=The x-coordinate of the widget counted in pixels from the left side of the screen.
visible=Should the widget be visible?
```

So the only change would be in the editor. It maps the property name to an optional tooltip.
The downside would be that it has no clue about the actual property.
For a property "name", it wouldn't really know if that's a widget name, group name, trace name, ...
Still, having to add this to every property all over the code base is a lot of work for something that we could live without in the last 30 years.

Well, if you have the spare cycles to do it, go ahead.

PR #3492, opened 2025-08-05
PVA: Open server info API

kasemir:
Allows for example a gateway to show connected clients


core/pva/src/main/java/org/epics/pva/client/PVAClientMain.java
```diff
@@ -351,6 +383,12 @@ else if (command.equals("put") && names.size() == 2)
                 request = "value";
             put(names.get(0), names.get(1));
         }
+        else if (command.equals("call") && names.size() > 0)
```
jacomago: Maybe it would be good to make "call", "get", ... an enum, PVACommand

kasemir: Results in 10 more lines, but if that rocks your boat, did it

PR #3493, opened 2025-08-06
Implement dynamic window title management based on active dock tabs

### LINKED ISSUE #3480: Feature Request: Option to Set Window Title to Active Tab Name
It would be great to have an option (maybe in the config file) to make the window title show the name of the currently active tab. This would really help when you have a bunch of windows open, since right now they all have the same name and it's hard to tell which is which at a glance.

One thing to consider is that a window can have multiple split panes, each with its own tab. Maybe just using the tab that was last focused would work fine. That seems like the most intuitive option.

This feature would be optional and off by default so it doesn’t change behavior unexpectedly. If this sounds like a good idea, I’d be happy to implement it and submit a pull request.

Thanks!

DKastelic:
As discussed, this pull request resolves #3480.

It adds a class that sets up listeners that listen to to tab focus events, which update the window title to include the focused tab. 
The format of the window title is configurable with the message resource `WindowTitleFormat=CS-Studio: %s`, with %s being replaced by the active tab name. In the edge case where the main window has no active tabs, it falls back to `FixedTitle`. 

It also slightly improves the logic determining which panel is currently active when splitting and merging them, as before it could lead to situations where a tab that no longer existed would still be deemed the "active tab".

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXStageRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXStageRepresentation.java
@@ -49,7 +49,6 @@ public class JFXStageRepresentation extends JFXRepresentation
     public Parent configureStage(final DisplayModel model, final Consumer<DisplayModel> close_request_handler)
     {
         final String name = model.getDisplayName();
-        stage.setTitle(name);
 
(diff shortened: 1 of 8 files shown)


kasemir (approved):

shroffk:
Do you think it would be too much trouble to make this a preference instead of a message? It is a bit easier to configure.

Also, a minor change would be to avoid wildcard imports.

Overall this is a useful feature.

PR #3462, opened 2025-07-09
Add smoothing and waveform overlaping to databrowser

minglan:
These two tools were presented at the EPICS Collaboration Meeting held in April 2025. The EPICS community has requested that I share the code. The code was initially rough, so I have made extensive modifications over the past months. This involved:

- Adding diverse algorithms;

- Optimizing algorithms for handling massive datasets to enhance computational efficiency and minimize loading latency;

- Implementing various UI components to improve usability.

- Extensive testing was also conducted to resolve bugs, ensuring the tools' stability and reliability.

Key modifications：

- Two classes are modified：
    1. org.csstudio.trends.databrowser3.ui.Perspective.java；
    2. org.csstudio.trends.databrowser3.Messages.java. 

- One Resource Bundle file is modified:       
    1. app\databrowser\src\main\resources\org\csstudio\trends\databrowser3\messages.properties

- Two icons are added to the app\databrowser\src\main\resources\icons folder
    1. smooth.png;
    2. overlap.png. 

- Two new packages are added: 
    1. org.csstudio.trends.databrowser3.ui.smoothview;
    2. org.csstudio.trends.databrowser3.ui.waveformoverlapview.

Any suggestions and bug reports are welcome to help improve the tools. Your feedback is invaluable for testing and refinement!

PR #3494, opened 2025-08-06
CSSTUDIO-3374 Set border of `model_root` in Display Editor when it is focused directly using CSS

abrahamwolk:
This pull request implements the setting of the blue border when Display Editor is focused directly using CSS instead of using an event handler to manually add and remove a style class.

The background for this pull request is that I am currently developing a new widget for Phoebus that displays a Waterfall Plot. When selecting an instance of the new Waterfall Plot Widget and then clicks on the background of the OPI (so that `model_root` is selected), the CPU usage goes constantly to over 100%. Through debugging, I have concluded that the cause is the event handler on `model_root` that adds and removes the CSS style class that sets the border of `model_root` in order to indicate that it is selected.

This pull request removes the event handler and replaces it with a CSS implementation of `widget_pane:focused`. I think this is an improvement over the implementation using an event handler, while it at the same time solves the performance issue I encounter.

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/DisplayEditor.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/DisplayEditor.java
@@ -199,7 +199,7 @@ public class DisplayEditor
     public Parent create ()
     {
         model_root = toolkit.createModelRoot();
-        model_root.getStyleClass().add("widget_pane_unfocused");
+        model_root.getStyleClass().add("widget_pane");
         autoScrollHandler = new AutoScrollHandler(model_root);
 
         final Group scroll_body = (Group) model_root.getContent();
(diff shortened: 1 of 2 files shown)


kasemir:
Looks good!

kasemir:
How complex is your waterfall plot widget? In the scene graph, is it one image, or is it a node tree with lines, labels, rectangles, ...?

jacomago:
> > How complex is your waterfall plot widget? In the scene graph, is it one image, or is it a node tree with lines, labels, rectangles, ...?
> 
> I think it's quite complex. It's based on the https://github.com/fair-acc/chart-fx/ library, which provides a waterfall plot. E.g., it contains axes, which contain labels, etc.

Maybe it overlaps partially with https://github.com/ControlSystemStudio/phoebus/pull/3462 ?

Issue #3502, opened 2025-08-12
Multi-line TextEntry widget does not show 'no_write' cursor


rjwills28:
I found an issue with the `TextEntry` widget when it is `disabled`/`no_write`. If the 'Multi-Line' property is true for this widget, then the `no_write` cursor (red 'x') does not show when you hover over the disabled widget. Instead it just appears on the border of the widget, which is not very obvious. 
See below where it does not appear when hovering over the main body of the widget (but does on the border)

<img width="69" height="48" alt="Image" src="https://github.com/user-attachments/assets/8073f775-397b-45e5-a93b-2bd35843c015" />
<img width="69" height="48" alt="Image" src="https://github.com/user-attachments/assets/407cf686-283b-41d3-9c96-1760b4316d11" />

This is because a `TextArea` is used to display a multi-line text entry widget. Setting the cursor for this node sets it for the `ScrollPane` only, hence is only shows around the borders (see discussion: https://stackoverflow.com/questions/25629938/javafx-textarea-setcursor-not-working/25630584#25630584). 
Following advice from this post (https://stackoverflow.com/questions/56628369/javafx-change-cursor-in-textarea/56630506), it looks like the cursor needs to be set for the _content_ of the TextArea.

PR with this fix is on its way.

PR #3503, opened 2025-08-12
Fix no_write cursor not showing for multi-line TextEntry widget

rjwills28:
PR to fix the case where a disabled `TextEntry` widget with 'multi-line' enabled does not show the 'no-write' cursor over the whole widget - see https://github.com/ControlSystemStudio/phoebus/issues/3502.

A fix for this is to set the cursor for the `TextArea` `content` node. To do this I have set a unique ID for each text entry widget and use this to get the content node from the scene and then set the cursor on that node.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/TextEntryRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/TextEntryRepresentation.java
@@ -9,6 +9,7 @@ package org.csstudio.display.builder.representation.javafx.widgets;
 
 import static org.csstudio.display.builder.representation.ToolkitRepresentation.logger;
 
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
@@ -51,6 +52,8 @@ public class TextEntryRepresentation extends RegionBaseRepresentation<TextInputC
      */
     private boolean active = false;
     private volatile boolean enabled = true;
+    private static final String NODE_ID = UUID.randomUUID().toString();;
+
 
     private final DirtyFlag dirty_size = new DirtyFlag();
     private final DirtyFlag dirty_style = new DirtyFlag();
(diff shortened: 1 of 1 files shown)


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/TextEntryRepresentation.java
```diff
@@ -51,6 +52,8 @@ public class TextEntryRepresentation extends RegionBaseRepresentation<TextInputC
      */
     private boolean active = false;
     private volatile boolean enabled = true;
+    private static final String NODE_ID = UUID.randomUUID().toString();;
```
kasemir: Why is the NODE_ID static? Don't you need a new, unique one for each TextEntryRepresentation?

rjwills28: Yes, thanks for catching that! Last minute change and wasn't thinking it through. Fixed in latest commits.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/JFXBaseRepresentation.java
```diff
@@ -369,12 +369,14 @@ public void updateOrder()
      * @param children list of children nodes under the parent widget
      */
     public void setDisabledLook(Boolean enabled, ObservableList<Node> children) {
-        jfx_node.setCursor(enabled ? Cursor.DEFAULT : Cursors.NO_WRITE);
-        if (children != null) {
-            for (Node node : children)
-            {
-                Styles.update(node, Styles.NOT_ENABLED, !enabled);
-            }   
+        if (!toolkit.isEditMode()) {
```
rjwills28, line 372: Adding this check as the disabled look should only be applied for the runtime display. 
Some widgets call this `setDisabled()` method from inside an `if (!toolkit.isEditMode()) {}` which means it behaves correctly but this is not true of all. To be safe we can do it inside this method here to catch all calls to it.

kasemir (approved):

PR #3507, opened 2025-08-13
Fix close checks in save&restore

georgweiss:
This PR fixes close checks in save&restore: while checks were called when closing individual tabs, closing the app or Phoebus did not run checks.

Also: some refactoring in the name of code maintainability.


app/save-and-restore/app/doc/index.rst
```diff
@@ -324,12 +324,25 @@ Prior to restore user has the option to:
...
+   :width: 80%
+
+
 Restore from context menu
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 User may invoke a restore operation (from client or from service) from context menu items in the tree
-view or in the search-and-filer view. In this case user will not have the possibility to unselect specific PVs.
-However, PV items configured as read-only will not be restored.
+view or in the search-and-filer view. In this case user will not have the possibility to deselect specific PVs.
```
abrahamwolk: Typo: `filer` -> `filter`.

app/save-and-restore/app/doc/index.rst
```diff
@@ -324,12 +324,25 @@ Prior to restore user has the option to:
...
+Filtering/exclusion based on PV name will also not be possible.
```
abrahamwolk: Should `will also not be` be `is not`?

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/SaveAndRestoreInstance.java
```diff
@@ -53,21 +55,59 @@ public SaveAndRestoreInstance(AppDescriptor appDescriptor) {
...
+     *
+     * @return <code>false</code> if any of the tabs contains unsaved data, otherwise <code>true</code>.
+     */
+    private boolean mayClose() {
+        if (Platform.isFxApplicationThread()) {
+            return saveAndRestoreController.doCloseCheck();
+        } else {
+            CompletableFuture<Boolean> completableFuture = new CompletableFuture<>();
+            Platform.runLater(() -> {
+                boolean okToClose = saveAndRestoreController.doCloseCheck();
```
abrahamwolk: Should this line be surrounded with a `try`-`catch` construct, that completes `completableFuture` in case an exception occurs? Otherwise, the future may never complete.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/configuration/ConfigurationController.java
```diff
@@ -512,24 +513,28 @@ public void loadConfiguration(final Node node) {
...
-     * Handles clean-up when the associated {@link ConfigurationTab} is closed.
      * A check is made if content is dirty, in which case user is prompted to cancel or close anyway.
      *
      * @return <code>true</code> if content is not dirty or user chooses to close anyway,
      * otherwise <code>false</code>.
      */
     @Override
-    public boolean handleTabClosed() {
+    public boolean doCloseCheck() {
         if (dirty.get()) {
```
abrahamwolk, line 524: `DockItemWithInput` already contains a field `AtomicBoolean is_dirty` with a setter `DockItemWithInput.setDirty()` and getter `isDirty()`. Should those be used instead? If they are used, the third argument to the constructor of `DockItemWithInput` should probably be set, see documentation about the parameter `save_handler`: https://github.com/ControlSystemStudio/phoebus/blob/master/core/ui/src/main/java/org/phoebus/ui/docking/DockItemWithInput.java#L95-L123

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/CompositeSnapshotController.java
```diff
@@ -423,17 +423,21 @@ public void loadCompositeSnapshot(final Node node) {
...
-    public boolean handleTabClosed() {
+    public boolean doCloseCheck() {
         if (dirty.get()) {
             Alert alert = new Alert(Alert.AlertType.CONFIRMATION);
-            alert.setTitle(Messages.closeTabPrompt);
+            alert.setTitle(Messages.closeCompositeSnapshotTabPrompt);
             alert.setContentText(Messages.closeCompositeSnapshotWarning);
+            DialogHelper.positionDialog(alert, borderPane, -200, -200);
             Optional<ButtonType> result = alert.showAndWait();
             return result.isPresent() && result.get().equals(ButtonType.OK);
```
abrahamwolk, line 440: If I understand this correctly, the user may be alerted several times, for each `ConfigurationController` and for each `CompositeSnapshotController`. Could these checks be consolidated into only one modal window?

app/save-and-restore/app/doc/index.rst
georgweiss: Will update

app/save-and-restore/app/doc/index.rst
georgweiss: Will update

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/SaveAndRestoreInstance.java
georgweiss: Close checks should be "safe", but sure, exception could be thrown by JavaFX.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/configuration/ConfigurationController.java
georgweiss: Decided to not use ```is_dirty``` in ```DockItemWithInput``` as the save&restore app maintains tabs which all would need to report back to ```DockItemWithInput```.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/CompositeSnapshotController.java
georgweiss: Yes, but having multiple unsaved tabs when closing the save&restore app should be a rare case.

PR #3508, opened 2025-08-13
Support delta value sorting on string PVs

georgweiss:
In the save&restore snapshot view, the delta columns supported sorting only on numeric scalar types. Deltas on string PVs where evaluated as zero diff even though they were treated as non-equal.

This PR enhances sorting to consider deltas for string scalar types.


shroffk:
I think we could create a group of comparators which are smart enough to work with some of our columns which might have multiple types.

PR #3514, opened 2025-08-19
Improvements to the Linear Meter widget

abrahamwolk:
This pull request implements several improvements to the Linear Meter widget:

1. Perhaps most noticeably, it implements a new mode: the indicator can now be drawn as a bar in addition to the existing functionality to be drawn as a needle. The following screenshot shows the bar-mode and the needle-mode side by side:screenshot
2. With the goal of improving performance and responsiveness on multicore processors, the concurrency of the Linear Meter has been made more fine-grained. Previously, concurrency issues were avoided by both reading and modifying state only on the JavaFX Application thread using `Platform.runLater()`. This pull request replaces that strategy by instead relying on a `ReentrantReadWriteLock`, allowing for simultaneous read-accesses while write-operations are sequential.
3. The widget property "Limits from PV" has been generalized: In addition to allowing for the options "Limits from PV" and "No limits from PV", the property now also allows for the options "Min & max from PV" and "Alarm limits from PV", allowing the user more fine-grained control over this option. This functionality has been implemented in a manner such that OPIs saved using the previous version of the Linear Meter will still be loaded correctly. However, an OPI created using the new version of the Linear Meter implemented in this pull request should not be loaded using an older version of Phoebus.
4. It is now possible to reduce the width of the widget to 10 pixels at minimum instead of 25 pixels.

### DIFF
--- a/app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/LinearMeterRepresentation.java
+++ b/app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/LinearMeterRepresentation.java
@@ -288,86 +288,88 @@ public class LinearMeterRepresentation extends RegionBaseRepresentation<Pane, Li
 
     private void valueChanged(WidgetProperty<?> property, Object old_value, Object new_value) {
         if (new_value instanceof VDouble) {
-            VDouble vDouble = ((VDouble) new_value);
-            double newValue = vDouble.getValue();
+            meter.withWriteLock(() -> {
+                VDouble vDouble = ((VDouble) new_value);
+                double newValue = vDouble.getValue();
 
-            if (!Double.isNaN(newValue)) {
-                if (Double.isNaN(observedMin) || newValue < observedMin) {
-                    observedMin = newValue;
-                    newObservedMinAndMaxValues = true;
-                }
+                if (!Double.isNaN(newValue)) {
+                    if (Double.isNaN(observedMin) || newValue < observedMin) {
+                        observedMin = newValue;
+                        newObservedMinAndMaxValues = true;
+                    }
 
-                if (Double.isNaN(observedMax) || newValue > observedMax) {
-                    observedMax = newValue;
-                    newObservedMinAndMaxValues = true;
-                }
-            }
-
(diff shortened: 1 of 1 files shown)


app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
```diff
@@ -1114,7 +1109,7 @@ else if (value >= high && value < hiHi) {
                                 gc.setPaint(needleColor);
                             }
                             // Draw the bar:
-                            gc.fillRect(marginLeft, currentIndicatorPosition, meterBreadth, linearMeterScale.getBounds().height-currentIndicatorPosition-marginBelow);
+                            gc.fillRect(marginLeft, currentIndicatorPosition, meterBreadth - 1, linearMeterScale.getBounds().height-currentIndicatorPosition-marginBelow);
```
georgweiss: Found in several commits:

Math symbols (minus) could be surrounded by space for better readability.

kasemir:
I meant to refer to the LinearMeterWidget.METER_VERSION:
```
/**
 * 1.0.0: Linear meter by Claudio Rosatti based on 3rd party library
 * 2.0.0: Simple linear meter, based on meter v 3.0.0, drawn in background
 */
public static Version METER_VERSION = new Version(2, 0, 0);
```

A new version 3.0.0 would allow the LinearMeterConfigurator or other tools that create displays to know if the bar mode and the additional limits-from-pv behavior is available.

As for minimum width, we do have some large and cryptic overview displays full of this:

Screenshot 2025-08-19 at 8 52 06 AM

They are of course only meaningful to those in the know, but they use down to 1-pixel wide bars to indicate information.

app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
abrahamwolk: Fixed: https://github.com/ControlSystemStudio/phoebus/pull/3514/commits/fb5ad0970d3e8551436b8942ebf6005d1fb70f30

PR #3520, opened 2025-08-25
CSSTUDIO-1355 Add Waterfall Plot widget to Phoebus.

abrahamwolk:
This pull request adds a new Waterfall Plot widget to Phoebus:

<p align="center">
screenshot
</p>

The intended usage of the widget is to display the evolution of a waveform or of a number of (compatible) PVs over time. The widget plots the values of the waveform or the list of PVs from a configurable amount of time in the past until the current moment. For simplicity, the time is always displayed in UTC in the initial implementation. The plot is updated once per second.

The implementation of the widget uses the ChartFX [1] library for its implementation.

I have added the widget to the Maven-build files by adding it to the `pom.xml` file for `phoebus-product`. At the moment, it is enabled by default in the pull request. Would it be preferable that it is disabled by default?

Unfortunately, I am not familiar withe the Ant build system, and I have therefore not added it to this build system. Help to add it to the Ant build system is welcome.

One limitation is that the widget currently does not allow for dynamically changing the connected PV(s) when it is running.

When testing the widget and the widget doesn't display any values, it may be helpful to check the setting "Z-axis min & max" to make sure it is correctly set.

[1] https://github.com/fair-acc/chart-fx

### DIFF
--- a/dependencies/phoebus-target/pom.xml
+++ b/dependencies/phoebus-target/pom.xml
@@ -558,6 +558,13 @@
       <artifactId>tika-core</artifactId>
       <version>3.2.0</version>
     </dependency>
+    
+    <!-- ChartFX library for the Waterfall Plot widget -->
+    <dependency>
+      <groupId>io.fair-acc</groupId>
+      <artifactId>chartfx</artifactId>
+      <version>11.3.1</version>
+    </dependency>
   </dependencies>
 
   <build>


app/display/waterfallplot/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+            <version>5.0.3-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.phoebus</groupId>
+            <artifactId>core-types</artifactId>
+            <version>5.0.3-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>io.fair-acc</groupId>
+            <artifactId>chartfx</artifactId>
```
kasemir, line 31: Can you add the new dependency to dependencies/phoebus-target/pom.xml?

abrahamwolk: It's possible, but isn't it preferable that only the module `waterfallplot` includes it? If one builds Phoebus without the Waterfall plot, then there is no need to include the new dependency in the build.

georgweiss (approved):

app/display/waterfallplot/pom.xml
kasemir: See https://github.com/ControlSystemStudio/phoebus/tree/master/dependencies#readme
This is about establishing a target platform, a list of all the dependencies, for use with Eclipse, ant, make, .. whatever you want to use. If you limit the build tools to maven, then you're of course correct.

abrahamwolk: I have added the dependency to `dependencies/phoebus-target/pom.xml`: https://github.com/ControlSystemStudio/phoebus/pull/3520/commits/7bc807541fa921d28498c7b0f75f62bb6ba5e760

PR #3521, opened 2025-08-27
Olog websockets

georgweiss:
This PR adds web sockets to Olog ES clients (Log Entry Table and Log Entry Calendar).

As the implementation uses Spring Framework web socket libs, two new dependencies are added, total ~2MB. If Olog is not used, site specific products should exclude ```core-websocket``` to avoid these dependencies.

A dedicated class ```WebSocketClientService``` is intended to be somewhat generic. It handles reconnections as well as keep-alive (aka ping/pong).

The idea is to skip polling for new log entries twice per minute, but rather rely on web socket messages dispatched from the associated Olog service.

**NOTE**: the Olog web client will through this PR depend on https://github.com/Olog/phoebus-olog/pull/220.


shroffk:
this is great,
can we wait till a new release before merging this since it has breaking changes

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryTableViewController.java
```diff
@@ -398,7 +413,7 @@ public void search() {
                 },
                 (msg, ex) -> {
                     searchInProgress.set(false);
-                    ExceptionDetailsErrorDialog.openError(Messages.LogbooksSearchFailTitle, ex.getMessage(), null);
+                    //ExceptionDetailsErrorDialog.openError(Messages.LogbooksSearchFailTitle, ex.getMessage(), null);
```
abrahamwolk: Commented-out code.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
```diff
@@ -23,74 +32,109 @@
...
+        webSocketClientService.connect(webSocketUrl);
     }
 
-    /**
-     * Utility method to cancel any ongoing periodic search jobs.
-     */
-    public void shutdown() {
-        cancelPeriodSearch();
+    @Override
+    public void handleWebSocketMessage(String message) {
```
abrahamwolk, line 213: If two new log entries are created more or less simultaneously, will this function be called in parallel, or will the two calls occur sequentially (in that case, presumably on the same thread)?

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
```diff
@@ -23,74 +32,109 @@
...
+        try {
+            WebSocketMessage webSocketMessage = objectMapper.readValue(message, WebSocketMessage.class);
+            if (webSocketMessage.messageType().equals(MessageType.NEW_LOG_ENTRY)) {
```
abrahamwolk, line 216: Should the two other message types (`MessageType.LOG_ENTRY_UPDATED` and `MessageType.SHOW_BANNER`) also be handled by this function?

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
```diff
@@ -544,6 +556,8 @@ public LogTemplate fromString(String name) {
 
         setupTextAreaContextMenu();
 
+        //doStompStuff();
```
abrahamwolk: Commented-out code.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
```diff
@@ -1066,4 +1080,40 @@ private void loadTemplate(LogTemplate logTemplate) {
             selectedLogbooks.forEach(l -> updateDropDown(logbookDropDown, l, false));
         }
     }
+
+    private void doStompStuff(){
```
abrahamwolk: This function is not called. (The only call to it (on line `559`) is commented out.)

core/websocket/src/main/java/org/phoebus/core/websocket/springframework/WebSocketClientService.java
```diff
@@ -0,0 +1,263 @@
...
+        if (path.endsWith("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
+        this.contextPath = path;
+        this.webSocketConnectUrl = scheme + "://" + host + (port > -1 ? (":" + port) : "") + this.contextPath + "/web-socket";
+        doConnect();
+    }
+
+    /**
+     *
```
abrahamwolk: Empty comment.

core/websocket/src/main/java/org/phoebus/core/websocket/springframework/WebSocketClientService.java
```diff
@@ -0,0 +1,263 @@
...
+            stompSession.send(contextPath + "/web-socket/echo", message);
+        }
+
+    }
+
+    /**
+     * Disconnects the STOMP session if non-null and connected.
+     */
+    public void disconnect() {
+        if(stompSession != null && stompSession.isConnected()) {
```
abrahamwolk: Does this class need to be made thread-safe, perhaps by means of synchronization? E.g., could `stompSession` be assigned to between the check `stompSession != null`, the check `stompSession.isConnected()`, and the call `stompSession.disconnect()`?

georgweiss: I do not think this needs to be synchronized. ```stompSession``` is ```null``` only initially, i.e. as long as API client has not called ```connect``` and the connection attempt succeeds. Hence the null check.

core/websocket/src/main/java/org/phoebus/core/websocket/springframework/WebSocketClientService.java
```diff
@@ -121,6 +121,14 @@ public void disconnect() {
         }
     }
 
+    /**
+     * Disconnects the socket if connected and terminates connection thread.
+     */
+    public void shutdown(){
+        disconnect();
+        attemptReconnect.set(false);
```
abrahamwolk, line 135: Should `attemptReconnect.set(false)` be called _before_ `disconnect()`, to prevent a reconnection from occurring immediately after `disconnect()` has finished but before `attemptReconnect.set(false)` has finished?

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCalenderApp.java
```diff
@@ -17,8 +17,8 @@ public class LogEntryCalenderApp implements AppResourceDescriptor {
 
     public static final Logger logger = Logger.getLogger(LogEntryCalenderApp.class.getName());
     static final Image icon = ImageCache.getImage(LogEntryCalenderApp.class, "/icons/logbook-16.png");
-    public static final String NAME = "logEntryCalender";
-    public static String DISPLAYNAME = "Log Entry Calender";
+    public static final String NAME = "logEntryCalendar";
+    public static String DISPLAYNAME = "Log Entry Calendar";
 
     private static final String SUPPORTED_SCHEMA = "logCalender";
```
abrahamwolk, line 23: `SUPPORTED_SCHEMA` should likely also be changed to `logCalendar`.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
```diff
@@ -39,22 +48,62 @@ public abstract class LogbookSearchController implements WebSocketMessageHandler
...
+        int port = uri.getPort();
+        String path = uri.getPath();
+        if (path.endsWith("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
+        String webSocketScheme = scheme.toLowerCase().startsWith("https") ? "wss" : "ws";
+        this.webSocketUrl = webSocketScheme + "://" + host + (port > -1 ? (":" + port) : "") + path;
+    }
+
+    protected void determineConnectivity(Runnable completionHandler){
```
abrahamwolk: 1. Given the name `determineConnectivity()`, I would expect the function to return the "connectivity".
2. The function takes as argument a continuation-like function `completionHandler()`, which in turn is passed an argument implicitly through `connectivityModeObjectProperty`. Could the type of `completionHandler` be changed from `Runnable` to `Consumer<ConnectivityMode>` and the connectivity mode be passed explicitly to `determineConnectivity()`? I think it would make the code easier to understand.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
```diff
@@ -175,8 +175,9 @@ private void cancelPeriodSearch() {
      * Utility method to cancel any ongoing periodic search jobs or close web socket.
      */
     public void shutdown() {
-        if (webSocketClientService != null) {
+        if (connectivityModeObjectProperty.get().equals(ConnectivityMode.WEB_SOCKETS_SUPPORTED) && webSocketClientService != null) {
```
abrahamwolk, line 183: What happens if `shutdown()` is called very quickly after the logbook application is opened, so that `shutdown()` is called before the connection has been established?

core/websocket/src/main/java/org/phoebus/core/websocket/springframework/WebSocketClientService.java
georgweiss: You have a point, these calls should be reversed.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogEntryCalenderApp.java
georgweiss: This is legacy code, will not change it.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
georgweiss: Then connectivity mode is ```NOT_CONNECTED```

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
georgweiss: Yes, why not...

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
abrahamwolk: But then the `webSocketClientService` will not be shutdown?

georgweiss: There would not be anything to shut down.

abrahamwolk: To me it seems there may be: `connectWebSocket()` is called asynchronously using `JobManager.schedule()` in `determineConnectivity()`. What if `connectWebSocket()` has not run yet when `shutdown()` is called?

georgweiss: You're right, this needs a bit of synchronization...

core/websocket/src/main/java/org/phoebus/core/websocket/springframework/WebSocketClientService.java
abrahamwolk: But the following `Thread` is launched, that (potentially) updates `stompSession`: https://github.com/ControlSystemStudio/phoebus/pull/3521/files#diff-8280f3eb08fcfd51f5c7f2b8c964e5277e353856254a4008c06d287ff966b6fcR159-R185

georgweiss: Potentially, yes, but highly unlikely. Added synchronization all the same.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/LogbookSearchController.java
shroffk: I guess we can just add `|| MessageType.LOG_ENTRY_UPDATED` since the desired behaviour is to simply query again right @georgweiss ?

PR #3539, opened 2025-09-10
Support search for referenced snapshots

georgweiss:
As per user request: provide a tool for finding composite snapshot referencing a snapshot or other composite snapshot. 

This has been implemented as a context menu item (see screen shot) which based on the selected node opens the search and filter view to perform a search. The search result list will reveal in what composite snapshots the slected node has been found.

Server side hence supports a new type of query: referenced=<node id>.

Screenshot 2025-09-10 at 14 16 55

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
@@ -86,6 +86,7 @@ import org.phoebus.applications.saveandrestore.ui.contextmenu.CopyUniqueIdToClip
 import org.phoebus.applications.saveandrestore.ui.contextmenu.CreateSnapshotMenuItem;
 import org.phoebus.applications.saveandrestore.ui.contextmenu.EditCompositeMenuItem;
 import org.phoebus.applications.saveandrestore.ui.contextmenu.ExportToCSVMenuItem;
+import org.phoebus.applications.saveandrestore.ui.contextmenu.FindReferencesMenuItem;
 import org.phoebus.applications.saveandrestore.ui.contextmenu.ImportFromCSVMenuItem;
 import org.phoebus.applications.saveandrestore.ui.contextmenu.LoginMenuItem;
 import org.phoebus.applications.saveandrestore.ui.contextmenu.NewCompositeSnapshotMenuItem;
@@ -232,6 +233,7 @@ public class SaveAndRestoreController extends SaveAndRestoreBaseController
                     }),
             new SeparatorMenuItem(),
             new EditCompositeMenuItem(this, selectedItemsProperty, this::editCompositeSnapshot),
+            new FindReferencesMenuItem(this, selectedItemsProperty, this::findReferences),
             new RenameFolderMenuItem(this, selectedItemsProperty, this::renameNode),
             copyMenuItem,
             pasteMenuItem,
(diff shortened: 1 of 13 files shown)


app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
```diff
@@ -1506,4 +1508,9 @@ public boolean doCloseCheck(){
         }
         return true;
     }
+
+    private void findReferences(){
+        SearchAndFilterTab searchAndFilterTab = openSearchWindow();
+        searchAndFilterTab.getController().findReferencesForSnapshot(treeView.getSelectionModel().getSelectedItems().get(0).getValue().getUniqueId());
```
abrahamwolk, line 1514: Could `treeView.getSelectionModel().getSelectedItems()` return an empty result?

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/contextmenu/FindReferencesMenuItem.java
```diff
@@ -0,0 +1,33 @@
...
+                                  Runnable onAction) {
+        super(saveAndRestoreController, selectedItemsProperty, onAction);
+        setGraphic(ImageCache.getImageView(ImageCache.class, "/icons/sar-search_18x18.png"));
+        setText(Messages.findSnapshotReferences);
+    }
+
+    @Override
+    public void configure() {
+        visibleProperty().set(selectedItemsProperty.size() == 1 &&
+                (selectedItemsProperty.get(0).getNodeType().equals(NodeType.SNAPSHOT) ||
```
abrahamwolk, line 30: Same question here as above:  could `selectedItemsProperty` be empty?

services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/search/SearchUtil.java
```diff
@@ -90,10 +106,9 @@ public SearchRequest buildSearchRequest(MultiValueMap<String, String> searchPara
                         for (String pattern : getSearchTerms(value)) {
                             String term = pattern.trim().toLowerCase();
                             // Quoted strings will be mapped to a phrase query
-                            if(term.startsWith("\"") && term.endsWith("\"")){
+                            if (term.startsWith("\"") && term.endsWith("\"")) {
```
abrahamwolk, line 109: Can `term` be equal to `"` (i.e., consists of only the character '`"`')? If so, `term.substring(1, term.length() - 1)` on the next line will throw an `IndexOutOfBoundsException`.

georgweiss: That's a good catch: of course user could specify a single ```"```. Will fix this.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/contextmenu/FindReferencesMenuItem.java
georgweiss: No, because this is called when user right-clicks on an item in the ```TreeView```, which implicitly selects that node.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
georgweiss: No, because this is called when user right-clicks on an item in the ```TreeView```, which implicitly selects that node.

services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/search/SearchUtil.java
georgweiss: Actually, this is already covered in the call to ```getSearchTerms```. So no changes.

georgweiss:
@shroffk, OK to merge this

PR #3542, opened 2025-09-10
Enable support for influxdb v2 as data source (PV)

flef:
The purpose of this PR is to enable the use of an InfluxDB v2 database as a data source.

InfluxDB is an open-source time series database designed to efficiently store and query large volumes of time-stamped data, such as monitoring metrics, events, and real-time analytics. It is optimized for high write and query performance, making it a popular choice for industrial applications and data monitoring solutions.

As @kasemir said : 
__________

As for InfluxDB, we had been working on support for that, see presentation under http://www.rri.kyoto-u.ac.jp/EPICS/program.html

It was faster than an RDB, but we eventually decided to not use it because

It's implemented in go, a somewhat uncommon language

- The data is transferred in JSON, and the actual implementation resulted in lost detail for small/large floating point values, which was a problem for our vacuum readings which are something like "1e-20" which turned into "0". Reported, https://github.com/influxdata/influxdb-java/issues/276, but closed as “wont-fix” which was at the time quite frustrating.

- While influx supports auto-aggregation, for example replacing older samples with the 1-hour-average to reduce the data size, how it does that exactly by storing the data under a different name wasn't convenient. The data retrieval would basically need to look in several different places: Is there recent data? Is there data reduced to the 1 hour avg? Is there data reduced to the daily avg?

- Compared to RDB systems with a track record of having been available for decades and keeping data for that long, influx is still quite new. Cool for system logs or your Arduino fishtank controller data from last month. But control system data going back to 2006??

Also, the two options we have are pretty good: RDB for those who prefer generality, Archive Appliance for those who prefer something optimized for control system  data.
___________

InfluxDB has made some questionable choices regarding its open source version in version 3. We are currently exploring alternatives (as [QuestDB](https://questdb.com/?utm_source=GoogleAds&utm_medium=CPC&utm_content=OSS&utm_campaign=Adoption&gad_source=1&gad_campaignid=20888431615&gclid=EAIaIQobChMIpZi2sr7OjwMVAj8GAB0jgyGgEAAYASAAEgLuw_D_BwE)) but we have no other option at the moment since certain existing systems rely on this database.

see also: [72 hours limit to fetch data](https://www.influxdata.com/blog/influxdb3-open-source-public-alpha-jan-27/?_gl=1*p89790*_ga*MTQwMDk1NDQ2My4xNzU3NTE3MjU3*_ga_CNWQ54SDD8*czE3NTc1MTcyNTckbzEkZzEkdDE3NTc1MTcyNzUkajU2JGwwJGgxMjg5NTM4MDMz*_gcl_au*Njc1MDc1MzEwLjE3NTc1MTcyNTk.)

### DIFF
--- a/app/trends/archive-reader/pom.xml
+++ b/app/trends/archive-reader/pom.xml
@@ -45,6 +45,11 @@
             <version>0.0.10</version>
         </dependency>
         <dependency>
+            <groupId>com.influxdb</groupId>
+            <artifactId>influxdb-client-java</artifactId>
+            <version>${influxdb2Client.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.phoebus</groupId>
             <artifactId>core-ui</artifactId>
             <version>5.0.3-SNAPSHOT</version>
--- a/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/influx2/InfluxArchiveReader.java
+++ b/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/influx2/InfluxArchiveReader.java
@@ -1,0 +1,582 @@
+/*******************************************************************************
+ * Copyright (C) 2025 Thales.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.phoebus.archive.reader.influx2;
+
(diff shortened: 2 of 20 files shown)


flef:
We encountered limitations with InfluxDB: when the database receives a large volume of data, the most recently written entry becomes readable only after a 10-second delay, which is unacceptable for our systems.

We are migrating to QuestDB, where our initial tests show no such limitations.
I will keep you informed of the ongoing benchmark and test results, and will propose an implementation in Phoebus.

PR #3546, opened 2025-09-12
CSSTUDIO-3420 Waterfall Plot widget optimization: Remove data points that are not plotted.

abrahamwolk:
This pull request implements the removal of data-points that are not plotted in the Waterfall Plot widget.

The motivation for this change is that for large waveform PVs (on the order of 10000 indices), I get OutOfMemory-exceptions when plotting the waveforms for 30 minutes to 1 hour.

A drawback of removing data-points is that detail may be lost when resizing the plot, which may occur either because the widget is resized or because one of the axes of the waterfall plot changes size (e.g., if an axis changes from being `[0.0, 1.0, ..., 9.0]` to `[0.0, 1.0, ..., 10.0]`, requiring one more digit to be displayed).


jacomago:
LTTB - Largest Triangle Three Bucket

Joao gave me:

https://www.base.is/flot/ 
https://github.com/sveinn-steinarsson/flot-downsample 
https://github.com/parkertomatoes/lttb-cpp

PR #3554, opened 2025-09-19
Login to all services

georgweiss:
As per user request...

The credentials manager app now offers a way to sign in to all services (e.g. logbook, save&restore) through a single action. 
In case any of the services rejects user's credentials (or is off-line), the dialog stays open (see screen shot) to indicate the problem.

Documentation updated accordingly.

Screenshot 2025-09-19 at 13 19 37


app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -126,51 +202,79 @@ public void updateItem(Void o, boolean empty) {
...
+            for (ServiceItem serviceItem : tableView.getItems()) {
+                serviceItem.username.set(loginToAllUsernameProperty.get());
+                serviceItem.password.set(loginToAllPasswordProperty.get());
+                login(serviceItem, tableView.getItems().size());
+            }
+            if (loggedInCount.get() == tableView.getItems().size()) {
+                stage.close();
+            }
+        } catch (Exception e) {
+            LOGGER.log(Level.WARNING, "Failed to delete all authentication tokens from key store", e);
```
abrahamwolk: Is this error message correct?

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -126,51 +202,79 @@ public void updateItem(Void o, boolean empty) {
...
-            serviceItem.getServiceAuthenticationProvider().authenticate(serviceItem.getUsername(), serviceItem.getPassword());
+            serviceItem.getServiceAuthenticationProvider().authenticate(serviceItem.getUsername().get(), serviceItem.getPassword().get());
             try {
                 secureStore.setScopedAuthentication(new ScopedAuthenticationToken(serviceItem.getAuthenticationScope(),
-                        serviceItem.getUsername(),
-                        serviceItem.getPassword()));
-                stage.close();
+                        serviceItem.getUsername().get(),
+                        serviceItem.getPassword().get()));
+                loggedInCount.set(loggedInCount.get() + 1);
```
abrahamwolk: Reading from and writing to `loggedInCount` in this way is not thread-safe.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -126,51 +202,79 @@ public void updateItem(Void o, boolean empty) {
...
             }
-        } catch (Exception exception) {
-            LOGGER.log(Level.WARNING, "Failed to login to service", exception);
-            ExceptionDetailsErrorDialog.openError(parent, "Login Failure", "Failed to login to service", exception);
+        } catch (ConnectException e) {
+            serviceItem.setLoginResultMessage(Messages.ServiceConnectionFailure);
+        } catch (ServiceAuthenticationException exception) {
+            serviceItem.setLoginResultMessage(Messages.UserNotAuthenticated);
         }
+        serviceItem.setLoginResult(LoginResult.FAILED);
```
abrahamwolk: Do I understand the code correctly here: is the login-result _always_ set to `LoginResult.FAILED` here when `login()` is called?

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -126,51 +202,79 @@ public void updateItem(Void o, boolean empty) {
...
+    private void logOut(ServiceItem serviceItem) {
+        if (serviceItem.userLoggedIn.not().get()) {
+            return;
+        }
         try {
-            secureStore.deleteScopedAuthenticationToken(scope);
-            updateTable();
+            serviceItem.serviceAuthenticationProvider.logout();
+            serviceItem.setLoggedOut();
+            loggedInCount.set(loggedInCount.get() - 1);
```
abrahamwolk: This assignment is not thread-safe.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -207,27 +313,36 @@ private void updateTable() {
...
-        private boolean loginAction = false;
+        private final StringProperty username = new SimpleStringProperty();
+        private final StringProperty password = new SimpleStringProperty();
+        private final BooleanProperty userLoggedIn = new SimpleBooleanProperty();
+        private final StringProperty buttonTextProperty = new SimpleStringProperty();
+        private final StringProperty loginResultMessage = new SimpleStringProperty();
+        private final ObjectProperty<LoginResult> loginResult = new SimpleObjectProperty<>(LoginResult.OK);
 
         public ServiceItem(ServiceAuthenticationProvider serviceAuthenticationProvider, String username, String password) {
             this.serviceAuthenticationProvider = serviceAuthenticationProvider;
```
abrahamwolk, line 302: I think it may be a good idea to change the type of `serviceAuthenticationProvider` from `ServiceAuthenticationProvider` to `Optional<ServiceAuthenticationProvider>` since `serviceAuthenticationProvider` may be `null`. (See lines `287`-`290`: https://github.com/ControlSystemStudio/phoebus/pull/3554/files#diff-8682c52d79413fbe3e61310d72e1daa306a698af4c7efe00f97ffa2e16952350R287-R290.) This way, the fact that the field may be empty is reflected in its type. See also line `358` [1] and line `364` [2].

[1] https://github.com/ControlSystemStudio/phoebus/pull/3554/files#diff-8682c52d79413fbe3e61310d72e1daa306a698af4c7efe00f97ffa2e16952350R358
[2] https://github.com/ControlSystemStudio/phoebus/pull/3554/files#diff-8682c52d79413fbe3e61310d72e1daa306a698af4c7efe00f97ffa2e16952350R364

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -181,19 +285,21 @@ private void updateTable() {
...
             }).collect(Collectors.toList());
             // Also need to add ServiceItems for providers not matched with a saved token, i.e. for logged-out services
             authenticationProviders.forEach(p -> {
                 Optional<ServiceItem> serviceItem =
                         serviceItems.stream().filter(si ->
-                                p.getAuthenticationScope().equals(si.getAuthenticationScope())).findFirst();
-                if(serviceItem.isEmpty()){
+                                p.getAuthenticationScope().getScope().equals(si.getAuthenticationScope().getScope())).findFirst();
+                if (serviceItem.isEmpty()) {
                     serviceItems.add(new ServiceItem(p));
```
abrahamwolk: This is not thread-safe.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -239,87 +354,128 @@ public AuthenticationScope getAuthenticationScope() {
...
-            textField.setOnKeyReleased(ke -> getTableRow().getItem().setUsername(textField.getText()));
+        public void setLoginResult(LoginResult loginResult) {
+            this.loginResult.set(loginResult);
         }
 
+        public void setLoggedOut() {
+            userLoggedIn.set(false);
+            username.set(null);
+            password.set(null);
+            loginResultMessage.set(null);
```
abrahamwolk: Would it be better to use the type `Optional<String>` instead of `String` for the fields `username`, `password`, and `loginResultMessage`?

app/credentials-management/src/main/resources/org/phoebus/applications/credentialsmanagement/CredentialsManagement.fxml
```diff
@@ -40,23 +40,31 @@
                 <Label text="%NoCredentialsFound" />
             </placeholder>
             <columns>
-                <TableColumn fx:id="scopeColumn" prefWidth="${parent.width * 0.2}" text="%Scope">
+                <TableColumn fx:id="scopeColumn" style="-fx-alignment: CENTER-lEFT;" text="%Scope">
```
abrahamwolk: Typo: `lEFT` should be `LEFT`.

app/credentials-management/src/main/resources/org/phoebus/applications/credentialsmanagement/CredentialsManagement.fxml
```diff
@@ -40,23 +40,31 @@
...
                 </TableColumn>
-                <TableColumn fx:id="passwordColumn" editable="true" prefWidth="${parent.width * 0.3}" text="%Password">
+                <TableColumn fx:id="passwordColumn" editable="true" prefWidth="110.0" text="%Password">
                     <cellValueFactory>
                         <PropertyValueFactory property="password" />
                     </cellValueFactory>
                 </TableColumn>
-                <TableColumn fx:id="actionButtonColumn" prefWidth="${parent.width * 0.2}" />
+                <TableColumn fx:id="actionButtonColumn" prefWidth="70.0" />
+                <TableColumn fx:id="loginResultColumn" editable="false" prefWidth="150.0" style="-fx-alignment: CENTER-lEFT;" text="%LoginResult">
```
abrahamwolk: Typo: `lEFT` should be `LEFT`.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
```diff
@@ -19,32 +19,39 @@
...
         try {
             OlogHttpClient.builder().build().authenticate(username, password);
-        } catch (Exception e) {
+        }
+        catch(ConnectException e){
+            throw e;
+        }
+        catch (Exception e) {
             Logger.getLogger(OlogServiceAuthenticationProvider.class.getName())
                     .log(Level.WARNING, "Failed to authenticate user " + username + " with logbook service", e);
```
abrahamwolk: Can the stack trace of `e` contain sensitive information? The exception originates from the call to `OlogHttpClient.builder().build().authenticate(username, password)` which contains `username` and `password`. I think it may be safer to not include the exception `e` in the log, and also not to re-throw it (https://github.com/ControlSystemStudio/phoebus/pull/3554/files#diff-1cc8e66eabcdb21075575c28e20439ca032c7b00a5770c4044c369028a6d2eadR44).

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreClientImpl.java
```diff
@@ -588,6 +588,8 @@ public UserData authenticate(String userName, String password) {
                     .build();
             HttpResponse<String> response = CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
             return OBJECT_MAPPER.readValue(response.body(), UserData.class);
+        } catch (ConnectException e) {
+            throw e;
         } catch (Exception e) {
             throw new RuntimeException(e);
```
abrahamwolk: Same as above: is it possible that the exception `e` contains sensitive information, since it originates in a call that contains `userName` and `password`? I think it may be safer if `e` is not re-thrown (both on this line and on line `592`).

core/security/src/main/java/org/phoebus/security/authorization/ServiceAuthenticationProvider.java
```diff
@@ -31,13 +34,19 @@ public interface ServiceAuthenticationProvider {
...
      * Signs out user from the service.
-     * @param token Username or other type of token (e.g. session cookie).
      */
-    void logout(String token);
+    default void logout(){
+        try {
+            SecureStore secureStore = new SecureStore();
+            secureStore.deleteScopedAuthenticationToken(getAuthenticationScope());
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
abrahamwolk, line 46: I think it's safer not to re-throw an exception here, since it originates from a call to `secureStore`.

core/security/src/main/java/org/phoebus/security/store/SecureStore.java
```diff
@@ -197,18 +199,21 @@ private List<ScopedAuthenticationToken> matchEntries(List<String> aliases) throw
...
                 username = get(tokens[0]);
                 password = get(PASSWORD_TAG);
             }
             else{
-                scope = AuthenticationScope.fromString(tokens[0]);
-                username = get(scope.getName() + "." + USERNAME_TAG);
-                password = get(scope.getName() + "." + PASSWORD_TAG);
+                scope = findAuthenticationScopeFromProviders(tokens[0]);
+                if(scope == null){
+                    throw new IllegalArgumentException("No authentication provider found matching scope \"" + tokens[0] + "\"");
```
abrahamwolk, line 213: I think it's not safe to throw exceptions from within a class called `SecureStore`, since exceptions may contain sensitive information.

Could a failure be handled in a type-safe way in the code by a suitable return-type? Perhaps the function could have the return type `Optional<List<ScopedAuthenticationToken>>` and return `Optional.empty()` in case it is not successful.

core/security/src/main/java/org/phoebus/security/authorization/ServiceAuthenticationProvider.java
```diff
@@ -31,13 +34,19 @@ public interface ServiceAuthenticationProvider {
      * @param username User name
      * @param password Password
      */
-    void authenticate(String username, String password);
+    void authenticate(String username, String password) throws ConnectException;
```
abrahamwolk: I suggest _not_ to throw exceptions in `authenticate()`, but instead to change its return-type to something that can represent failure. Perhaps `boolean` would work: `true` indicates success and `false` indicates failure to login. The motivation is to avoid accidental printing of sensitive information in stack traces or similar.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: Good catch, copy paste mistake

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: Why does it need to be thread safe? All login/logout API calls are synchronous.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: You're right, there is a return statement missing.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: See previous comment

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
georgweiss: No. But I'll remove the exception anyway.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: In practice the ```ServiceAuthenticationProvier``` should never be ```null``` as saved tokens are associated only with existing ```ServiceAuthenticationProvier```s. I will update the callee code to handle potential ```null```s.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreClientImpl.java
georgweiss: No. But I'll remove the exception anyway.

core/security/src/main/java/org/phoebus/security/authorization/ServiceAuthenticationProvider.java
georgweiss: The case where exceptions could be thrown is when calling the Java keystore APIs. It would be quite surprising to see those APIs leak sensitive information.

core/security/src/main/java/org/phoebus/security/store/SecureStore.java
georgweiss: See previous comment.

core/security/src/main/java/org/phoebus/security/authorization/ServiceAuthenticationProvider.java
georgweiss: The idea is to catch different type of exceptions and differentiate messages to user in the UI, e.g. "User not authenticated" vs "Service offline".

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreClientImpl.java
georgweiss: Exception not logged.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: Login (via authenticate) and logout APIs are synchronous, so why should this be thread safe?

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: Login (via authenticate) and logout APIs are synchronous, so why should this be thread safe?

core/security/src/main/java/org/phoebus/security/authorization/ServiceAuthenticationProvider.java
georgweiss: Decided to introduce an ```AuthenticationResult``` enum to differentiate between bad credentials and failure to connect. Will update PR shortly.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
abrahamwolk: What if two concurrent calls to `login()` get interleaved? I think that if a computer is slow to respond (maybe it is running out of RAM and/or the CPU is busy), then the UI may become unresponsive and the user may click on "Login" several times, and the calls may be interleaved.

georgweiss: That's a good point. Will disable login buttons such that new login attempts cannot be invoked until ongoing has completed.

abrahamwolk: Disabling the button is a good idea, I think, but I think the calls need to be made synchronous also, since state is being modified. `setDisable()` may also not finish until the button has already been pressed twice.

georgweiss: Sure, that could happen. Will secure concurrency.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreClientImpl.java
```diff
@@ -577,21 +576,17 @@ public List<Node> deleteTag(TagData tagData) {
...
-                    .header("Content-Type", CONTENT_TYPE_JSON)
-                    .POST(HttpRequest.BodyPublishers.ofString(OBJECT_MAPPER.writeValueAsString(new LoginCredentials(userName, password))))
-                    .build();
-            HttpResponse<String> response = CLIENT.send(request, HttpResponse.BodyHandlers.ofString());
-            return OBJECT_MAPPER.readValue(response.body(), UserData.class);
-        } catch (ConnectException e) {
-            throw e;
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+    public void authenticate(String userName, String password) throws Exception {
```
abrahamwolk: The declared exception that may be thrown is now more general (`Exception` instead of `ConnectException`) than before (i.e., there is less static information about it): is this correct?

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -175,7 +170,8 @@ public TableCell<ServiceItem, ServiceItem> call(final TableColumn<ServiceItem, S
                         btn.getStyleClass().add("button-style");
                         btn.setOnAction((ActionEvent event) -> {
                             ServiceItem serviceItem = getTableRow().getItem();
-                            if (serviceItem != null && serviceItem.userLoggedIn.get()) {
+                            if (serviceItem != null && (serviceItem.authenticationStatus.get().equals(AuthenticationStatus.AUTHENTICATED) ||
```
abrahamwolk, line 173: I think the action of this button also needs to be synchronized.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -314,33 +290,51 @@ private void updateTable() {
...
-    public static class ServiceItem {
+    public class ServiceItem {
         private final ServiceAuthenticationProvider serviceAuthenticationProvider;
         private final StringProperty username = new SimpleStringProperty();
         private final StringProperty password = new SimpleStringProperty();
-        private final BooleanProperty userLoggedIn = new SimpleBooleanProperty();
         private final StringProperty buttonTextProperty = new SimpleStringProperty();
         private final StringProperty loginResultMessage = new SimpleStringProperty();
-        private final ObjectProperty<LoginResult> loginResult = new SimpleObjectProperty<>(LoginResult.OK);
+        private final ObjectProperty<AuthenticationStatus> authenticationStatus = new SimpleObjectProperty<>();
```
abrahamwolk, line 298: Should the `SimpleObjectProperty` be initialized with an initial value?

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
```diff
@@ -368,28 +362,30 @@ public String getDisplayName() {
...
+
+        public AuthenticationStatus login() {
+            AuthenticationStatus authenticationStatus = serviceAuthenticationProvider.authenticate(username.get(), password.get());
+            if (authenticationStatus.equals(AuthenticationStatus.AUTHENTICATED)) {
+                try {
+                    secureStore.setScopedAuthentication(new ScopedAuthenticationToken(getAuthenticationScope(),
+                            username.get(),
+                            password.get()));
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
abrahamwolk: I think it's better _not_ to re-throw an exception here (both username and password are arguments to the call in the try-block). (Also `Exception e` can maybe be generalized to `Throwable t`, if the clause is intended to catch all errors.)

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
```diff
@@ -31,22 +32,29 @@ public class OlogServiceAuthenticationProvider implements ServiceAuthenticationP
...
     @Override
-    public void authenticate(String username, String password) throws ConnectException{
+    public AuthenticationStatus authenticate(String username, String password) {
         try {
             OlogHttpClient.builder().build().authenticate(username, password);
-        }
-        catch(ConnectException e){
-            throw e;
+            return AuthenticationStatus.AUTHENTICATED;
+        } catch (ConnectException e) {
```
abrahamwolk: Is the `catch`-clause for `ConnectException` still relevant after the changes in this commit? (The catch-all clause `Exception e` should remain, of course).

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
```diff
@@ -31,22 +32,29 @@ public class OlogServiceAuthenticationProvider implements ServiceAuthenticationP
...
+
+    public OlogServiceAuthenticationProvider() {
         ologAuthenticationScope = new OlogAuthenticationScope();
     }
 
     @Override
-    public void authenticate(String username, String password) throws ConnectException{
+    public AuthenticationStatus authenticate(String username, String password) {
         try {
             OlogHttpClient.builder().build().authenticate(username, password);
```
abrahamwolk: Could the function `authenticate()` return the authentication status, instead of throwing exceptions when failing?

app/logbook/olog/client-es/src/main/java/org/phoebus/olog/es/api/OlogHttpClient.java
```diff
@@ -362,7 +364,6 @@ public void groupLogEntries(List<Long> logEntryIds) throws LogbookException {
      * @throws Exception if the login fails, e.g. bad credentials or service off-line.
      */
     public void authenticate(String userName, String password) throws Exception {
```
abrahamwolk: I suggest to return the authentication status as a value instead of throwing exceptions when there's an error. This makes it much easier to follow the control flow.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
```diff
@@ -31,22 +32,29 @@ public class OlogServiceAuthenticationProvider implements ServiceAuthenticationP
...
-            throw e;
+            return AuthenticationStatus.AUTHENTICATED;
+        } catch (ConnectException e) {
+            logger.log(Level.WARNING, "Unable to connect to logbook service");
+            return AuthenticationStatus.SERVICE_OFFLINE;
+        } catch (ServiceAuthenticationException e){
+            logger.log(Level.WARNING, "User " + username + " not authenticated");
+            return AuthenticationStatus.BAD_CREDENTIALS;
         }
         catch (Exception e) {
```
abrahamwolk: This catch-all clause can be improved by changing it from `Exception e` to `Throwable t`.

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/client/SaveAndRestoreClientImpl.java
georgweiss: That is correct. The change is about intercepting a true authentication response and turn that into an ```ServiceAuthenticationException```. The ```SaveAndRestoreAuthenticationProvider``` then analyses exceptions to return the appropriate ```AuthenticationStatus```

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: The ```login``` and  ```logout``` methods are synchronized.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: I do not see the need.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
georgweiss: See previous comment on how exceptions are handled.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
georgweiss: That is what it does.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
georgweiss: Will not change it, see also https://www.baeldung.com/java-catch-throwable-bad-practice

app/logbook/olog/client-es/src/main/java/org/phoebus/olog/es/api/OlogHttpClient.java
georgweiss: You have a point. Will move the exception handling to the http client classes.

app/credentials-management/src/main/java/org/phoebus/applications/credentialsmanagement/CredentialsManagementController.java
georgweiss: Will update to just log the problem. But I fail to see how credentials could be leaked.

app/logbook/olog/client-es/src/main/java/org/phoebus/applications/logbook/authentication/OlogServiceAuthenticationProvider.java
kasemir: Could still include the exception in the log message to show the reason and stack trace:
```
logger.log(Level.WARNING, "User " + username + " not authenticated", e);
```

PR #3559, opened 2025-09-23
github: add a Pull Request template

minijackson:
This is intended to serve as a reminder for contributors, and hopefully should help keeping the documentation and release notes up-to-date.

This should also save time for reviewers by automatically asking common questions to contributors.

I've only included some common generic questions in the checklist. The PR template doesn't start with a title because GitHub automatically prepend the PR template with the commit message body (so the title would be after that).


georgweiss:
Agree that the template should indicate whether unit tests and/or integration tests have been added.

Issue #3555, opened 2025-09-22
Updated null-check for value_prop limits types that can be in value.


kasemir:
@abrahamwolk, there's another follow-up problem with #3522, #3474.

We used to have
```
final Object value = value_prop.getValue();
```

You changed that into
```
final VType value = value_prop.getValue();
```

We have scripts that set local PVs to arrays etc., so their type is for example "PyList" and now we get this cast exception:
```
WARNING [org.csstudio.display.builder.model] Property update error for 'connected' = true
java.lang.ClassCastException: class org.python.core.PyList cannot be cast to class org.epics.vtype.VType 
      at org.csstudio.display.builder.representation.javafx.widgets.RegionBaseRepresentation.connectionOrValueChanged(RegionBaseRepresentation.java:264)
```

PR #3561, opened 2025-09-24
CSSTUDIO-3463 Bugfix: Don't assume that value_prop.getValue() returns a value of type VType.

abrahamwolk:
This pull request fixes https://github.com/ControlSystemStudio/phoebus/issues/3555.


kasemir (approved):
Thanks, this should restore the previous behavior that allowed scripts to place arbitrary data in local PVs.

PR #3567, opened 2025-09-26
docs: create a Sphinx domain extension to document preference settings

minijackson:
This changes how the preference settings are fetched and put into the documentation.

Instead of creating the `preference_properties.rst` file in `conf.py`, this PR creates a Sphinx extension that parses the preference files, and outputs "objects", like Sphinx would do for documenting a Python or a C API.

Some more information can be found in the ["Adding a reference domain" Sphinx tutorial](https://www.sphinx-doc.org/en/master/development/tutorials/adding_domain.html).

This is quite a bit more Python code than the previous implementation, but gives the following benefits:

- A structured, more readable output than code blocks of properties files

- More readable preference descriptions, now that they are with markup

- A more complete table of contents in the side bar

- Referenceable packages and preferences in the documentation,
  by using the new `preferences:pack` and `preferences:pref` roles

- If a document references a preference that doesn't exist, the Sphinx build will output a warning. This helps quite a bit keeping the documentation up-to-date.

- "[source]" links to GitHub for viewing the properties file

- Indices of preferences, in the "genindex.html" and
  "preferences-index.html" pages

I've been running something similar to this for EPNix, and I found that once done, there's not much maintenance to do.

I have another PR ready after this one that changes the doc-comments inside properties files so that they have proper markup, but I wanted to propose a PR as small as possible (and it's already quite big).


minijackson:
@georgweiss Yes, apparently I developed the extension for a more recent version of Sphinx, so I had to update the Sphinx version in the `requirements.txt`.

I just pushed it, you might have checked out the PR before that.

But somehow the ReadTheDocs output doesn't show the preferences, and I'm currently trying to fix that too…

jacomago:
Is it possible to have the source link to a file hosted in the read the docs? I think we host the docs inside of phoebus and we might want an airgapped network.

minijackson:
@jacomago technically, it should be possible, but I'd argue there is no need: the entirety of the properties file is rendered on the page. I think the `[source]` link is more useful to developers than Phoebus users.

minijackson:
Updated to fix merge conflict.

PR #3569, opened 2025-09-30
Fix tags selection not shown in text field

georgweiss:


shroffk (approved):
thank you for cleaning up the logbook text creation too

Issue #3577, opened 2025-10-03
Implement Formula Function for Array Downsampling Using LTTB Algorithm


shroffk:
This function should efficiently reduce the number of data points in large arrays while preserving the overall visual characteristics of the original dataset.

PR #3578, opened 2025-10-03
array down sampling with LTTB

shroffk:
Addresses #3577

Includes basic unit tests
Test are passing


shroffk:
downsampled from 1000 to 100 points

georgweiss:
This is cool!
Guess the purpose of the formula function is to off-load plot rendering code, even though raw data must be processed anyway.

georgweiss (commented):
Since by default function arguments are initialized as ```VDouble``` in ```FormulaPV``` L66, ```ArraySampleWithLTTBFunction#compute()``` will throw an exception. At the same time ```ArrayScalarDivisionFunction``` returns ```DEFAULT_NAN_DOUBLE_ARRAY```. Should this be aligned?

shroffk:
A single line log message? a warning

georgweiss:
Maybe a warning including "....maybe due to mismatch in initial type" or similar

PR #3579, opened 2025-10-06
Fix annotation moving to the left when X axis PV changes

mnabywan:
When on plot the X axis PV changes the way that here is no anymore the same x value, the annotation was choosing the last index before. We found the issue while using NTPVs on X and Y axis, which was recalculated every time in the IOC - so we never could find the same sample.

When X axis PV is changing every time, it was causing the issue that annotation was slowly going to the left and finally ending at the beginning.

Before:
image

Finding the closest sample instead should fix the issue (or at least limit it).


mnabywan:
Actually, I was testing annother approach and it may be simpler - instead of searching for closest sample, we can just alternate between less-or-equal and greated-or-equal sample.
 Feel free to take a look.

georgweiss (approved):

PR #3585, opened 2025-10-09
Implement opening File Browser from Top Resources

### LINKED ISSUE #3584: Feature Request: Add the ability to open File Browser from Top Resources
We have a number of network locations for our common files, for instance PLT's, which we access via projects in Eclipse. 

Having a look at doing the same with Phoebus, the best idea I could come up with was accessing them through Top Resources. I have trialed some changes to allow File Browser to be opened via Top Resources. 

The main change is to remove the query component before passing the URI to FileBrowserApp. This means the path to the repo can be added to settings.ini, under top_resources:
```org.phoebus.ui/top_resources= … | file:///path_to_common_folder/Common_PLTs?app=file_browser, Common PLT | …```
<br />

I have also added code to show the current folder name in the tab. Having a few File Browsers open can lead to clicking on a few to find the tab you want. The tab name is updated when the user sets a new path, either via right click or by entering a new path. 

I wasn't sure the best way to format the tab name - the only other tab I could see that displays the app name and another name is Display Editor. I put forward a few suggestions on formats to my colleagues and they suggested that 'File Browser (foldername)' looked the best out of the options. 

Image<br />

As it stands, this is how it looks:    <br />
Top Resources: 
Image<br />  

 Tab:  
Image <br /> 

I will open a PR with the modified code. 

Thanks.

Gir4ffe:
Main change is code to remove the query component before passing the URI to FileBrowserApp. This means the path to the repo can be added to settings.ini, under top_resources. Example: 
```org.phoebus.ui/top_resources= … | file:///path_to_common_folder/Common_PLTs?app=file_browser, Common PLT | … ```

Added a listener to listen to path changes and update the tab name. 
Added the function to get the filebrowser icon to display in top resources. 

Rebuilt using maven and tested that changes work. 

closes #3584  

## Checklist

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/filebrowser/src/main/java/org/phoebus/applications/filebrowser/FileBrowser.java
+++ b/app/filebrowser/src/main/java/org/phoebus/applications/filebrowser/FileBrowser.java
@@ -55,19 +55,50 @@ public class FileBrowser implements AppInstance
             content = new Label("Cannot load UI");
         }
 
+        // final DockItem tab = new DockItem(this, content);
+        // DockPane.getActiveDockPane().addTab(tab);
+
+        // if (controller != null  &&  file != null){
+        //     if(file.isDirectory()){
+        //         controller.setRoot(file);
+        //     }
+        //     else{
+        //         controller.setRootAndHighlight(file);
+        //     }
+        // }
+
+        // tab.addClosedNotification(controller::shutdown);
         final DockItem tab = new DockItem(this, content);
         DockPane.getActiveDockPane().addTab(tab);
 
-        if (controller != null  &&  file != null){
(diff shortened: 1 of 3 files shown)


kasemir (approved):
As Georg commented, "File Browser (folder)" makes for a rather large tab text. That could be abbreviated or use a folder icon. But that might require a 'docking' update because we already place a Label in the tab to support drag/drop, so let's make that a separate update, if desired.
This PR looks good as far as opening the file browser from top resources is concerned, and showing the folder in the tab is basically also good.

Issue #3590, opened 2025-10-14
Allow a custom 'not_enabled' css styling


rjwills28:
Would it be possible to add a property for facilities to customize the look of 'not_enabled'/'no_write' widgets?

At the moment, the default is to set the opacity to 40% (as per the disabled look in JavaFX). 

I opened a ticket a while back showing how this wasn't always ideal - see https://github.com/ControlSystemStudio/phoebus/issues/3368 discussion and below is an example of where the opacity is set to 40% but gives a strange look on the widget as the green background is coming through:
<img width="156" height="163" alt="Image" src="https://github.com/user-attachments/assets/002421e3-7244-48ac-b7ac-7c72cbf895f1" />

I explored alternative ways of displaying 'not_enabled' widgets but couldn't find anything that was better _and_ just worked for all widgets. 

Adding a property where users can override the default behaviour (i.e. setting the opacity to 40%) seems like a nice alternative. In the case above we could for example keep the opacity at 1.0.

PR #3591, opened 2025-10-14
Add property to configure the 'not_enabled' css styling for widgets

rjwills28:
I have implemented the changes required to allow the 'not_enabled' css style to be configured by facilities using a Phoebus property. This would allow users to override the default 40% opacity if this doesn't suit their needs.

This PR addresses the feature request in issue https://github.com/ControlSystemStudio/phoebus/issues/3590.

Note: if the property is not set then the default styling will be used so current users will see no changes.

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - [x] The feature is documented - comment in the properties file, which will get automatically picked up when generating preferences documentation.
    - [x] The documentation is up to date
    - Release notes:


rjwills28:
Thanks for the feedback @kasemir - much appreciated. Yes, that was a good catch - it needs to be able to remove the custom styling when the widget becomes enabled again without having to reload the page.

I have updated the code to do so. I've switched around some of the statements to make sure that we handle all cases. I've also surrounded the custom 'not_enabled' styling with comments in the css string signalling the start and end of that styling so that I can ensure that only this bit gets removed from the styling when the widget is enabled.

If you could take another look and let me know what you think that would be much appreciated.

kasemir:
In setSceneStyle, we add the path to the bundled csstudio.css.
As you noted, it's a resource URL.

THere's nothing that prevents adding another `file:/path/to/my/site/adjustments.css`.
You could get that URL, be it a `file:` or `http`, from a new "site_styles" preference.

In there, you can re-define the `not_enabled` style class, or leave it at the default.

This would allow each site to overwrite more than just `not_enabled`.
Surely everybody has their own preferred design for the `tracker_handle`, so that would now be up for local adjustments.

kasemir:
Enthusiastic Yay!

As for a missing `*.css` style sheet, on one hand it seems to be an industry standard to ignore them. My web browser doesn't quit because of a missing style sheet, there's no popup, you need to dig real deep to notice.

But I think we should have a warning in the log. 
Brings up the question how to check... In #3466 and earlier related issues, we tried to optimize via for example `new URL(...).openConnection().connect()` but that would connect fine even for cases where the document doesn't exist. Really need to check via `new URL(...).openStream()` to see if we can read the document. It's somewhat expensive to always read the document, we should not do that inside `setSceneStyle` each time that's called.
Maybe check once inside the static `org.phoebus.ui.Preferences` initializer:
```
    static
    {
        AnnotatedPreferences.initialize(Preferences.class, "/phoebus_ui_preferences.properties");

        // Check once if custom style sheet is accessible
        if (! custom_css_styling.isBlank())
        {
             try
             {
                 new URL(custom_css_styling).openStream().close();
             }
             catch (Exception ex)
             {
                    logger.log(Level.WARNING, custom_css_styling + " is inaccessible", ex);
                    custom_css_styling = "";
             }
     }
```

rjwills28:
Ah that's nice - thanks for the code snippet. I have implemented that as suggested.

rjwills28:
Resolved merge conflicts so this should be ready for re-review and merging if all are happy.

PR #3595, opened 2025-10-20
improve SaR snapshot UI behavior

DenSinH:
Basically, the SaR snapshot controller was a bit unintuitive. The UI suggested you needed to enter a name / description before hitting take snapshot / save, but hitting take snapshot would clear out the metadata entered by the user. I changed it so that the name and the comment are no longer cleared. However, if you start from an existing snapshot, this would preserve the name / description from that existing snapshot (which would be fine IMO, you can start editing from there) but the name needs to be unique. The error message shown to the user on a non-unique node name contained exception names because it was a nested exception. This is no longer the case.

## Checklist

- Testing: Tested new behavior locally

- Documentation: I don't think this is significant enough to warrent a release not entry, nor is it breaking


shroffk:
@georgweiss I have merged the node setting PR... can I merge this and then we can test it again or do you th ink it is safer to update this branch with master and test again

PR #3614, opened 2025-10-30
CSSTUDIO-3524 Fix equal spacing of widgets in Display Builder

abrahamwolk:
This pull request implements a bugfix to the functionality to space a set of widgets equally in either the horizontal or the vertical direction in Phoebus by changing the calculation of the new widget positions to use the type `double` instead of `int`. This prevents rounding errors (caused by the use of integer arithmetic), which, when the set of widgets grows larger, compounds and can become noticeable. I have tested the change manually in the Display Builder.

### DIFF
--- a/app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
+++ b/app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
@@ -506,7 +506,7 @@ public abstract class ActionDescription
                                           .mapToInt(w -> w.propWidth().getValue())
                                           .sum();
 
-            final double offset = ( (double) max - (double) min - (double) totalWidth ) / ( (double) N - 1.0 );
+            final double offset = (double) (max - min - totalWidth) / (double) (N - 1);
             final List<Widget> sortedWidgets = widgets.stream()
                                                       .sorted(( w1, w2 ) ->
             {
@@ -550,7 +550,7 @@ public abstract class ActionDescription
                 //  Equal gap distribution...
                 //  ------------------------------------------------------------
                 Widget widget = sortedWidgets.get(0);
-                double location = widget.propX().getValue();
+                double location = (double) widget.propX().getValue();
                 int width = widget.propWidth().getValue();
 
                 for ( int i = 1; i < N - 1; i++ )
@@ -582,13 +582,13 @@ public abstract class ActionDescription
                 width = widget.propWidth().getValue();
 
                 final int leftCenter = location + width / 2;
-                final double coffset = ( rightCenter - leftCenter ) /  ( (double) N - 1.0 );
+                final double coffset = (rightCenter - (double) leftCenter) / (double) (N - 1);
 
                 for ( int i = 1; i < N - 1; i++ )
                 {
                     widget = sortedWidgets.get(i);
                     width = widget.propWidth().getValue();
-                    undo.execute(new SetWidgetPropertyAction<>(widget.propX(), (int) Math.round(( leftCenter + i * coffset ) - (double) width / 2.0)));
+                    undo.execute(new SetWidgetPropertyAction<>(widget.propX(), (int) Math.round((leftCenter + (double) i * coffset - (double) width / 2.0))));
                 }
             }
         }
@@ -620,7 +620,7 @@ public abstract class ActionDescription
                                            .mapToInt(w -> w.propHeight().getValue())
                                            .sum();
 
-            final double offset = ( (double) max - (double) min - (double) totalHeight ) / ( (double) N - 1.0 );
+            final double offset = (double) (max - min - totalHeight ) / (double) (N - 1);
             final List<Widget> sortedWidgets = widgets.stream()
                                                       .sorted(( w1, w2 ) ->
             {
@@ -710,12 +710,12 @@ public abstract class ActionDescription
                 height = widget.propHeight().getValue();
 
                 final double topCenter = (double) location + (double) height / 2.0;
-                final double coffset = ( bottomCenter - topCenter ) /  ( (double) N - 1.0 );
+                final double coffset = (bottomCenter - topCenter) / (double) (N - 1);
                 for ( int i = 1; i < N - 1; i++ )
                 {
                     widget = sortedWidgets.get(i);
                     height = widget.propHeight().getValue();
-                    undo.execute(new SetWidgetPropertyAction<>(widget.propY(), (int) Math.round(( topCenter + (double) i * coffset ) - (double) height / 2.0)));
+                    undo.execute(new SetWidgetPropertyAction<>(widget.propY(), (int) Math.round((topCenter + (double) i * coffset - (double) height / 2.0))));
                 }
             }
         }


kasemir (approved):
> use .. double instead of int

That brings up a pain point...
Until JavaFX, every graphics library known to mankind used integer coordinates, and that's perfect because the display is pixelated.
When we started the display builder, we naively kept that view of widget positions as integers.

JavaFX not only uses floating point, but also makes the whole numbers refer to the gridlines __between__ pixels. The center of the top-left pixel is (0.5, 0.5), not (0,0).
See https://docs.oracle.com/javase/8/javafx/api/javafx/scene/Node.html, "Coordinate System".
When you draw a line from (1,1) to (10,10), you'll see that it's somewhat fuzzy because you're drawing the line along the gridline between pixels. Should for example have used (1.5, 1.5) to (9.5, 9.5). See also https://stackoverflow.com/questions/11886230/how-to-draw-a-crisp-opaque-hairline-in-javafx-2-2, https://stackoverflow.com/questions/11881834/what-are-a-lines-exact-dimensions-in-javafx-2, https://dlsc.com/2014/04/10/javafx-tip-2-sharp-drawing-with-canvas-api/

Bottom line, most of our drawing commands may be off by 0.5

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
```diff
@@ -620,7 +620,7 @@ public void run(final DisplayEditor editor, final boolean selected)
                                            .mapToInt(w -> w.propHeight().getValue())
                                            .sum();
 
-            final int offset = ( max - min - totalHeight ) / ( N - 1 );
+            final double offset = ( (double) max - (double) min - (double) totalHeight ) / ( (double) N - 1.0 );
```
georgweiss: Casting to double should not be necessary as N - 1.0 will implicitly yield a double, which in turn will return a double in the division.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
```diff
@@ -703,19 +703,19 @@ else if ( offset < 0 )
                 int location = widget.propY().getValue();
                 int height = widget.propHeight().getValue();
 
-                final int bottomCenter = location + height / 2;
+                final double bottomCenter = (double) location + (double) height / 2.0;
```
georgweiss, line 706: Cast to double should not be necessary.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
```diff
@@ -703,19 +703,19 @@ else if ( offset < 0 )
...
 
                 widget = sortedWidgets.get(0);
                 location = widget.propY().getValue();
                 height = widget.propHeight().getValue();
 
-                final int topCenter = location + height / 2;
-                final int coffset = ( bottomCenter - topCenter ) /  ( N - 1 );
+                final double topCenter = (double) location + (double) height / 2.0;
+                final double coffset = ( bottomCenter - topCenter ) /  ( (double) N - 1.0 );
```
georgweiss: Cast to double should not be necessary.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
```diff
@@ -703,19 +703,19 @@ else if ( offset < 0 )
...
                 for ( int i = 1; i < N - 1; i++ )
                 {
                     widget = sortedWidgets.get(i);
                     height = widget.propHeight().getValue();
-                    undo.execute(new SetWidgetPropertyAction<>(widget.propY(), ( topCenter + i * coffset ) - height / 2));
+                    undo.execute(new SetWidgetPropertyAction<>(widget.propY(), (int) Math.round(( topCenter + (double) i * coffset ) - (double) height / 2.0)));
```
georgweiss: Cast to double should not be necessary.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
abrahamwolk: That's true, but I prefer explicit casts: I think it makes the intent of the code clearer, and another advantage is that if the code is re-factored and the cast is implicit, then it's easy to introduce a bug.

georgweiss: But would it then not make sense to declare variables like ```max``` and ```min``` as doubles?

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
```diff
@@ -506,7 +506,7 @@ public void run(final DisplayEditor editor, final boolean selected)
                                           .mapToInt(w -> w.propWidth().getValue())
                                           .sum();
 
-            final int offset = ( max - min - totalWidth ) / ( N - 1 );
+            final double offset = ( (double) max - (double) min - (double) totalWidth ) / ( (double) N - 1.0 );
```
jacomago: You probably only want to cast the numerator and denominator to doubles, not every variable.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
jacomago, line 492: As @georgweiss suggested you could just cast once here.

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
shroffk: why not do

`final double bottomCenter = (double) (location + height / 2.0);`

the intent is clear
everything will be promoted to double
the code is more readable without a lot of casts

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
shroffk: in this case I think the subtraction would still happen as int without the explicit cast... then the division would promote the numerator to double and the result in double

I think george's suggestion of just making the min max double makes sense... are there any other ramifications? like we don't want users to think of min/max/height in terms of anything other than ints?

abrahamwolk: > But would it then not make sense to declare variables like max and min as doubles?

I disagree: `max` and `min` _are_ of type `int` and they can only assume integer values. By casting them we throw away this type information. It is precisely `offset` and `location` that can assume non-integer values, and the types in the pull request reflect this. I think precision in the types is preferable over minimizing the number of casts.

georgweiss (approved):

app/display/editor/src/main/java/org/csstudio/display/builder/editor/actions/ActionDescription.java
abrahamwolk: You are right! I have now implemented this as part of https://github.com/ControlSystemStudio/phoebus/pull/3614/commits/a0227c66ef376ccec3a22c7c19a268ef5d631df0.

PR #3615, opened 2025-10-31
CSSTUDIO-3523 Bugfix: Use `layoutBoundsWithoutScrollbars` instead of `layoutBoundsWithScrollbars` in if-condition

abrahamwolk:
This pull request fixes a bug introduced in https://github.com/ControlSystemStudio/phoebus/pull/3489: when determining whether the width and height specified in `outline` or in `model` should be used for computing the zoom-level when zooming, `layoutBoundsWithoutScrollbars` should be compared against instead of `layoutBoundsWithoutScrollbars`. I have tested this change manually in the Display Builder.

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java
@@ -516,7 +516,7 @@ public class JFXRepresentation extends ToolkitRepresentation<Parent, Node>
                 // correct width/height properties.
                 final double zoomXWithoutScrollbars, zoomYWithoutScrollbars;
                 final double zoomXWithScrollbars, zoomYWithScrollbars;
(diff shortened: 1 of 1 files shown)


georgweiss (approved):

jacomago (approved):
Would be nice if commit explained "why" the change rather than just the "what".

abrahamwolk:
> Would be nice if commit explained "why" the change rather than just the "what".

I don't have a full understanding of the underlying cause, but this pull request is intended to restore the logic described here: https://github.com/ControlSystemStudio/phoebus/blob/4d9b2e4dac890a2a1e18ab16b5a2701b29bfc80e/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L508-L516

In https://github.com/ControlSystemStudio/phoebus/pull/3489, I had changed the logic (as part of a larger change), so that it was not the case that the full viewport was considered in the comparison.

Empirically, it also seems that the fix works.

PR #3616, opened 2025-10-31
CSSTUDIO-3537 Bugfix: Fix check for non-zero-length array (For ESS-specific navigator)

abrahamwolk:
This pull request fixes a bug when detecting the presence of a divider when running "Split Left/Right" when running the ESS-specific Navigator application.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
+++ b/core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
@@ -658,7 +658,7 @@ public class DockPane extends TabPane
             final SplitPane parent = (SplitPane) dock_parent;
             // Remove this dock pane from BorderPane
             Optional<Double> dividerPosition;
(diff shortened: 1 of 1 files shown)


georgweiss (approved):

core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
```diff
@@ -658,7 +658,7 @@ else if (dock_parent instanceof SplitPane) // "dock_parent instanceof SplitPane"
             final SplitPane parent = (SplitPane) dock_parent;
             // Remove this dock pane from BorderPane
             Optional<Double> dividerPosition;
-            if (parent.getDividerPositions().length > 1) {
+            if (parent.getDividerPositions().length > 0) {
```
jacomago, line 661: Could use Arrays.isNullOrEmpty

PR #3622, opened 2025-11-06
Scheduled scans in scan server

DenSinH:
I hope this change does not seem too strange, we are actually kind of curious how other facilities are using the scan server, as we think we might be using it in a strange way (more like a "scripting server"). Basically, we want to be able to schedule certain tasks (which are usually started by clicking a button, which triggers remote procedures through PV writes, but potentially also other more complex tasks), and we figured the scan server does almost everything that we want to do, besides scheduling. We added a new "scan type" (ScheduledScan), which behaves like a queued ExecutableScan until its execution time has come, when it is "resubmitted" to the scan engine as a normal ExecutableScan, and executed (either queued or not depending on what the user configured). So long as the scans are still scheduled, they are _not_ blocking. I also made it so the progress indicates the leftover waiting time, and the execution time is shown in the "Command" field.

I decided to put the ScheduledScans in the same queue as the ExecutableScans and the LoggedScans, so the "usual" operations still work on them. It required an extra ScanState, which I also added to `scan.db`  in the examples folder as the comment seemed to suggest (?). I am not sure how to test if that still works. This change also adds some dependencies from the display runtime on the scan server (to add a "schedule task" context menu to ActionButtons with WritePV-type actions).

Again, we are very curious if this is something other facilities might like, and how the scan server is being used in general. Perhaps you have another suggestion to execute scheduled tasks with something that is already in phoebus, if you do not deem this a good fit for the scan server. The commits labeled `[phoebus]` are just fixes / improvements that may be cherrypicked into main anyway, and are unrelated to the scheduling, though related to the scan server.

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - [x] The documentation is up to date
    - Release notes:


kasemir:
Yes, you can add the SPI into your site-specific product.
But again what you want to hardcode into an action is basically this:
```
from scan import *
client = ScanClient('localhost')
cmds = [ Set('pvname', 3.14) ]
client.submit(cmds)
```
You can easily grow that script to get the value or PV name from somewhere else.
Or put it into your site library as
```
def OurScanBasedSet(pv_name, value):
  from scan import *
  client = ScanClient('localhost')
  cmds = [ Set(pv_name, value) ]
  client.submit(cmds)
```
and now your script becomes a one-liner.

DenSinH:
I implemented the fixes / improvements you suggested and removed the dependencies. How we are going to add the task scheduling from the display runtime internally we shall see, perhaps with a python script, perhaps with some introspection magic, but it seems that it is not a great fit to upstream this in the end. I can squash some commits if you prefer?

By the way, LocalDateTime is still used in other places (for example internally in the DateTimePane), but I didn't want to break anything.

DenSinH:
Apparently there are some time zone issues that arise in the scan submission, I think the same may be going on with the `deadline` parameter, working on a fix

DenSinH:
So, I think the same issue happens to the `deadline` parameter but it may be "breaking" to implement it this way. In the end, the server runs in UTC and has no knowledge of the time zone a client runs in, so either all datetimes have to be submitted in UTC, or as timestamps...

PR #3624, opened 2025-11-07
Modified the behavior of spinner widget to react differently with dif…

wmliuanl:
…ferent key combinations

arrow up/down: change by 1 increment
page up/down: change by 10 increment
Alt + arrow up/down change by 5 increment
Alt + page up/down change by 50 increment

## Checklist

- Testing:
    - [ ] Tests were run
    - The features have been tested at our facility.  The old spinner widget only increase or decrease by 1 increment when arrow up/down keys are pressed.  But our operators are used to the our old inhouse control system programs which has edit controls that can response with different increments for different key combinations.  To smooth up the transition from old inhouse control system to EPICS control system, I modified the spinner widget to provide similar behavior from CSS Phoebus OPIs.

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SpinnerRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/SpinnerRepresentation.java
@@ -128,12 +128,7 @@ public class SpinnerRepresentation extends RegionBaseRepresentation<Spinner<Stri
                     break;
                 case PAGE_UP:
                     if (!active) {
-						if(event.isControlDown()) {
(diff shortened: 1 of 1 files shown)


georgweiss:
For Mac users a bit of clarification may be needed:
```Alt + arrow up``` is better phrased as ```Option + arrow up```
Page up/down is actually ```fn + arrow up/down```. So change by 50 needs three keys ```Option + fn + arrow up/down```

wmliuanl:
the combination of Ctrl + Page Up/Down doesn't work when test and thus been removed.  
The working combinations are
    Ctrl + Up/Down arrow keys with 5x step size 
    Ctrl + Left/Right arrow keys with 10x step size

PR #3631, opened 2025-11-10
Fix timestamp being invalid if nanoseconds component is zero

### LINKED ISSUE #3463: Timestamp formatting fails when the nanoseconds component is zero
I am hosting a PV with p4p that has a timestamp set in the past. If I view this PV with a "Text Update" widget and check the tool-tip it shows the correct timestamp. However if I view the timestamp using the new `timestamp` function it does not show the timestamp from the `timeStamp` field of the PV, it instead shows the timestamp of when the PV was first loaded by the display. I have seen the same behavior on the "PV Table" application but it is not consistently reproducible.

rosstitmarsh:
Closes #3463

I am not familiar enough with Java and this project to add a test for it but I verified that it fixed the bug and that all the existing tests passed. Please add them if you think it is needed before merging.

## Checklist

- Testing:
    - [ ] Tests were run
        - Used the PV that I noticed the bug with that always has a 0 nanoseconds component to see if the timestamp shown correctly

- Documentation:
    - Release notes:


kasemir (approved):
Thanks!

PR #3637, opened 2025-11-11
Swap to pbrawclient 0.2.2

jacomago:
Update the pbrawclient to the latest

Has dependency on google protobuf 4.33 now

## Checklist

- Testing:
    - [ x] Tests were run (Tested by me and @georgweiss on our archivers.)

- Documentation:
    - Release notes:


georgweiss (commented):
By the way...
You need to update protobuf version in ```dependencies/phoebus-target/pom.xml```

shroffk:
@jacomago since you are working on publishing this jar to maven central... in the future we should be able to complete get rid of the install-jar nonsense I had to cook up right?

jacomago:
> @jacomago since you are working on publishing this jar to maven central... in the future we should be able to complete get rid of the install-jar nonsense I had to cook up right?

Yeah hopefully. But it looked a lot of work to setup. So I thought I woudl do this upgrade first.

PR #3647, opened 2025-11-14
Kafka update

shroffk:
Create a new compose file to deploy a kafka 4 cluster

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:


kasemir:
See also first rule of clustering: Simply enabling "the cluster" is more likely to add problems than increase reliability.

PR #3649, opened 2025-11-17
CSSTUDIO-3538 Bugfix: Redraw Linear Meter widget when the range or alarm limits are updated

abrahamwolk:
This pull request fixes a bug in the Linear Meter: before this pull request, the Linear Meter would only be redrawn when the indicator position had changed at least one pixel. In particular, an update to the range or to the alarm limits in the connected PV did not force a redrawing of the Linear Meter widget.

I have tested the bugfix manually using a soft IOC.

### DIFF
--- a/app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/LinearMeterRepresentation.java
+++ b/app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/LinearMeterRepresentation.java
@@ -170,7 +170,7 @@ public class LinearMeterRepresentation extends RegionBaseRepresentation<Pane, Li
                 boolean validRange = Double.isFinite(new_value) && Double.isFinite(model_widget.propMaximum().getValue());
                 meter.setRange(new_value, model_widget.propMaximum().getValue(), validRange);
                 if (toolkit.isEditMode() && validRange) {
-                    meter.setCurrentValue((new_value + model_widget.propMaximum().getValue()) / 2.0);
+                    meter.setCurrentValue((new_value + model_widget.propMaximum().getValue()) / 2.0, true);
                 }
             });
 
@@ -183,7 +183,7 @@ public class LinearMeterRepresentation extends RegionBaseRepresentation<Pane, Li
                 boolean validRange = Double.isFinite(new_value) && Double.isFinite(model_widget.propMinimum().getValue());
                 meter.setRange(model_widget.propMinimum().getValue(), new_value, validRange);
                 if (toolkit.isEditMode() && validRange) {
-                    meter.setCurrentValue((new_value + model_widget.propMinimum().getValue()) / 2.0);
+                    meter.setCurrentValue((new_value + model_widget.propMinimum().getValue()) / 2.0, true);
                 }
             });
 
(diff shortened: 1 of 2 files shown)


app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
```diff
@@ -777,7 +777,7 @@ private Optional<Integer> computeIndicatorPosition(double value) {
         });
     }
 
-    private void drawNewValue(double newValue) {
+    private void drawNewValue(double newValue, boolean forceRedraw) {
```
jacomago, line 780: Shouldn't this method be renamed now?

app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
```diff
@@ -735,7 +735,7 @@ public void setShowWarnings(boolean showWarnings) {
     private boolean isValueWaitingToBeDrawn = false;
     private double valueWaitingToBeDrawn = Double.NaN;
     /** @param newValue Current value */
-    public void setCurrentValue(double newValue)
+    public void setCurrentValue(double newValue, boolean forceRedraw)
```
jacomago, line 738: I'm not a fan of boolean flags, I would write an enum

enum DrawAction {
   ForceRedraw
   Draw
   NoRender
}

or something like this.

app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/LinearMeterRepresentation.java
```diff
@@ -336,12 +338,15 @@ private void valueChanged(WidgetProperty<?> property, Object old_value, Object n
                                     && controlRange.getMaximum() - controlRange.getMinimum() > 0.0) {
                                 if (meter.linearMeterScale.getValueRange().getLow() != controlRange.getMinimum() || meter.linearMeterScale.getValueRange().getHigh() != controlRange.getMaximum() || !meter.getValidRange()) {
                                     meter.setRange(controlRange.getMinimum(), controlRange.getMaximum(), true);
+                                    linearMeterScaleHasChanged = true;
```
jacomago, line 341: This makes me feel like this should be a property of the linearmeter, so this should be something like:

meter.setScaleUpdated()

abrahamwolk: I disagree: I prefer explicit passing of values (in contrast to implicit passing of values through update of state). In particular, `linearMeterScaleHasChanged` is passed to `meter.setCurrentValue()` on line 386: https://github.com/ControlSystemStudio/phoebus/pull/3649/files/c37e695e063c8014201fceb374da86198dbf2aa3#diff-99f66c9c7631b64fc8906945f212e452266b923ac7c5a7214eb7f604df210cbfR386

app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
abrahamwolk: An enum could be used, but since the data is still essentially only `true` and `false` and the parameter name of `meter.setCurrentValue()` describes the meaning well, I prefer to use a boolean.

app/display/linearmeter/src/main/java/org/csstudio/display/extra/widgets/linearmeter/RTLinearMeter.java
abrahamwolk: No: the method draws a new value. Before this pull request, it was "optimized too much": `drawNewValue()` only actually drew a new value if it determined that the indicator had moved at least one pixel, which didn't account for the case where the range or the alarm limits had changed.

PR #3655, opened 2025-11-28
add full path of the bob file as a tab name, if no name indicated

agaget:
Currently if no name is indicated in the bob file properties, only `[Edit]` is displayed, it can be confusing during edition.
image

Now thanks to the DisplayModel information, the complete path of the file is displayed instead following the example of the runtime instance.

- Testing:
Result
_9AA0CC66-BC4C-4968-8A08-C94154C4C930_

also tested by @lcaouen

### DIFF
--- a/app/display/model/src/main/resources/examples/initial.bob
+++ b/app/display/model/src/main/resources/examples/initial.bob
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <display version="2.0.0">
-  <name>Display</name>
+  <name></name>
   <widget type="label" version="2.0.0">
     <name>Label</name>
     <class>TITLE</class>


shroffk:
So there are a few factors that contributing to the above

When screens are converted from .opi or .edl then the Display name is an empty string... this might be the issue @agaget is primarily trying to address here.

When you create a new display then the editor uses a template... the default being `examples:/initial.bob` in this template the Display name is "display"

When *running* old .opi screens when they don't have a name the runtime uses the complete file path as the tab name.

While the complete file path seems very informative it gets very messy very quickly... usually users will have many tabs open and if each tab is deploying the complete path which might be hundreds or chars the tab heading becomes very large.
At NSLS2 we had a script to actually go replace the display name with just the file name.

I would hope we can come up with a holistic and consistent solution here and I would vote for having just the file name... the complete path can be accessed via the display info ( yes it is 2 more clicks.. but it keeps the tabs more manageable )

agaget:
Hi,
This correction was asked by a colleague so I'm not exactly sure what is the use case but as @shroffk said I think it comes from old *.opi files.
The modification was made also to be consistent with runtime where the full path is displayed. 

I would agree with @shroffk , just the file name is probably enough, but in this case to be consistent I can change it also in the runtime instance.
 Tell me

agaget:
Ok, so I have changed the code a little bit.
- I have  changed in the model the function `getDisplayName()` to return the filename and not the fullPath. That way Instance stay unchanged, and in the editor I just call getDisplayName()
- I have changed ìnitial.bob` to put empty value by default

Do you think changing the model will interfer with other functionnality ?

shroffk:
> I have changed in the model the function getDisplayName() to return the filename and not the fullPath. That way Instance stay unchanged, and in the editor I just call getDisplayName()

This looks much cleaner.

I checked the usage and so far the `getDisplayName()` I don't think it should have any adverse consequences.

If the file path is needed then the right way should be using `model.getUserData(DisplayModel.USER_DATA_INPUT_FILE` as is the case with the selectionInfo

```
return new SelectionInfo(model.getUserData(DisplayModel.USER_DATA_INPUT_FILE),
                                 model.getDisplayName(),
                                 model.propMacros().getValue(),
                                 false,
                                 model_parent);
```

@kasemir can you think of any possible regressions

Issue #3605, opened 2025-10-24
Warn on unknown preference setting


minijackson:
Has this ever happened to you? You configure a service URL in the Phoebus preference settings and it doesn't work. You spent way too much time checking every assumption, such as networking issues, service version mismatch, etc., just to find out you wrote `org.phoebus.application.alarm/server = my-server:9092` instead of `org.phoebus.applications.alarm/server = my-server:9092`.

Ideally, Phoebus should make it clear when it doesn't recognize a preference settings:

- when a preference setting is from an unknown package
- when a preference setting is unknown in that package

I think printing a warning in the logs would be nice, but the logs are a bit crowded. I think it would be good to also have an unmissable visual indication, like a [banner](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.Banner.html), and/or a message in the "About > Preference Settings" tab.

PR #3661, opened 2025-11-28
show in red incorrect properties in settings.ini

lcaouen:
Shows in the About dialog box the incorrect settings from settings.ini.
It compares the settings from the settings.ini to the settings from the preferences.properties
The incorrect ones are displayed in red

This is related to this issue :
https://github.com/ControlSystemStudio/phoebus/issues/3605

- Testing:
    - tested manually by @lcaouen @minijackson 

- Documentation:
    - A message is displayed at the top of the view in the about dialog saying the settings in red are incorrects

### DIFF
--- a/core/framework/src/main/java/org/phoebus/framework/preferences/PropertyPreferenceWriter.java
+++ b/core/framework/src/main/java/org/phoebus/framework/preferences/PropertyPreferenceWriter.java
@@ -19,16 +19,16 @@ import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
+import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


georgweiss:
Unfortunately not able to get the intended behavior.
Starting without any path to a settings file I get:

Screenshot 2025-12-01 at 09 16 20

Starting with a custom settings file (correct settings) specified through ```-settings /path/to/settings/file.ini``` I get:

Screenshot 2025-12-01 at 09 39 03

lcaouen:
I have updated the source code.
If there is only one jar in the classpath, it opens it to get the paths of the other jars.
I have also added the "misssing code"
- a monospace font
- escape the html tags in the properties
- a message at the top saying properties in red are incorrects

image

lcaouen:
Thanks.
I have just downloaded the files.
Also I have just pushed a new version which should work with IntelliJ and exclude some keys from the check.

Now I'm gonna have a look to the snap_widgets and alarm/server keys

lcaouen:
Do you use phoebus.sh(.bat) in your phoebus-product to run it ?
Do you add any argument like -settings to give your ini file ?

image

When I run phoebus.bat without any argument (so no -settings given), I get this screen, which is correct as no settings file is loaded.
image

Is that what you get or something different ?

PR #3669, opened 2025-12-08
Save&Restore comparison of array data

georgweiss:
A per user request:

Comparisons in the save&restore snapshot view are difficult for arrays with more than a dozen or so elements. This PR adds the possibility to launch a dialog showing saved, delta and live value in a table view, see screenshots. A threshold spinner like the one in the snapshot view can be used to remove highlight of deltas smaller than the user defined level.

The dialog controller will connect to the PV in order to show actual live value.

Differences in array length are treated such that elements missing from either array are simply rendered as "---", see screen shots.

This PR is a first iteration, the idea is to expand this to also support NTTable types.

**NOTE** "Click to compare" is used only for non-scalar types.

Screenshot 2025-12-08 at 12 50 09

Screenshot 2025-12-08 at 09 16 26

## Checklist

- Testing:
    - [ X] Tests were run

- Documentation:
    - [ X] The feature is documented

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/VDeltaCellEditor.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/VDeltaCellEditor.java
@@ -96,6 +96,8 @@ public class VDeltaCellEditor<S, T> extends VTypeCellEditor<S, T> {
                             comparisonDialog.show();
                         });
                     } else {
+                        // Do not handle mouse clicked, e.g. if live PV is disconnected.
+                        setOnMouseClicked(e -> {});
                         String percentage = Utilities.deltaValueToPercentage(pair.value, pair.base);
                         if (!percentage.isEmpty() && showDeltaPercentage) {
                             Formatter formatter = new Formatter();


app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/snapshot/VDeltaCellEditor.java
```diff
@@ -34,7 +40,7 @@
  * @param <T>
  * @author Kunal Shroff
  */
-public class VDeltaCellEditor<T> extends VTypeCellEditor<T> {
+public class VDeltaCellEditor<S, T> extends VTypeCellEditor<S, T> {
```
abrahamwolk, line 43: Why is the type parameter `S` added to `VTypeCellEditor`?

georgweiss: In order to be able to use ```VTypeCellEditor``` for both ```TableEntry``` and ```ColumnEntry```

jacomago:
> @jacomago, updated such that row where any of the values (base or live) is not available (i.e. array length not equal), sorting on the delta column will order such rows next to rows with non-zero delta. This way user can easily find differences in array length without having to scroll to bottom of the table view.

Looks better, I would still do the summary info. Number of rows different, number of rows matching. Its a tricky feature for UX wise obviously.

PR #3674, opened 2025-12-12
CSSTUDIO-3605 Always use the type "long" when converting to hexadecimal notation

abrahamwolk:
This pull request changes the method `FormatOptionHandler.formatNumber()` so that the type `long` is always used when converting a number to hexadecimal notation.

`FormatOptionHandler.formatNumber()` is used by the Text Entry widget and the Text Update widget when the widget property "Format" is set to "Hexadecimal". If, furthermore, the widget property "Precision" is set to `-1` (the default value), then `FormatOptionHandler.formatNumber()` will be called with its argument `precision` set to `3` when called with a value of type `double`, with the consequence that, prior to this pull request, the type (signed) `int` would be used to convert a number represented by a `double` to hexadecimal notation, potentially truncating the input.

To me, it seems incorrect that the argument `precision` is used to determine the number of digits. For the decimal notation, the parameter refers to the number of digits after the period, and in this context, the default value of `3` makes sense for a value of type `double`. I think that for the property to mean something different for the hexadecimal notation is confusing, and I think the default (if any) should be a different value. Should this be changed, e.g., by introducing a new widget property for the number of digits before the period to display?

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/vtype/FormatOptionHandler.java
+++ b/core/ui/src/main/java/org/phoebus/ui/vtype/FormatOptionHandler.java
@@ -18,6 +18,10 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.epics.util.array.ListNumber;
+import org.epics.util.number.UByte;
+import org.epics.util.number.UInteger;
+import org.epics.util.number.ULong;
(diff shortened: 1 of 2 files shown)


georgweiss (commented):
A unit test would maybe make sense?

georgweiss (approved):

PR #3680, opened 2025-12-19
Individual colors on navtabs

gabriel-desmarchelier:
This PR implements an option to allow each tab from the `NavigationTabs` widget to have its own colors.

Here is the use case that drives me to suggest this feature :    
I want to be able to use rules to apply a specific color to a tab, depending on the status of the device associated to this tab.
Let's say I manage several power supply units and I want the tab color to show me which one is up, which one is ramping up, which one is in fault.

What it looks like when I'm looking into `Tab0` but something happened in `Tab2` :

image

Of course, I also want to know that I'm looking into `Tab0`, even when everything goes wrong :

image

So I implemented : 

- new tab properties as part of the Tab StructuredWidgetProperty
    - `individual_selected_color`
    - `individual_deselected_color`
- new widget property : `enable_individual_tab_colors`
    - if `False` : the usual  `selected_color` and `deselected_color` are used
    - if `True` : the new `individual_selected_color` and `individual_deselected_color` are used
    - defaults to `false`, avoiding any breaking change

I also changed the name of `tabs` attribute and of `setTabs` method to clarify the fact that they refer to the tab names.

Here is a .bob with various configurations of NavigationTabs that use this feature : 
[nav_tabs_colors.tar.gz](https://github.com/user-attachments/files/24258391/nav_tabs_colors.tar.gz)

Let me know if you think such a feature is relevant of if you think this not the right way to implement it.

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/NavigationTabsWidget.java
+++ b/app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/NavigationTabsWidget.java
@@ -8,7 +8,6 @@
 package org.csstudio.display.builder.model.widgets;
 
 import static org.csstudio.display.builder.model.ModelPlugin.logger;
-import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.newBooleanPropertyDescriptor;
 import static org.csstudio.display.builder.model.properties.CommonWidgetProperties.propDirection;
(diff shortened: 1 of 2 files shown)


gabriel-desmarchelier:
> And please also implement this for the web runtime, https://github.com/ornl-epics/dbwr

I'll have a look at this during our next internal codeathon (since I won't attend the official one at Diamond)

PR #3682, opened 2026-01-06
Expand Alarm Configuration Actions

shroffk:
## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:


kasemir:
Is there a way to suppress the context menu? Maybe that's already in the contribution support, I forgot. Or somehow tie this to authorization for the alarm configuration?

In any case, for our setup I don't want to have "Add to alarm tree" in every PV context menu.
Adding an alarm requires some thought. Authorization is limited to those who check if the PV is useful, what the guidance should be, suitable related display etc.

shroffk:
by simplifying this workflow I am hoping to avoid other programmatic/scripted solutions which have been necessary when adding a large number of PV's...a workflow which is even future away from the thoughtful user actions :)

The authorization part is a valid issue which might need a bit more thought

shroffk:
exactly... I did not want to redo the whole Viewer again using fxml

PR #3692, opened 2026-02-03
Align web socket client implementation between Olog and save&restore

georgweiss:
This is to align to the same web socket client implementation for Olog and save-and-restore. Server side has also been updated to use the same Spring Boot APIs.

New modules have been created to enable use of common APIs and thus reduce code duplication.

**NOTE!!!**
Sadly I had to change the Olog ES client implementation such that "Olog" is prepended to all endpoints. Reason is that the Olog service by default deploys to the context "/", but all endpoints start with /Olog. At the same time the Olog client (erroneously) handled this "Olog" prefix as the context root. 
**ACTION NEEDED**
The preference setting ```org.phoebus.olog.es.api/olog_url``` must omit the "Olog" part of the URL!

## Checklist

- Testing:
    - [ X] The feature has automated tests
    - [ X] Tests were run

- Documentation:
    - Release notes:


app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreBaseController.java
```diff
@@ -33,11 +32,11 @@
 public abstract class SaveAndRestoreBaseController {
 
     protected final SimpleStringProperty userIdentity = new SimpleStringProperty();
-    protected final WebSocketClientService webSocketClientService;
+    //protected final WebSocketClientService webSocketClientService;
```
abrahamwolk: Commented-out code. (Also on line `39`).

shroffk (approved):

PR #3695, opened 2026-02-06
Use VType display for formatting data browser cursor values

DenSinH:
Use VType display for formatting data browser cursor marker values. Previously, they were formatted using the formatting specified by the trace (which seemed to be obtained automatically, and depends on the axis labels, which can also not be configured manually it seems). I changed this to where this can be implemented in the `PlotDataItem` implementation. The default one is the same as the one that was implemented before, and I specified it for the `PlotSample` class, where it uses the `FormatOptionHandler` with the `DEFAULT` format to format the VType from the selected sample. This makes it so that the value formatted in the cursor marker and the value in the Traces table are all formatted using the VType's format, which is more consistent with the rest of the UI.

I am however not sure if I missed some configurable option, or if other sites possibly depend on the "old" formatting.  

## Checklist

- Testing:
    - [ ] Tests were run
    - Checked whether the newly formatted values properly showed up in the data browser

- Documentation:
    - Release notes:

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PlotSample.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/model/PlotSample.java
@@ -10,7 +10,9 @@ package org.csstudio.trends.databrowser3.model;
 import java.time.Instant;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.csstudio.javafx.rtplot.Trace;
 import org.csstudio.javafx.rtplot.data.PlotDataItem;
+import org.csstudio.javafx.rtplot.internal.YAxisImpl;
 import org.epics.vtype.Alarm;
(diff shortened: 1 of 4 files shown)


georgweiss:
@DenSinH, I have put this to the test, but honestly I cannot really see what it changes. Can you please clarify and provide an example (e.g. screen shots)?

DenSinH:
It's about this label:

image

which in the old situation would have its own formatting rules, which are formatted in the way defined in `PlotDataItem`, which takes the values from the `YAxisImpl`. Back when I worked on this I traced it back to how this was determined, and it boiled down to always being something like 3 decimals... This caused for inconsistencies in the UI.

In general this is kind of part of a bigger problem we are sometimes facing though: formatting of PVs is fairly inconsistent throughout Phoebus at times, even though the way PV values should be formatted is generally available in the `VType`.

PR #3698, opened 2026-02-11
Split Queue Server into modular network, monitor, and edit-control packages

JMit-dev:
## Summary
- Split monolithic queue-server app into 3 Maven sub-modules: `network`, `monitor`, `edit-control`
- Queue Monitor and Edit & Control Queue are now independent Phoebus applications
- Removed the left-side tab switcher - each app opens as its own dock tab
- Added connection manager widget to Edit & Control Queue layout
- Extracted ConnectionManager singleton so both apps share one connection
- Added reference counting to AppLifecycle so shared state only resets when the last app closes


app/queue-server/src/main/java/org/phoebus/applications/queueserver/util/JythonScriptExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+                for (String key : contextVars.keySet()) {
+                    python.set(key, null);
+                }
+            }
+
+            // Convert PyObject to Java object
+            return result != null ? result.__tojava__(Object.class) : null;
+
+        } catch (Exception e) {
+            LOG.log(Level.FINE, "Jython script execution failed: " + e.getMessage());
```
kasemir: Instead of
`LOG.log(Level.FINE, "Jython script execution failed: " + e.getMessage());`
it's generally better to use
`LOG.log(Level.FINE, "Jython script execution failed", e);`

This will show the message, if there is one, plus the complete stack trace.
On the other hand, e.getMessage() may return null and without a stack trace you then wonder where/what/why happened.

shroffk: @JMit-dev could you fix the logging as suggested above :)

PR #3700, opened 2026-02-12
Change how AlarmTableUI handles selecting items when state changes to no longer clear manually selected items

Peyton-Hill-CLS:
Currently, whenever the alarm table state changes or when the user searches for something, any selected rows get cleared and it creates a new selection based of the search input. If there is no search input, then all selections are cleared. This means that any items selected with a users mouse get cleared when an alarm is added, removed, or is modified (such as the pv severity changing). This can cause user frustration, particularly when a pv severity is flickering in and out of different states.

This PR attempts to fix this by recording all the pv's that were previously selected that are still in the new alarm list when the state changes. After the rows are updated with the new alarms, create a new selection with this pv list.


shroffk:
similar to:
https://github.com/ControlSystemStudio/phoebus/pull/3667

PR #3710, opened 2026-02-25
Add support for DBEMask

jacomago:
Allows opening different MonitorRequests
In particular for use in the archiver using DBE_ARCHIVE

Added a new pv to the demo.db that changes alarm frequently to test the DBE_ALARM in ClientDemo

## Checklist

- Testing:
    - [x ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/FieldRequest.java
+++ b/core/pva/src/main/java/org/epics/pva/client/FieldRequest.java
@@ -53,74 +53,45 @@ class FieldRequest
      */
     public FieldRequest(final String request)
     {
-        this(0, false, request);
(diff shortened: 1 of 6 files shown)


jacomago:
> Have you tested this with both pvAccessCPP/QSRV and PVXS/QSRV2? Do both support the DBE.. options?

I tested on PVXS only, @mdavidsaver specifically did not want to fix pvAccessCPP to support it.

kasemir:
I agree to not adding anything in pvAccessCPP.
It's OK when older IOCs with pvAccessCPP ignore DBE, but need to test that the older IOCs don't crash.

PR #3717, opened 2026-02-27
Improve email preferences

MJGaughran:
This is for the Codeathon task: https://github.com/ControlSystemStudio/phoebus-codeathon/tree/master/projects#phoebus-ui-005-default-email-address-preferences

There are a couple of differences to that project description:

- Preference is labelled 'to', in line with the existing 'from' field
- Username macro would be 'user.name' since it uses the system properties

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/email/ui/pom.xml
+++ b/app/email/ui/pom.xml
@@ -23,7 +23,31 @@
             <artifactId>core-ui</artifactId>
             <version>5.0.3-SNAPSHOT</version>
         </dependency>
-    </dependencies>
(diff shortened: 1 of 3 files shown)


GDH-ISIS:
Built our Windows Phoebus client using the Fork/Branch - looks good to me. To and From email addresses appeared as expected. Used $(USERNAME) for the from address. All populated as expected.

PR #3720, opened 2026-03-02
Change setModified to destructively overwrite target in OpenDisplayAction

vincent-hve:
Middle or control click on a widget supporting Actions, changed the target field permanently in the associated OpenDisplayAction, causing this operation to be executed henceforth, even without the modifier keys, or middle mouse button.

This PR fixes that, such that the target is only active for as long as the modifier is set.

### DIFF
--- a/app/display/actions/src/main/java/org/csstudio/display/actions/OpenDisplayAction.java
+++ b/app/display/actions/src/main/java/org/csstudio/display/actions/OpenDisplayAction.java
@@ -32,7 +32,8 @@ public class OpenDisplayAction extends ActionInfoBase {
     /**
      * Default {@link Target} is new tab.
      */
-    private Target target = Target.TAB;
+    private Target target = Target.REPLACE;
+    private Target modifierTarget = null;
 
     private Macros macros;
 
@@ -116,7 +117,7 @@ public class OpenDisplayAction extends ActionInfoBase {
     }
 
     public Target getTarget() {
-        return target;
+        return modifierTarget != null ? modifierTarget : target;
     }
 
     public Macros getMacros() {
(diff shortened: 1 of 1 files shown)


georgweiss:
You're right, my bad. Depending on the wanted default target configured on the action, one needs to honor that when no modifier keys are used.

Thanks for adding the missing case (REPLACE). 

I would maybe suggest:

```
- By holding the 'Control' key, the display will open in a new tab.
- By holding the 'Shift' key or press middle mouse button, the display will open in a new window.
- By holding both 'Control' and 'Shift' key, the display is replaced
On Mac, use Command key instead of Control.
``` 

This text is in https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/model/src/main/resources/examples/controls_action_buttons.bob

kasemir:
> TAB is in a sense the default

No. Replacement should be the default, just like links on a web page will by default replace the current page with the linked one.
The documentation is correct:
```
By holding the 'Control' key, the display will open in a new tab.
By holding the 'Shift' key, the display will open in a new workspace window.
By holding both 'Control' and 'Shift' key, the display will open in a standalone window.
```
Without any key modifiers, the default should be the default target of actions, which is replacement.

shroffk:
Ok,

So the documentation is the desired ( correct ) behaviour. 
We should update the

vincent-hve:
From what I can tell STANDALONE is deprecated. So, should we still support this? 
```
    /**
         * Open standalone window
         *
         * @deprecated Was only used in RCP version.
         */
        @Deprecated
        STANDALONE(Messages.Target_Standalone);
```
I'm more inclined to make the CTRL+SHIFT operation REPLACE.  This would allow you to force any of the 3 non-deprecated options. For now I have modified it in accordance to https://github.com/ControlSystemStudio/phoebus/pull/3720#issuecomment-3998312080 .

Issue #3723, opened 2026-03-04
Make the databrowser min max lines configurable along with the main line


jacomago:
<img width="3431" height="2103" alt="Image" src="https://github.com/user-attachments/assets/9e101808-e422-4864-b969-53ff6004ad7c" />

Green here is the raw values, red is the optimized call.

From here I see a faint line that the red max matches the max of the raw values, but it looks strange.

I can configure the width and style of the main line but not of the min/max lines. I think it would be nice to have the min max configurable, and it would make comparing raw vs optimized (and any future downsampling we support) better.

kasemir:
The idea is that raw vs. optimized give the same outline.

Optimized should provide around N min/max/average buckets, with N close to or larger than the plot width in pixels.
When simply drawing lines between all the raw samples vs lines from min0 to max0 to min1 to max1 to min2 to max2, ..., that should look the same. For the optimized, we instead draw a line from avg0 to avg1 to avg2 ... and fill the min/max outline with a lighter color. That way, you get the filled outline that looks like the raw data (albeit lighter), PLUS you get the average so you can tell if the value tended to be smack in the middle vs for example mostly at the upper end.

So bottom line, optimized should result in faster plot updates because there are only (pixel width) x (min, max, avg) values to handle, it gives the same outline, plus you see where the average is within the outline.
Downsampling would hide detail, and then users wonder why a brief spike in the data is hidden.
Asking for much more than (pixel width) buckets simply wastes resources.

<img width="728" height="1075" alt="Image" src="https://github.com/user-attachments/assets/d6f476c6-58e6-4bbe-91d2-974c7cda7437" />

PR #3725, opened 2026-03-05
Add support for modifying the opacity and area color

jacomago:
Adds two new columns to the traces:

- AreaOpacity Which sets the opacity for the area, keeping the color from the trace
- AreaColor Which is by default the trace color with the opacity from the AreaOpacity

This allows the user when looking at more spiky data to easily see the min, max peaks. Right now mostly they resort to swapping to raw data.

## Checklist

- Testing:
    - [ ] Tests were run
  Opened the data browser and tested a few traces.
- Documentation:
    - Release notes:


jacomago:
fixes #3723

georgweiss:
@kasemir, what is your take on globally configurable area color? Changing opacity to 100% will make the statistics-related data indistinguishable.

kasemir:
I'm for keeping it simple. One color per trace. The area uses a lighter/transparent variant of the trace's color.

The original issue was that different people have different preferences for the lightness or transparency. So make that one new configuration parameter.  For example, in the "Misc." tab below Foreground and Background color, add "Area Opacity:" with 0-100%.

For both the end user and the maintainer of the source code, adding a new AreaColor plus AreaOpacity per trace seems complete overkill. But I think I stated that above, so do whatever.

jacomago:
Closing as @georgweiss taking over.

PR #3731, opened 2026-03-09
Added file drag-n-drop for opening them

### LINKED ISSUE #3730: [Feature] Drag and drop files from explorer into Phoebus to open them
In my workflow I would like to drag-'n-drop files from Windows explorer into my Phoebus instance to open them.  
This is useful when e.g. I'm working out of three folders with very different paths. It's a lot of navigating otherwise.

Currently I can drag files into an open editor. This will for example quickly make an embedded widget, which is quite cool.

I would suggest a feature that dragging a file into these sections causes it to open it:
- An empty home screen
- The startup 'welcome' screen
- The top status and toolbars (always)
- An active, rendered display (any, always)

Image

I'm not entirely sure if is mentioned in the issues already. While searching I found more specific bugs, but I don't think about opening files.

RobertoRoos:
Added feature to allow drag-and-drop of files to open them.

Resolves #3730 
- #3730 

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
+++ b/core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
@@ -53,6 +53,8 @@ import javafx.scene.input.KeyCodeCombination;
 import javafx.scene.input.KeyCombination;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.input.MouseEvent;
+import javafx.scene.input.Dragboard;
+import javafx.scene.input.TransferMode;
 import javafx.scene.layout.BorderPane;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.HBox;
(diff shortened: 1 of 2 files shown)


core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
```diff
@@ -487,6 +489,27 @@ private void startUI(final MementoTree memento, final JobMonitor monitor) throws
...
+
+            if (db.hasFiles()) {
+                for (File file : db.getFiles()) {
+                    this.openResource(file.toURI(), true);
+                }
+            }
+
+            e.setDropCompleted(true);
+            e.consume();
+        });
```
RobertoRoos: This works, but causes the dragged-file icon to still show on top of the UI file the 'Choose editor' dialog is open. 
What is a neat way to delay the `openResource()` call by a little bit?

core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
```diff
@@ -487,6 +489,27 @@ private void startUI(final MementoTree memento, final JobMonitor monitor) throws
         toolbar = createToolbar();
         createTopResourcesMenu();
 
+        menuBar.setOnDragOver(e -> {
```
RobertoRoos: What would be a good location to put this event method? Because the event-handler should be shared, as I would like it to be called from other panels too.

core/ui/src/main/java/org/phoebus/ui/application/PhoebusApplication.java
kasemir: Instead of calling `this.openResource(...)` right there, try deferring it to the next UI cycle via
```
 Platform.runLater(() -> this.openResource(...));
```

PR #3754, opened 2026-03-30
Support add/remove attachments when editing Olog entry

georgweiss:
**NOTE** This PR requires commit 53a9ca03c8a50c0a89955f46976523a00891e106 or later from https://github.com/Olog/phoebus-olog. Or release 6.0.3 of the Olog service.

This PR adds support for adding and removing attachments when editing an Olog entry. Note though that removing attachments only removes the references in the updated entry. Such attachments are not deleted from storage, they are still accessible when inspecting the log entry history.

- Testing:
    - [X] Tests were run
    
- Documentation:


georgweiss:
@shroffk , can we merge this?

shroffk (approved):

PR #3757, opened 2026-03-31
Fix core-websocket-client source directory layout

emilioheredia-source:
The two source files were placed in a JPMS-style dotted directory:

  src/main/java/org.phoebus.core.websocket.client/

instead of the standard Maven/Java nested directory:

  src/main/java/org/phoebus/core/websocket/client/

Maven compiled the files but output the .class files into a dotted directory that the JVM cannot resolve as a package.  The module still produced a valid jar (the jar plugin repackages the classes correctly), so any build that used the cached .m2 jar succeeded.  However, builds that compiled against the reactor's target/classes directory directly (e.g. any -pl/--also-make partial build, or a downstream module built in the same reactor run before the jar was installed) would fail with:

  package org.phoebus.core.websocket.client does not exist

Fix: move both files to the standard nested directory layout. No changes to package declarations, imports, or any other code.


emilioheredia-source:
Hi, I work at the Canadian Light Source and recently started compiling and adding little fixes and enhancements to Phoebus, in the hope they can be merged with the oficial Phoebus release, so we won't need to maintain a separated fork. This is the first of a handful of PRs I want to submit for approval, this one dealing with something minor I found while compiling.
Thanks for taking the time to look into this, and thanks for Phoebus. It's currently replacing most other GUI frameworks for EPICS at the CLS, hence the interest. Cheers.
Emilio.
Controls Analyst.
Canadian Light Source.

PR #3759, opened 2026-03-31
feat: split DockPane on edge drag-and-drop

emilioheredia-source:
Dragging a tab to an edge zone (outer 25% of width/height) now splits the pane in that direction instead of floating the tab.  Dropping in the centre zone, or on the tab header strip, merges the tab into the target pane as before.
To float a tabs, drop it outside the Main Window, or use right click "Detach" on the tab context menu.
Colored outlines indicate the type of drop action (green rectangle = merge, blue edge = split).

## What changed (DockPane.java only)

- Add DropZone enum (CENTER/LEFT/RIGHT/TOP/BOTTOM)
- DRAG_OVER registers as a capture-phase filter so drops onto content-heavy panes (e.g. file browser) are accepted even when child nodes consume the event; previously those drops would float the tab
- Tab header strip always maps to CENTER (merge, not split): boundary measured via getBoundsInParent() and cached in handleTabChanges() to avoid a repeated CSS scene-graph walk on every pointer-move event
- Directional blue border feedback via inline CSS: only the edge where the new pane will appear is highlighted (LEFT/RIGHT/TOP/BOTTOM); inline style is used because it sits at the highest cascade priority and is not overridden by the application stylesheet
- mergeTabIntoPaneDeferred / splitAndPlaceTabAsync defer the remove+add to the same UI pulse so the source pane is never transiently empty, preventing spurious mergeEmptyAnonymousSplit calls that caused the application close button to require two clicks
- mergeEmptyAnonymousSplit guards against running when the pane is no longer empty by the time the deferred call executes
- split() restores setActiveDockPane(this) after constructing the new empty pane, preventing a null active-item notification to listeners during a live drag
- Window.requestFocus() after drop re-asserts window focus so the first mouse action after a drag is not swallowed as a focus-click

The new behavior was tested manually by trigin to cover most or all use cases.

### DIFF
--- a/core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
+++ b/core/ui/src/main/java/org/phoebus/ui/docking/DockPane.java
@@ -25,6 +25,7 @@ import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.scene.control.Alert;
 import javafx.scene.control.ButtonType;
+import javafx.scene.control.Control;
 import javafx.scene.control.ContextMenu;
 import javafx.scene.control.MenuItem;
 import javafx.scene.control.SeparatorMenuItem;
@@ -53,6 +54,7 @@ import javafx.scene.input.TransferMode;
 import javafx.scene.layout.Border;
 import javafx.scene.layout.BorderPane;
 import javafx.scene.layout.StackPane;
+import javafx.scene.paint.Color;
 import javafx.stage.Stage;
 
 /** Pane that contains {@link DockItem}s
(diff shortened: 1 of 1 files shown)


emilioheredia-source:
Ok, this is the first real change. I know it's an invasive one, but I assume given most of the old behavior is still there, and this PR just adds new options, most users won't mind. And I really miss rearranging tabs by drag and drop, like the original CS Studio does.
If it helps with his adoption, I could make the new "drag and drop close to edges to split" feature, into a option so it can be turned on/off.
Thanks for considering this.

PR #3765, opened 2026-04-03
Fix bug found after the PR #3412

lcaouen:
Fix bug found by @georgweiss after the PR https://github.com/ControlSystemStudio/phoebus/pull/3412

Tested with bob file given here
https://github.com/ControlSystemStudio/phoebus/pull/3412#issuecomment-4126237545
No more java exception and local pv is written correctly


georgweiss (approved):
Seems to resolve the issue.

PR #3766, opened 2026-04-04
perf(rtplot): add parallel_rendering preference for RTTank UpdateThrottle

emilioheredia-source:
### Problem

`RTTank` serialises all rendering on `UpdateThrottle.TIMER` — a single
global background thread shared across every `RTTank` instance in the
application.  On a display with many Tank or Progress Bar widgets each
render queues behind the previous one.  With 100 widgets, a 50 ms dormant
period, and ~5–20 ms per render, the effective refresh rate collapses well
below the 4 Hz target even though each individual widget could keep up.

The bottleneck is thread contention, not rendering cost per widget.

### Fix

Add a `parallel_rendering` boolean preference to
`org.csstudio.javafx.rtplot`.  When `true`, each `RTTank` is assigned to
the module's existing `Activator.thread_pool` (one thread per CPU core)
instead of `UpdateThrottle.TIMER`, so tanks on separate threads render
concurrently.

The default is `false`, preserving the original serialised behaviour.
Site-local `settings.ini` can enable it with one line:

`org.csstudio.javafx.rtplot/parallel_rendering=true`

Requires restart to take effect.

### Performance context

*(screen recording to be added)*

On a representative screen with 100 progress-bar widgets updating at 4 Hz:
- `parallel_rendering=false`: visible widgets update far below 4 Hz; one
  CPU core saturated from queue backlog.
- `parallel_rendering=true`: all widgets track the 4 Hz target; total CPU
  rises from ~3 % to ~6 % spread across cores; peak per-core load drops
  markedly.

### Files changed

| File | Change |
|---|---|
| `Activator.java` | `@Preference boolean parallel_rendering` field + javadoc |
| `RTTank.java` | constructor uses `thread_pool` or `TIMER` per preference |
| `rt_plot_preferences.properties` | new entry, default `false`, docs note restart required |

### Notes
- No dependency on any other open PR.
- No automated test exists; manual validation: open a Tank-heavy display
  with both values of the preference and observe refresh rate.

### DIFF
--- a/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/Activator.java
+++ b/app/rtplot/src/main/java/org/csstudio/javafx/rtplot/Activator.java
@@ -30,6 +30,13 @@ public class Activator
     @Preference(name="shady_future") private static int[] rgba;
     public static final Color shady_future;
 
+    /** When true, RTTank renders on the shared thread pool (one thread per CPU
+     *  core) so many simultaneous Tank / ProgressBar instances update in
+     *  parallel.  When false, all renders serialise on a single global thread
+     *  (the pre-fix behaviour). Controlled by the {@code parallel_rendering}
+     *  preference. */
+    @Preference(name="parallel_rendering") public static boolean parallel_rendering;
+
     /** Thread pool for scrolling, throttling updates
      * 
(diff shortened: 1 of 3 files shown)


kasemir (approved):

kasemir:
@shroffk can we ignore the doc build failure and merge this?

Issue #2841, opened 2023-10-12
Integration of the Scale option in widget progress bars


XavSPM:
Hello everyone,

I am in the process of gradually migrating our interfaces from CS-Studio to Phoebus.

I am encountering an issue on one of the operator pages. This page allows us to display the current value measured by the BLMs. Here is what it looks like:

![273912353-174d19b8-55ac-40b1-8acf-31f586b4ed63](https://github.com/ControlSystemStudio/phoebus/assets/104097021/3d3227e5-3a86-435b-9c0e-eb9219443d52)

With CS-Studio, the bar graph is created using "Progress Bars" with the "Show Scale" option enabled.

However, in Phoebus, this option is no longer available for the "Progress Bars". Therefore, I have opted for "Tanks" to create the bar graphs. Nevertheless, it seems that the display speed of the "Tanks" is slower than that of the "Progress Bars". Given that the current reading can be very fluctuating, the "Progress Bars" widgets would be more suitable for our needs.

Videos:

- cs-studio :  

https://github.com/ControlSystemStudio/phoebus/assets/104097021/8831e3ea-9a60-4157-8154-94f34a186392

- phoebus : 

https://github.com/ControlSystemStudio/phoebus/assets/104097021/8cd4bb42-5d46-4342-9e28-3b81fe41434c

Would it be possible to integrate the "Show Scale" option into the "Progress Bars" as is the case in CS-Studio?

Thank you in advance.

Xavier

georgweiss:
Sorry for late feed-back.

The Progress Bar widget is essentially a wrapper around the native JavaFX ProgressBar, which does not support tick marks or scale. There are good reasons to keep it lightweight.

Recently we added the Linear Meter widget. It has more customization options compared to the Progress Bar. Please give it a try to see if it meets your needs.

XavSPM:
I had thought about using LinearMeter, but I am encountering two problems with this tool.

The first is that it does not handle a minimum value higher than the maximum value well. I am now using this feature since the modification made to ProgressBar.

The second problem is that it is less visible to the user. It would be useful to be able to fill the LinearMeter to make it more similar to ProgressBar.

The simplest solution for my use case would be to add a scale to ProgressBar and to display the PV value inside it. I have started to explore this possibility, but not being experienced in Java, and lacking time, I have not yet been able to dedicate myself to it.
[...]

PR #3768, opened 2026-04-04
feat (display): RTTank-based Progress Bar backend — scale, alarm limits, and stacking support (opt-in via preference)

emilioheredia-source:
> **Depends on #3767.** This branch is stacked on `display/rtscaled-tank-refactor`: the first 3 commits here are #3767, only the last 3 commits are this PR. Please review and merge #3767 first; this PR will then show only its own diff. Until then the ReadTheDocs check may fail to compile `RTProgressBarRepresentation`, which extends `RTScaledWidgetRepresentation` from #3767.

## Update (Sep 2026)

Rebased onto current master and reworked after re-reading it:

- **The stock `ProgressBarRepresentation` is no longer touched.** An earlier revision of this PR edited it and, as a side effect, dropped its handling of inverted ranges (`minimum > maximum`, added in 2023). That is gone: with the default preference the widget is byte-for-byte the code on master, including the recent fix for vertical bars inside a Group (#3833).
- The `parallel_rendering` changes that used to be part of this PR were merged separately as #3766 and are dropped here.
- Default widget size and defaults of the pre-existing properties are unchanged.
- The new renderer honours inverted ranges the same way the stock one does (fill from the opposite end), which relies on the small `RTTank` fix in #3929. Without that fix an inverted range is ignored, as it is for the Tank on master today.
- The border is persisted as `tank_border_width`, the name the Tank already uses, for all scaled widgets: every legacy BOY `.opi` widget carries a `<border_width>` element, and a property of that name would have turned it into a bar border on import. `ProgressBarWidgetUnitTest` has a BOY import test for exactly that, and the importer now also maps BOY's `level_hi`/`level_lo`.
- `TankRepresentation` uses the same shared listener/look helpers as the new representation, so the Tank and the Progress Bar really share the code.

## Summary

This PR adds a second rendering backend for the existing `ProgressBar` widget, based on `RTTank`, the engine behind the Tank widget. It is **opt-in** through a preference:

    org.csstudio.display.builder.representation/progressbar_scale_mode=false

- `false` (default): the stock JavaFX `ProgressBarRepresentation` is used. **Nothing changes for existing displays.**
- `true`: `RTProgressBarRepresentation` is used for all Progress Bar widgets of that Phoebus instance.

The RTTank backend uses a different rendering path than the stock control (Java2D offscreen, blitted to a Canvas), which costs more CPU while the widget updates. For a discussion with side-by-side recordings see the comments in #3766. The widget also looks slightly different from the stock control (font, scale, fill), so making it the default would change every existing screen. Hence the preference.

## Screenshots

New renderer (`progressbar_scale_mode=true`), same `.bob` file:

progressbar_new_renderer

Stock renderer (default), same `.bob` file; the scale properties are simply ignored:
progressbar_stock_renderer

## What the new renderer adds

- Numeric scale next to the bar (`scale_visible`, off by default so a bar looks like a bar until asked otherwise), with `format`, `precision`, `major_ticks`/`minor_ticks`, `show_minor_ticks`, `perpendicular_tick_labels`
- Optional second scale on the opposite side (`opposite_scale_visible`)
- `show_scale_labels=false` draws tick marks without text, for stacking several bars under one labelled scale (see screenshot)
- Alarm limit lines LOLO / LO / HI / HIHI, from the PV or from `level_*` properties; PV limits are solid, manual ones dashed
- `log_scale` with a proper log axis
- `tank_border_width` and `inner_padding` (gap between widget edge and bar, default 3 px to match the stock CSS inset)
- `font` and `foreground_color` for the scale

## Model changes

`ProgressBarWidget` now extends `ScaledPVWidget` (shared with `TankWidget` since #3760), which is where the range, format and alarm limit properties come from. The scale look properties (font, fill color, log scale, scale visibility, tick options, border width) are defined once in `ScaledPVWidget.defineScaleLookProperties()` so the Thermometer PR can reuse them; the Tank keeps its own definitions and its released `tank_border_width` name. Existing `.bob` files load unchanged: `fill_color`, `background_color`, `horizontal`, `limits_from_pv`, `minimum`, `maximum` and `log_scale` keep their names. Older Phoebus versions ignore the new elements.

The property panel hides the scale-only properties while the stock renderer is selected (`ProgressBarWidget.SCALE_MODE_PROPS`), so operators do not see settings that have no effect. This is a deliberate, small coupling between the editor and the renderer preference; if you would rather keep the editor widget-agnostic, that check can simply be dropped and the properties shown.

The BOY `.opi` importer picks up `show_scale` and `scale_font`; the EDM converter maps the `activeBar` scale flag, range, precision and border.

## Design note

Two renderers for one widget type, selected globally, is a compromise. A new widget type would force every existing `.bob` to be migrated; a per-widget renderer property would leak the representation choice into the model. The preference keeps the model untouched, is reversible, and lets sites try the new renderer on real screens without editing files. If it proves out, it can become the default in a later release.

## Files changed

| Area | Files |
|---|---|
| Widget model | `ProgressBarWidget.java`, `ScaledPVWidget.java` (shared scale look properties), `TankWidget.java` (uses them), `Messages.java`, `messages*.properties`, `ProgressBarWidgetUnitTest.java`, `TankWidgetUnitTest.java` |
| Representation | `RTProgressBarRepresentation.java` (new), `RTScaledWidgetRepresentation.java` (shared listener/look helpers), `TankRepresentation.java` (uses them), `BaseWidgetRepresentations.java` |
| RTTank | `RTTank.java` (`inner_padding`, flat track) |
| Preference | `Preferences.java`, `display_representation_preferences.properties` |
| Editor | `PropertyPanelSection.java` |
| Converter | `Convert_activeBarClass.java` |

## Testing

Default:
1. Open displays with Progress Bar widgets without the preference: identical to before, including `minimum > maximum` bars.

With `progressbar_scale_mode=true`:
2. Existing `.bob` files open normally; the property panel shows the new properties.
3. `scale_visible` on/off, `minimum`/`maximum`, `major_ticks`/`minor_ticks`, `show_minor_ticks`, `perpendicular_tick_labels`.
4. `format`/`precision`: labels update.
5. `limits_from_pv` with a PV that carries LOLO/LO/HI/HIHI (e.g. `sim://sine(0,100,2)`) and `show_alarm_limits=true`: lines at the right positions; `alarm_limits_from_pv=false` with `level_*` gives dashed lines.
6. `show_scale_labels=false` on stacked bars: ticks stay, text goes.
7. `inner_padding` 0..20, `tank_border_width` 0..5, `foreground_color` for the scale.
8. `horizontal=false`; `minimum > maximum` in both orientations (needs #3929).
9. Tank widgets on the same display are unaffected.

## Checklist
- Testing:
    - [x] The feature has automated tests (`ProgressBarWidgetUnitTest`: defaults, legacy file loads unchanged, XML round trip of the new properties, default widget writes no new elements, hidden-property list matches real properties)
    - [x] Tests were run (`mvn install` with tests on the touched modules, plus the manual checks above)
- Documentation:
    - Release notes:


emilioheredia-source:
The new progress bar backend allows for looks like this, for instance: 
image

georgweiss:
To be honest, I think this overlaps quite substantially with the Linear Meter.

emilioheredia-source:
Hi Georg, I get your point: once the scale is on, the two do end up in similar territory.

Honestly, the reason I went this way is the migration path. Every legacy bar with a scale lands on the Progress Bar: EDM `activeBarClass`, MEDM `bar`, and BOY progress bars through the alias. Their scale settings (`show_scale`, `scale_font`, `scale_format`, level markers, ...) quietly disappear on import today. The Linear Meter has no alias and no converter points at it, so getting a scale back on a converted screen means replacing each bar by hand. I did try the meter first, after your suggestion in #2841, but if a display uses an inverted range it reports "Value < Min", and a converted row of bars should keep looking like a row of bars.

progressbar_vs_linearmeter


Nothing changes for anyone who doesn't opt in: with `progressbar_scale_mode=false`, the default, `ProgressBarRepresentation` is the unchanged file from master. With it on, and the scale off, the new renderer draws the bar the way the JavaFX style sheet does. It isn't pixel-identical, only a pixel or two at the edges and a few at the end of the fill, but colors, sizes and fill direction are the same, so the existing behaviour is still there either way:

progressbar_backends


I did think about pointing the converters at the Linear Meter instead, but I was a bit worried about existing sites: screens converted before and after the change would end up with different widget types, and rules or scripts that refer to the bar might need a second look. I also feel a scale is one of the things people miss most when they compare the bar with what they had in CS-Studio or EDM, so it seemed worth offering, at least as an option. The Linear Meter would of course keep everything that makes it different, like the needle, the warning regions, units and tolerance. For the Thermometer (#3930) I couldn't find another widget that covers it, which is part of why I went this way.

That said, you know the code base and its direction much better than I do, so I'm happy to go whichever way you think is best. If you'd prefer to keep the bar simple, I could close this and help add the missing pieces to the Linear Meter instead, together with the converter mappings. Just let me know what you'd like.

Cheers.

georgweiss:
@emilioheredia-source, I admit I did not think of the EDM conversion aspect. We need to consider that, of course.

However, I am concerned about the added maintenance burden having two widgets with overlapping functionality. I would advocate this to be discussed during the upcoming codathon at BNL. A possible strategy would be to extend the Linear Meter with any feature it lacks to be the preferred mapping from EDM bar widgets.

BoredDadDev:
Hi Georg, that is fair, and discussing it at the codeathon works for me. Keeping the scope manageable is a real concern, so let me put some numbers on the maintenance side and say where I stand on the Linear Meter.

On maintenance: these PRs do not add a second implementation. All three widgets render through RTTank and share one RTScaledWidgetRepresentation (357 lines) that holds the scale, the range and alarm limit handling and the update scheduling. What is specific to each widget is small: 57 lines for the Tank, 69 for the Progress Bar, 67 for the Thermometer, plus the property list in each widget class. So after #3767 the marginal cost of a widget on this backend is on the order of 70 lines, and a fix in the scale code fixes all three at once. That is also why the RTTank fixes in #3929 are a separate PR: they are bugs in code that already ships.

On making the Linear Meter the EDM target: I am happy to go that way if that is the direction you prefer, and to do the work. Two things to weigh first. The meter would need inverted ranges, which today produce "Value < Min", and it would need a plain bar look without the needle and the zones for converted screens, so one of the two widgets ends up with a second rendering mode either way. The question is mainly which one carries it. There is also a compatibility wrinkle: pointing the converters at a different widget type means screens converted before and after the change get different types, and rules or scripts that refer to the bar would need a second look.

Worth noting that nothing changes for existing users in the meantime. With progressbar_scale_mode=false, the default, ProgressBarRepresentation is the unchanged file from master, so this is opt-in in the strict sense, and we can treat it as a trial period and revisit once there is feedback from people who turned it on.

I can prepare a short feature comparison of the Progress Bar, the Linear Meter and the EDM/MEDM/BOY bar sources before the codeathon, so the discussion has something concrete to work from. I will not be at BNL, but I am glad to join that session remotely if that is useful. For what it is worth, CLS started using Phoebus last year and maintains a fork, so these widgets will be maintained on our side regardless. I would much rather spend that time maintaining them upstream than resolving merge conflicts on every pull, and I am happy to be the contact for them.

One request while this is open: #3929 is an independent fix for two regressions in RTTank and does not touch the Progress Bar, and #3767 is approved and is the refactor the Tank needs on its own. Could those two go in regardless of how the bar discussion ends? That would leave only #3768 and #3930 for the codeathon.

Thanks for taking the time on this.

Emilio

emilioheredia-source:
Hi Georg, that is fair, and discussing it at the codeathon works for me. Keeping the scope manageable is a real concern, so let me put some numbers on the maintenance side and say where I stand on the Linear Meter.

On maintenance: these PRs do not add a second implementation. All three widgets render through RTTank and share one RTScaledWidgetRepresentation (357 lines) that holds the scale, the range and alarm limit handling and the update scheduling. What is specific to each widget is small: 57 lines for the Tank, 69 for the Progress Bar, 67 for the Thermometer, plus the property list in each widget class. So after https://github.com/ControlSystemStudio/phoebus/pull/3767 the marginal cost of a widget on this backend is on the order of 70 lines, and a fix in the scale code fixes all three at once. That is also why the RTTank fixes in https://github.com/ControlSystemStudio/phoebus/pull/3929 are a separate PR: they are bugs in code that already ships.

You can notice the similarity among the three widgets, the shared code:
shared_scale_three_widgets


On making the Linear Meter the EDM target: I am happy to go that way if that is the direction you prefer, and to do the work. Two things to weigh first. The meter would need inverted ranges, which today produce "Value < Min", and it would need a plain bar look without the needle and the zones for converted screens, so one of the two widgets ends up with a second rendering mode either way. The question is mainly which one carries it. There is also a compatibility wrinkle: pointing the converters at a different widget type means screens converted before and after the change get different types, and rules or scripts that refer to the bar would need a second look.

Worth noting that nothing changes for existing users in the meantime. With progressbar_scale_mode=false, the default, ProgressBarRepresentation is the unchanged file from master, so this is opt-in in the strict sense, and we can treat it as a trial period and revisit once there is feedback from people who turned it on.

I can prepare a short feature comparison of the Progress Bar, the Linear Meter and the EDM/MEDM/BOY bar sources before the codeathon, so the discussion has something concrete to work from. I will not be at BNL, but I am glad to join that session remotely if that is useful. For what it is worth, CLS started using Phoebus last year and maintains a fork, so these widgets will be maintained on our side regardless. I would much rather spend that time maintaining them upstream than resolving merge conflicts on every pull, and I am happy to be the contact for them.

One request while this is open: https://github.com/ControlSystemStudio/phoebus/pull/3929 is an independent fix for two regressions in RTTank and does not touch the Progress Bar, and https://github.com/ControlSystemStudio/phoebus/pull/3767 is approved and is the refactor the Tank needs on its own. Could those two go in regardless of how the bar discussion ends? That would leave only https://github.com/ControlSystemStudio/phoebus/pull/3768 and https://github.com/ControlSystemStudio/phoebus/pull/3930 for the codeathon.

Thanks for taking the time on this.

Emilio

Issue #3763, opened 2026-04-02
Can no longer run phoebus at SNS because preference check limits the size of all preferences


kasemir:
#3740 added this to take a snapshot of all settings in the `Launcher`:

```
Preferences.userRoot().put(SETTINGS_SNAPSHOT, prefsBuf.toString());
```

That bombs like this
```
java.lang.IllegalArgumentException: Value too long: # Preference settings<br/>
...
# End.<br/>
	at java.prefs/java.util.prefs.AbstractPreferences.put(AbstractPreferences.java:248)
	at org.phoebus.product.Launcher.loadSettings(Launcher.java:233)
	at org.phoebus.product.Launcher.main(Launcher.java:140)
```

because java.util.prefs.AbstractPreferences.put(AbstractPreferences.java:248) looks like this
```
   public void put(String key, String value) {
...
        if (value.length() > MAX_VALUE_LENGTH)
            throw new IllegalArgumentException("Value too long: "+value);
...
```

with a MAX_VALUE_LENGTH of 
```
   /**
     * Maximum length of string allowed as a value (8192 characters).
     */
    public static final int MAX_VALUE_LENGTH = 8*1024;
```

shroffk:
A potential idea would be to store the `SETTINGS_SNAPSHOT` not in the Preferences but maybe in a temporary file... not really a fan of this idea honestly

lcaouen:
> A potential idea would be to store the `SETTINGS_SNAPSHOT` not in the Preferences but maybe in a temporary file... not really a fan of this idea honestly

Not sure it's a good idea because you'll need to find a directory where you are sure Phoebus has the rights to write and which will work for any OS (Linux, Windows, Mac, ...).

We need to find a way to store the "orginal" settings in memory (just a String stored somewhere is enough) or a flag to say this setting is an "original" one and this one is a "hard-coded" one.

Let's brainstorm on Thursday

PR #3771, opened 2026-04-06
Revert 3740

shroffk:
Revert the changes from #3740 util we can come up with a permanent solutions

#3763

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:


kasemir (approved):

lcaouen (approved):
Tested and it looks like the revert is ok.
I can see settings in red and the "memory settings" too.

Issue #3773, opened 2026-04-07
readthedoc build failing with time out


shroffk:
It seems like the recent readthedocs build are filing with a timeout.
https://app.readthedocs.org/projects/control-system-studio/builds/32119186/

It seems like it the maven build step takes too long when running all the tests...I would like to disable the tests, I don't think they are needed for generating the documentation and they are run as part of the CI built on github actions

PR #3774, opened 2026-04-07
Modify maven build to skip tests when building the readthedocs

shroffk:
addressing #3773

We don't need to run all the unit tests when building documentation

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:


lcaouen (approved):
Looks good to me.
By the way I'm wondering if there is a maven goal faster than "clean install" to build only the documentation without building all the binaries...

jacomago (approved):

PR #3780, opened 2026-04-15
Fix double-layered DER decoding when reading X.509 certificate extension OIDs

george-mcintyre:
## Problem

`X509Certificate.getExtensionValue(oid)` returns a DER-encoded `OCTET STRING` whose *payload* is itself another DER-encoded value (the actual extension content). The previous code only unwrapped one DER layer, so the certificate status PV name stored under OID `1.3.6.1.4.1.37427.1` was never correctly extracted.

## Fix

Unwrap both DER layers: first strip the outer `OCTET STRING` tag/length, then strip the inner `OCTET STRING` tag/length to reach the raw UTF-8 string payload. This is required to correctly read any X.509v3 extension value via `getExtensionValue()`.

This fix is only relevant when the client needs to locate the cert-status PV name from the peer's certificate; it has no effect on connections that do not use certificate status gating.

## Files Changed

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/common/SecureSockets.java
+++ b/core/pva/src/main/java/org/epics/pva/common/SecureSockets.java
@@ -274,8 +274,13 @@ public class SecureSockets
     {
         if (der_value == null)
             return "";
-        // https://en.wikipedia.org/wiki/X.690#DER_encoding:
-        // Type 4, length 0..127, characters
+        // X509Certificate.getExtensionValue() returns a DER OCTET STRING
+        // that wraps the actual extension content.
+        // The extension content itself is a DER-encoded string
+        // (OCTET STRING 0x04 or UTF8String 0x0C), so we must unwrap two layers:
+        //   Outer: 0x04 <len> <inner DER>
+        //   Inner: 0x04|0x0C <len> <actual string bytes>
+        // https://en.wikipedia.org/wiki/X.690#DER_encoding
         if (der_value.length < 2)
             throw new Exception("Need DER type and size, only received " + der_value.length + " bytes");
         if (der_value[0] != 0x04)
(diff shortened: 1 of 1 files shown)


kasemir (approved):

kasemir:
@george-mcintyre for what it's worth, this doesn't work.

When I run a pva client against a softIocPVX, where I used to see the "Status PV: CERT:STATUS:.."...
```
FINE [org.epics.pva] Server name: 'ioc'
FINE [org.epics.pva] * O=diane.ornl.gov, C=US, CN=ioc
FINE [org.epics.pva]   - Server/IOC CN
FINE [org.epics.pva]   - Expires Thu Nov 26 13:43:21 EST 2026
FINE [org.epics.pva]   - Status PV: 'CERT:STATUS:ed890ce4:12942841478827205582'
```

I now get

```
FINE [org.epics.pva] Server name: 'ioc'
FINE [org.epics.pva] * O=diane.ornl.gov, C=US, CN=ioc
FINE [org.epics.pva]   - Server/IOC CN
FINE [org.epics.pva]   - Expires Thu Nov 26 13:43:21 EST 2026
WARNING [org.epics.pva] Error while logging result of handshake with server
java.lang.Exception: Expected inner DER string type 0x04, 0x0C, or 0x16, got 0x43
	at org.epics.pva.common.SecureSockets.decodeDERString(SecureSockets.java:324)
	at org.epics.pva.common.SecureSockets.createClientSocket(SecureSockets.java:279)
	at org.epics.pva.client.ClientTCPHandler.initializeSocket(ClientTCPHandler.java:134)
	at org.epics.pva.common.TCPHandler.receiver(TCPHandler.java:272)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1623)
```

Reverting this commit fixes it for me

kasemir:
> `Expected inner DER string type 0x04, 0x0C, or 0x16, got 0x43`

0x43 is ASCII 'C', the first character of "CERT:STATUS:..."

The plain string is in there, no double-layering

kasemir:
@george-mcintyre @mdavidsaver Turns out this is a change somewhere between the 1.3 and 1.5 version of pvxs-tls.

The  "1.3.6.1.4.1.37427.1" extension used to be a DER encoded OCTET STRING (0x04) with value "CERT:STATUS:....".

Now it's a OCTET STRING which in turn holds an IA5String (0x16) as inner type, and that one has a value "CERT:STATUS:...".

This is not a globally standardized extension. We can put anything into it, and we decided to place the CERT:STATUS PV name there. Fine. But is there a reason to change from `OCTET STRING { "CERT_STATUS:..." }` to 
`OCTET STRING { IA5String{"CERT_STATUS:..."}}`.
 Seems to make it incompatible just for fun.

george-mcintyre:
Michael told me to change it. Not for fun 😀Sent from my iPhoneOn 4 Jun 2026, at 17:18, Kay Kasemir ***@***.***> wrote:﻿kasemir left a comment (ControlSystemStudio/phoebus#3780)
@george-mcintyre @mdavidsaver Turns out this is a change somewhere between the 1.3 and 1.5 version of pvxs-tls.
The  "1.3.6.1.4.1.37427.1" extension used to be a DER encoded OCTET STRING (0x04) with value "CERT:STATUS:....".
Now it's a OCTET STRING which in turn holds an IA5String (0x16) as inner type, and that one has a value "CERT:STATUS:...".
This is not a globally standardized extension. We can put anything into it, and we decided to place the CERT:STATUS PV name there. Fine. But is there a reason to change from OCTET STRING { "CERT_STATUS:..." } to
IA5String{ OCTET STRING{"CERT_STATUS:..."}}? Seems to make it incompatible just for fun.

mdavidsaver:
> Michael told me to change it. Not for fun

For [standards](https://datatracker.ietf.org/doc/html/rfc5280) compliance.

> ... Each extension includes an OID and an ASN.1 structure.  When an
   extension appears in a certificate, the OID appears as the field
   extnID and the corresponding ASN.1 DER encoded structure is the value
   of the octet string extnValue.  ...

The key point is that each extension value must itself be a DER encoded structure.  So yes, double encoded.

kasemir:
Got it, thanks!

PR #3782, opened 2026-04-15
Add XDG standard path auto-discovery for PVA TLS keychains

george-mcintyre:
## Motivation

PVXS (the reference C++ PVA implementation) stores its keystores at the XDG standard location:

```
$XDG_CONFIG_HOME/pva/1.5/client.p12   (client keychain)
$XDG_CONFIG_HOME/pva/1.5/server.p12   (server keychain)
```

where `$XDG_CONFIG_HOME` defaults to `~/.config` on Linux and the platform-appropriate equivalent on macOS/Windows.

The Java client had no equivalent discovery logic: `EPICS_PVA_TLS_KEYCHAIN` had to be set explicitly, making interoperability with a PVXS-configured system painful in development environments and CI.

## Change

`PVASettings` now probes the XDG path for `client.p12` / `server.p12` at startup when `EPICS_PVA_TLS_KEYCHAIN` / `EPICS_PVAS_TLS_KEYCHAIN` are not explicitly set. If a keychain is found at the XDG location it is used automatically, matching PVXS behaviour and requiring zero configuration when the user has already set up PVXS.

Explicit env-var settings always take precedence; the XDG probe is only a fallback.

## Files Changed

- `core/pva/src/main/java/org/epics/pva/PVASettings.java`

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/PVASettings.java
+++ b/core/pva/src/main/java/org/epics/pva/PVASettings.java
@@ -289,7 +289,7 @@ public class PVASettings
         EPICS_PVA_TCP_SOCKET_TMO = get("EPICS_PVA_TCP_SOCKET_TMO", EPICS_PVA_TCP_SOCKET_TMO);
         EPICS_PVA_MAX_ARRAY_FORMATTING = get("EPICS_PVA_MAX_ARRAY_FORMATTING", EPICS_PVA_MAX_ARRAY_FORMATTING);
         EPICS_PVAS_TLS_KEYCHAIN = get("EPICS_PVAS_TLS_KEYCHAIN", EPICS_PVAS_TLS_KEYCHAIN);
-        if (EPICS_PVAS_TLS_KEYCHAIN.isEmpty())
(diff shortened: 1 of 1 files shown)


kasemir (approved):

kasemir (changes requested):
Actually, that might now work.
`if (EPICS_PVAS_TLS_KEYCHAIN.isEmpty())`, it auto-discovers. How do I disable TLS? With the C++ implementation, I understand you can
```
export  EPICS_PVAS_TLS_KEYCHAIN=""
```
so now it's defined but empty, and that disables TLS. 

Right now it's a little hard to distinguish between not-defined and defined-as-empty....

george-mcintyre (commented):
Good catch, thanks Kay. Fixed in a9d5e46 — added `isDefined()` which checks `System.getProperty()` and `System.getenv()` directly to distinguish "variable absent" from "variable explicitly set to empty".

The XDG auto-discovery now only runs when the variable is **absent** (not defined in env or as a Java property). Setting `export EPICS_PVA_TLS_KEYCHAIN=""` (or `export EPICS_PVAS_TLS_KEYCHAIN=""`) explicitly to blank will suppress discovery and keep TLS disabled, matching the C++ behaviour you described.

PR #3783, opened 2026-04-15
Add EPICS_PVA_TLS_KEYCHAIN_PWD_FILE for file-based keychain password

george-mcintyre:
## Motivation

The existing `EPICS_PVA_TLS_KEYCHAIN` setting embeds the keystore password directly in the path string (e.g. `path/to/client.p12|password`). This is convenient for development but unsuitable for production environments where secrets must not appear in environment variables, process listings, or configuration files.

Container orchestration platforms such as Kubernetes mount secrets as files (e.g. via `Secret` volumes or external secret operators like Vault Agent). There was no way to supply the keychain password via a file path.

## Change

Two new environment variables are introduced:

| Variable | Purpose |
|---|---|
| `EPICS_PVA_TLS_KEYCHAIN_PWD_FILE` | Path to a file containing the client keychain password |
| `EPICS_PVAS_TLS_KEYCHAIN_PWD_FILE` | Path to a file containing the server keychain password |

When set, `SecureSockets` reads the password from the referenced file instead of parsing it from the keychain path string. The inline `path|password` syntax continues to work; the `_PWD_FILE` variable takes precedence when both are present.

This enables secure password injection for Kubernetes deployments using `secretKeyRef` volume mounts without patching the keychain path.

## Files Changed

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/PVASettings.java
+++ b/core/pva/src/main/java/org/epics/pva/PVASettings.java
@@ -153,6 +153,22 @@ public class PVASettings
      */
     public static String EPICS_PVAS_TLS_KEYCHAIN = "";
 
+    /** Path to a file containing the password for the PVA server keychain.
+     *
+     *  <p>Alternative to embedding the password in {@link #EPICS_PVAS_TLS_KEYCHAIN}
+     *  using the "/path/to/file;password" syntax.
+     *  When set, the password is read from this file instead, with leading and trailing
+     *  whitespace stripped.
+     *  Takes precedence over an inline password in {@link #EPICS_PVAS_TLS_KEYCHAIN}
+     *  when no ";" separator is present in that setting.
+     *
+     *  <p>Intended for environments where secrets are mounted as files,
+     *  for example Kubernetes pods using a {@code Secret} volume.
+     *
+     *  <p>When empty, no password file is used.
+     */
+    public static String EPICS_PVAS_TLS_KEYCHAIN_PWD_FILE = "";
+
     /** Secure server options
      *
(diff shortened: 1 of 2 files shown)


kasemir (changes requested):
Like all PVA settings, the new `EPICS_PVAS_TLS_KEYCHAIN_PWD_FILE` and `EPICS_PVA_TLS_KEYCHAIN_PWD_FILE` should be defined and documented in `src/main/java/org/epics/pva/PVASettings.java`, next to the existing definition and documentation of `PVASettings.EPICS_PVAS_TLS_KEYCHAIN` and `PVASettings.EPICS_PVA_TLS_KEYCHAIN`

george-mcintyre (commented):
Done in cd72fb5. Both `EPICS_PVAS_TLS_KEYCHAIN_PWD_FILE` and `EPICS_PVA_TLS_KEYCHAIN_PWD_FILE` are now declared as `public static String` fields in `PVASettings`, documented in the same style as `EPICS_PVAS_TLS_KEYCHAIN` / `EPICS_PVA_TLS_KEYCHAIN` immediately adjacent to them, and initialised in the static block right after their respective keychain peers. `SecureSockets.readKeychainPassword()` now reads from the `PVASettings` fields directly instead of calling `PVASettings.get()` ad hoc.

kasemir:
Looks good, except there's now a minor merge conflict

kasemir:
As described in #3879, `EPICS_PVA(S)_TLS_KEYCHAIN_PWD_FILE` has been removed from PVXS and there's thus no longer any need to add it to the Java version.

PR #3784, opened 2026-04-15
Expose TLS connection details through PV → Channel → TCPHandler chain

george-mcintyre:
## Motivation

There was no API for applications or display widgets to inspect TLS connection details (whether TLS is active, the authenticated server identity, the client identity, and the remote address) for a live PVA connection. This information is needed both for diagnostics and for the PVA Security Widget.

## Change

Adds read-only accessors through the full PVA client stack:

| Class | New methods |
|---|---|
| `ClientTCPHandler` | `isTLS()`, `getServerX509Name()`, `getClientX509Name()`, `getAuthenticationInfo()`, `getRemoteAddress()` |
| `PVAChannel` | Delegates the above through to its `ClientTCPHandler` |
| `PVA_PV` (core/pv-pva) | Delegates the above through to its `PVAChannel` |

All methods are safe to call at any point in the channel lifecycle; they return `false` / `null` / empty string when the connection is not established or TLS is not in use.

## Files Changed

- `core/pva/src/main/java/org/epics/pva/client/ClientTCPHandler.java`
- `core/pva/src/main/java/org/epics/pva/client/PVAChannel.java`
- `core/pv-pva/src/main/java/org/phoebus/pv/pva/PVA_PV.java`

### DIFF
--- a/core/pv-pva/src/main/java/org/phoebus/pv/pva/PVA_PV.java
+++ b/core/pv-pva/src/main/java/org/phoebus/pv/pva/PVA_PV.java
@@ -207,6 +207,36 @@ public class PVA_PV extends PV
         return channel.write(true, name_helper.getRequest(), new_value);
     }
 
+    /** @return <code>true</code> if the connection to the server uses TLS */
+    public boolean isTLS()
+    {
+        return channel.isTLS();
+    }
+
+    /** @return Common Name from the server's X.509 certificate, or <code>null</code> */
+    public String getServerX509Name()
+    {
+        return channel.getServerX509Name();
+    }
+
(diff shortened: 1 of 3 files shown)


kasemir (approved):
It's fine to expose this info from core-pva.

I'm ambivalent about exposing it from pv-pva. The PV layer is on purpose designed as a common denominator. Generic PV tools don't know if the PV is local, simulated, CA (JCA) or CA (Jackie), PVA, Tango, MQTT, ..
We do NOT expose channel access details, not even the IOC's IP address, to decouple the generic tools. Tools that now check `pv instanceof PVA_PV` will break the abstraction.

So this is "well, OK, but don't actually use it"

PR #3785, opened 2026-04-15
Add PVA Security Widget showing runtime TLS and authentication details

george-mcintyre:
## Overview

Adds a new **PVA Security** widget to the Display Builder. When placed on a display and configured with a PV name, the widget shows live TLS and authentication metadata for that PV's underlying PVA connection:

- **TLS active** — whether the connection is secured with TLS
- **Authentication method** — e.g. `x509`, `ca`, or anonymous
- **Server identity** — X.509 subject name of the server certificate
- **Client identity** — X.509 subject name presented by the client

The widget forces the `pva://` protocol prefix on the configured PV name at runtime so it always interrogates the actual PVA connection, regardless of the default protocol configured in the environment.

## Implementation

Three new classes following the standard Display Builder widget pattern:

| Class | Module | Purpose |
|---|---|---|
| `PVASecurityWidget` | `app/display/model` | Widget model: properties, XML persistence |
| `PVASecurityRepresentation` | `app/display/representation-javafx` | JavaFX rendering |
| `PVASecurityRuntime` | `app/display/runtime` | PV connection logic, forces `pva://` prefix |

Registered via the existing SPI `BaseWidgetsService`, `BaseWidgetRepresentations`, and `BaseWidgetRuntimes`.

Depends on the TLS accessor API added in the companion PR "Expose TLS connection details through PV → Channel → TCPHandler chain".

## Files Changed

- `app/display/model/src/main/java/org/csstudio/display/builder/model/BaseWidgetsService.java`
- `app/display/model/src/main/java/org/csstudio/display/builder/model/widgets/PVASecurityWidget.java` *(new)*
- `app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/BaseWidgetRepresentations.java`
- `app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/PVASecurityRepresentation.java` *(new)*
- `app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/internal/BaseWidgetRuntimes.java`
- `app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/internal/PVASecurityRuntime.java` *(new)*
- `app/display/runtime/pom.xml`

### DIFF
--- a/app/display/model/.classpath
+++ b/app/display/model/.classpath
@@ -1,12 +1,61 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
-	<classpathentry kind="src" path="src/main/resources"/>
-	<classpathentry kind="src" path="src/test/java"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17">
+		<attributes>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" output="target/classes" path="src/main/java">
+		<attributes>
+			<attribute name="optional" value="true"/>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources">
+		<attributes>
+			<attribute name="maven.pomderived" value="true"/>
+			<attribute name="optional" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" output="target/test-classes" path="src/test/java">
+		<attributes>
+			<attribute name="optional" value="true"/>
+			<attribute name="maven.pomderived" value="true"/>
+			<attribute name="test" value="true"/>
+		</attributes>
+	</classpathentry>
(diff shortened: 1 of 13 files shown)


kasemir:
This is problematic because it makes the generic display depend on pv-pva and core-pva.
The display was designed to only depend on the abstracted core-pv. It can be used with any combination of simulated, local, CA, PVA, MQTT, Tango, ... PVs.
We do not check `pv instanceof PV_CA` and likewise should not add a check `pv instanceof PV_PVA` that makes it depend on any specific PV type.

This does mean that we knowingly drop the option of showing PV-specific details like a CA server IP. We leave that to `cainfo` on the command line, outside of the generic CS-Studio tools.

To still allow access to the PVA connection detail in displays, there are some other options:

 * Move the PVA security widget to an outside module. For example, a new `app/display/pva`
 
 * Instead of creating a new widget, add formula functions like ```pvaTLSstate(`SomePVName`)```, ```pvaServerIdentity(`SomePVName`)``` which can then be used with existing TextUpdate or LED or .. widgets. Since those formulas would again depend on pv-pva, they need to be in a separate module, not core-formula

kasemir:
Third option: The `pv-pva` module right now serves `pva://SomePvaPvName`. In addition, it could serve the PVA connection detail via a separate PV schema like `pvainfo://SomePvaPvName`. That way, the PVA info appears just like any of the other PVs. You can show it in probe, displays, data browser etc. Would also right away work in the web runtime.

I think that's the most promising option. The hardest part would be
 1) Come up with a good schema name. `pvainfo://`? `pvai://`?
2) That new `pvainfo://SomePvaPvName` can provide a structure, and we can then for example show the server ID as
`pvainfo://SomePvaPvName/serverID`.  Should it be one structure? Or does that get too large, want separate simpler PVs?

PR #3786, opened 2026-04-15
Add certificate status gating for PVA TLS connections

george-mcintyre:
## Overview

After a TLS handshake, if the peer certificate contains a certificate status PV name (stored in the custom OID `1.3.6.1.4.1.37427.1`), the connection now waits for that PV to report `VALID` before allowing data operations. This implements the PVA TLS certificate status protocol, matching PVXS behaviour.

## Behaviour

| Scenario | Result |
|---|---|
| Status PV reports `VALID` within timeout | Connection proceeds normally |
| Status PV not confirmed within `EPICS_PVA_CERT_STATUS_TMO` (default 30 s) | Degraded mode: data operations released with a warning logged |
| Status PV reports `REVOKED` or `EXPIRED` | Connection is immediately shut down |
| Certificate has no status PV OID | No gating; connection proceeds immediately |

## Client side

- `ClientTCPHandler` starts a `CertificateStatusMonitor` after a successful TLS handshake.
- Data operations are gated on a `CompletableFuture` that resolves when the status PV responds.
- The monitor uses a `PVAClient(tls_disabled=true)` to avoid infinite recursion (cert-status monitoring cannot itself require cert-status monitoring).
- OCSP responses are tracked with expiry timers; when an OCSP response expires the connection re-validates.

## Server side

- `ServerTCPListener` and `ServerTCPHandler` start a `CertificateStatusMonitor` for the client certificate after TLS handshake.
- `CreateChannelHandler` gates its reply on the client certificate status being `VALID`.

## Configuration

| Variable | Default | Description |
|---|---|---|
| `EPICS_PVA_CERT_STATUS_TMO` | `30` | Seconds to wait for cert status PV before entering degraded mode |

## Dependencies

This branch includes the prerequisite commits for XDG keychain discovery, keychain password file support, CA certificate extraction, and DER double-unwrap, since the certificate status monitor relies on `SecureSockets` and `PVASettings` helpers introduced in those changes. Those changes are also submitted as standalone PRs:
- Fix CA certificate extraction from PKCS12 key entry chains (#3779)
- Fix double-layered DER encoding in certificate extension decoding (#3780)
- Add XDG standard path auto-discovery for PVA TLS keychains (#3782)
- Add EPICS_PVA_TLS_KEYCHAIN_PWD_FILE for file-based keychain password (#3783)

## Files Changed

- `core/pva/src/main/java/org/epics/pva/PVASettings.java`
- `core/pva/src/main/java/org/epics/pva/client/ClientTCPHandler.java`
- `core/pva/src/main/java/org/epics/pva/common/CertificateStatus.java`
- `core/pva/src/main/java/org/epics/pva/common/CertificateStatusMonitor.java`
- `core/pva/src/main/java/org/epics/pva/server/CreateChannelHandler.java`
- `core/pva/src/main/java/org/epics/pva/server/ServerTCPHandler.java`
- `core/pva/src/main/java/org/epics/pva/server/ServerTCPListener.java`

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/PVASettings.java
+++ b/core/pva/src/main/java/org/epics/pva/PVASettings.java
@@ -21,13 +21,6 @@ import java.util.logging.Logger;
  */
 public class PVASettings
 {
-    /** PVA protocol version for XDG path construction.
-     *
-     *  <p>Matches PVXS versionString() so that the Java client
-     *  and PVXS share the same well-known keychain locations.
-     */
-    private static final String PVA_VERSION = "1.5";
-
     /** Common logger
      *
      *  Usage of levels:
@@ -267,6 +260,16 @@ public class PVASettings
 
 
 
+    /** Timeout [seconds] for waiting for certificate status PV to confirm VALID
+     *
+     *  <p>After a TLS handshake, if the certificate includes a status PV extension
+     *  (OID 1.3.6.1.4.1.37427.1), the connection will wait for the status PV
+     *  to report VALID before allowing data operations.
+     *  If not confirmed within this timeout, the connection enters degraded mode:
+     *  data operations are released with a warning.
+     */
+    public static int EPICS_PVA_CERT_STATUS_TMO = 30;
+
     /** Whether to allow PVA to use IPv6
      *
      *  <p> If this is false then PVA will not attempt to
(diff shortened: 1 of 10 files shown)


kasemir:
I would defer this because it is a significant change with no immediate need.

This Java implementation aims to be compatible with the C++ (PVXS) implementation. This includes the PVXS security additions which as of now have not been released. Java servers and clients can use the same certificates as PVXS servers and clients.
For a PVA server, it is essential to know the identity of the PVA client. That way, write access to PVs can be controlled based on the trusted client authentication. While the Java PVA server so far sees very little use (https://github.com/controlSystemStudio/pvaify is the only known utility that depends on it), the Java PVA server does already implement the client certificate verification based on optional CERT:STATUS:.. PVs provided by PVACMS.
For the client, there is no immediate need to verify the server identity because there is no mechanism to use that information. PV names are not tied to specific IOCs. Clients do not keep a white list of approved servers.
The client-side check of a server certificate should thus be delayed until the PVXS implementation settles into a release, and maybe even further until clients start to adopt mechanisms that require the server cert validation.

PR #3789, opened 2026-04-20
Allow setting record options externally

jacomago:
## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/client/RecordOptions.java
+++ b/core/pva/src/main/java/org/epics/pva/client/RecordOptions.java
@@ -47,21 +47,25 @@ public class RecordOptions {
         private boolean completion;
         private int pipeline;
         private DBEMask dbeMask = null;
+
         public Builder() {
         }
 
-        Builder  completion(boolean completion) {
+        public Builder completion(boolean completion) {
             this.completion = completion;
             return this;
         }
-        Builder pipeline(int pipeline) {
+
+        public Builder pipeline(int pipeline) {
             this.pipeline = pipeline;
             return this;
         }
-        Builder dbeMask(DBEMask dbeMask) {
+
+        public Builder dbeMask(DBEMask dbeMask) {
             this.dbeMask = dbeMask;
             return this;
         }
+
         public RecordOptions build() {
             return new RecordOptions(completion, pipeline, dbeMask);
         }


shroffk:
Is this a requirement from archiver appliance

kasemir (approved):

jacomago:
> Is this a requirement from archiver appliance

Yes! But also maybe alarms...

kasemir:
> ...also maybe alarms...

We started out running the alarm server with
```
org.phoebus.pv.ca/monitor_mask=ALARM
```

Seemed logical, after all the alarm server only cares about changes in the alarm state.
But that means you won't see the current value of PVs when you list them in the alarm server console, since only changes that come with different alarm states will get through.
Also means that expressions for alarm filters will likely not get the expected behavior since they miss most value updates. OK, alarm filters were a bad idea anyway, should create appropriate alarm trigger PVs on the IOC in the first place. Still, switching monitors from VALUE to ALARM tends to create emails from surprised users.

jacomago:
@mdavidsaver mentioned to me that he though the IOC should send the value updates as well and said he would aim to add that to pvxs. @mdavidsaver did you make an issue ticket for that?

 Or of course we could define a new mask ALARM_WITH_VALUE?

kasemir:
With channel access, the raw masks are bits which can be combined.
At a higher level, when we select "VALUE" for our preference, that turns into `Monitor.VALUE | Monitor.ALARM` on the CAJ level. "ALARM" is meant to reduce the updates so really just `Monitor.ALARM`, but as mentioned that's likely counterproductive, the default is better.

https://github.com/ControlSystemStudio/phoebus/blob/7e3fbdf1dd49489e5f9aa7ef88228b60e4b224dc/core/pv-ca/src/main/java/org/phoebus/pv/ca/JCA_Preferences.java#L88-L98

Should be the same with PVA: Default "VALUE" gets you any changes in the data. Certainly value changes, but also changes where only the status/severity change. "ALARM" would reduce it to only alarm changes. That option is pretty useless, but we keep it available.

"ARCHIVE" is like the default but at a reduced value change threshold.

mdavidsaver:
> @mdavidsaver mentioned to me that he though the IOC should send the value updates as well and said he would aim to add that to pvxs. @mdavidsaver did you make an issue ticket for that?

I believe this is related to the DBE mask for monitors.  See https://github.com/epics-base/pvxs/issues/97 .  This DBE mask has the same meaning as with CA.  So both dimensions of what (which fields) and when (which events) conditions can be specified in the pvRequest for a subscription.

Issue #3169, opened 2024-10-17
High CPU usage with multiple detached windows


rjwills28:
I have been testing the **performance** of Phoebus, measuring the CPU and memory usage. I have found that the CPU usage is very high when opening multiple screens in detached windows in Phoebus. I have tested opening 10 windows from the command line using the following command:
 ```
./phoebus.sh -resource file:/path/to/bob/testBob0.bob?target=window -resource file:/path/to/bob/testBob1.bob?target=window -resource file:/path/to/bob/testBob2.bob?target=window ...
 ```

 The **CPU usage** of Phoebus was ~120%, while opening the same number of standalone windows in CS-Studio was of order 40%.

 In addition I see a lot of threading related **exceptions** thrown when initially opening 6 or more detached windows in this way. Below is a snippet but I have also attached the full log: [phoebus_thread_exceptions.txt](https://github.com/user-attachments/files/17409307/phoebus_thread_exceptions.txt)

```
2024-10-16 12:50:25 SEVERE [org.phoebus.ui.application.PhoebusApplication] UI Freezeup
...
*********************************
*** JavaFX Application Thread ***
*********************************
"JavaFX Application Thread" prio=5 Id=34 WAITING on java.util.concurrent.locks.ReentrantLock$NonfairSync@68d8eaf8 owned by "QuantumRenderer-0" Id=29
	at java.base@21.0.4/jdk.internal.misc.Unsafe.park(Native Method)
	-  waiting on java.util.concurrent.locks.ReentrantLock$NonfairSync@68d8eaf8
	at java.base@21.0.4/java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
	at java.base@21.0.4/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
	at java.base@21.0.4/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:990)
	at java.base@21.0.4/java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
	at java.base@21.0.4/java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
	at app//com.sun.javafx.tk.quantum.QuantumToolkit.runWithRenderLock(QuantumToolkit.java:440)
[...]

PR #3792, opened 2026-04-27
Updatethrottle singleton

rjwills28:
A little while we were looking at the performance of Phoebus when there are multiple windows open (see issue https://github.com/ControlSystemStudio/phoebus/issues/3169), specifically that CPU usage was very high with multiple windows with PVs updating at 10Hz. 

After some further investigations, including testing a standalone JavaFX application, we found that a lot of the bad performance stems from JavaFX itself and the way it handles repainting multiple windows (as also suggested by Kay in the above issue). 

However, we continued to look for any ways that we could improve performance on the Phoebus side, which we address in this PR. Currently Phoebus creates an `RepresentationUpdateThrottle` instance for _each_ window. This class is in charge of collecting updates and running them on the UI thread. This means that in the case of 10 windows, there are 10 `RepresentationUpdateThrottle` all scheduling jobs on the single UI thread. 

Instead we found that it was slightly more efficient to turn the `RepresentationUpdateThrottle` class into a singleton and only have this one thread to collect updates across all windows and hence there is only one thread trying to run jobs on the UI thread. In testing, with this change, we saw CPU usage drop by about 10%, e.g. from 165% to 155%. 

There is only one UI thread that can repaint the display and so I can't see any ill effects of only having one thread to collect updates and call this UI thread, but I would appreciate some further thoughts on this. Is there anything we haven't thought of that this might affect? 

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation/src/main/java/org/csstudio/display/builder/representation/RepresentationUpdateThrottle.java
+++ b/app/display/representation/src/main/java/org/csstudio/display/builder/representation/RepresentationUpdateThrottle.java
@@ -38,6 +38,9 @@ import java.util.logging.Level;
 @SuppressWarnings("nls")
 public class RepresentationUpdateThrottle
 {
+    /** Reference counter to determine when to safely shutdown */
+    private static final AtomicInteger reference_count = new AtomicInteger();
+
     /** Period in seconds for logging update performance */
     private static final int performance_log_period_secs = Preferences.performance_log_period_secs;
 
@@ -83,6 +86,7 @@ public class RepresentationUpdateThrottle
         if(instance == null) {
             instance = new RepresentationUpdateThrottle(gui_executor);
         }
+        reference_count.incrementAndGet();
         return instance;
     }
 
@@ -231,7 +235,13 @@ public class RepresentationUpdateThrottle
     /** Shutdown the throttle thread and wait for it to exit */
     public void shutdown()
     {
+        // Only shutdown if this is the last reference
+        if (reference_count.decrementAndGet() != 0){
+            return;
+        }
+
         run = false;
+        instance = null;
         synchronized (updateable)
         {
             updateable.notifyAll();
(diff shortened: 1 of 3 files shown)


georgweiss (commented):
@rjwills28 , this looks like a nice optimization.

Please check warnings issued by SonarCloud.

rjwills28:
Thanks for taking a look @georgweiss . I will fix the SonarQube warning before this is ready.

I came across an issue when testing which I have tried to fix but I could do with some input on the design pattern here. What I had missed was that there is a `shutdown()` method that is called when a window/tab is closed. Because I hadn't modified this, if you shut a window it killed the update thread and none of the other displays or a new display could get update.

There are 2 ways of fixing this:

1. Keep a reference count on the number of windows/tabs using the `RepresentationUpdateThrottle` and when this drops to 0, then we know it is not being used and we can perform the shutdown. This is what I have tried in this PR but there is the issue that I am setting the `instance = null` during shutdown, which is a static variable called in a non-static method. But we would need to set it to `null` so that when a display is next opened, a new `RepresentationUpdateThrottle` can start. This means that this class is no longer a singleton as such.
2. Remove the `shutdown()` entirely so that once the `RepresentationUpdateThrottle` has been created (i.e. when you open your first BOB screen) it is only destroyed when you shut down Phoebus. The only downside to this is that if you close all your BOB screens and then use some of the other phoebus applications (e.g. databrowser) that do not use the `RepresentationUpdateThrottle`, then you have this resource still running when it doesn't need to be. Although I guess it will not be doing anything as there are no displays to update.

Any input would be much appreciated!

kasemir:
I suggest you keep the reference count and at the same time keep the one `instance`.
So new displays increment the reference count, `shutdown` will decrement, but keep the `instance` set to the one and only. The reference count becomes a debug tool, it's not used to release the instance, but might be useful to debug overall display startup and shutdown. Yes, this means that the one and only RepresentationUpdateThrottle will remain alive after the last display closes, but that's OK since it's not doing anything. It will be ready when the next display is opened.

rjwills28:
Thanks for the advice @kasemir - very helpful. I have implemented that behaviour in my latest commit and also fixed the SonarCloud issues.

kasemir (approved):

PR #3795, opened 2026-04-29
Scan server: Enforce loop limits

### LINKED ISSUE #3794: Scan loop command exceeds end value
Try this loop:
```
<commands>
  <loop>
    <device>loc://x(0)</device>
    <start>1.0</start>
    <end>1.1</end>
    <step>5.0</step>
    <wait>false</wait>
    <tolerance>0.5</tolerance>
    <body>
      <log><devices><device>loc://x(0)</device></devices></log>
    </body>
  </loop>
</commands>
```

Yes, the step size is too large.
One could argue if it should execute steps 1.0 and then 1.1, or just 1.0 and then no second step.
Instead, we get steps 1.0 and 6.0, which wasn't the case in the original implementation.

At first glance, the change could come from one of these commits

commit a9875c13e101771979f84a4e4b54191bd8fb264c
Date:   Tue Mar 4 09:13:11 2025 +0100

    don't use functions with side effects and update getWorkUnits

commit 0e2df3abc6abdd67a91bdaf06b6baa87cc5743fb
Date:   Tue Mar 4 09:05:35 2025 +0100

    add extra step for including bounds

commit 6f102973a0ba573fb5efe6b277024044c2697756
Date:   Mon Mar 3 16:38:29 2025 +0100

    make sure alternating scans work with predictable steps

commit 30964db684d0844dad76ef34f87b7a21d3b55884
Date:   Mon Mar 3 15:48:11 2025 +0100

    make number of steps in loop command predictable

kasemir:
Fixes #3794

The scan server loop command started out as basically
`for (double value=start; value<=end; value += step)`.
While this is quite a normal loop, with equivalent loops available in every programming language, there can be subtle rounding errors. A loop from 0 to 10.0 for example might end at 9.9999 instead of exactly 1.0, with details depending on the step size.
In PR #3305, commits  30964db684d0844dad76ef34f87b7a21d3b55884 and 0e2df3abc6abdd67a91bdaf06b6baa87cc5743fb tried to make it more predictable by avoiding rounding errors. Unfortunately, this disregarded the loop end points. A loop from `1.0` to '2.0' in steps of `2.0` would now yield values `1.0` and `3.0`, with the latter clearly being beyond the loop end of `2.0`.

A basic `for (double value=1.0; value<=2.0; value += 2.0)` loop would only give a value of `1.0` and then end because the next potential value of `1.0+2.0=3.0` is beyond the loop end condition.
I could argue that this original loop command behavior was good and didn't need to be "fixed" in the first place, but in keeping with the spirit of 0e2df3abc6abdd67a91bdaf06b6baa87cc5743fb to include the bounds, this update will will enforce the loop limits, resulting in a loop with values `1.0` and `2.0`.

## Checklist
- Testing:
    - [X] Tests were run

### DIFF
--- a/services/scan-server/.classpath
+++ b/services/scan-server/.classpath
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+    <classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
 	<classpathentry kind="src" output="target/classes" path="src/main/java"/>
 	<classpathentry kind="src" path="src/main/resources"/>
 	<classpathentry kind="src" path="src/test/java"/>
-	<classpathentry combineaccessrules="false" kind="src" path="/phoebus-target"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-framework"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-util"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-pv"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-pv-ca"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/core-pv-pva"/>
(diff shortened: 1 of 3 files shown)


shroffk:
The `computeStop` seems like a reliable way to address the multitude of cases... and thank you for the unit tests.

PR #3796, opened 2026-05-06
Implement a dedicated action to open the data browser

rjwills28:
In CS-Studio we used to have an ActionButton, which would run a script to open the data browser and plot specific PVs. This was done using an `executeEclipseCommand`, e.g.

```
ScriptUtil.executeEclipseCommand(command_id, ['pvnames', 'XXX:YYY'])
```
Obviously this is no longer valid in Phoebus.

I am aware that you can launch the data browser from the right click context menu but we would like to be able to be able to do this from a left click action and plot multiple PVs on the same plot. We could not find another suitable way to reproduce this behaviour (e.g. using a command or script).

Instead we thought it would be cleaner to provide a new action to open the data browser. The inputs include a single or list of PVs to plot as well as the timeframe to show. Potentially we could configure other parts of the data browser but this seemed like a good starting point.

We would appreciate any feedback you may have.

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


georgweiss:
Works as expected. 

Since in the data browser users may specify a formula, would it make sense to support it here as well? If yes, then parsing list of PVs would be affected as space chars may be used in formulas.

kasemir:
In your implementation, the OpenDataBrowserActionHandler in display-runtime uses classes from org.csstudio.trends.databrowser3.*.
In principle, it should be possible to open another application from the display without a direct dependency on that application.
For example, the “open file action” can already open any *.plt file using the generic application service.
“open file action” has no direct dependencies to any databrowser or PV table classes, but it can nevertheless open *.plt or *.pvs files via the application service, locating the appropriate application based on the *.plt or *.pvs file association.

rjwills28:
Thanks for the idea @kasemir  - that is a good point. I did try this by calling:
```
toolkit.openFile("pv://?pvname1&pvname2&app=databrowser");
```
Indeed it does work for opening the data browser showing the list of PVs, however it does not by default add the archive data source, which means that no data actually gets display. I also then can't configure the timeframe.

kasemir:
Well, since we made the DataBrowserWidget one of the common, always included, widgets, there is already a dependency from the display builder to the data browser, so you're not adding any new dependency, only using one that's already in place.

In your OpenDataBrowserActionHandler, you might want to call `instance.getModel().setSaveChanges(false)`. When users invoke your new action, it opens the data browser with the desired PVs and time range. When users then close it, they should not be prompted to "Save" because they can easily get back to that plot by invoking your new action. No need to have them save a config file each time they then close it.

rjwills28:
Ah yes, good point/idea. Thanks @kasemir  - I've made that change.

rjwills28:
@georgweiss  - thank you for testing and for the feedback. I've been looking into how formulas would work. With my current implementation I am building a URI to configure the data browser instance. This is essentially similar to how you might launch the data browser to display PVs from the terminal i.e.
```
phoebus.sh -resource "pv://?SOME.PV&app=databrowser"
```

I believe that through this interface you cannot have a space within a formula. I found that the following was parsed correctly:
```
phoebus.sh -resource "2*pv://?SOME.PV&app=databrowser"
```
However it did not work when I added a space into the formula, e.g. 
```
phoebus.sh -resource "2 * pv://?SOME.PV&app=databrowser"
```
Is there another syntax I should be using to get this to work?
The only other delimiter I think we could use would be an `&` and then the PV list input would look something similar to the syntax you would use to launch from the terminal? The only problem with that is it potentially doesn't look all that 'user' friendly.

So I think there are 2 ways we could go with this:
- Continue with a 'space' separated list and maintain that a formula cannot contain a space (which I've found is also true when launching the data browser from a terminal).
- Use a `&` instead of a space. It is still true that it will break if there is a space in the formula but I could add some logic to the handler to catch this and remove the spaces.

kasemir:
> phoebus.sh -resource "2*pv://?SOME.PV&app=databrowser"

This one works for opening "sim://sine":

    -resource 'pv://?sim://sine&app=probe' 

When I try to open an equation "eq://2+`sim://sine`", you'd think that this does is:

    -resource 'pv://?eq://2+`sim://sine`&app=probe' 

Doesn't work. Well, maybe need to html-escape into this:

    -resource 'pv://?eq://2%2B%60sim:/sine%60&app=probe'

Still doesn't work.

This constant equation "eq://2" succeeds:

    -resource 'pv://?eq://2&app=probe'

So the trouble starts with the "+" and/or the backticks to quote a PV inside the equation.

georgweiss:
@rjwills28, sorry for the delayed response.
Since formulas will introduce issues for various characters, I'd maybe conclude that they are not supported for an action (or ```-resource``` value). If that is the case then it should be documented.

rjwills28:
Hi @georgweiss - yes, I think this is the case as it stands. As per your suggestion I have added a comment to the 'formulas' section of the documentation to make this clear.

PR #3804, opened 2026-05-18
Attempt reconnect on IllegalStateException in WebSocketClientService

georgweiss:
In some cases (e.g. during a debug session), the service may fail to send heartbeat messages in a timely fashion. On the client side the Spring STOMP library will close the web socket connection, in which case a reconnect thread must be started.

Interval for heartbeat messages has also been changed from 30 to 60 s to avoid disconnection when debugging server code.

- Testing:
    - [ X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/core/websocket/client/src/main/java/org/phoebus/core/websocket/client/WebSocketClientService.java
+++ b/core/websocket/client/src/main/java/org/phoebus/core/websocket/client/WebSocketClientService.java
@@ -19,6 +19,7 @@ import org.springframework.web.socket.client.WebSocketClient;
 import org.springframework.web.socket.client.standard.StandardWebSocketClient;
 import org.springframework.web.socket.messaging.WebSocketStompClient;
 
+import javax.websocket.DeploymentException;
 import java.lang.reflect.Type;
 import java.net.URI;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


core/websocket/client/src/main/java/org/phoebus/core/websocket/client/WebSocketClientService.java
```diff
@@ -142,7 +143,7 @@ public void connect() {
         ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
         threadPoolTaskScheduler.initialize();
         stompClient.setTaskScheduler(threadPoolTaskScheduler);
-        stompClient.setDefaultHeartbeat(new long[]{30000, 30000});
+        stompClient.setDefaultHeartbeat(new long[]{60000, 60000});
```
jacomago, line 146: Can we not make this a configuration variable?

georgweiss: Not sure it would make sense as it is not obvious on how to set it.

shroffk: Let us merge this... while I do like the idea of a configuration variable for now we can just hard code it unless we need to change this again

PR #3805, opened 2026-05-19
More efficient check of references in save&restore

georgweiss:
Recently added: annotation of save&restore snapshot and composite snapshots referenced in any composite snapshot. The implementation however is very inefficient as it generates queries to the service whenever JavaFX determines that the ```TreeCell``` must be re-rendered.

This PR abandons the annotation strategy to instead disable the Delete menu item if references are found. This way the call to the service happens only when context menu is launched, and only for the selected node, so user will be able to determine if a snapshot or composite snapshot can be deleted before attempting the operation.

- Testing:
    - [ X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
+++ b/app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
@@ -134,6 +134,8 @@ import java.util.ResourceBundle;
 import java.util.ServiceLoader;
 import java.util.Stack;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
```diff
@@ -1405,10 +1407,41 @@ public void configureContextMenu(ContextMenuEvent e) {
...
+            return false;
+        }
+        AtomicBoolean hasReferences = new AtomicBoolean(false);
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        JobManager.schedule("Find references", monitor -> {
+            MultivaluedMap<String, String> multivaluedMap = new MultivaluedHashMap<>();
+            multivaluedMap.put("referenced", List.of(node.getUniqueId()));
+            try {
+                hasReferences.set(saveAndRestoreService.search(multivaluedMap).getHitCount() > 0);
+            } catch (Exception e) {
```
jacomago, line 1432: Should this method be in the service? 

Like a /snapshot/:id/references endpoint?

jacomago (approved):

app/save-and-restore/app/src/main/java/org/phoebus/applications/saveandrestore/ui/SaveAndRestoreController.java
georgweiss: Maybe, but it would invoke the exact same search.

georgweiss: A new endpoint would also require re-deployment of a new service version.

jacomago: Can be a follow up for an api version 2 :D

PR #3807, opened 2026-05-26
Archiver reader lock refactor

jacomago:
Refactors for ArchiveFetchJob, ApplianceValueIterator, ApplianceArchiveReader.

Refactors:

- Remove IteratorListener (didn't seem necessary to me)
- Add more logging
- Make fetch data in ArchiveFetchJob asynchrounous

Fixes:

- Make lock in ApplianceValueIterator per object rather than global

Feature:

- Add a timeout archive_read_timeout_ms to databrowser_preferences, so down ArchiveSources don't block fetching for others.

Testing:

- Add tests for the fix
- Add more generic tests for the involved classes

## Checklist

- Testing:
    - [ x] The feature has automated tests
    - [ x] Tests were run

- Documentation:
    - [x ] The feature is documented
    - Release notes:

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/archive/ArchiveFetchJob.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/archive/ArchiveFetchJob.java
@@ -9,14 +9,17 @@ package org.csstudio.trends.databrowser3.archive;
 
 import static org.csstudio.trends.databrowser3.Activator.logger;
 
+import java.io.IOException;
 import java.text.MessageFormat;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 
@@ -82,6 +85,9 @@ public class ArchiveFetchJob implements JobRunnable
         /** Archive reader that's currently queried */
         private AtomicReference<ArchiveReader> reader = new AtomicReference<>();
 
+        /** Future for the per-source fetch currently in progress */
+        private final AtomicReference<Future<List<VType>>> currentFetch = new AtomicReference<>();
+
         /** @return Message that somehow indicates progress */
         public String getMessage()
         {
@@ -92,6 +98,10 @@ public class ArchiveFetchJob implements JobRunnable
         public void cancel()
         {
             cancelled = true;
+
+            final Future<List<VType>> pending = currentFetch.getAndSet(null);
+            if (pending != null)
+                pending.cancel(true);
 
             final ArchiveReader the_reader = reader.get();
             if (the_reader != null)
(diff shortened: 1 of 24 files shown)


jacomago:
Not sure whether to split this up into multiple MRs:

- Lock changes
- Timeout addition
- Asynchronous fetching

kasemir:
> Make fetch data in ArchiveFetchJob asynchrounous

The archive fetch job is already a background thread, and `concurrent_requests` controls how many are started.
Why add one more layer of threading?

kasemir:
Is the intent of the added level of threading only for the timeout?

Before: 
One ArchiveFetchJob iterates over data sources, reads from one source, merges with data-fetched-so-far?

With this PR:
One ArchiveFetchJob iterates over data sources, starts fetch in background to read from one source, 'gets' the future to obtain the result, merges with data-fetched-so-far?

app/databrowser/src/main/java/org/csstudio/trends/databrowser3/archive/ArchiveFetchJob.java
```diff
@@ -117,51 +125,53 @@ public void run()
...
-                    catch (UnknownChannelException e)
-                    {
-                        // Do not immediately notify about unknown channels. First search for the data in all archive
-                        // sources and only report this kind of errors at the end
-                        archives_without_channel.add(archive);
-                    }
-                    finally
+                }
+                catch (UnknownChannelException ex)
+                {
```
kasemir, line 138: I'd like to keep this comment:
```
catch (UnknownChannelException e)
{
    // Do not immediately notify about unknown channels. First search for the data in all archive
    // sources and only report this kind of errors at the end
   archives_without_channel.add(archive);
```

jacomago:
Refactored so the timeout handling is in ArchiveFetchJob.run rather than in ArchiveFetchJob.WorkerThread.run.

Also now uses the existing cancel rather than some explicit other way of stopping the job.

jacomago:
@georgweiss Note I didn't add a notification for the timeout in this PR. I'm not sure how to do it, but I thought would be better as a follow up.

kasemir (approved):
For the `ArchiveFetchJob`, this PR basically changes
```
if (monitor.isCanceled())
    worker.cancel();
```
into
```
if (monitor.isCanceled()  ||  we_have_exceeded_archive_read_timeout_ms))
    worker.cancel();
```

The `ArchiveReader` is otherwise created and used the same way,
so this should be perfectly safe for the different archive reader
implementations.

kasemir:
> Also now uses the existing cancel rather than some explicit other way of stopping the job.

Yes, that is good.
The different archive reader implementations all have their own way of cancelling.
RDB implementation tries to cancel the JDBC transaction, archive appliance might try to close the web service connection. The ArchiveFetchJob should simply call 'ArchiveReader.cancel' so the reader implementation can then do what's most appropriate. The ArchiveFetchJob should not interrupt the thread and potentially put the reader into a bad state.

georgweiss (approved):
Agree that we may skip user notification on timeouts for now.

kasemir:
>  I didn't add a notification for the timeout in this PR

That just bit us. The default archive_read_timeout_ms of 30 seconds is too short for queries used to create yearly performance summaries here at SNS. Some queries take ~5 minutes, and the silent failure took about 2 days to figure out what's going on.

The ArchiveFetchJob should log when it's canceled based on the archive_read_timeout_ms

PR #3809, opened 2026-05-27
Support enable date on non-PV node in alarm config

georgweiss:
This  PR implements a feature requested by my users: support enable date on non-PV nodes in alarm config. 

I've updated the dialog to support two different UIs, which are now in fxml. The dialog will for a non-PV node looks like:

Screenshot 2026-05-27 at 13 58 50

This is supposed to work as follows:

- The Enable checkbox will mimic behavior of the enable/disable context menu items for a non-PV node.
- With a non-null enable date, all PV nodes in the subtree will be updated, regardless of current enable state.
- If any PV node in the subtree has a non-null enable date, setting a new enable date is not possible as I think this could cause confusion and unexpected behavior. I could be wrong here, though...

Documentation has been updated. It was lacking information on enable date, by the way...

- Testing:
    - [ X] Tests were run

- Documentation:
    - [ X] The feature is documented
    - Release notes:


app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/config/ItemConfigDialog.java
```diff
@@ -0,0 +1,77 @@
...
+
+            // ── OK-button validation filter ───────────────────────────────────────
+            final Button ok = (Button) getDialogPane().lookupButton(ButtonType.OK);
+            ok.addEventFilter(ActionEvent.ACTION,
+                    event -> ((ConfigDialogController)fxmlLoader.getController()).validateAndStore());
+
+            getDialogPane().setContent(root);
+
+        } catch (Exception ex) {
+            throw new RuntimeException("Failed to load LeafConfigDialog.fxml", ex);
```
kasemir: > "Failed to load LeafConfigDialog.fxml"

Seems like it might have failed to load LeafConfigDialog.fxml or ComponentConfigDialog.fxml based on item type. Include the fxmlLoader location in the message?

georgweiss: That's a good catch!

georgweiss:
Closing as I found a bug.

georgweiss:
Reopening as bug is addressed.

Also: all UI elements of the alarm config dialog now fxml. All string externalized and may be localized.

shroffk (approved):

shroffk:
Tested with multiple combinations of parents/child/grandchild
When a node is empty then this is basically a NOP - not really an issue I think

With all other combinations it works as expected

georgweiss:
We could add a check: if node is empty then suppress rendering of elements related to disable.

shroffk:
This would be for the context menu right.. I don't think we need to change the dialog right now

georgweiss:
Not really, user may still want to configure guidance etc.

PR #3816, opened 2026-06-02
Writing plt file must consider ordering of PVs vs formulas

georgweiss:
Fixes the following issue:

User adds PV to a data browser plot and then a formula referencing those PVs. Additional PVs are added and the formula is updated to reference the added items. The saved plot would write the added PVs last, which breaks parsing of the formula tag.

Note: junit test class generated by Claude. Some tweaks were necessary. Also added a test case not spotted by Claude.

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/persistence/XMLPersistence.java
+++ b/app/databrowser/src/main/java/org/csstudio/trends/databrowser3/persistence/XMLPersistence.java
@@ -7,24 +7,10 @@
  ******************************************************************************/
 package org.csstudio.trends.databrowser3.persistence;
 
-import static org.csstudio.trends.databrowser3.Activator.logger;
-
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.time.Duration;
-import java.time.temporal.TemporalAmount;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Optional;
-import java.util.logging.Level;
-import java.util.stream.Collectors;
-
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.stream.XMLOutputFactory;
-import javax.xml.stream.XMLStreamWriter;
-
+import javafx.scene.paint.Color;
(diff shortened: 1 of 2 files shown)


app/databrowser/src/main/java/org/csstudio/trends/databrowser3/persistence/XMLPersistence.java
```diff
@@ -558,9 +559,21 @@ public static void write(final Model model, final OutputStream out) throws Excep
             writer.writeEndElement();
 
             // PVs (Formulas)
+            // All PVs must appear before formulas
             writer.writeStartElement(TAG_PVLIST);
-            for (ModelItem item : model.getItems())
-                item.write(writer);
+
+            List<PVItem> pvItems =
+                    model.getItems().stream().filter(i -> i instanceof PVItem).map(PVItem.class::cast).toList();
```
jacomago, line 529: You could do this as a forEach rather than a map then you don't need the lower for loop. Same for the formulaItems.

jacomago (approved):

kasemir:
This is indeed a problem; a plt file will fail to load if it lists formulas before all their inputs are defined.
As for the implementation of this fix, it hard to see the actual change because it modified almost every line in the source code. Still, sorting plain channels vs. formulas with the latter placed last on save does fix it.
There is the side effect that it changes the order of items, and since traces are rendered in the order in which they are listed in the model, a trace that used to be "on top" of another one might now appear "behind". I think it's fine to simply accept that.

app/databrowser/src/main/java/org/csstudio/trends/databrowser3/persistence/XMLPersistence.java
georgweiss: But then you would need a try/catch block inside the forEach, which impairs readability.

georgweiss:
Yes, sorry about the auto-formatting @kasemir.

georgweiss:
In a sense the issue reported may be a corner case... Guess one would "normally" add all PVs before creating a formula where all are referenced.

PR #3817, opened 2026-06-02
Update TableWidget editable properties consistently

georgweiss:
NOTE: this PR may be irrelevant in case I have misunderstood the Enabled property on the TableWidget. In any case...

When marking a TableWidget as non-editable, all columns should also be marked as non-editable. Conversely, if all columns are marked non.editable, then the table should also be marked as non-editable.

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


kasemir (approved):

kasemir:
I think you understand correctly, there is some interplay, and the overall table flag has precedence:

Table not editable: The per-column flags don't matter, the table is read-only.
Table editable: The per-column flags decide if a specific column really is editable.
All columns non-editable: Would be nice to mark the table as non-editable as well, since it de facto is.

As far as I know, all the tables we use here are either generally editable or not.
The use case where the table per se is editable but some columns are not hasn't been used here, but maybe somebody has tables configured that way so now it's too late to remove the per-column 'editable' flag.

PR #3818, opened 2026-06-03
Support simplified access to save&restore root node

georgweiss:
The save&restore root node is created with a hard coded unique ID. Clients cannot be expected to know the value of this id, so this PR adds the ability to query for the root node using a GET call to ```<context>/root``` or ```<context>/node/root```. To get child nodes of the root client may use a GET call to ```<context>/node/root/children```.

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


shroffk:
This is a small but very very useful addition

Issue #3821, opened 2026-06-04
Selection in Table widget


georgweiss:
As per user request: investigating if it would be possible to copy value(s) from a Table widget on cell level. The idea would be to add another context menu item to put only the selected value(s) onto the clipboard.

Came across ```SelectionVTable``` that constructs a ```VTable``` based on user selection/mouse click in the table. However, the way it works currently is to create a VTable for the entire row on which user has right-clicked, even if the ```int[] cols``` argument only specifies one element.

Is this an oversight or does it cover a use case I am unfamiliar with?

kasemir:
The table allows complete rows or single cells

<img width="1167" height="367" alt="Image" src="https://github.com/user-attachments/assets/d51c9f8e-e32d-4258-b84d-9fabeba9cf84" />

You may select multiple cells (or rows when in row selection mode)

<img width="1126" height="366" alt="Image" src="https://github.com/user-attachments/assets/495e2f76-5124-438e-a877-b1f07a12e4db" />

In any case, the selection PV is updated with a table that lists all the selected cells (list of rows and columns) plus it contains the complete data of the selected rows.

So all sorts of selection options are available, and the selection PV gets a table with the complete information.
You can then attach a script to the selection PV to use the row & column indices or the values from the selected cells or whatever. What you do with the selection info is highly specific to the case at hand, hence you likely need a script, but all the info is there.

kasemir:
> entire row on which user has right-clicked, even if the int[] cols argument only specifies one element.

Yes

> Is this an oversight 

No. Reason is that the layout of the table in the selection PV shouldn't change. It always has the same number of columns, namely "row", "column", .. and then the columns from the data table.
If we only added the selected cells, then the selection table layout would change as you select multiple cells within a row.
Now the number of rows in the selection table changes as you add/remove cells or rows to the selection, but the table layout stays the same which makes it a little easier to write scripts to handle the table selection

kasemir:
Select 1 cell:

<img width="1115" height="286" alt="Image" src="https://github.com/user-attachments/assets/596a35bf-69c9-473b-a923-1bfe331dc7f3" />

Select 2 cells:

<img width="1127" height="224" alt="Image" src="https://github.com/user-attachments/assets/9a26a7b9-b78d-4264-8cda-e4185dd27a10" />


--> Selection info table has 1 or 2 rows, same number of columns, same column names

georgweiss:
[...]

PR #3826, opened 2026-06-09
Utility API for putting string on clipboard

georgweiss:
Follow up on #3821:

API to put a string on the system clipboard, e.g. from (embedded) script. For the sake of simplicity, no types besides ```java.lang.String``` are supported.

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/ScriptUtil.java
+++ b/app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/script/ScriptUtil.java
@@ -24,6 +24,9 @@ import java.util.logging.Logger;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 
+import javafx.application.Platform;
+import javafx.scene.input.Clipboard;
+import javafx.scene.input.ClipboardContent;
 import org.csstudio.display.builder.model.ChildrenProperty;
 import org.csstudio.display.builder.model.DisplayModel;
 import org.csstudio.display.builder.model.Widget;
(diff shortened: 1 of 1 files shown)


kasemir:
This adds a direct dependency from the ScriptUtil to javafx (Platform, ...).

Originally, we had a separation of display-model, display-runtime, display-representation.
The representation was designed as base classes/interfaces with two specializations: representation-swt and representation-jfx. The runtime or model had no idea about SWT, JavaFX, or whatever graphics lib we'll use in the future. You can still see evidence of that in ScriptUtil.showMessageDialog, which redirects to the abstract ToolkitRepresentation.getToolkit -> showMessageDialog:
```
    public static void showMessageDialog(final Widget widget, final String message)
    {
        try
        {
            ToolkitRepresentation.getToolkit(widget.getDisplayModel()).showMessageDialog(widget, message);
```

The toolkit-specific implementation is then in representation-jfx and representation-swt:

https://github.com/kasemir/org.csstudio.display.builder/blob/2a0ba03576cb76b1b6346ca35791354384fa75e9/org.csstudio.display.builder.representation.javafx/src/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L810
https://github.com/kasemir/org.csstudio.display.builder/blob/2a0ba03576cb76b1b6346ca35791354384fa75e9/org.csstudio.display.builder.representation.swt/src/org/csstudio/display/builder/representation/swt/SWTRepresentation.java#L135

Regrettably, I think we've burned enough bridges by now to sustain this approach. The display editor is tied at the hip to JFX, might as well directly call JFX from anywhere and then completely start over when JFX needs to be replaced.

kasemir (approved):
See comment on `import javafx....` dependencies unlikely worth fixing at this point in time.

georgweiss:
Sure, I could add a method in ```ToolkitRepresentation```...

PR #3827, opened 2026-06-09
Fix ant build after 3796

kasemir:
PR 3796 added a dependency from display-runtime to the databrowser

- Testing:
Before this fix, ant build fails with
```
app-display-runtime:
    [javac] ...OpenDataBrowserActionHandler.java:9: error:
            package org.csstudio.trends.databrowser3 does not exist
    [javac] import org.csstudio.trends.databrowser3.DataBrowserApp;
```
   With this fix, it completes the build process.

### DIFF
--- a/app/display/runtime/build.xml
+++ b/app/display/runtime/build.xml
@@ -12,6 +12,7 @@
         <pathelement path="../representation/${build}/app-display-representation-${version}.jar"/>
         <pathelement path="../representation-javafx/${build}/app-display-representation-javafx-${version}.jar"/>
         <pathelement path="../actions/${build}/app-display-actions-${version}.jar"/>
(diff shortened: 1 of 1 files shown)


georgweiss:
For save&restore actions I have put the implementation in that module. We could consider the same for this action, i.e. put in some data browser module.

kasemir:
display-representation already depended on databrowser because of the DBwidget.
Now display-runtime depends on DB as well.
Not sure how important it is to neaten the dependency graph within the core and app modules. In any case, we can do that at our own pace.
I'm more worried about the ~250 external dependencies in `dependencies/phoebus-target/target/lib`. Something in there could require an update because of some severe bug, and if that is then incompatible with some other external dependency, you're stuck.

PR #3829, opened 2026-06-11
don't re-expand SaR tree nodes if they are already expanded

DenSinH:
## Checklist

Before: opening a save and restore snapshot through a display builder OPEN_SAR_NODE action would (on first click) open the folders up to the snapshot's parent, but NOT the snapshot's parent itself. On second click, it would open the proper node. The issue was that both `locateNode` and the treeview event handler:

```
treeView.getRoot().addEventHandler(TreeItem.<Node>branchExpandedEvent(), e -> expandTreeNode(e.getTreeItem()));
```

called the `expandTreeNode` handler, which caused it to "expand" the parent folder twice, making it closed on the second call.

(Example trying to open `acc/accelerator/oper`)

image

After: this is fixed, now the node is opened right away.

image

- Testing:

The bug is no longer present


georgweiss:
Will check this, there may be use cases that need extra consideration.

DenSinH:
It did seem like a bug, basically the handler was being called twice for all parent nodes: once when the node was requested and the node was actually opened, and then again when the callback for actually opening the (parent) nodes happened, which was of course triggered by the action itself...

DenSinH:
Turns out this doesn't always work, as for some reason some tree nodes are already marked as expanded...

PR #3833, opened 2026-06-15
Vertical progress bar causes Group widget to extend further than it should

rjwills28:
We discovered a fairly obscure bug when placing a vertical progress bar in a Group widget positioned on the far right edge of a screen. We were opening the window from the command line specifying the height and width so that it opening at the exact size of the screen, however we continued to get scroll bars on the window. We also found that if you right clicked to the right of the group, i.e. outside of the group, you would still get the option to view 'Group' information indicating that the Group had been drawn beyond its defined size.

I found that the cause of this was the vertical progress bar, which by default is horizontal and so a translation is applied to the node. However the width of the Progress bar node before the translation (i.e. the length of the bar) seems to remain the same width after the translation. So when it is placed in a Group widget it thinks that the width of the progress bar is much larger than it is and draws the Group out to the length of the progress bar.

E.g. Maybe easier to see in this example - note the right click menu option and the scroll bars

Screenshot from 2026-06-15 16-00-01

I have also attached an example BOB file: 
[progressbar_group.bob.txt](https://github.com/user-attachments/files/28966843/progressbar_group.bob.txt)

If you open this with
```
phoebus.sh  -resource "file:/path/to/bob/progressbar_group.bob?target=window@210x370+50+50"
```
you will see that the right click menu extends to the right beyond the group and that you continued to get scroll bars even when the window is large enough to encompass the visible screen.

Something similar has been seen before as a potential bug in the JFX Progress bar, see https://bugs.openjdk.org/browse/JDK-8295341. 

I have found a workaround for this issue which involves wrapping the progress bar in a JFX Pane() allowing us to rotate the progress bar as before but then set the correct width/height of the pane so that the Group widget doesn't think that this child element is bigger than it is. This fix gives the desired effect in that we can open the window to fit the screen without scroll bars appearing, e.g. 
Screenshot from 2026-06-15 16-03-00

## Checklist

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ProgressBarRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ProgressBarRepresentation.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.csstudio.display.builder.representation.javafx.widgets;
 
+import javafx.scene.layout.Pane;
 import org.csstudio.display.builder.model.DirtyFlag;
 import org.csstudio.display.builder.model.UntypedWidgetPropertyListener;
 import org.csstudio.display.builder.model.WidgetProperty;
@@ -28,7 +29,7 @@ import javafx.scene.transform.Translate;
  *  @author Amanda Carpenter
  */
 @SuppressWarnings("nls")
-public class ProgressBarRepresentation extends RegionBaseRepresentation<ProgressBar, ProgressBarWidget>
+public class ProgressBarRepresentation extends RegionBaseRepresentation<Pane, ProgressBarWidget>
 {
     private final DirtyFlag dirty_look = new DirtyFlag();
     private final DirtyFlag dirty_value = new DirtyFlag();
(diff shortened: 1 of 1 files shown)


kasemir:
While this is wrapping the JFX progress bar to better support rotating it, #3768 proposes a complete replacement. Which way to go here?

rjwills28:
Ah, I hadn't seen that PR. The replacement progress bar with an RTTank bank end much closer resembles the progress bar in CS-Studio and looks to be a nice improvement to the current one, which has few configurable properties. It's also nice that the use of it can be switched on/off making it backwards compatible and minimizing impact on existing screens.

In my opinion I think we should pursue getting https://github.com/ControlSystemStudio/phoebus/pull/3768 approved and merged. What do others think? I am can start testing it with some of our current converted screens.

georgweiss:
We've seen the same (or very similar) issue for the vertical progress bar. Bug reported, see: https://bugs.java.com/bugdatabase/JDK-8295341

In short: after transformation the ProgressBar width and height properties are unchanged (unlike for instance the Label control).

rjwills28:
Maybe in the short term we just get this issue fixed here and then look at the refactoring of the progress bar?

georgweiss:
@kasemir , @rjwills28 , @shroffk : I'd be in favor of merging #3768 as a replacement, but then again I think we need to test that PR a bit more.

rjwills28:
Just to note that I've just seen the exact same issue with a Label widget in a Group. If you apply a rotation of 90 or -90 degrees then again the width and height properties do not get switched correctly and the group gets painted out further than it should. I'll work on a similar fix for that now.

kasemir:
> same issue with a Label widget in a Group

This makes me lean towards merging the rotation fix for the progress bar as well as the label in groups soon.
The progress bar replacement then becomes a separate, follow-up.

georgweiss (approved):
Solves the issue as advertized.

PR #3838, opened 2026-06-17
Change Data Browser area opacity in Properties Misc tab

georgweiss:
Supersedes #3725.

The opacity of the "area" in a plot was read from properties (default 40%). 

This PR adds the option to change the opacity with a slider control in the Misc tab, e.g. for better visibility.

Also: Transformed Misc tab in Properties view to fxml. 

Default opacity:
Screenshot 2026-06-16 at 19 14 18

100% opacity:
Screenshot 2026-06-16 at 19 14 30

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


kasemir (approved):
Not a personal fan of converting to fxml, you'll need to maintain that from now on since I'm not fluent, but that's OK.

jacomago (approved):

georgweiss:
@kasemir, my point is that fxml is easier to maintain than a Java-only UI.

PR #3842, opened 2026-06-18
GitHub ci extend

jacomago:
Add testing to github ci.

- Refactor the ci to be per type of action
- Add reusable sub workflows, unit-test, integration-test, ui-test, etc
- Rename JavaFX tests (they fail unless access to some kind of ui) have them run with ui-tests workflow
- Remove some unnecessary relying on Elasticsearch Config for some integration tests in save and restore
- Add dependabot for the github actions (could be expanded for the pom.xml)
- Remove the dead travis ci (was tied to a very old version of ubuntu)

Open questions:

- Add dependabot for pom.xml?
- Remove 'tests were run from PR template?

## Checklist

- Testing:
    - [ x] The feature has automated tests
    - [x ] Tests were run
    - If not, explain how you tested your changes

- Documentation:
    - [ ] The feature is documented
    - [ ] The documentation is up to date
    - Release notes:
        - [ ] Added an entry if the change is breaking or significant
        - [ ] Added an entry when adding a new feature

PR #3843, opened 2026-06-18
Ci audit

shroffk:
Following the advice from #3842 adding zizmor

## Checklist

- Testing:
    - [x] The feature has automated tests
    - [ ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/.github/workflows/ci-audit.yml
+++ b/.github/workflows/ci-audit.yml
@@ -5,6 +5,10 @@ on:
     branches: ["master"]
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:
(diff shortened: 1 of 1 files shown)


shroffk:
Oh God!!!!
What have we unleashed

jacomago (commented):
I would rather do this in my pr so it can be a subworkflow pulled into the pr workflow. 

Also what about using pre-commit for this? Then it can be checked locally more easily

minijackson:
TBH, I'm not sure I'd recommend using `ci-doctor`, it seems to be a project that has [only 2 stars](https://github.com/depmedicdev-byte/ci-doctor/) and created [2 months ago](https://github.com/depmedicdev-byte/ci-doctor/commits).

It seems to provide interesting lints, but using such a young, unknown project seems a bit dangerous to me.

shroffk:
closing this.
feel free to address it in the appropriate PR.

PR #3851, opened 2026-06-25
Update: replace removed Thread.stop() with stream-close cancellation

gian21391:
Download cancellation in `Update` relied on `Thread.stop()`, which is removed in
recent JDKs and already throws `UnsupportedOperationException` at runtime on the
JDK 21 target, so cancelling was a silent no-op. The download watcher now closes
the source stream on cancel to abort the in-progress `Files.copy`. This is a step
towards improving support for JDK 26, where the old code no longer compiles.

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


kasemir (approved):

georgweiss:
@kasemir, any reason to not merge?

kasemir:
No, go ahead

PR #3839, opened 2026-06-17
Jdk25

shroffk:
## Checklist

- Testing:
    - [ ] The feature has automated tests
    - [x] Tests were run
    - If not, explain how you tested your changes

- Documentation:
    - [ ] The feature is documented
    - [ ] The documentation is up to date
    - Release notes:
        - [ ] Added an entry if the change is breaking or significant
        - [ ] Added an entry when adding a new feature

PR #3852, opened 2026-06-25
Add jdk-jsobject so javafx.web (WebView) builds on JDK 26

gian21391:
`MergedLogEntryDisplayController` uses `netscape.javascript.JSObject` to bridge the
Olog WebView to Java. That class was provided by the JDK's `jdk.jsobject` module,
which is removed in JDK 26, so `javafx.web` no longer compiles or resolves there.

Since JavaFX 24 that module is shipped with JavaFX as the upgradable
`org.openjfx:jdk-jsobject` artifact; this adds it to the classpath. It is versioned
independently of `openjfx.version` because the standalone artifact only exists from
25 onwards. On JDK 25 and earlier the platform module still takes precedence, so the
dependency is inert; on JDK 26 it supplies `netscape.javascript`.

Together with #3851 (which replaces the removed `Thread.stop()`), this clears the
second and final compile blocker, so the project builds and runs on JDK 26.

## Checklist
- Testing:
    - [X] Tests were run
        - Built the full reactor and ran the test suite on both the JDK 21 target
          and JDK 26 (2332 tests, 0 failures on each). On JDK 21 the JDK's own
          `jdk.jsobject` still takes precedence (behavior unchanged); on JDK 26
          `netscape.javascript` resolves from the bundled artifact and the embedded
          `WebView` renders a page.
- Documentation:
    - Release notes:

### DIFF
--- a/core/ui/pom.xml
+++ b/core/ui/pom.xml
@@ -46,6 +46,13 @@
       <artifactId>javafx-web</artifactId>
       <version>${openjfx.version}</version>
     </dependency>
+    <!-- netscape.javascript for javafx.web; previously provided by the JDK's
+         jdk.jsobject module, which was removed in JDK 26. -->
+    <dependency>
+      <groupId>org.openjfx</groupId>
+      <artifactId>jdk-jsobject</artifactId>
+      <version>${jsobject.version}</version>
+    </dependency>
     <dependency>
       <groupId>org.openjfx</groupId>
       <artifactId>javafx-swing</artifactId>
--- a/core/ui/pom.xml
+++ b/core/ui/pom.xml
@@ -46,13 +46,6 @@
       <artifactId>javafx-web</artifactId>
       <version>${openjfx.version}</version>
     </dependency>
-    <!-- netscape.javascript for javafx.web; previously provided by the JDK's
-         jdk.jsobject module, which was removed in JDK 26. -->
-    <dependency>
-      <groupId>org.openjfx</groupId>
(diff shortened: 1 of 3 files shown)


gian21391:
Note: this overlaps with #3839 (JDK 25 bump) — that PR moves `openjfx.version` to 25, which already pulls in `jdk-jsobject` transitively. This PR adds it explicitly so it's guaranteed on the classpath for JDK 26.

georgweiss:
Adding the ``org.openjfx:jdk-jsobject`` will on pre-JDK26 result in a build where (for instance) ``JSObject`` will be available from two jar files. Moreover, the one added through ``org.openjfx:jdk-jsobject`` has a byte code major version of 68, which is Java24, so a pre-Java24 class loader will fail to load it.

I will make an effort to completely remove the dependency to ``JSObject``. In the meanwhile, can we instead add this dependency in a Maven profile?

georgweiss (commented):
Addition of ```jdk-jsobject``` should be put in a Maven profile.

gian21391:
Thanks @georgweiss! That is indeed a cleaner solution. 
For the record, the previous solution worked also on JDK < 26 versions by virtue of module shadowing. The platform jdk.jsobject module owns netscape.javascript, so even with the jar on the classpath the class is always resolved from the module and the Java-24 class in the jar is never loaded. That's fragile and relies on a non-obvious classpath/module interaction, so the JDK-26-activated profile is clearly the better approach.

georgweiss (approved):

PR #3853, opened 2026-06-26
Web Browser widget: add "Resize with Window" property

gian21391:
The Web Browser widget keeps its fixed design-time size at runtime, like every other widget. For displays built around a single browser filling the screen, that's awkward — the embedded page doesn't follow the window when it is resized.

This PR adds an opt-in `resize_with_window` property (Behavior category, default off). When enabled, the browser grows and shrinks to fill the runtime window/tab. The default is off, so existing displays are unaffected, and the property is left out of the saved file unless it is set (older versions ignore the unknown tag).

It is intended for top-level, browser-dominant displays: the widget should be a direct child of the display root. When the browser is nested in a Group, Tabs, or embedded display the property is a no-op and logs a warning.

## Checklist

- Testing:
    - [X] Tests were run
    - Added unit tests for the model property (default value, XML round-trip) and for the fit calculation; ran `mvn clean install` on JDK 21.
- Documentation:
    - Release notes:
        - [X] Added an entry when adding a new feature

### DIFF
--- a/app/display/editor/doc/widgets_properties.rst
+++ b/app/display/editor/doc/widgets_properties.rst
@@ -37,3 +37,26 @@ path may not be portable between hosts, in particular not between Windows and Li
 
 File paths may be specified using both forward slash (/) or backslash (\\) as path separator as both will work interchangeably
 between Windows and Linux/Mac.
+
+Web Browser Widget
+==================
+
+The Web Browser widget embeds a web view in a display. In addition to the common properties it provides
+the following widget specific properties:
+
+ - "url" the address that is loaded when the display starts.
+ - "show_toolbar" whether to show a navigation toolbar (back, forward, reload and an address bar) above the
+   page. When turned off, only the page itself is shown.
+ - "resize_with_window", described below.
+
+Resize with Window
+------------------
+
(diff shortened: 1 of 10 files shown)


kasemir:
Well, the web browser widget has acknowledged limitations. Web pages with javascript may not work as expected and CPU load tends to be higher than in a "real" web browser. If your goal is to have a browser that fills the window, like a normal web browser, why don't you simply use an action button that opens the URL in a normal web browser? That way your users get the expected normal web browser behavior.

gian21391:
The idea is to have tabs containing documentation or specific tools that are HTML based, rather than a generic browser. We have a specific use case for such a feature. Having HTML pages in an external browser is an option but not ideal at all.

kasemir:
Sure, I can see `resize_with_window` being nice for images, plots, tables that are at the 'bottom' of a display so that they then extend downwards to fill the window. Also opens the door to widgets at the right growing to fill the available window widths.

gian21391:
You are right. I did not see the value in other widgets but it could be a nice to have everywhere. Should I edit this PR to generalize this? Otherwise this could be implemented on each widget separately with distinct PRs.

PR #3854, opened 2026-06-26
Make Slider UI elements larger when widget size increases

georgweiss:
The Scaled Slider widget relies on the native JavaFX Slider. On high resolution monitors the default size of the track and knob may become a bit too small. While increasing the font is supported in the widget, other UI elements stay at their defaults even if widget is made taller/wider.

This PR adds a bit of logic to increase the width of the track, the size of the knob and the length of the major tick marks when user increases the height (horizontal) or width (vertical) of the widget. Unfortunately I was not able to figure out how to make both major and minor tick marks longer.

A minimum and maximum is applied to keep this reasonably useful:

Screenshot 2026-06-26 at 15 41 47
Screenshot 2026-06-26 at 15 42 02
Screenshot 2026-06-26 at 15 42 11

Decided that new widget properties for track, knob and major tick mark dimensions would be an overkill.

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
@@ -10,7 +10,9 @@ package org.csstudio.display.builder.representation.javafx.widgets;
 import javafx.beans.value.ObservableValue;
 import javafx.geometry.Insets;
 import javafx.geometry.Orientation;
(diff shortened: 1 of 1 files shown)


kasemir (approved):
Sad to see how carefully aligned code, typed by a human with pride in his/her work like this:

Image

... is replaced by mechanical formatting like this:

Image

After all, this PR is all about cosmetics, the size of a slider button, so style does seem to matter at some level.

Still, I like the end result, so whatever

georgweiss:
@kasemir, sorry for hurting your pride.  I will revert.

PR #3855, opened 2026-06-28
Make Slider UI elements larger when widget size increases

georgweiss:
Supersedes #3854 to deal with code formatting and readability.

- Testing:
    - [ X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
@@ -10,7 +10,9 @@ package org.csstudio.display.builder.representation.javafx.widgets;
 import javafx.beans.value.ObservableValue;
 import javafx.geometry.Insets;
 import javafx.geometry.Orientation;
+import javafx.scene.Node;
 import javafx.scene.control.Slider;
+import javafx.scene.control.skin.SliderSkin;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
@@ -35,6 +37,7 @@ import org.epics.vtype.VType;
 
 import java.text.DecimalFormat;
 import java.time.Instant;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 
 import static org.csstudio.display.builder.representation.ToolkitRepresentation.logger;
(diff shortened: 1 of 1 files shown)


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
```diff
@@ -566,4 +574,32 @@ private void openConfigurationPanel()
...
+     * a workaround for visibility issues when the JavaFX native {@link Slider} is rendered on a
+     * high-resolution monitor.
+     * <p>
+     *     Applied sizes are based on the height (horizontal) or width (vertical), and
+     *     clamped between the default size and an upper ad-hoc determined limit.
+     * </p>
+     */
+    private void adjustSizes(){
+        SliderSkin skin = (SliderSkin) slider.getSkin();
+        if (skin != null) {
```
jacomago: If you return for skin == null you avoid the heavy nesting.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
```diff
@@ -566,4 +574,32 @@ private void openConfigurationPanel()
...
+            double size = isHorizontal.get() ? jfx_node.getHeight() : jfx_node.getWidth();
+            for (Node node : skin.getChildren()) {
+                if (node.getStyleClass().contains("thumb")) {
+                    // 7.6 seems to be default padding
+                    node.setStyle("-fx-padding: " + Math.clamp(size * 0.3, 7.6, 20));
```
jacomago: Can you define all these numbers as constants with descriptive names?

Then you don't really need the comments

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
georgweiss: "Heavy" is maybe not how I would describe it, but I'll update.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
georgweiss: Good point

jacomago (approved):

PR #3858, opened 2026-06-29
Target platform: Fix testfx version

kasemir:
#3842 changed testfx version in some modules but didn't update target platform, breaking the ant build


kasemir:
@shroffk , this needs a real maven expert!

shroffk (approved):

Issue #3543, opened 2025-09-11
Option to open windows in 'standalone' mode


rjwills28:
Currently Phoebus supports opening new screens (BOB files) in new windows. These new windows are tied to the main window in that they adopt the same settings, i.e. display tabs, display toolbar. 

At Diamond, a lot of our CS-Studio screens make use of the CS-Studio standalone mode where screens display in a separate windows with no other features other than a title bar with the minimize, maximize and close button, e.g.:
<img width="357" height="433" alt="Image" src="https://github.com/user-attachments/assets/341da264-1484-44cb-baf2-a65ddf78d775" />

It would be very useful if we could replicate this same sort of look in Phoebus.

I can see from the code that the 'standalone' target is still available on an ActionButton and that there are some deprecated method stubs that simply open the screen as a new window. For example see [OpenDisplayAction](https://github.com/ControlSystemStudio/phoebus/blob/b32391bdbf75a80507e8abb94dc1196788dc70ed/app/display/actions/src/main/java/org/csstudio/display/actions/OpenDisplayAction.java#L69) and [ToolkitRepresentaton](https://github.com/ControlSystemStudio/phoebus/blob/b32391bdbf75a80507e8abb94dc1196788dc70ed/app/display/representation/src/main/java/org/csstudio/display/builder/representation/ToolkitRepresentation.java#L255). Presumably this is just there at the moment to support OPI->BOB conversions that reference 'standalone'.

I have had a quick play myself at leveraging these existing methods and implementing a way of opening standalone windows, which basically involves ensuring the toolbar and tabs do not show and that they also are not displayed when enabled from the main window options.

To minimize impact on other users we would suggest introducing some Phoebus preferences that enable the use of this standalone window mode and otherwise ignores it and maintains the current behaviour of just opening as a new window with a toolbar and tabs configured as standard. 

We wanted to gauge whether the community would be open to having this additional feature if we were to spend some time properly developing the standalone window implementation?

epicDirk:
We have the same problem at ESS Lund.

Our operators use "New Window" (popup) to show detail views for equipment from the overview displays. The problem is that from inside a popup there is no possibility to navigate back to the main Phoebus window:

Replace changes the content inside the popup (and does not resize the popup automatically)
New Tab opens a new tab inside the popup
New Window opens another popup window on top
During shifts this creates many open windows and the operators lose the overview which window is which. Especially in stressful situations this is a real problem.

[...]

Issue #3436, opened 2025-06-18
Screen window sizing is not respected in JavaFX 21


rjwills28:
I have noticed a problem when using the command line to open a screen in a new window which specifies the **size** of that screen.

The command I use is:
``` 
phoebus.sh -resource  file:/home/xxx/bob/standalone1.bob?target=window@600x600+800+150
```

I would expect the size to be 600x600 but instead it always opens much smaller. No matter what value I specify for the screen size, this is what I get:

![Image](https://github.com/user-attachments/assets/1d10d8fc-0b07-4bfd-99cf-e4f8ddf75f27)

Interestingly I do _not_ see this issue in Phoebus v4.7.3. I have pinned it down to an issue in JavaFX that was not there in JavaFX 19 but is in JavaFX 21. It also depends on the gnome version being used - I have tested on Ubuntu24.04 and Fedora41 - both using later versions of gnome. Yet, the problem is not there in any version of Phoebus on RedHat 7 (using a much older version of gnome).

Here you can find the **bug report**: https://bugs.openjdk.org/browse/JDK-8316423 and the associated fix in JavaFX: https://github.com/openjdk/jfx/pull/1249. You can see it is consistent with my finding that it affects later versions of gnome as the underlying change came from an update to mutter.

**FIX OPTIONS**:
- The above fix is available in JavaFX 22 but not 21, therefore one solution to fix this in Phoebus would be to update to JavaFX22 which still requires JDK17 (JavaFX 23 would required JDK21).
- Alternatively, if this isn't an option, a workaround is to set the stage height and width _after_ the stage has been created, i.e. after `stage.show()` at line 173: https://github.com/ControlSystemStudio/phoebus/blob/0332a7fc0f1a6bd98ea16f46d94f94a36a4fd73e/core/ui/src/main/java/org/phoebus/ui/docking/DockStage.java#L173 
add:
   ```
   stage.setWidth(geometry.width);
   stage.setHeight(geometry.height);
   ```

What do people think is the best way forward here?

rjwills28:
And you don't think it will be a problem that then to use the latest version of Phoebus we will require users to update to JDK25? I only ask as I'm not sure how quick people are to adopt the latest version.

georgweiss:
Thing is there are other JavaFX bugs that require (if I recall correctly) at least JDK 24, which is not LTS.

PR #3859, opened 2026-06-29
Implement the 'standalone' window mode

rjwills28:
The main purpose of this PR is to add back in the standalone window implementation that existed in CS-Studio. See Issue https://github.com/ControlSystemStudio/phoebus/issues/3543 for a full description of our motivations.

A standalone window is essentially a window with no toolbars or tabs and so _just_ displays the BOB screen (again, see above issue for an example screen).

A summary of what has been done in this PR:
- Configure a standalone window:
  - No toolbars or tabs:
    - Remove right click menu option to show toolbars
  - All applications launched from a standalone window will open up in the 'main' Phoebus window (including the editor). Focus will also shift to the main window.
- Allow displays to be opened in a standalone window from an 'Open Display' action:
  - Additional option added.
  - Standalone windows are sized as close to their display dimensions (i.e. width x height from BOB file) as possible and need less padding than the 'original windows' as we do not need to allow room for toolbars, tabs etc.
- If a standalone window contains an action to open a display in the 'replace' mode, then the standalone window is resized to the new display dimensions.
- Configure the memento to save the standalone window mode
- Launch a window in standalone mode from the terminal:
  - These windows will open with the correct width & height specified from the BOB file (if no dimensions are given from the command line).

All of these changes only apply to standalone windows meaning that the **original 'New Window' implementation has not been touched and still functions as it did before**. These changes will be transparent to users unless they change their actions to open in the standalone mode or use the command line argument to launch them.

We have been actively testing this implementation at Diamond and so have ironed out most of the issues with behaviour. 

Note: some of the resizing of windows can be done in a cleaner way once Phoebus has moved to JFX25. This is because there is a bug in older versions of JFX that mean that any sizing of the stage before it is 'shown' (i.e. `stage.show()`) are not applied (see https://github.com/ControlSystemStudio/phoebus/issues/3436). This means that currently we have to do all of our resizing after `stage.show()`, which can cause some flickering of screens when they first launch as they pop up at one size and then very quickly get resized. I have a fix for this but propose postpone implementing that until the basic standalone window implementation has been agreed and Phoebus moves to JFX25.

I appreciate that this is a reasonably big PR but most of the commits should be fairly self contained with a descriptive message. Please let me know if you have any questions/feedback.

I anticipate a few Sonarcloud issues so I will address those asap. 

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


app/display/runtime/src/main/java/org/csstudio/display/builder/runtime/app/ContextMenuSupport.java
```diff
@@ -91,7 +93,25 @@ public void handleContextMenu(final Widget widget, final int screen_x, final int
                     DisplayRuntimeInstance displayRuntimeInstance = DisplayRuntimeInstance.ofDisplayModel(displayModel);
                     DockItem dockItem = displayRuntimeInstance.getDockItem();
                     DockPane dockPane = dockItem.getDockPane();
-                    setFocus = () -> DockPane.setActiveDockPane(dockPane);
+                    if (dockPane.isStandaloneWindow()) {
+                        // If in a standalone window, set the active dock pane
+                        // to be the 'main' Phoebus pain instead of the current dock pane
```
shroffk: Freudian slip :)

kasemir: It's an abbreviation: Main pane -> pain

rjwills28: Ha, yes my bad - I'll fix that ;)

shroffk:
There is work that started on JDK25.
Do you want to wait for that to be completed, switch to javafx25 and then complete this PR or do you feel like this should be merged with a separate effort to cleanup code post javafx25 upgrade.

rjwills28:
@shroffk thanks for the feedback. Personally I think it would be good to get this merged now as it already contains a fair few changes and then I'll create another PR to clean up the window sizing once we've moved to JFX25. If that's OK with you? I just feel like that might be a bit cleaner?

Otherwise, I've fixed the spelling mistake in the comment (thanks for catching).

In terms of SonarCloud, I've fixed all the issues that I think I can, including refactoring some fairly complex nested if-else statements. 

The big one that remains is '**Duplicated Lines of code**'. I'm not sure how this can be avoided as the two affected classes both need to override the base class's `isOpenAction()` method to `return true` instead of the default value of false. The only way to remove this duplication would be to create a new class that has the `isOpenAction(){return true}` and have these two classes implement that, but that seems like overkill in this case. Correct me if I'm wrong, otherwise I am going to have to ignore this.
Finally I have followed the naming convention of variables that is used by each class, which in some cases is not the Sonar cloud recommended camel case. I think trying to be consistent within a class is more important than sticking to the camel case standard and so I have left these warnings too. Just let me know if you disagree.

shroffk:
@rjwills28 

do not fixate on the duplicate lines of code.

following the existing variable naming convention is the correct thing to do. We should add an exception for sonarcube.

I understand that these changes are isolated to the standalone window and should not affect existing setup... are there any particular changes that we should test rigorously?

rjwills28:
As you said, most of the behavior changes are contained to standalone windows. Below are some areas to test.

In terms of checking **'normal' windows** still behave the same I would test out:
- Launching with the `target=window` from the terminal, e.g. `phoebus.sh -resource file:/xxx/xxx/xxx.bob?target=window`
Expect a window to show up with tabs, toolbar etc. You can still hide/show toolbar from context menu

- Check creating an ActionButton with the action to open a display in a window.
Expect display to open in a window.

- Right clicking and selecting any context menu application (e.g. 'Open in Editor', 'Databrowser') 
Expect that the application opens in a tab in the _same_ active window


To test the **standalone window** implementation:
- Test launching from terminal with `target=standalone`, e.g. `phoebus.sh -resource file:/xxx/xxx/xxx.bob?target=standalone`
Expect the window to be the same size as the BOB display.

- Check creating an ActionButton with the action to open a display in a standalone.
Expect it to open a window that fits the size of the BOB display.

- Right clicking and selecting any context menu application (e.g. 'Open in Editor', 'Databrowser') 
Expect it should open that application in a tab in the _main_ Phoebus window and not the current one.

georgweiss (approved):
Works as advertised.

I would however suggest one change: the options "New Window" and "New Standalone Window" may be a bit confusing. After all, they both open a new window, and the definition of "standalone" is open for interpretation. Should we consider "New Undecorated Window" or similar?

rjwills28:
Thanks for taking the time to review @georgweiss - much appreciated.

Yes I see your point - I am happy to change the option name. Besides your suggestion of "New Undecorated Window" (which I think would work), the only other idea I can think of would be "New Pop-Up Window"?

georgweiss:
@rjwills28, I'd argue that both options could be labelled "pop-up windows".

How about:
- New Window
- New Window (undecorated)

That would make it clear that these options are essentially the same, save for the toolbar and other UI elements.

rjwills28:
Yes, I like it. I've made that change in the latest commit.

georgweiss:
@shroffk, I think we can merge this

PR #3861, opened 2026-06-29
Fix issues with unc paths

### LINKED ISSUE #3837: Opening a file in edit mode on windows that is on a mounted network drive fails
The file can be opened without an issue in display mode. The file browser also has no issues

Image

The issue is that the filepath seems to remove the mount's hostname, and instead appends the drive letter that the Phoebus installation lives on.

```
java.io.FileNotFoundException: D:\AlmaLinux-9\home\jwlodek\Workspace\ADKeyenceSR\KeyenceSRApp\op\bob\ADKeyenceSR_Top.bob (The system cannot find the path specified)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:185)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:139)
	at org.csstudio.display.builder.editor.EditorGUI.lambda$loadModel$9(EditorGUI.java:462)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

This is easily reproducible with WSL2. From within wsl, clone/copy a bobfile. Then use `explorer.exe .` to open a windows file explorer in that folder (which will automatically mount the folder as a network drive). Then, copy the file path, and paste it into the filebrowser in phoebus. Right click on the display -> Open With -> Display, works fine, do the same thing for Edit and it will fail with the above error.

Image

jwlodek:
Update path resolution logic to handle UNC style paths (i.e. \\server\share etc.). This seems to work on my end loading files from WSL's network drive on windows. I have not extensively tested in other configurations. Should hopefully fix #3837 

## Checklist

- Testing:
    - [X] Tests were run
    - Tested manually opening files in display mode from a WSL share.

- Documentation:
  N/A - this is a change that should be transparent to users.

### DIFF
--- a/app/3d-viewer/src/main/java/org/phoebus/applications/viewer3d/Viewer3dPane.java
+++ b/app/3d-viewer/src/main/java/org/phoebus/applications/viewer3d/Viewer3dPane.java
@@ -16,6 +16,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.phoebus.framework.jobs.JobManager;
+import org.phoebus.framework.util.ResourceParser;
 import org.phoebus.ui.javafx.ImageCache;
 
 import javafx.application.Platform;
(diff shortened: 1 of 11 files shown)


georgweiss:
I think this calls for extensive testing in "other configurations".

PR #3868, opened 2026-07-06
Rotated Label causes Group widget to extend further than it should

rjwills28:
This is the same issue as seen with a vertical progress bar in PR: https://github.com/ControlSystemStudio/phoebus/pull/3833. 

Essentially when the JFX label widget gets rotated, the width and height properties do not get switched and so the group containing the label gets drawn out further than it should be. 
Screenshot from 2026-07-06 14-24-32

Another impact of this is that it can cause widgets next to the group to become un-clickable due to the group being painted wider than it should be and covering neighboring widgets without users noticing. I am attaching an example BOB screen to reproduce this issue:
[label_group.bob.txt](https://github.com/user-attachments/files/29706294/label_group.bob.txt)

The fix for this is the same as that for the progress bar, e.g. wrap the JFX label widget in a JFX pane 

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.csstudio.display.builder.representation.javafx.widgets;
 
+import javafx.scene.layout.Pane;
 import org.csstudio.display.builder.model.DirtyFlag;
 import org.csstudio.display.builder.model.UntypedWidgetPropertyListener;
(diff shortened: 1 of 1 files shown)


kasemir (approved):
This appears to make sense. There's a caveat that displays tend to contain MANY labels. Pretty much every text entry, text update, ... has a label next to it. Changing from a low-level `Label` to a `Pane` might significantly raise the JFX scene graph complexity.

--> We should watch for performance changes and potentially revert this.

If this needs to be reverted, it might require an approach similar to the code related to `was_ever_transformed`. Right now it only touches the `Transforms` if we actually need a rotation. Similarly, it could only add the `Pane` wrapper when there are `Transforms`.
But I'd leave that to later and first see how this simpler overall `Pane` addition works out.

PR #3870, opened 2026-07-07
Increase size of trace anchor circle

milenaschedle:
We changed the hard coded radius value of the annotation point in the data browser from 4 to 6, as it might be too small to see and drag on modern high resolution monitors.
See screenshots for comparison.
If needed, the radius can be made customizable.

radius_4

---> radius = 4

vs

radius_6

---> radius = 6

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:


kasemir (approved):
6 is so much better than 4 since it's divisible by 2 as well as 3!

app/databrowser/doc/index.rst
```diff
@@ -154,6 +154,25 @@ behavior, you already activated it, so you need to use "Save As" from the File m
 the ``*.plt`` file with the un-checked "Save Changes" option.
 
 
+Context Menu
+^^^^^^^^^^^^
+
+One or more traces in the trace tab of the property panel can be selected using *Shift* + click or *Control* + click.
```
kasemir, line 160: On a Mac, the *Command* propeller button is used instead of *Control*, but we can assume that Macizens will know that by heart.

georgweiss:
You're right, let's clarify for those Mac users out there.

PR #3874, opened 2026-07-08
added property name to display name

milenaschedle:
The Property Name of each Tree Item is now displayed as a bold prefix in the tree.

Screenshot 2026-07-08 151803

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/channel/channelfinder/src/main/java/org/phoebus/channelfinder/ChannelUtil.java
+++ b/app/channel/channelfinder/src/main/java/org/phoebus/channelfinder/ChannelUtil.java
@@ -98,8 +98,6 @@ public class ChannelUtil {
         return propertyValues;
     }
 
-    //public static void getPropertyValueByDisplayName
-
     /**
      * Returns all the channel Names in the given Collection of channels
      * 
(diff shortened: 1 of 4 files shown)


app/channel/channelfinder/src/main/java/org/phoebus/channelfinder/ChannelUtil.java
```diff
@@ -98,6 +98,8 @@ public static Collection<String> getPropValues(Collection<Channel> channels, Str
         return propertyValues;
     }
 
+    //public static void getPropertyValueByDisplayName
```
jacomago: Remove commented out code

app/channel/views/pom.xml
```diff
@@ -38,5 +38,11 @@
       <artifactId>core-types</artifactId>
       <version>6.0.0-SNAPSHOT</version>
     </dependency>
+      <dependency>
+          <groupId>org.eclipse.jetty</groupId>
```
jacomago: Why a new dependency?

georgweiss: Hmmm, cannot say where this comes from... will check.

jacomago (approved):

PR #3876, opened 2026-07-10
Use authenticated user name when creating log entries from save&restore

milenaschedle:
moved SaveAndRestoreAuthenticationScope to model module and fixed user name for restored log entry

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
+++ b/app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
@@ -388,10 +388,9 @@ public class LogEntryEditorController {
 
         textArea.textProperty().bindBidirectional(descriptionProperty);
         switch(editMode){
-            case NEW_LOG_ENTRY -> descriptionProperty.set("");
+            case NEW_LOG_ENTRY -> descriptionProperty.set(logEntry.getDescription()!= null ? logEntry.getDescription(): "");
             case UPDATE_LOG_ENTRY, NEW_LOG_ENTRY_FROM_SELECTION -> descriptionProperty.set(logEntry.getSource());
         }
-//        descriptionProperty.set(logEntry.getSource());
         descriptionProperty.addListener((observable, oldValue, newValue) -> isDirty = true);
 
         Image tagIcon = ImageCache.getImage(LogEntryEditorController.class, "/icons/add_tag.png");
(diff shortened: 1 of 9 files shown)


app/save-and-restore/logging/src/main/java/org/phoebus/applications/saveandrestore/logging/SaveAndRestoreEventLogger.java
```diff
@@ -84,12 +88,25 @@ public void snapshotRestored(Node node, List<RestoreResult> failedPVs, Consumer<
...
+        try {
+            SecureStore store = new SecureStore();
+            ScopedAuthenticationToken scopedAuthenticationToken = store.getScopedAuthenticationToken(new SaveAndRestoreAuthenticationScope());
+            if (scopedAuthenticationToken != null) {
+                return scopedAuthenticationToken.getUsername();
+            }
+        } catch (Exception e){
+            logger.log(Level.WARNING, "Unable to get Authentication Token.", e);
+        }
+        return "<Not Available>";
```
abrahamwolk, line 110: If the retrieval of the authentication token is not successful, shouldn't that be an error, and the log message should not be written?

@georgweiss @kasemir I am not familiar with the way authentication is handled; could you provide your input?

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
```diff
@@ -388,10 +388,9 @@ public void initialize() {
 
         textArea.textProperty().bindBidirectional(descriptionProperty);
         switch(editMode){
-            case NEW_LOG_ENTRY -> descriptionProperty.set("");
+            case NEW_LOG_ENTRY -> descriptionProperty.set(logEntry.getDescription()!= null ? logEntry.getDescription(): "");
             case UPDATE_LOG_ENTRY, NEW_LOG_ENTRY_FROM_SELECTION -> descriptionProperty.set(logEntry.getSource());
```
abrahamwolk, line 392: Should this line also have an additional check for `null` (analogous to the one added on the preceding line) of `logEntry.getSource()`?

app/save-and-restore/logging/src/main/java/org/phoebus/applications/saveandrestore/logging/SaveAndRestoreEventLogger.java
georgweiss: In this case an error dialog would be a bit confusing in the context of the workflow as it is not obvious _why_ it is shown. Moreover, the exception would be a corner case as logging of an action is relevant only when a user has been signed-in.

app/logbook/olog/ui/src/main/java/org/phoebus/logbook/olog/ui/write/LogEntryEditorController.java
georgweiss: Actually there is no need for a null check.

jacomago (approved):

PR #3878, opened 2026-07-10
PVA: harden handling of protocol sizes

kasemir:
Protocol handling was prone to memory exhaustion because of this pattern:
```
int count = buffer.getInteger();
Whatever[] data = new Whatever[count];
.. then read data from buffer,
.. which may actually be too small for 'count' elements
```

A misconfigured package can contain a huge `count`. This update checks the `count` against the remaining package size, refusing packets that are obviously too small or rather the `count` is obviously too large for the packet size.

Along the same lines, one could introduce an upper packet size limit similar to `EPICS_CA_MAX_ARRAY_BYTES` in channel access, but at this time, confirmed in a 2026-07-10 EPICS core telecon, PVA does not impose any such size limits, using all available memory.

### DIFF
--- a/core/pva/src/main/java/org/epics/pva/common/SearchRequest.java
+++ b/core/pva/src/main/java/org/epics/pva/common/SearchRequest.java
@@ -173,15 +173,23 @@ public class SearchRequest
         boolean tcp = search.tls = false;
         int count = Byte.toUnsignedInt(buffer.get());
         String unknown_protocol = "<none>";
-        for (int i=0; i<count; ++i)
+        try
         {
-            final String protocol = PVAString.decodeString(buffer);
-            if ("tls".equals(protocol))
-                search.tls = true;
-            else if ("tcp".equals(protocol))
-                tcp = true;
-            else
-                unknown_protocol = protocol;
+            for (int i=0; i<count; ++i)
+            {
+                final String protocol = PVAString.decodeString(buffer);
+                if ("tls".equals(protocol))
+                    search.tls = true;
+                else if ("tcp".equals(protocol))
+                    tcp = true;
+                else
+                    unknown_protocol = protocol;
+            }
+        }
+        catch (Exception ex)
+        {
+            logger.log(Level.WARNING, "PVA Client " + from + " sent search #" + search.seq + " with invalid protocol", ex);
+            return null;
         }
 
         // Loop over searched channels
@@ -200,18 +208,26 @@ public class SearchRequest
                 return null;
             }
             search.channels = new ArrayList<>(count);
-            for (int i=0; i<count; ++i)
+            try
             {
-                final int cid = buffer.getInt();
-                final String name = PVAString.decodeString(buffer);
-                logger.log(Level.FINER, () -> "PVA Client " + from + " sent search #" + search.seq + " for " + name + " [cid " + cid + "]"
-                                            + ", reply addr " + orig_response_addr
-                                            + (orig_response_addr.equals(search.client) ? "" : ", using " + search.client)
-                                            + (search.tls               ? " (TLS)" : "")
-                                            + (search.unicast           ? " (unicast)" : "")
-                                            + (search.reply_required    ? " (reply required)" : "")
-                                            + (search.reply_to_src_port ? (origin == null ?  " (reply to source port)"  : " (reply to source port ignored because of origin tag)") : ""));
-                search.channels.add(new Channel(cid, name));
+                for (int i=0; i<count; ++i)
+                {
+                    final int cid = buffer.getInt();
+                    final String name = PVAString.decodeString(buffer);
+                    logger.log(Level.FINER, () -> "PVA Client " + from + " sent search #" + search.seq + " for " + name + " [cid " + cid + "]"
+                            + ", reply addr " + orig_response_addr
+                            + (orig_response_addr.equals(search.client) ? "" : ", using " + search.client)
+                            + (search.tls               ? " (TLS)" : "")
+                            + (search.unicast           ? " (unicast)" : "")
+                            + (search.reply_required    ? " (reply required)" : "")
+                            + (search.reply_to_src_port ? (origin == null ? " (reply to source port)" : " (reply to source port ignored because of origin tag)") : ""));
+                    search.channels.add(new Channel(cid, name));
+                }
+            }
+            catch (Exception ex)
+            {
+                logger.log(Level.WARNING, "PVA Client " + from + " sent damaged search #" + search.seq, ex);
+                return null;
             }
         }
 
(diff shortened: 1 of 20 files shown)


core/pva/src/main/java/org/epics/pva/data/PVAByteArray.java
```diff
@@ -114,6 +114,8 @@ public void encodeType(ByteBuffer buffer, BitSet described) throws Exception
     public void decode(final PVATypeRegistry types, final ByteBuffer buffer) throws Exception
     {
         final int size = PVASize.decodeSize(buffer);
+        if (size < 0  ||  size > buffer.remaining())
+            throw new Exception("Array size " + size + " with only " + buffer.remaining() + " bytes in buffer");
```
jacomago: Why not throw this inside PVASize.decodeSize? Then you don't need to update every single type.

jacomago (commented):
I think making custom exceptions is a good suggestion by sonarlint.

kasemir:
> I think making custom exceptions is a good suggestion by sonarlint.

Never understood that.

Exceptions are not status codes where based on the status code you would know exactly what to do.
If a method may throw several different exceptions and you want to catch specific ones because you know exactly how to react, those are basically status codes and shouldn't be exceptions in the first place.

An exception indicates that something happened about which we can't do anything.
Exceptions are for situations where there is no good solution other than to catch and report the exception at a higher level, and move on with something else.

core/pva/src/main/java/org/epics/pva/data/PVAByteArray.java
kasemir: Decoding the size doesn't fail. It parses the size just fine.
The calling code then checks if that size is possible within the context. If not, it throws an exception because we're stuck in a situation with no good way out other than give up and move on with the next message at a higher level. The exception text gives some potential hint to experts, showing if this is about an array size or structure size or string size or ...

jacomago:
> > I think making custom exceptions is a good suggestion by sonarlint.
> 
> Never understood that.
> 
> Exceptions are not status codes where based on the status code you would know exactly what to do.
> If a method may throw several different exceptions and you want to catch specific ones because you know exactly how to react, those are basically status codes and shouldn't be exceptions in the first place.
> 
> An exception indicates that something happened about which we can't do anything.
> Exceptions are for situations where there is no good solution other than to catch and report the exception at a higher level, and move on with something else.
> 

It's not just about catching custom exceptions (although I don't really understand your point about status codes, dB exception for EntityNotFoundException would makes sense there). Custom exceptions give you data structure when debugging and logging, for example you repeated the same exception message over and over in this pr. If you just make an ArrayBufferOutOFRangeException where you pass in the min, max and size you can write the message once, and format it once. Its also clear in the logs this was the type of exception that failed, I can also quickly find all places that exception is raised rather than exception I have to search for the message string. 

There are many other advantages I won't bother to list them all, but using raw Exception class is a very bad code smell to me.

core/pva/src/main/java/org/epics/pva/data/PVAByteArray.java
jacomago: Ok, then a tiny method 

checkRemainingSize(min, max, remaining) 

So that the code isn't repeated and you don't repeat yourself unnecessarily.

core/pva/src/main/java/org/epics/pva/common/SearchResponse.java
```diff
@@ -97,6 +97,9 @@ public static SearchResponse decode(final int payload, final ByteBuffer buffer)
         result.found = PVABool.decodeBoolean(buffer);
 
         final int count = Short.toUnsignedInt(buffer.getShort());
+        if (count*Integer.BYTES > buffer.remaining())
+            throw new Exception("PVA Server sent search reply #" + result.seq + " for " + count + " CIDs " +
```
jacomago, line 101: Can we have a unit test that checks this exception is raised.

kasemir:
Might need an alternate approach

shroffk:
Why do we need an alternate approach...while having structured exception do help (not during runtime but logging/debugging) but the main fix is far more important.

I have reopened this PR since I do not want to loose the work to address the memory exhaustion issue.

shroffk:
As I mentioned before the count vs buffer mismatch checks are a important catch

I have simply wrapped the exception in a dedicated `PVAArraySizeException`

shroffk (approved):

shroffk:
@kasemir can I merge this

kasemir:
Well, the way it replaces “PVA Client “ with a constant is ridiculous. Shouldn’t it then also replace each occurrence of “PVA” with a constant? Or internationalize them? I’d say no. Log and exception messages should remain in English. No matter where you run it, you need to rely on the same internal messages. That eases debugging. After all, we don’t replace class, function and filenames based on locale.

My other concern is custom exceptions. Bloch's “Efficient Java” suggests to use standard exceptions.
I'd use `IndexOutOfBoundsException` instead of a custom `PVAArraySizeException` and so on. 

But this should be decided by whoever will maintain PVA going forward, so forget what I just typed.

core/pva/src/main/java/org/epics/pva/common/SearchRequest.java
```diff
@@ -26,6 +26,9 @@
 @SuppressWarnings("nls")
 public class SearchRequest
 {
+    private static final String LOG_PVA_CLIENT = "PVA Client ";
```
jacomago, line 29: Rather than this I think the template for a format string makes sense? i.e. 

private static final String LOG_SENT_SEARCH_FORMAT = "PVA Client %s sent search #%s" ?

jacomago (commented):
Still missing a test.

jacomago:
> Well, the way it replaces “PVA Client “ with a constant is ridiculous. Shouldn’t it then also replace each occurrence of “PVA” with a constant? Or internationalize them? I’d say no. Log and exception messages should remain in English. No matter where you run it, you need to rely on the same internal messages. That eases debugging. After all, we don’t replace class, function and filenames based on locale.
> 

Looks like the class is doing structured logging, without defining the structure. Might be worth doing something like logSearchDetails() or something similar rather than the constants. I presume @shroffk added them to satisfy a lint saying don't repeat a string more than 3 times or something in a class.

> My other concern is custom exceptions. Bloch's “Efficient Java” suggests to use standard exceptions. I'd use `IndexOutOfBoundsException` instead of a custom `PVAArraySizeException` and so on.

I think you mean Effective Java?

Maybe https://docs.oracle.com/javase/8/docs/api/java/nio/BufferUnderflowException.html is more appropriate?

kasemir:
Yes, Effective. Looks like you know the book so this is in good hands. Carry on!

shroffk:
The constant was to primarily address the linter complain

I did contemplate the use of the `IndexOutOfBoundsException`, however there are other places where we are currently throwing a simple `Exception`, the `PVAProtocolException` in preparation for those exceptions.

I will cleanup the constant usage and merge this... again I think the key is capturing the situation which can lead to memory exhaution

PR #3880, opened 2026-07-13
added date of disabled pvs to structural element

milenaschedle:
If all the leafs of a structural element in the alarm tree have the same enable date, it is displayed next to the name of the component

Screenshot 2026-07-13 093428

- Testing:
    - [x ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
+++ b/app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
@@ -26,9 +26,7 @@ import org.phoebus.applications.alarm.ui.Messages;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 
 /** TreeCell for AlarmTreeItem
  *  @author Kay Kasemir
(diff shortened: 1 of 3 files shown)


app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeViewCell.java
```diff
@@ -127,22 +125,27 @@ protected void updateItem(final AlarmTreeItem<?> item, final boolean empty)
...
 
                     if (leavesDisabledStatusBooleanPair.getKey().equals(LeavesDisabledStatus.AllDisabled)) {
                         if (leavesDisabledStatusBooleanPair.getValue()) {
-                            disabledTimerIndicator.setText("(" + Messages.disabled + "; " + Messages.timer + ")");
-                        }
-                        else {
+                            Set<AlarmClientLeaf> withEnableDate = new HashSet<>();
+                            boolean test = DisableAction.checkEnableDates(node.getChildren(), new HashSet<>(), withEnableDate);
+                            if (test){
+                                disabledTimerIndicator.setText("(" + Messages.disabled + "; " + Messages.timer + ": " + withEnableDate.iterator().next().getEnabledDate() + ")");
```
abrahamwolk, line 137: The comment describing `DisableAction.checkEnableDates()` states:

> Returns true when all leaf items of the same structure either have no enable dates or all the same

Therefore, `withEnableDate.iterator().next().getEnabledDate()` not throwing an exception here (in the case that there are no enable dates) implicitly depends on `leavesDisabledStatusBooleanPair.getValue()` being `true` meaning that there is at least one timer in the sub-tree.

In practice, it seems that `DisableAction.checkEnableDates()` is used to determine whether there exists _one shared enable date_ for a sub-tree: I suggest to consider changing the signature of `DisableAction.checkEnableDates()` to something like the following to reflect this:
```
Optional<LocalDateTime> checkEnableDates(AlarmTreeItem<?>) 
```
Returning:

1. `Optional.empty()` if there is no shared enabled date.
2. `Optional.of(sharedEnabledDate)` if there is a shared enabled date.

This would remove the implicit dependency, also.

abrahamwolk, line 137: I suggest to format the enabled date with second-precision only, i.e., to remove the milliseconds, etc.

I don't know whether it would be a good idea to also implement logic to simplify the date depending on the _current date_: e.g., if the current date is `2026-07-15` and the enable date is `2026-07-15T10:00:00`, perhaps it would simplify the display to only display `10:00:00`, or similar.

georgweiss: ```DisableAction.checkEnableDates()``` is actually used also for another purpose, which has been a bit lost in the refactoring done in this commit. 
I will update to split the required business logic a bit to improve readability.

georgweiss: Yes, second precision is sufficient. But I prefer to always use the same format for the sake of consistency.

PR #3881, opened 2026-07-13
Optimizations in ScaledSlider

georgweiss:
As suggested by Mr Claude: do not adjust sizes of the slider UI elements on value change, only on size change.

Also tweaked default sizes a bit to make a fresh ScaledSlider resemble the native JavaFX slider with default paddings.

- Testing:
    - [ X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
@@ -7,10 +7,12 @@
  *******************************************************************************/
 package org.csstudio.display.builder.representation.javafx.widgets;
 
+import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.geometry.Insets;
 import javafx.geometry.Orientation;
 import javafx.scene.Node;
+import javafx.scene.control.Skin;
 import javafx.scene.control.Slider;
 import javafx.scene.control.skin.SliderSkin;
 import javafx.scene.input.KeyEvent;
@@ -37,7 +39,6 @@ import org.epics.vtype.VType;
 
 import java.text.DecimalFormat;
 import java.time.Instant;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 
 import static org.csstudio.display.builder.representation.ToolkitRepresentation.logger;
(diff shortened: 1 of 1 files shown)


app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
```diff
@@ -595,24 +597,39 @@ private void openConfigurationPanel()
      * </p>
      */
     private void adjustSizes(){
-        SliderSkin skin = (SliderSkin) slider.getSkin();
```
jacomago, line 598: Why remove this line? You are now calling this method twice.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
```diff
@@ -595,24 +597,39 @@ private void openConfigurationPanel()
...
+    }
+
+    /**
+     * Custom {@link ChangeListener} for the purpose of setting slider sizes once there is a {@link Skin}
+     * available to manipulate.
+     */
+    private class SkinChangeListener implements ChangeListener<Skin<?>>{
+        @Override
+        public void changed(ObservableValue<? extends Skin<?>> observableValue, Skin<?> oldValue, Skin<?> newValue) {
+            if(oldValue == null && newValue != null){
```
jacomago, line 631: Isn't it better to check if they are different?

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
georgweiss: Refactored this a bit. In general avoiding explicit cast if not necessary.

app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/ScaledSliderRepresentation.java
georgweiss: In my view better readability showing intent: want to intercept when skin property goes from null to non-null.

jacomago (approved):

Issue #3677, opened 2025-12-17
Execute Display window ignores width/ height


limesTech:
Similar to the Issue from 2023 see https://github.com/ControlSystemStudio/phoebus/issues/2729:

Version 5.0.2 still has this issue at our site. Any suggestion how to fix it?

In the Display editor, everything works fine and the dropdown width/height/all in a separate or docked window will scale according to the displayed red border. However, every time we use the execute display button, these autoscaling functions ignore the set display size (red border) in the file. This is more obvious in the docked window state.

Seems like Display does not get correct viewport sizes of the current window and in addition the display ignores the set display sizes.

Does anyone else can replicate it?

Editor Screenshot:
<img width="2395" height="1380" alt="Image" src="https://github.com/user-attachments/assets/50d5769b-d1bf-490b-8592-87ec3a7708ab" />

Executed Display manually zoomed out: (Background is greater than the editor borders) 
<img width="2399" height="1530" alt="Image" src="https://github.com/user-attachments/assets/dbf9b7c1-e1ac-4aa9-912d-066f5a6ebe91" />

Executed Display after 'height' zoom function: (mismatch of height to viewport (scrollbar))
<img width="2394" height="1525" alt="Image" src="https://github.com/user-attachments/assets/99adb44c-9b2d-41b9-871e-1e4678a66483" />

PR #3887, opened 2026-07-14
Stop scrollbars from flickering on/off when resizing windows

rjwills28:
We noticed that when you resize a window with the mouse you can see scroll bars flicker on and off. See attached screencast showing this:
[scrollbar_flicker.webm](https://github.com/user-attachments/assets/399a12ac-2928-4b22-9e3d-e14004f5491e)

You will also see that even when the window is made smaller than the screen in one direction, you still get scroll bars appear in both x and y even though they are only needed in one. 

The issue is caused by code that attempts to resize the widget pane to fill the parent scroll pane:https://github.com/ControlSystemStudio/phoebus/blob/11eb814d610cf978c41029604d00d8f0170ff689/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L651

This is a workaround to another issue of setting the background of a ScrollPane, see comments in code here: https://github.com/ControlSystemStudio/phoebus/blob/11eb814d610cf978c41029604d00d8f0170ff689/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L939 Essentially you can not set the background of the ScrollPane without it propagating down to the children node. Instead, the background was set on the widget pane and then a listener is added so that whenever the scrollpane is resized, the widget pane gets resized using its `setMinSize()` method to fill the ScrollPane. This must be causing some race condition where it momentarily thinks the pane is too small for the content and so puts on scroll bars before realising it is not and then removing them. 

The fix for the scroll bar issue described above is to remove this listener and the resizing of the widget pane. However this then leaves us with this problem of setting the background. I have been able to fix this by:
- setting the background on the Scroll pane's parent BorderPane.
- then setting the ScrollPane background to be transparent so we see the BorderPane's background underneath.
- Set the widget pane background as well to work around a JFX bug where a transparent ScrollPane causes text to render differently (see  https://bugs.openjdk.org/browse/JDK-8118303 and https://bugs.openjdk.org/browse/JDK-8095008).

I have tested this with all widgets and noticed some different behavior for the Tabs widget where the tab label text was white, as was the border. I fixed this by explicitly setting them in the code.

I think my implementation also goes some way to fixing https://github.com/ControlSystemStudio/phoebus/issues/3677, or at least when you zoom out the background still gets painted all the way out so you wouldn't see the grey background in this example. 

Manual tests to run:
- Check that you do not see scroll bars flicker on and off as you resize a window
- Check that scroll bars only appear in the dimensions where the window is too small
- Change the background of the display in the editor and check that this is reflected in both the runtime and editor. 
   - Expand the windows to make sure it is still painted out to the boundaries. 
   - Try zooming out - is the background still painted out to the boundaries
- Close Phoebus and open again so that it loads from the memento - is the background still drawn out to the boundaries.
- Open some existing BOB files - are the widgets all visible and look the same.
- Test creating a new BOB with a Tab widget. Does this behave as normal.

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java
@@ -225,9 +225,12 @@ public class JFXRepresentation extends ToolkitRepresentation<Parent, Node>
         model_root = new ScrollPane(scroll_body);
         model_root.setId(MODEL_ROOT_ID);
 
-        final InvalidationListener resized = prop -> handleViewportChanges();
-        model_root.widthProperty().addListener(resized);
-        model_root.heightProperty().addListener(resized);
+        // Update background of parent when it changes or, for example,
+        // becomes not null
+        final InvalidationListener background = prop -> {
+            updateBackground();
+        };
+        model_root.parentProperty().addListener(background);
 
         // Middle Button (Wheel press) drag panning started
         final EventHandler<MouseEvent> onMousePressedHandler = evt ->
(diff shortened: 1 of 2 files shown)


shroffk:
this is great
have you had a chance to check this issue with javafx25?

rjwills28:
Yes, I've been testing on both JFX21 and JFX25. The scroll bar flickering issue is there in both and my fix works on both.

shroffk (approved):

Issue #3891, opened 2026-07-20
After recent change, labels and the text from text update widgets disappear


kasemir:
After a recent change, labels and textupdate widgets have their text disappear because their foreground color turns transparent .

<img width="754" height="631" alt="Image" src="https://github.com/user-attachments/assets/777d4d87-888d-4e25-967a-2c1ddab911aa" />

<img width="519" height="562" alt="Image" src="https://github.com/user-attachments/assets/9a1eda55-0ba5-40c0-965b-aedc0e26d1f8" />

kasemir:
The problem is somehow related to the change in `JFXRepresentation` from #3887.

When I disable the `updateBackground()` change, labels show up fine.
As `JFXRepresentation.updateBackground ()` is written in #3887, labels and text update use transparent foreground colors.

kasemir:
It comes down to this one line from the `JFXRepresentation` change in https://github.com/ControlSystemStudio/phoebus/pull/3887

https://github.com/ControlSystemStudio/phoebus/blob/51a9bf6e4f882b21830f94531ba795a42f915c58/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/JFXRepresentation.java#L964-L965

```
// Now make the scroll pane transparent to see the parent background
model_root.setStyle("-fx-background: transparent; -fx-background-color: transparent; ");
```

With `setStyle ... transparent`, the text in labels and text update widgets becomes transparent as shown in the screenshot above or here:

<img width="1026" height="416" alt="Image" src="https://github.com/user-attachments/assets/62f87db8-144f-4f57-9e0c-2e4f9c9f33bb" />

When I comment that line, the text shows up as expected:

<img width="1013" height="376" alt="Image" src="https://github.com/user-attachments/assets/55805443-e85d-4653-a2a4-8f665ff8544d" />

rjwills28:
@kasemir - thanks for catching this! I was wondering why I hadn't seen this as I did a significant amount of testing. It turns out I had tested that my PR fixed the scroll bar issues in both JFX21 and JFX25, BUT my tests on how it affected the widgets was done on JFX 25 only. 

This issue is not there with JFX25, that is to say that the foreground text does render correctly and is not invisible. That's my bad for not fully testing it against JFX21 - sorry. How do we want to proceed - revert the commits until Phoebus moves to JFX25 - how far away from that are we?

kasemir:
> This issue is not there with JFX25

[...]

PR #3892, opened 2026-07-21
Workaround to JFX21 bug causing transparent labels

rjwills28:
See description and discussion of bug here: https://github.com/ControlSystemStudio/phoebus/issues/3891.

Bug in JFX21 means that setting the foreground color of the label widget using the setter fails and the label is rendered transparent. 

JFX21 bug: https://bugs.openjdk.org/browse/JDK-8336097

This bug has been fixed in JFX24. 

The workaround is to set the style using inline CSS. 

Once we have moved to JFX25 this workaround should be removed - I have added a TODO for this in the code.

## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/BaseLEDRepresentation.java
+++ b/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/BaseLEDRepresentation.java
@@ -229,6 +229,7 @@ abstract class BaseLEDRepresentation<LED extends BaseLEDWidget> extends RegionBa
         {
             final Color color = JFXUtil.convert(model_widget.propForegroundColor().getValue());
             label.setTextFill(color);
+            setTextFillColorStyle(label, JFXUtil.webHex(model_widget.propForegroundColor().getValue()));
             label.setFont(JFXUtil.convert(model_widget.propFont().getValue()));
 
             led.setStyle("-fx-stroke: " + JFXUtil.webRgbOrHex(model_widget.propLineColor().getValue()));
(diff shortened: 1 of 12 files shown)


kasemir (approved):

kasemir:
Looks like the Label in Group boxes also needs this fix.

kasemir:
And the labels in Tabs

kasemir:
.. and Radio Buttons as well as Check boxes.

Now I'm not sure any more if reverting until JFX25 comes out is more practical

kasemir (commented):
Fixes label and text update...

rjwills28:
Thanks for noting those @kasemir  - you're right - everywhere that we do a `setTextFill()` we need to apply the patch.

I have just done that for this branch - mostly to show what the solution would look like. I created a helper method to do this to save repeating the same code too much. But we have now modified 12 files...

**Note**, I rebased the branch to keep the changes to a single commit so that we can easily revert it when we move the JFX25. If you have previously checked out the branch there may be problems so I would advise switching back to master, deleting your local version of this branch (fix_jfx21_transparent_labels) and then check it out again. Something like:
```
git checkout master
git branch -D fix_jfx21_transparent_labels
git checkout fix_jfx21_transparent_labels
```

All that said, if we still think it is simpler to revert the original commits in the PR that caused this issue in JFX21, https://github.com/ControlSystemStudio/phoebus/pull/3887, then I am happy for that to be the strategy as long as we make a point to test and merge them again when we have moved to JFX25.

kasemir:
Once we get to JFX25, I think we agree that all will be fine.

But until that time, what do you prefer? Revert, or update those 12 files? Either way, it would be temporary to keep the main branch usable until JFX25

rjwills28:
To be honest I guess it feels like it would be easier to revert the previous PR commits as that only impacts 2 files instead of changing 12...

kasemir:
> easier to revert

Then do that, and re-apply when we're on JFX25.

kasemir:
Replaced by #3893

PR #3907, opened 2026-08-06
Spring4

shroffk:
## Checklist

- Testing:
    - [x] Tests were run

- Documentation:
    - Release notes:


services/save-and-restore/src/main/java/org/phoebus/service/saveandrestore/web/config/WebSecurityConfig.java
```diff
@@ -255,17 +256,17 @@ public LdapAuthoritiesPopulator authorities(BaseLdapPathContextSource contextSou
...
-            configurer.userDnPatterns(ldap_user_dn_pattern);
-        }
-        if (ldap_user_search_filter != null && !ldap_user_search_filter.isEmpty()) {
-            configurer.userSearchFilter(ldap_user_search_filter);
-        }
-        if (ldap_user_search_base != null && !ldap_user_search_base.isEmpty()) {
-            configurer.userSearchBase(ldap_user_search_base);
-        }
-        configurer.contextSource(contextSource);
+//        LdapAuthenticationProviderConfigurer configurer = new LdapAuthenticationProviderConfigurer();
```
jacomago, line 261: Rather delete than comment out. Always have git history to fallback on.

services/alarm-logger/src/test/java/org/phoebus/alarm/logging/rest/SearchControllerTest.java
```diff
@@ -4,28 +4,31 @@
...
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
-import static org.junit.jupiter.api.Assertions.*;
 
-@ExtendWith(SpringExtension.class)
-@ContextConfiguration(classes = ControllersTestConfig.class)
-@WebMvcTest(SearchController.class)
+/**
+ * Unit test for {@link SearchController} using standalone MockMvc setup.
+ * Migrated from @WebMvcTest (removed in Spring Boot 4.x) to
```
jacomago, line 20: Don't need this comment.

services/save-and-restore/src/test/java/org/phoebus/service/saveandrestore/web/config/ControllersTestConfig.java
```diff
@@ -131,7 +134,9 @@ public ExecutorService executorService() {
 
     @Bean
     public SnapshotUtil snapshotUtil() {
-        return new SnapshotUtil();
+        // Return a mock so tests can stub it with when(...); previously @MockBean was used
```
jacomago, line 150: unneeded comment

services/save-and-restore/src/test/java/org/phoebus/service/saveandrestore/web/controllers/HelpResourceTest.java
```diff
@@ -19,20 +19,36 @@
...
 
     @Autowired
-    protected MockMvc mockMvc;
+    private WebApplicationContext webApplicationContext;
+
+    private MockMvc mockMvc;
+
+    @BeforeEach
+    void setUp() {
+        mockMvc = MockMvcBuilders.webAppContextSetup(webApplicationContext).apply(springSecurity()).build();
```
jacomago, line 50: Feels weird this is repeated everywhere, feels like either we need a parent class for these tests or we're doing something wrong with how to use MockMvc.

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogSearchUtil.java
```diff
@@ -276,16 +279,53 @@ public static List<AlarmLogMessage> search(ElasticsearchClient client,
         }
         SearchRequest searchRequest = searchRequestBuilder.build();
         try {
-            SearchResponse<JsonNode> strResponse = client.search(searchRequest, JsonNode.class);
-            return strResponse.hits().hits().stream().map(hit -> {
-                JsonNode jsonNode = hit.source();
+            // Build the search request body as JSON for low-level API
```
jacomago: Why swap to the low level api?

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogSearchUtil.java
```diff
@@ -394,4 +471,49 @@ public static int getDateSpanInDays(String indexDateSpanUnits){
                 return -1;
         }
     }
+
+    /**
+     * Helper method to convert SearchRequest to JSON string for low-level API
+     * This works by using Jackson to serialize the SearchRequest object tree to JSON
+     * NOTE: We exclude 'index' from the JSON since indices are specified in the URL path
+     */
+    private static String buildSearchJson(SearchRequest searchRequest) throws IOException {
```
jacomago, line 394: Please add a unit test for this method.

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
```diff
@@ -73,15 +69,14 @@ public class ElasticClientHelper {
...
-    private final JsonMapper mapper = new JsonMapper();
-
     private ElasticClientHelper() {
         try {
             Runtime.getRuntime().addShutdownHook(new Thread(() -> {
                 logger.info("Shutting down the ElasticClientHelper.");
                 if (client != null) {
                     try {
-                        client.shutdown();
+                        // Do not call client.shutdown() with Rest5 transport because some
```
jacomago, line 78: Don't need this comment

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
```diff
@@ -97,19 +92,28 @@ private ElasticClientHelper() {
...
                         esPort.isEmpty() ? 9200 : Integer.parseInt(esPort));
                 esHttpHosts = new HttpHost[] {http_host};
             } else {
                 if (!esHost.isEmpty() || !esPort.isEmpty()) {
                     logger.warning("Only one of es_urls or es_host and es_port can be specified, ignoring es_host and es_port.");
                 }
-                esHttpHosts = Arrays.stream(esUrls.split(",")).map(HttpHost::create).toArray(HttpHost[]::new);
+                esHttpHosts = Arrays.stream(esUrls.split(","))
+                        .map(url -> {
+                            try {
```
jacomago, line 105: This looks a bit ugly, not sure of a correct fix though, maybe just use a normal for loop instead.

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
```diff
@@ -246,42 +259,89 @@ public void run() {
...
+                return;
+            }
+
+            Request request = new Request("POST", "/_bulk");
+            request.addParameter("refresh", "true");
+            request.setJsonEntity(bulkPayload.toString());
+
+            try {
+                int statusCode = restClient.performRequest(request).getStatusCode();
+                if (statusCode >= 200 && statusCode < 300) {
```
jacomago, line 335: Does a response not come with a 'success' or failure method?

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogSearchUtil.java
jacomago, line 1: Better to squash this with previous commit.

jacomago (commented):
First commits for spring boot are a nice commit structure, but not so much the elastic ones. I think better to pull out the elastic changes into a separate PR. 

I don't really know why you are swapping to the low-level client? If this is really needed I think you should make a ElasticResponseObjectMapper  and ElasticRequetObjectMapper to map between the objects with a strong set of unit tests. 

Some comments in the code I think would be better as part of the commit messages as they explain why something has changed not why it is now.

shroffk:
The low level elastic is need to ensure consistent functionality with the 9.x elastic client and 8.x version of elastic service

The low level client seems more resilient.

Using the old client conflicts with the upgrade to jackson3 and using the newer client fails with older service.... a lot of object mapping wrapping would be needed to continue using the high level client

That having been said... @georgweiss and I feel like for now we should move the main code base over to the newest elastic java client and use the low level rest client for now and address the elastic migration as a separate effort in the future

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/ElasticClientHelper.java
shroffk: no

services/alarm-logger/src/main/java/org/phoebus/alarm/logging/rest/AlarmLogSearchUtil.java
shroffk: https://github.com/ControlSystemStudio/phoebus/pull/3907#issuecomment-5253938508

jacomago:
> The low level elastic is need to ensure consistent functionality with the 9.x elastic client and 8.x version of elastic service
> 
That seems out of scope.

> Using the old client conflicts with the upgrade to jackson3 and using the newer client fails with older service.... a lot of object mapping wrapping would be needed to continue using the high level client
> 

That's annoying.

> That having been said... @georgweiss and I feel like for now we should move the main code base over to the newest elastic java client and use the low level rest client for now and address the elastic migration as a separate effort in the future

Yeah, I agree.

shroffk:
> That's annoying.

Tell me about it...I had no desire to include this work but it is the fastest and cleanest way to get to jdk15,spring4, jackson3

I think I have one of my first codeathon project ideas (elastic upgrades + high level client) here and I think I also have a volunteer.

georgweiss (commented):
A few things found here:

- The core/websocket/client module references Spring Boot 2.7.1, a version set in top level pom. Service modules reference 4.1.0. Suggest we have this version in the top level pom only. Thing is that core/websocket/client needs a few code changes when moving to 4.1.0.
- Jackson libs (core, databind, datatype-yaml) will be present in the build with varying versions (3.x, 2.2.x, 2.1.x). For instance, the elastic java client and the kafka streams clients pull in 2.x. Should we add exclusions to make sure we only get the 3.x version? I have not confirmed that this would work in runtime...

shroffk:
Thanks @georgweiss 
We should move the websocket module to springboot 4.1... I had moved each project over one by one instead of using the parent spring version was to break down the work into more manageable peices. With the unit and integration tests passing now we should revert back to a single parent spring version.

regarding jackson we should certainly add exceptions to ensure we don't have a hodge podge or versions pulled in. I think I already do that in some of the services but I understand we should do a through double check.

georgweiss:
@shroffk, would you like me to create a separate PR for the Spring Boot 4 migration for web sockets?

georgweiss:
@shroffk, as for jackson libs... I was able to add exclusions such that the build would include only 3.x. But did not test how that works when actually running services and clients. Let me know if you'd like me to work more on this and create a PR.

shroffk:
I have moved everything to `springboot 4.1.0`

@georgweiss how best to test the websockets? would save and restore be the best way to verify this.

With save a restore client I have verified that the switch to the new websocket client was able to successfully create the websocket and receive the update messages

georgweiss:
Yes, web sockets test is as simple as adding a folder or other object to save-and-restore and then check if the tree view is updated. The change to the data is communicated over web sockets.

shroffk:
Done

```
2026-08-18 14:16:18 INFO [org.phoebus.core.websocket.client.WebSocketClientService] Attempting web socket connection to ws://localhost:8080/save-restore/web-socket
2026-08-18 14:16:18 INFO [org.phoebus.core.websocket.client.WebSocketClientService] Connected to web socket
2026-08-18 14:16:40 INFO [org.phoebus.core.websocket.client.WebSocketClientService] Handling subscription frame: {"messageType":"NODE_ADDED","payload":"f33a2f79-90ed-4ea9-b4ad-096dc834c0a5"}
```

georgweiss:
@shroffk, turns out jackson 2.x libs are needed for springdoc, i.e. for the services. However, when building the product these are also pulled in due to the phoebus-target being included in the product pom. This I think is not necessary. Removing the phoebus-target dependency builds a package with only 3.x jackson libs.

Still need to verify/test a bit.

georgweiss:
@shroffk, there are additional changes needed in ```WebSocketClientService``` due to the async nature of the connect call.

Will update this PR...

georgweiss:
@shroffk, fiddling with exclusions to suppress addition of jackson 2.x libs is in my view not the correct approach. Instead we should remove the phoebus-target dependency in the product pom. 

As far as I understand it, phoebus-target is just a way to compile a list of 3rd party libs for non-maven build tools (ant?). If that is the case then including phoebus-target in the product pom just blurs the actual dependency graph.

Currently phoebus-target cannot be removed from the product pom as that has side effects (I will spare you the details).

For the time being I think we need to bite the bullet and accept multiple jackson libs. Going forward we need to make an effort to cleanup in the product pom.

PR #3910, opened 2026-08-19
Changes to render full information on alarm tree non-leaf nodes. Also…

georgweiss:
This is a follow up on #3880, aiming at reducing code complexity while also updating the alarm tree view to provide more information on non-leaf nodes.

Suggestion is to show full information on non-leaf nodes:
* The total PV (leaf) count in sub-structure.
* Number of leaves disabled without enable date, if any
* Number of leaves disabled with enable date, if any
* The enable date if the same for all leaves with enable date.

Screenshots below. 

The view may become a bit cluttered as shown in the first example. However, the second example should probably reflect a more realistic scenario.

Screenshot 2026-08-19 at 08 57 42

Screenshot 2026-08-19 at 09 29 03

- Testing:
    - [ X] The feature has automated tests
    - [ X] Tests were run

- Documentation:


jacomago (approved):

georgweiss:
@shroffk, any objections to merging this?

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeHelper.java
```diff
@@ -104,18 +108,105 @@ public static boolean validateNewPath(String path, AlarmTreeItem<?> root)
...
 		    }
 		}
 
 		return true;
 	}
+
+	protected static Set<AlarmClientLeaf> getLeafItems(List<AlarmTreeItem<?>> items){
+		Set<AlarmClientLeaf> leaves = new HashSet<>();
+		items.forEach(item -> leaves.addAll(getLeafItems(item)));
+		return leaves;
```
abrahamwolk: I think this results in much copying and allocation on the heap: every leaf is copied once per layer from the leaf in question to the root.

I think this and the next method (which takes as argument a `AlarmTreeItem<?> root`) can be combined into a method that takes only a `final AlarmTreeItem<?> root` as its argument, plus a helper method for traversing the tree, along the following lines:

```
    protected static Set<AlarmClientLeaf> getLeafItems(final AlarmTreeItem<?> root) {
        return streamLeafItems(root).collect(Collectors.toSet());
    }

    private static Stream<AlarmClientLeaf> streamLeafItems(final AlarmTreeItem<?> alarmTreeItem){
        if (alarmTreeItem instanceof AlarmClientLeaf alarmClientLeaf){
            return Stream.of(alarmClientLeaf);
        }
        else {
            return alarmTreeItem.getChildren().stream().flatMap(child -> streamLeafItems(child));
        }
    }
```

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeHelper.java
```diff
@@ -104,18 +108,105 @@ public static boolean validateNewPath(String path, AlarmTreeItem<?> root)
...
+	}
+
+	protected static Set<AlarmClientLeaf> getLeafItems(final AlarmTreeItem<?> root){
+		Set<AlarmClientLeaf> leaves = new HashSet<>();
+		if(root instanceof AlarmClientLeaf){
```
abrahamwolk: The cast can be made type safe:
```
if(root instanceof AlarmClientLeaf alarmClientLeaf) {
```

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeHelper.java
georgweiss: > every leaf is copied once 

What does "copied" here mean? No new ```AlarmClientLeaf``` objects are created. 
But maybe you mean new ```Set``` objects are created and ```AlarmClientLeaf``` references are copied over?

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeHelper.java
georgweiss: Will udate.

shroffk (approved):

app/alarm/ui/src/main/java/org/phoebus/applications/alarm/ui/tree/AlarmTreeHelper.java
abrahamwolk: You are correct: "copying" was inaccurate. I should have written "re-insertion" instead: a reference to every leaf is inserted into a `HashSet` once per layer from the leaf in question to the root, and each re-insertion (1) hashes the element and (2) allocates a node for the hash set.

abrahamwolk (approved):

PR #3911, opened 2026-08-19
Add section detailing performance improvements to documentation

rjwills28:
We ran some performance tests on Phoebus using multiple windows displaying 100s of PVs updating at 10Hz. We discovered that the displayed update rate was much slower than the actual PV update rate. After some further investigations we found that there are 2 environment variables that can be set which drastically improved the performance of the displays when running using hardware acceleration. 

I have added a section to the Phoebus documentation detailing these variables and how to use them, just in case others should come across this issue and might benefit from this advice.

I think having this in the 'Starting CS-Studio/Phoebus' makes sense but if anyone has other ideas then please just let me know and I can move it.

## Checklist

- Testing:
    - [ ] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/docs/source/running.rst
+++ b/docs/source/running.rst
@@ -124,11 +124,11 @@ will contact the already running instance and have it open the requested file.
 Performance
 -----------
 
-Phoebus displays are built on JavaFX, which will use hardware acceleration wherever possible. If your screens display PVs with high update rates but you notice that the displayed update rates are not keeping up, you may benefit from using the following environment and Java options::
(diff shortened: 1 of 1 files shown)


kasemir:
Does this apply to any operating system or only to for example Linux?

rjwills28:
Ah good question. Tested on Linux RedHat and Ubuntu OSs. Not tested on Windows. I can clarify that in the text.

shroffk (approved):

PR #3912, opened 2026-08-21
Updated comparison strategy for VEnum in save&restore

georgweiss:
To resolve a (corner) case where labels are not set or changed, the comparison of ```VEnum```s will consider both index and labels. For instance, if a label is updated (e.g. due to typo), stored and live value will be considered different even if index stays the same.

Additionally ```VEnum```s will also be rendered to show both index and label. This should be helpful to determine the cause of inequality, and also help find PVs where labels have not been defined.

Screenshot 2026-08-21 at 12 47 25

- Testing:
    - [ X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
+++ b/app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
@@ -264,7 +264,7 @@ public final class Utilities {
             return VStringArray.of(list, alarm, time);
         } else if (type instanceof VBooleanArray) {
             String[] elements = data.split(",");
-            List<String> list = Arrays.stream(elements).map(String::trim).collect(Collectors.toList());
+            List<String> list = Arrays.stream(elements).map(String::trim).toList();
             boolean[] booleans = new boolean[list.size()];
             for (int i = 0; i < list.size(); i++) {
                 booleans[i] = Integer.parseInt(list.get(i)) > 0;
@@ -354,7 +354,7 @@ public final class Utilities {
             StringBuilder sb = new StringBuilder(size * 15 + 2);
             sb.append('[');
             Pattern pattern = Pattern.compile(",");
-            NumberFormat formatter = ((SimpleValueFormat) FORMAT.get()).getNumberFormat();
+            NumberFormat formatter = (FORMAT.get()).getNumberFormat();
             if (type instanceof VDoubleArray) {
                 for (int i = 0; i < size; i++) {
                     sb.append(pattern.matcher(formatter.format(list.getDouble(i))).replaceAll(".")).append(COMMA)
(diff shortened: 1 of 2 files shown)


app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
```diff
@@ -426,26 +426,14 @@ public static String valueToString(VType type, int arrayLimit) {
             return sb.toString().trim();
         } else if (type instanceof VNumber) {
             if (type instanceof VDouble) {
```
abrahamwolk, line 428: Can be cast in a type safe way as `type instanceof VDouble vDouble)`. (Analogously also in other branches.)

app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
```diff
@@ -648,20 +636,22 @@ public static VTypeComparison valueToCompareString(VType value, VType baseValue,
...
-        } else if (value instanceof VEnum && baseValue instanceof VEnum) {
+        } else if (value instanceof VEnum vEnumValue && baseValue instanceof VEnum vEnumBaseValue) {
             String str = valueToString(value);
-            String b = ((VEnum) value).getValue();
-            String c = ((VEnum) baseValue).getValue();
-            int diff = b == null ? (c == null ? 0 : 1) : (c == null ? -1 : b.compareTo(c));
-            return new VTypeComparison(str, diff, diff == 0, Math.abs(diff));
+            int diff = Integer.compare(vEnumValue.getIndex(), vEnumBaseValue.getIndex());
+            // Index diff may be 0, but labels may still differ
+            if(diff == 0){
```
abrahamwolk, line 651: The comparison of (1) the index _and_ (2) the associated label when the indices match, is implemented in three different places: here, in lines `860`-`864`:
https://github.com/ControlSystemStudio/phoebus/blob/7ac7bc69a5d3bb290ad8e56291bc3994fc595b14/app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java#L860-L864

and in lines `1020`-`1021`:
https://github.com/ControlSystemStudio/phoebus/blob/7ac7bc69a5d3bb290ad8e56291bc3994fc595b14/app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java#L1020-L1021

I would suggest to factor out a helper function that implements the comparison in one place, since it will make updating the definition less error-prone.

app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
georgweiss: This has not been updated in this PR. 

Once we move to Java25, the entire class will be revisited to simplify nested if statements. Any other improvements will be considered at that time.

app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
georgweiss: Once we move to Java25, the entire class will be revisited to simplify nested if statements. Any other improvements will be considered at that time.

abrahamwolk: Why wait for Java 25?

georgweiss: A change done now may need to be changed again during upcoming refactoring.

abrahamwolk: I would still think it's good to factor out the definition of the comparison, since it should be shared. A future re-factoring may or may not notice that the definition should be factored out.

georgweiss: I disagree as the type of refactoring you propose is applicable to several parts of ```Utilities``` and will for sure be considered.

abrahamwolk: OK!

abrahamwolk (approved):

app/save-and-restore/util/src/main/java/org/phoebus/saveandrestore/util/Utilities.java
shroffk: yes,
this should all be redone with jdk15 migration so for I would like to close this.

shroffk (approved):

PR #3915, opened 2026-08-24
Add preference to use the 'optimized with last sample' AA postprocessor

rjwills28:
A while back we introduced a new post processor to the Archive Appliance that better represents the data called 'Optimized with last sample'.

The history and detailed discussion of why this pp was required is here: https://github.com/ControlSystemStudio/cs-studio/issues/2483. 

In brief, say a PV has a flurry of activity, then goes to a value (say 0) and does not change for some time. With the Optimized PP, you would get the average during this flurry, say 50 and then the next bins will have no events, in which case the Optimized algorithm just returns the mean value of the last bin that had events, e.g. it would return 50. This means that data browser would show the value '50' for the entire duration that the PV does not change, even though the PV's last value was 0. 

The optimized with last sample algorithm is based on the optimized one but if a bin does not have any events then it uses the last value in the previous bin that had events and hence in this example would return 0 as desired.

This new post processor was introduce to the CS-Studio data browser as a preference that could be switched on. In this PR I have added a similar preference to use the `optimLastSample_` option.

At the moment I have left the default as 'false', i.e. it is not used but I wonder, given that this post processor gives a better representation of the data, whether it should be the default to use (i.e. make the variable true) - thoughts?  
## Checklist

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:

### DIFF
--- a/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
+++ b/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
@@ -69,7 +69,7 @@ public class ApplianceOptimizedValueIterator extends ApplianceValueIterator {
         String optimizedOperator = ApplianceArchiveReaderConstants.OP_OPTIMIZED;
         if (AppliancePreferences.ppOptimizedWithLastSample)
             optimizedOperator = ApplianceArchiveReaderConstants.OP_OPTIMIZED_WITH_LAST_SAMPLE;
-        
+
         String optimized = new StringBuilder().append(optimizedOperator)
                 .append(requestedPoints).append('(').append(pvName).append(')').toString();
         super.fetchDataInternal(optimized);
--- a/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
+++ b/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
@@ -66,7 +66,11 @@ public class ApplianceOptimizedValueIterator extends ApplianceValueIterator {
      */
     @Override
     protected void fetchDataInternal(String pvName) throws ArchiverApplianceException {
-        String optimized = new StringBuilder().append(ApplianceArchiveReaderConstants.OP_OPTIMIZED)
+        String optimizedOperator = ApplianceArchiveReaderConstants.OP_OPTIMIZED;
+        if (AppliancePreferences.ppOptimizedWithLastSample)
+            optimizedOperator = ApplianceArchiveReaderConstants.OP_OPTIMIZED_WITH_LAST_SAMPLE;
+        
+        String optimized = new StringBuilder().append(optimizedOperator)
                 .append(requestedPoints).append('(').append(pvName).append(')').toString();
         super.fetchDataInternal(optimized);
     }
--- a/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/AppliancePreferences.java
+++ b/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/AppliancePreferences.java
@@ -20,6 +20,7 @@ public class AppliancePreferences {
     @Preference static boolean useStatisticsForOptimizedData;
     @Preference static boolean useNewOptimizedOperator;
     @Preference static boolean useHttps;
+    @Preference static boolean ppOptimizedWithLastSample;
 
     static {
     	AnnotatedPreferences.initialize(AppliancePreferences.class, "/appliance_preferences.properties");
--- a/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceArchiveReaderConstants.java
+++ b/app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceArchiveReaderConstants.java
@@ -52,6 +52,8 @@ public class ApplianceArchiveReaderConstants {
     public static final String OP_MEAN = "mean_";
     /** Operator for the optimized post processor */
     public static final String OP_OPTIMIZED = "optimized_";
+    /** Operator for the optimized with last sample post processor */
+    public static final String OP_OPTIMIZED_WITH_LAST_SAMPLE = "optimLastSample_";
 
     /** The schema delimiter: ca://pvName or pva://pvname */
     static final String SCHEMA_DELIMITER = "://";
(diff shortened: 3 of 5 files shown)


app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
```diff
@@ -66,7 +66,11 @@ public ApplianceOptimizedValueIterator(ApplianceArchiveReader reader, String nam
      */
     @Override
     protected void fetchDataInternal(String pvName) throws ArchiverApplianceException {
-        String optimized = new StringBuilder().append(ApplianceArchiveReaderConstants.OP_OPTIMIZED)
+        String optimizedOperator = ApplianceArchiveReaderConstants.OP_OPTIMIZED;
+        if (AppliancePreferences.ppOptimizedWithLastSample)
```
jacomago, line 70: switch maybe makes more sense

app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/AppliancePreferences.java
```diff
@@ -20,6 +20,7 @@ public class AppliancePreferences {
     @Preference static boolean useStatisticsForOptimizedData;
     @Preference static boolean useNewOptimizedOperator;
```
jacomago, line 21: Shouldn't we instead be changing this to an enum?

jacomago (commented):
Isn't it a bit limiting to make this a configuration parameter rather than replace the 'use_raw' on the trace to be an enum of 'optimized', 'optimizedWithLastSample', 'raw'. Do we really want for every PV optimizedWithLastSample over optimized? Or is it a case by case basis? I always think of a temperature sensor vs switch triggered, one is a continuous value that we sample in time whereas a switch triggered is an event that happens. Those two usually have a different way you want to think about 'last sample' included.

Also optimized is a bad name, we should try avoid keeping it. I think we should consider a better name that describes what it does, displayMinMaxMean or something. 'Optimized' doesn't say what it is optimized for.

rjwills28:
Thanks for taking the time to look at this and for the ideas. We could definitely expand the 'Request' option to be an enum and include the optimLastSample instead of just having essentially optimized = On/Off. A couple of points/questions though:

- In response to whether we would always want `optimLastSample` given the two cases of a temperature sensor and a switch:
  - Wouldn't the switch be represented by an enum (e.g. On/Off) in which case the optimized/optimLastSample option doesn't get used anyway as it either falls back to raw or 'nth_' . So whether we're using optimized/optimLastSample wouldn't matter in this case. 
  - And then the temperature sensor is a case where we might want to use optimLastSample to catch cases where this might not change for a while? 

  It's not clear to me whether there is a case where you want 'Optimized' over 'OptimLastSample' but maybe I haven't missed something? The only difference in behvaiour between the two is when there are no samples in a bin. 
- Optimized name change: I agree it's not obvious what you are going to get when you select 'Optimized'  but I thought the original use of the `Optimized` name was to signify that the data browser requests a reduced set of samples, i.e. it was more to do with limiting how much data gets sent back from the request in terms of memory usage? In terms of an alternative name, whether the min/max/stddev actually get displayed would depend on the 'Trace Type' so maybe that shouldn't be in the name of the 'Request'. Another option could be 'Binned'? If we did go down the route of making the 'Request' an enum then I'm not sure what you would call the 'optimLastSample' version - 'BinnedWithLastSample' isn't very clear.

jacomago:
> Thanks for taking the time to look at this and for the ideas. We could definitely expand the 'Request' option to be an enum and include the optimLastSample instead of just having essentially optimized = On/Off. A couple of points/questions though:
> 
You convinced me optimLastSample is just better 😀  Configuration parameter is then good. We can leave the renaming/documenting what 'optimized' means to another PR.

app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/AppliancePreferences.java
rjwills28: We have decided to leave this configuration parameter as is so that it can replace the original 'optimized' post processor.

app/trends/archive-reader/src/main/java/org/phoebus/archive/reader/appliance/ApplianceOptimizedValueIterator.java
rjwills28: I assume this was referencing when we might have changed the request options to be an enum and could have used a switch statement to switch between the cases. As we have decided to leave the configuration parameter as a replacement to the 'Optimized' request we will not need a switch-statement -> resolving.

kasemir:
Long term, something to consider in a larger update would be the "plot binning" as implemented in the original channel archiver, see "2.7.5 Plot-Binning" in https://controlssoftware.sns.ornl.gov/chanarch/manual.pdf

It silently switched from returning raw samples to a sequence of `{ initial sample in bin, minimum, maximum, final sample }` for each plot bin. When simply connecting all those values with lines, you basically get the same plot for raw vs. optimized (binned) samples while reducing the amount of data that needs to be in the plot.
You get the final sample to fix the problem mentioned in this PR, plus the initial to solve similar issues with samples that come in bursts, long times without change and then suddenly many changes that should be "optimized".

The downside of the original implementation was doing it silently. You couldn't tell if you're looking at  `sample1 , sample2, ... ` or `initial , minimum, maximum, final, ...`. Now we show via the data type, `Double` vs. `VStatistics` how we automatically switch to optimized samples.  An (updated) `VStatistics ` data type could identify the initial and final value within the bin as such, allowing you to draw the line to the initial and then on from the final sample, while using a background area color to indicate the min/max range and so on.

shroffk:
> It's not clear to me whether there is a case where you want 'Optimized' over 'OptimLastSample' but maybe I haven't missed something? The only difference in behvaiour between the two is when there are no samples in a bin.

I can't either... I would also vote for just having "OptimizedWithLastSample" 

People have trouble with the 2 options I don't want to create a long list of data decimation options which no one will understand ( not bother to read up on ) and then complain that the archiver is broken.

shroffk (approved):

jacomago:
> Long term, something to consider in a larger update would be the "plot binning" as implemented in the original channel archiver, see "2.7.5 Plot-Binning" in https://controlssoftware.sns.ornl.gov/chanarch/manual.pdf
> 

At least the archiver side already has support: https://epicsarchiver.readthedocs.io/en/stable/user/userguide.html#processing-of-data

rjwills28:
Ah yes, going to require test changes - let me fix that.

rjwills28:
@shroffk - this should be ready to go now with the optimLastSample set as the default for the 'optimized' request.

PR #3921, opened 2026-09-10
Read Display information from each sample for raw requests

georgweiss:
In order to support Display information on archiver appliance sample level (as opposed to payload info), this PR updates the way the Display object is constructed. It will be based on field values if present (raw requests), otherwise fallback to payload info kicks in.

- Testing:
    - [X] Tests were run

- Documentation:
    - Release notes:


georgweiss:
Closing this (temporarily?). Does not work as intended.

PR #3925, opened 2026-09-17
feat: add Rotation function for embedded display

agaget:
I have added the function to rotate the embedded display.

# context
PLC Guy are creating their own widget using Phoebus Polyline (no SVG and no special Java class they are not comfortable with java). Instead of creating one different for each angle we want, they want to be able to rotate the object

image

The base idea was to add the rotation function for the polyline widget, but fact is adding the rotation field to embedded display cover more case because I can create a bob with whatever I want in it.

# How it works
Just added a Double rotation field in the widget, and it rotates the layout in the embedded display. It's a double because in javaFX the function setRotate of the Rotate object is a double
image

# test
Tested with the different behaviour as seen above, seems to work fine. (stretch, resize etc ...)

# Advantage
- Every widget or drawing, can now be rotate
- For sure you can do some marvellous animation to open a new embedded displat associating rotation and zoom if you want now

# Disavantage
- if the embedded diplay is not square, when rotating it goes beyond the border... I don't see a case where this is a problem
- During edition there is this red border appearing, not exactly clear what is it.
- People are now able to do Powerpoint transition with the embeddedDisplay like in the early 2000's 😨 

@lcaouen


georgweiss:
@agaget, at ESS we use the Image or Symbol widget for these use cases. Both can be rotated. Using an embedded display, and then rotate the content seems a bit odd.

agaget:
> @agaget, at ESS we use the Image or Symbol widget for these use cases. Both can be rotated. Using an embedded display, and then rotate the content seems a bit odd.



- You are creating your svg/png file on another software then import it in a container Symbol or Image.

- I'm creating my bob file on Phoebus then import it in a container Embedded Display.

For me it's the same thing 😄  except I'm all Phoebus, it's vectorized and no need to find and masterize another software to create SVG at the correct SVG format (I know for sure every svg files are not compatible with Phoebus), and I can use rules and script

georgweiss:
A free tool like Inkscape can be used to create and manage SVGs. Indeed, support for SVG in Phoebus has limitations, but keep in mind that not all SVG tools create correct SVGs.

Maintaining a SVG-based symbol in a dedicated app should in most cases be simpler than editing in Phoebus, which does not claim to be a first class drawing tool.

kasemir:
I'd echo the "is this really necessary" concern.

If it is, consider limiting the rotation to the `RotationStep` option used in several other widgets, and check the usage of `was_ever_transformed` in for example the Label implementation, https://github.com/ControlSystemStudio/phoebus/blob/master/app/display/representation-javafx/src/main/java/org/csstudio/display/builder/representation/javafx/widgets/LabelRepresentation.java
That way, instead of adding a 1:1 and thus superfluous scaling and rotation to all the non-rotated widgets, those transformations are only added for instances that actually need them.

As for '... embedded diplay is not square, when rotating it goes beyond the border.....':
The restriction to `RotationStep` options makes it easier to properly handle the width and height.
Might also make it easier to implement in the web runtime and the next display tool.

As for `During edition there is this red border appearing, not exactly clear what is it.`:
Add Scenic View to your runtime to debug the scene graph and determine where it adds the red pane.
https://github.com/ControlSystemStudio/phoebus/blob/d095b2873ba621fb2fef9770801cc2fff3ad00ab/app/display/Readme.md?plain=1#L443-L448

agaget:
> I'd echo the "is this really necessary" concern.

Yes I think it is. I could also argue that adding rotation to EmbeddedDisplay cover every future case for any need of a rotation, that nor you nor me have imagined yet. 
Moreover among other examples I  can change by using rules, scripts or color.def, the color of my rotating object more easily, that using Symbol or Image where I need to create differents objects each time. 

> If it is, consider limiting the rotation to the `RotationStep` option used in several other widgets,

Unfortunately I've already faced accelerator with angle being 22.5°, so I think it is usefull to get full range. Maybe I can limit it to -180° to 180° ?


>check the usage of `was_ever_transformed` 

will do

PR #3926, opened 2026-09-18
During edition, widget not really invisible anymore 

agaget:
When invisible selected, during edition widget get really invisible resulting to the fact that this is hardly selectionnable (click on it doesn't select it, we have to select it in a selection area) and we tend to forgot that there is an invisible widget and let it forever after that. We can only see it throught the widget tree.

image

In CSS Boy it was handle better in my opinion.
I propose a little modification to make it only invisible and desaturate (like a ghost). Maybe the effect is not that clear compared to the disable effect that's why it's more transparent and desaturate.

image

For now it's only a draft, before applying it to all widget (some doesn't extend the function from JFXBaseRepresentation), I wanted to be sure that you are ok with that proposition.

What do you think ?


georgweiss:
@agaget, could we consider a property/setting for this? Maybe an opacity property that defaults to zero to make this "backwards compatible"?

agaget:
> @agaget, could we consider a property/setting for this? Maybe an opacity property that defaults to zero to make this "backwards compatible"?

Yes i can consider that :) 

I understand the need for backward compatibility but I fear that putting it to 0 by default will just bury the functionnality for the eyes of every user but me ^^. 
I don't really see a case where this is not an improvement. At runtime no difference so operation is not disturb and at edition, if it disturbs the developer, she.he could (finally) handle this invisible object the way she.he wants.

But I will do as you please.

PR #3927, opened 2026-09-18
Add underline style font

lcaouen:
Add the underline style font to some widgets.

List of the widgets concerned by this PR :
- ActionButtonRepresentation
- BaseLEDRepresentation
- BoolButtonRepresentation
- CheckBoxRepresentation
- ChoiceButtonRepresentation
- LabelRepresentation
- RadioRepresentation
- SlideButtonRepresentation
- TextSymbolRepresentation
- TextUpdateRepresentation

Testing has been done manually by checking the widgets in a bob file.

image


georgweiss:
I think this looks OK.
@shroffk, any reason to not merge?

georgweiss:
@lcaouen, I spotted a potential issue...

If a display where a widget has selected one of the new styles is loaded into a version of Phoebus where this is not supported, the editor will show an error dialog and the widget will be highlighted as problematic, see below.

Maybe we should consider a fallback, i.e. if the style cannot be found, fallback to ```WidgetFontStyle.REGULAR```, and maybe just log the problem.

Screenshot 2026-09-18 at 14 39 34

```
java.lang.IllegalArgumentException: No enum constant org.csstudio.display.builder.model.properties.WidgetFontStyle.
	at java.base/java.lang.Enum.valueOf(Enum.java:293)
	at org.csstudio.display.builder.model.properties.WidgetFontStyle.valueOf(WidgetFontStyle.java:15)
	at org.csstudio.display.builder.model.properties.FontWidgetProperty.readFromXML(FontWidgetProperty.java:102)
	at org.csstudio.display.builder.model.WidgetConfigurator.configureAllPropertiesFromMatchingXML(WidgetConfigurator.java:136)
	at org.csstudio.display.builder.model.WidgetConfigurator.configureFromXML(WidgetConfigurator.java:104)
	at org.csstudio.display.builder.model.widgets.ActionButtonWidget$ActionButtonConfigurator.configureFromXML(ActionButtonWidget.java:181)
	at org.csstudio.display.builder.model.persist.ModelReader.createWidget(ModelReader.java:326)
	at org.csstudio.display.builder.model.persist.ModelReader.readWidget(ModelReader.java:298)
```

shroffk:
>Maybe we should consider a fallback, i.e. if the style cannot be found, fallback to WidgetFontStyle.REGULAR, and maybe just log the problem.

+1
I think we should add the fallback

Other than that I think this is a good addition

lcaouen:
> @lcaouen, I spotted a potential issue...
> 
> If a display where a widget has selected one of the new styles is loaded into a version of Phoebus where this is not supported, the editor will show an error dialog and the widget will be highlighted as problematic, see below.
> 
> Maybe we should consider a fallback, i.e. if the style cannot be found, fallback to `WidgetFontStyle.REGULAR`, and maybe just log the problem.
> Screenshot 2026-09-18 at 14 39 34
> 
> ```
> java.lang.IllegalArgumentException: No enum constant org.csstudio.display.builder.model.properties.WidgetFontStyle.
> 	at java.base/java.lang.Enum.valueOf(Enum.java:293)
> 	at org.csstudio.display.builder.model.properties.WidgetFontStyle.valueOf(WidgetFontStyle.java:15)
> 	at org.csstudio.display.builder.model.properties.FontWidgetProperty.readFromXML(FontWidgetProperty.java:102)
> 	at org.csstudio.display.builder.model.WidgetConfigurator.configureAllPropertiesFromMatchingXML(WidgetConfigurator.java:136)
> 	at org.csstudio.display.builder.model.WidgetConfigurator.configureFromXML(WidgetConfigurator.java:104)
> 	at org.csstudio.display.builder.model.widgets.ActionButtonWidget$ActionButtonConfigurator.configureFromXML(ActionButtonWidget.java:181)
> 	at org.csstudio.display.builder.model.persist.ModelReader.createWidget(ModelReader.java:326)
> 	at org.csstudio.display.builder.model.persist.ModelReader.readWidget(ModelReader.java:298)
> ```

Yes I've seen this issue too.
I can add the fallback, however it won't solve the issue as it appears only in versions of Phoebus which do not implement the underline, and the fallback will be in the version which implements it. So the modification will only solve problems if we implement new font features in the future.

georgweiss:
@lcaouen, you are right that he fallback will not help unless it's implemented.

May I suggest to make this a two-step solution: first would be to have a generic fallback, which would then be included in the next release that we publish. Once that has been done we could add support for underline.

I expect we will do a release soon (ping @shroffk) as we want to move on to Java 25.

lcaouen:
> @lcaouen, you are right that he fallback will not help unless it's implemented.
> 
> May I suggest to make this a two-step solution: first would be to have a generic fallback, which would then be included in the next release that we publish. Once that has been done we could add support for underline.
> 
> I expect we will do a release soon (ping @shroffk) as we want to move on to Java 25.

Sounds good to me, I gonna create a new PR for the generic fallback.

shroffk:
I see the new PR has been created and merged... So I will go ahead and make a release.

I understand that we might still see the exceptions for people using older versions of Phoebus.. but that is ok...

georgweiss:
@shroffk, the idea was to build a release with the fallback only.

shroffk:
I understand that...but if people are still using some version of Phoebus 5.x.x then they will encounter the issue you highlighted.

georgweiss:
Yes of course, but we would be able to point to a release where this is handled.

In any case, I do not really have a strong opinion on this.

